PR #6, opened 2026-01-08
chore: remove required fields from JSON schemas

nickcheng:


iceboundrock (commented):
Let's keep IDs and Created/UpdatedAt required. I will take from here.

iceboundrock:
I updated the schemas and test data.

PR #8, opened 2026-02-01
fix(cdc): align flush marking with snapshot

QuakeWang:
## Problem
 CDC flush used to mark change_log.flushed_at only by row_id, without enforcing changed_at <= snapshot.
If the same row_id was updated during the export, the new change could be incorrectly marked as flushed, causing it to be skipped by subsequent CDC runs (data loss).

## Fix
 - Gate marking by the snapshot boundary (changed_at <= snapshot) so only the exported window is marked
 - Update the manifest with the actually-flushed row_ids to avoid count drift
 - Add an integration test for the concurrent update scenario

### DIFF
--- a/internal/cdc/flusher.go
+++ b/internal/cdc/flusher.go
@@ -330,21 +330,30 @@ func executeFlush(
 				return
 			}
 			flushedAt := time.Now().UnixMilli()
-			rowsUpdated, err := MarkFlushedIDs(ctx, db, tableName, schemaID, sub, flushedAt)
+			updatedIDs, err := MarkFlushedIDsAtSnapshot(ctx, db, tableName, schemaID, sub, snapshot, flushedAt)
 			if err != nil {
 				logger.Sugar().Errorw("mark flushed failed", "err", err)
 				return
 			}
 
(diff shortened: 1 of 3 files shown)


internal/cdc/flusher.go
```diff
@@ -430,3 +449,21 @@ func updateManifest(
 	logger.Sugar().Infow("manifest updated", "schema_id", schemaID, "manifest_path", manifestPath, "file_path", filePath, "tier", tier)
 	return nil
 }
+
+func minMaxRowID(rowIDs []uuid.UUID) (string, string) {
```
iceboundrock, line 453: Could we return `(uuid.UUID, uuid.UUID)` instead of a `string` pair? Also, we likely don’t need to call `.String()` on each item in `rowIDs`; since they’re `UUID7`, we can compare them using `.Time()`.

iceboundrock:
Thanks for the PR, it does address a real data loss scenario, nice catch. I will look into it.

iceboundrock:
no worries, I will merge this PR and address my comments myself.

PR #9, opened 2026-02-09
fix(query): enhance anchor filter semantics in change log query

QuakeWang:
Keep the anchor condition consistent across union branches in the advanced query.
 
 This prevents change_log rows from bypassing original predicates (e.g., type=call) when flushed_at = 0.

### DIFF
--- a/internal/advanced_query_template.go
+++ b/internal/advanced_query_template.go
@@ -19,21 +19,27 @@ var optimizedQuerySQLTemplate = template.Must(template.New("optimizedQuery").Fun
 	            UNION
 	            -- Include real-time buffer rows from change_log (flushed_at = 0)
 	            -- while preserving the same anchor filter semantics.
-	            {{- if .UseMainTableAsAnchor }}
 	            SELECT cl.row_id
 	            FROM {{.ChangeLogTable}} cl
-	            INNER JOIN {{.MainTable}} m
-	                ON m.ltbase_schema_id = {{.SchemaID}}
-	                AND m.ltbase_row_id = cl.row_id
-	            WHERE cl.schema_id = {{.SchemaID}} AND cl.flushed_at = 0 AND {{.Anchor.Condition}}
-	            {{- else }}
-	            SELECT DISTINCT cl.row_id
-	            FROM {{.ChangeLogTable}} cl
-	            INNER JOIN {{.EAVTable}} t
-	                ON t.schema_id = {{.SchemaID}}
-	                AND t.row_id = cl.row_id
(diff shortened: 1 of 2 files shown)


QuakeWang:
@iceboundrock PTAL

internal/advanced_query_template.go
```diff
@@ -5,24 +5,37 @@ import "text/template"
...
+	            WHERE cl.schema_id = {{.SchemaID}} AND cl.flushed_at = 0 AND {{.Anchor.Condition}}
+	            {{- else }}
+	            SELECT DISTINCT cl.row_id
+	            FROM {{.ChangeLogTable}} cl
+	            INNER JOIN {{.EAVTable}} t
+	                ON t.schema_id = {{.SchemaID}}
+	                AND t.row_id = cl.row_id
+	            WHERE cl.schema_id = {{.SchemaID}} AND cl.flushed_at = 0 AND {{.Anchor.Condition}}
+	            {{- end }}
+	            {{- end }}
```
iceboundrock, line 43: Nice fix on keeping `anchor` semantics consistent across the `change_log` union branch.

One small optimization suggestion: in the `change_log` branch, replacing `INNER JOIN` (and `DISTINCT` on the EAV path) with a semi-join via `EXISTS` can reduce row amplification and dedup overhead while keeping the same filter semantics.

```sql
{{- if .ChangeLogTable }}
UNION
-- Include real-time buffer rows from change_log (flushed_at = 0)
-- while preserving the same anchor filter semantics.
SELECT cl.row_id
FROM {{.ChangeLogTable}} cl
WHERE cl.schema_id = {{.SchemaID}}
  AND cl.flushed_at = 0
  {{- if .UseMainTableAsAnchor }}
  AND EXISTS (
      SELECT 1
      FROM {{.MainTable}} m
      WHERE m.ltbase_schema_id = {{.SchemaID}}
        AND m.ltbase_row_id = cl.row_id
        AND {{.Anchor.Condition}}
  )
  {{- else }}
  AND EXISTS (
      SELECT 1
      FROM {{.EAVTable}} t
      WHERE t.schema_id = {{.SchemaID}}
        AND t.row_id = cl.row_id
        AND {{.Anchor.Condition}}
  )
  {{- end }}
{{- end }}
```

QuakeWang:
@iceboundrock  Great suggestion, thanks. I have updated the change_log branch to use `EXISTS` (semi-join) instead of `INNER JOIN`/`DISTINCT`, for both main-anchor and EAV-anchor paths, while preserving the same
  anchor filter semantics. And added regression tests and verified with `go test ./... -count=1`.

PR #10, opened 2026-02-24
refactor: enhance compactor logic for manifest updates

QuakeWang:
## Summary

Fix compaction no-op/metadata drift issues in `internal/compaction`:

1. Rewrite-only path (`needsRewrite=true`, `needsPromotion=false`) now **does not** update manifest.
2. Delta promotion now uses a case-insensitive tier match (`DELTA` and `delta` both work).
3. `FileProvider.SaveManifest` contract is documented: successful save must advance manifest metadata.

## Why These Changes

- Rewrite-only execution previously could advance manifest metadata without any real compaction effect, creating a false “compaction completed” signal.
- Tier handling was inconsistent: decision logic was case-insensitive, but promotion mutation was case-sensitive, which could produce `needsPromotion=true` but actually no-op.
- Compactor relies on provider-side metadata advancement during save, but this responsibility was not documented at the interface level.

### DIFF
--- a/internal/compaction/compactor.go
+++ b/internal/compaction/compactor.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/lychee-technology/forma/internal/cdc"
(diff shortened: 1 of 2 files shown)


iceboundrock:
Let me fix the CI first.

iceboundrock:
Merged, thanks @QuakeWang

PR #13, opened 2026-03-01
fix(compaction): harden SaveManifest metadata contract

### LINKED ISSUE #11: Harden FileProvider.SaveManifest metadata contract
## Background
In PR #10, `Compactor` no longer mutates manifest metadata (`Version`, `UpdatedAtMs`) directly and now relies on `FileProvider.SaveManifest` to advance those fields on successful save.

## Risk
This requirement currently exists as an interface comment only. A new or custom `FileProvider` implementation can silently violate the contract, which may lead to:
- stale manifest metadata after compaction,
- misleading compaction completion signals,
- subtle cross-provider behavior drift.

## Proposal
1. Add contract coverage for `FileProvider.SaveManifest` behavior.
2. Add a defensive runtime check in compaction flow:
- before save, capture `Version`/`UpdatedAtMs`;
- after save, verify they advanced;
- if not advanced, surface an explicit error (or at minimum a high-signal warning + metric).
3. Keep the contract documented in code, but back it with executable verification.

## Acceptance Criteria
- At least one contract test fails when a provider saves successfully but does not advance metadata.
- Compactor behavior is deterministic across provider implementations.
- Failure mode is explicit and observable (test assertion + log/metric).

QuakeWang:
close #11

### DIFF
--- a/internal/compaction/compactor.go
+++ b/internal/compaction/compactor.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	telemetry "github.com/lychee-technology/forma/internal"
 	"github.com/lychee-technology/forma/internal/cdc"
 	"github.com/lychee-technology/forma/internal/manifest"
 	"go.uber.org/zap"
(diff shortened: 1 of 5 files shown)


QuakeWang:
@iceboundrock PTAL

PR #160, opened 2026-07-07
fix: fail fast when a federated query has no schema metadata cache (#151)

iceboundrock:
## Summary

The no-metadata-cache branch of `injectSchemaProjections` (`internal/federated/duckdb_query.go`) emitted a fixed **toy-schema** projection — `fallbackOuterSelect` produced `name`/`age`/`tag` columns plus columns absent from the descriptors (`bigint_04`, `boolean_01`, …), with matching stale `S3SourceSelect`/`PGSourceSelect`/`PGGroupBy`.

That column set does not align with `model.EntityMainColumnDescriptors`, so it violates the **#147 positional-scan contract** (`streamDuckDBRows` / `duckDBScanBuffers` scan by descriptor order). If this path were ever hit with real DuckDB rows, `rows.Scan` would fail on a target-count mismatch, or — if counts happened to coincide — silently misplace values.

## Decision: fail fast (issue option 1)

Call-site evidence shows there is no *correct* fixed projection for a cacheless non-benchmark schema — the entity_main column order is only recoverable from the schema's attribute cache (and `BuildSchemaProjection` over an empty cache likewise degrades to a toy `name`/`version` shape). So rather than maintain a second, drift-prone column-order source, the cacheless path now errors explicitly.

- `injectSchemaProjections` returns `(bool, error)`; the `len(cache)==0` branch returns `federated query for schema_id N requires a schema metadata cache, but none is loaded`, propagated by `buildDuckDBQueryWithPlan`.
- Removed the dead `fallbackOuterSelect` helper.
- **Benchmark schemas (IDs 100–102) are unaffected** — they're handled before the cache check and never consult it.

## Tests

- New `TestInjectSchemaProjectionsNoCacheFailsFast` pins the fail-fast error, that no projection is injected on that path, and the benchmark-schema exemption.
- Three engine tests (`DuckDBFailureWithDegradedModeFallsBackToPostgres`, `DuckDBRouteUsesInjectedExecutorAndDirtyFetcher`, `ExecutionPlanPopulated`) were driving the render path with a `nil` metadata cache and inert fake executors — they silently relied on the toy fallback. They now register a real schema-7 cache via a `testMetadataCacheSchema7` helper, exercising the genuine projection path.

## Verification

- `go test ./internal/federated/` ✅
- `make test` ✅ (all packages)
- `make lint` ✅ (golangci-lint v1.64.8, exit 0)
- `go vet -tags e2e ./internal/e2e_harness/federated/... ./internal/federated/...` ✅

Refs #147, #149. Out of scope: benchmark projections (fixed in #149) and the cache-loading strategy itself.

### DIFF
--- a/internal/federated/duckdb_query.go
+++ b/internal/federated/duckdb_query.go
@@ -224,7 +224,10 @@ func (e *DBFederatedQueryEngine) buildDuckDBQueryWithPlan(
 	}
 
 	// Compute schema-driven projections for the template
-	projectionCacheHit := e.injectSchemaProjections(sqlParams, q.SchemaID, cache)
+	projectionCacheHit, err := e.injectSchemaProjections(sqlParams, q.SchemaID, cache)
+	if err != nil {
+		return "", nil, 0, err
+	}
 	planCtx.recordProjectionCache(projectionCacheHit)
 
 	// Determine if the EAV pivot can be skipped entirely (all filter/sort
@@ -391,7 +394,7 @@ func (e *DBFederatedQueryEngine) schemaProjection(schemaID int16, cache forma.Sc
 	})
 }
 
-func (e *DBFederatedQueryEngine) injectSchemaProjections(sqlParams map[string]any, schemaID int16, cache forma.SchemaAttributeCache) (projectionCacheHit bool) {
+func (e *DBFederatedQueryEngine) injectSchemaProjections(sqlParams map[string]any, schemaID int16, cache forma.SchemaAttributeCache) (projectionCacheHit bool, err error) {
 	if isBenchmarkSchemaID(schemaID) {
 		// Benchmark schemas: use hardcoded benchmarks projections that match
 		// the benchmark parquet shape exactly (flat columns for column-bound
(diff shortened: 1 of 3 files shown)


iceboundrock:
## Review Findings

### Spec / Behavior Risk

**Medium: no-cache “fail fast” can still be hidden by degraded-mode fallback**

Location: `internal/federated/duckdb_query.go:420`, `internal/federated/engine.go:127-130`, `internal/federated/engine_test.go:287`

The PR returns the new clear error from `injectSchemaProjections` when a non-benchmark schema has no metadata cache:

> `federated query for schema_id %d requires a schema metadata cache, but none is loaded`

However, the public `Query` path still treats that error like any other DuckDB failure. When `AllowPartialDegradedMode` is true, `engine.go:127-130` falls back to Postgres and does not surface the schema-cache-required error.

Issue #151 asks that when the path has no production reach, the fallback branch should be removed and “metadata cache missing” should explicitly error / fail fast, with an error message indicating that schema cache is required. The current implementation satisfies this at the helper level, but not necessarily at the public query behavior level under degraded mode.

Impact: a missing metadata cache can be masked as a degraded Postgres-only result instead of surfacing the configuration/data-contract problem that #151 intended to make loud.

Suggested fix: treat the missing-schema-metadata-cache error as non-degradable, or add an explicit test documenting that degraded mode is allowed to swallow this error if that is the intended behavior.

## Assessment

The PR correctly removes the stale `fallbackOuterSelect` path and chooses the fail-fast option allowed by #151. Benchmark schemas remain exempt, which is consistent with the PR description and issue scope.

The main remaining concern is whether the fail-fast behavior should apply at the externally visible `Query` boundary even when degraded mode is enabled.

## Observations

- No clear scope creep found.
- No descriptor-order fallback implementation issues found, because the PR chose the fail-fast option rather than keeping a fallback projection.
- The new unit test pins direct `injectSchemaProjections` behavior, but does not cover the public no-cache query path with `AllowPartialDegradedMode: true`.

## Verification

Review checked the PR diff between `origin/main...origin/pr/160` and issue #151 acceptance criteria.

Standards sources reviewed:

- `.golangci.yml`
- `go.mod`
- `Makefile`

No `CLAUDE.md`, `AGENTS.md`, `CONTRIBUTING.md`, `CONTEXT*.md`, `docs/adr`, or style/standards docs were found in the repo.

## Summary

- Standards findings: 0
- Spec findings: 1
- Worst issue: fail-fast schema-cache error can be swallowed by degraded-mode fallback.

iceboundrock:
Fixed in `37731d9`. The finding is correct — the fail-fast error was surfaced at the helper level but the public `Query` path swallowed it under `AllowPartialDegradedMode`, degrading to a Postgres-only partial result and masking the missing-cache config problem.

A missing schema metadata cache is a configuration / data-contract error, not a transient DuckDB failure, so it's now **non-degradable**:

- The error wraps a new `ErrSchemaMetadataCacheRequired` sentinel.
- `Query` degrades to Postgres-only only when the DuckDB error is **not** that sentinel (`errors.Is`); otherwise it returns the error.
- New `TestDBFederatedQueryEngine_MissingSchemaCacheNotDegradable` pins it (asserts `errors.Is` and `pg.queryCalls == 0`, i.e. no silent degradation). The existing `DuckDBFailureWithDegradedModeFallsBackToPostgres` test still passes, confirming genuine transient DuckDB failures still degrade as before.

`make test` / `make lint` / e2e vet green.

Issue #163, opened 2026-07-07
truth-pass oracle undercounts for pure-EAV filters absent from the hardcoded harness maps (orderChannel)


iceboundrock:
> Split out of #156 after the batching characterization proved this is not a batching artifact.

## Summary

The truth-pass oracle **undercounts** for a workload filtering on a pure-EAV attribute that is not in the harness's hardcoded attribute maps. `eav-low-selectivity-page` (`orderChannel = "web"`) produces oracle `expected = 65` while the production/service query returns `actual = 74`, which matches the generator truth (~75 `web` rows minus deletes). The ~9 missing are hot-tier `web` records.

## Root cause

The truth-pass oracle derives visibility from the harness's hand-written `ExecuteFederatedQuery` (`internal/e2e_harness/federated/query.go`). That query can only filter the fixed attribute set `{symbol, exchange, region, tradeType, tradeTime}` — via `benchmarkQueryColumn` (cold/parquet), `targetedHotFilterExpression` + the hot pivot meta (`attrIDList`/`pivotColumns`), and `benchmarkParquetProjection`. `orderChannel` is absent from all of them, so its filter is mishandled and hot-tier candidates are dropped from the truth set, undercounting the oracle.

## Not a batching artifact (evidence)

#156 replaced the per-candidate visibility probe with a single paginated sweep (`collectVisibleRowIDs`). The characterization test `TestTruthPassBatchEqualsPerCandidate` compares batch membership against the per-candidate probe for every candidate (167 across truth-pass workloads) and they are **identical**. So the batched and per-candidate oracles undercount `eav-low-selectivity-page` in exactly the same way — this is a harness truth-query correctness bug, orthogonal to the #156 speedup.

## Fix direction (to be validated)

Make the harness truth query filter **arbitrary** EAV attributes rather than the hardcoded set — e.g. a general EAV pivot/projection driven by the schema attribute cache, so any attribute (including `orderChannel`) filters correctly across hot and cold tiers. (Alternatively, derive truth from the production federated semantics.)

## Acceptance Criteria

- [ ] `eav-low-selectivity-page` truth-pass oracle matches the generator truth / production actual (no undercount).
- [ ] A regression test pins a pure-EAV, non-hardcoded-attribute filter at small scale.
- [ ] Remove `eav-low-selectivity-page` from `knownFailingAssertions` in `benchmark_workload_execution_test.go` once fixed.

## References

[...]

PR #148, opened 2026-07-05
perf: plan cache keyed by (schema_id, query_shape_hash) — full seven-phase implementation

iceboundrock:
## Summary

按 #142 修订版执行规划（含设计评审五项必须修订）完整实施，取代 #146 的简化版（已关闭）。九个 commit 分阶段落盘，每阶段独立编译测试绿。

## 七阶段对照

| Phase | 交付 | 关键测试 |
|---|---|---|
| 1 | `internal/queryplan`：shape hash（属性+canonical op+logic 结构+排序键+锚点+分页模式；值/游标/limit/offset 不参与）、`Key{Kind,SchemaVersion,SchemaID,ShapeHash,ScopeHash}`、并发 Cache | 10 项 hash 敏感性矩阵 + key 隔离 + 代际失效 + `-race` |
| 2 | schemameta copy-on-register + `SchemaFingerprint`（排序确定性 sha256，失效杠杆）+ 嵌套 RLock 修复 | 注册后改原 map 不影响快照；指纹随内容变化 |
| 3 | `BuildSchemaProjection` 去输入 mutation（局部副本） | 原 cache 不再被注入 name/version |
| 4 | **dual-clause plan + binder（评审必须项 2）**：值核心从三个叶子发射器单源抽取，bind-only walker 复用同一决策路径——arg 顺序 by construction | 等价矩阵：Plan+Bind ≡ ToDualClauses（11 场景含 OR-veto/嵌套/重复谓词） |
| 5 | **DuckDB 编译计划缓存（评审必须项 5）**：骨架一次渲染（dirty CSV 哨兵拼接）、`Bind` 复刻 advanced 模板参数交织（Duck+PgMain+Duck+keyset+tpl） | 字节平价矩阵（dirty×keyset 四组合）+ 跨请求复用 + 引擎级平价回归测试 |
| 6 | PG optimized 模板骨架缓存迁移共享 `queryplan.Cache`，`WithPlanCache` 注入 | pgxmock 端到端「同 shape 恰渲染一次」 |
| 7 | **缓存生命周期高于瞬态引擎（评审必须项 1）**：factory 与 benchmark runner 各构造一个共享 cache 注入全部 repo/engine；`RunResult` artifact 新增 `plan_cache_hits/_misses` | 引擎测试用三个瞬态引擎共享一个 cache 验证 benchmark 生命周期 |

[...]

Issue #104, opened 2026-05-04
Tune baseline-page-1 projection overhead and investigate concurrency tail latency


iceboundrock:
## Summary

Address two tunable bottlenecks identified in the post-optimization bottleneck catalog:
1. `baseline-page-1` latency regression introduced by projection negotiation overhead in PR #77 and PR #78
2. Concurrency tail latency plateau under fixed-concurrency benchmark execution (PR #94)

### baseline-page-1 regression

Medium gate benchmarks for optimization waves 1 and 2 both showed `baseline-page-1` regressing:
- PR #77 medium: `avg_latency_delta=+32.2ms`
- PR #78 medium: `avg_latency_delta=+92.1ms`

The fast gate was clean, suggesting the overhead amortizes poorly on small pages in the medium (zipf) distribution. Suspected cause: projection negotiation (`needsEAVJoin`, `buildFederatedQueryCountSQLDynamic`) adds per-request resolution cost that is disproportionate for page-1 queries.

### Concurrency tail latency plateau

PR #94 added fixed-concurrency execution and concurrency stability assertions but did not publish benchmark deltas for concurrent workloads. Suspected ceiling: DuckDB `threads=4` hardcoded in template + Postgres pool contention under concurrent mixed-tier queries.

## Scope

1. Profile `baseline-page-1` across fast (uniform) and medium (zipf) distributions
2. Profile projection negotiation (`needsEAVJoin`, routing decision, projection selection) to locate the per-request overhead
3. Cache or precompute projection decisions per schema where safe
4. Run small-live benchmarks with `Concurrency=2,4,8`, report p50/p95/p99 deltas
5. Test per-worker DuckDB isolation vs shared-instance contention
6. Make DuckDB thread count configurable through template params (replace hardcoded `PRAGMA threads=4`)
7. Validate that baseline-page-1 regressions are eliminated and concurrency tail latency is measurable

## Acceptance Criteria

- `baseline-page-1` no longer regresses in medium gate vs pre-wave baseline
- Concurrent benchmark runs produce publishable p50/p95/p99 deltas at Concurrency=2,4,8
- Route-stability assertion results are reported
- Projection negotiation overhead is either eliminated or documented as acceptable

## References

- `docs/federated-query/federated-query-post-optimization-bottleneck-catalog.md` (sections 1.2, 3.1, 4.3)
- PR #77, PR #78 (benchmark diffs showing baseline-page-1 regression)
- PR #94 (routing and concurrency wave)
[...]

Issue #158, opened 2026-07-07
Refactor oversized benchmark files: runner.go (2069) and report.go (1405) exceed the 500-line standard


iceboundrock:
## Summary

Two files in the federated-query benchmark package are far over the project's 500-line-per-file standard (`coding-standard.md` §1) and predate #100:

- `internal/e2e_harness/federated/benchmark/runner.go` — **2069 lines**, 69 functions
- `internal/e2e_harness/federated/benchmark/report.go` — **1405 lines**, 37 functions

PR #157 (#100) split the CLI's oversized `main.go`/`main_test.go` back under the limit, but deliberately left these two out of scope — a split of files this size is an independent refactor with its own review surface, and #100 only added a small fraction of their bulk.

## Why

- Both files bundle several unrelated responsibilities, which makes them hard to hold in context and slows review (the #100 review flagged both).
- The standard rejects changes that grow files past 500 lines; every future change to these files inherits a standing violation.

## Proposed split (behavior-preserving, no logic changes)

These are cohesive seams found by grouping the existing top-level functions — adjust during implementation:

### `runner.go` → keep core `Runner` (NewRunner/Run/RunWithHarness) + extract:
- **`oracle.go`** — expected-result construction: `buildExpectedResults`, `resolveRunOracleMode`, `buildExpectedWorkloadResults*`, `buildExpectedWorkloadResultFromFederatedTruth`, `buildTruthPassExpected`, `expectedVisibleRecords`, `filterExpectedRecordsForWorkload`, `semanticsForWorkload`, `sortExpectedRecordsForWorkload`, `expectedPageRowIDs`, and their helpers.
- **`assertions.go`** — outcome validation: the `validate*` family (`validateExpectedWorkloadOutcome`, `validateBasicWorkloadAssertions`, `validatePaginationTransition`, `validateConcurrent*`, `validateResultLevelAssertions`, `validatePushdownAssertions`, `validateSortOrder`, …) plus `recordMatchesFilter`, `benchmarkRecordTextValue`, `countFailedAssertions`.
- **`execute.go`** — service/query execution: `executeServiceWorkload`, `executeKeysetServiceQuery`, `executeServiceQuery*`, `queryRequestForWorkload`, `queryOptionsForWorkload*`, `conditionForWorkload`.
- **`loaded_state.go`** — `buildLoadedStateSnapshot`, `loadHotStateRecords`, `scanLoadedHotRecord`.

[...]

PR #164, opened 2026-07-07
perf(benchmark): batch truth-pass oracle visibility into one paginated sweep (#156)

### LINKED ISSUE #156: Speed up truth-pass oracle construction: per-candidate federated queries take 35+ minutes
## Summary

Truth-pass oracle construction issues one federated query per candidate record (`buildExpectedWorkloadResultFromFederatedTruth`, `internal/e2e_harness/federated/benchmark/runner.go`). At small-live scale this is already ~24k sequential queries for `eav-selective-page` — ~35 minutes on an idle machine, 3-4 hours under load (#147 measurements). This dominates live-benchmark wall clock and is pure setup cost, not measured workload time.

## Why

- `small-live` / `medium-live` runs spend most of their runtime building oracles, not benchmarking.
- #100 (heavy-live) bounds the cost with a sampling cap, but even the capped spot check (~10k candidates) costs ~15-30 minutes per selective workload at ~11 queries/s.
- Faster oracle construction would let the sampling cap be raised (stronger verification) at the same time budget.

## Candidate Directions (to be validated)

- **Batch the visibility check**: instead of one `ExecuteFederatedQuery` per candidate (`Limit: 1`, filter by single `RowID`), verify candidates in batches — e.g. one query per N row IDs (IN-list / row-ID set filter) and diff the returned set against the batch. Turns ~24k round trips into ~24k/N.
- **Parallel truth-pass workers**: the per-candidate checks are independent reads; a bounded worker pool could overlap round trips. Interacts with #104 resource single-source — DuckDB threads/memory are instance-global.
- **Plan reuse**: the per-candidate query shape is identical; confirm the plan cache (#148) is actually hit on this path, and if not, why.

Directions are not mutually exclusive; batching is likely the biggest single win since it removes round trips rather than overlapping them.

## Non-Goals

- Changing oracle semantics (what is asserted). This is purely about construction cost.
- Changing measured-workload execution paths.

## Acceptance Criteria

- Truth-pass oracle construction for small-live `eav-selective-page` (~24k candidates) completes in single-digit minutes on an idle machine.
- Oracle results are byte-identical to the current per-candidate implementation (characterization test on a small dataset comparing both paths).
- Live-preset docs updated if expected runtimes change materially.

## References

- #147 / PR #149 — truth-pass cost measurements and the per-candidate query loop
- #100 — heavy-live baseline; sampling cap bounds but does not fix this cost
- #104 — DuckDB resource single source (constrains the parallelism direction)
- #148 — plan cache (relevant to the plan-reuse direction)

iceboundrock:
## Summary

`buildExpectedWorkloadResultFromFederatedTruth` checked truth-pass visibility **per candidate** — one `ExecuteFederatedQuery` (`Limit:1` + `RowID` + workload conditions) each. At small-live scale that's ~24k sequential federated queries for `eav-selective-page` (~35 min idle, 3–4 h under load), dominating benchmark wall-clock as pure setup cost.

This replaces the per-candidate probe with a single **paginated sweep** (`collectVisibleRowIDs`) that pulls the full row-id set matching the workload conditions; visibility becomes an in-memory map lookup.

## Why it's a faithful equivalent

Membership is provably identical: a row is in the swept set iff a `RowID`-filtered query for it under the same conditions would return a row. Only the *computation* of visibility changes — `buildTruthPassExpected` (sampling, `TruthPassSampleCap`, oracle-mode reporting) is untouched, so heavy-live's sampled mode and `TestBenchmarkTruthPassSampledSpotCheck_RunWithHarness` behave identically. Query count drops from **O(candidates)** to **O(hits / pageSize)** (1–2 for selective workloads).

## Characterization

New e2e test `TestTruthPassBatchEqualsPerCandidate` (in-package, reuses the harness load path) compares batch membership against the retired per-candidate probe for **every** candidate — 167 across the truth-pass workloads — and they are **identical**. This is the byte-identical proof the AC asks for.

## Evidence

- The two e2e guard tests (`TestBenchmarkWorkloadExecution_RunWithHarness` + `TestBenchmarkTruthPassSampledSpotCheck_RunWithHarness`) drop from **~66s to ~13s** wall-clock — the truth-pass oracle is no longer the bottleneck.
- All green workloads stay green; the known-failing workloads stay excluded (no result changes).

## Bug 1 (folded-in orderChannel) — split out

Because batching is a faithful equivalent, it does **not** change `eav-low-selectivity-page`'s result. The characterization proved the undercount is an independent harness truth-query bug (pure-EAV `orderChannel` absent from the hardcoded `benchmarkQueryColumn` / `targetedHotFilterExpression` / pivot / parquet-projection maps), so it's split to **#163**. The exclusion in `benchmark_workload_execution_test.go` is re-pointed from #156 → #163.

## Verification

- `TestTruthPassBatchEqualsPerCandidate` (e2e) ✅ — 167 candidates, batch ≡ per-candidate
- e2e guard tests ✅ (green stay green, known-failing stay excluded)
- `make test` ✅, `make lint` ✅ (v1.64.8), `go vet -tags e2e ./internal/e2e_harness/federated/...` ✅

Refs #147, #148, #104. **Closes #156.** (Duplicate #152 will be closed alongside.)

### DIFF
--- a/docs/federated-query/federated-query-benchmark-ci-and-ops-guide.md
+++ b/docs/federated-query/federated-query-benchmark-ci-and-ops-guide.md
@@ -167,8 +167,12 @@ Runs the full small-live preset at `Concurrency=1,2,4,8` and aggregates the
 four runs into `.artifacts/benchmark/concurrency/concurrency-report.md`
 (overall and per-workload p50/p95/p99/QPS per level, plus the
 `concurrent-run-*` stability assertion pass rates from PR #94). Each level is
-a complete live run (~35 minutes on an idle machine, hours under load):
-operator-initiated only, never CI. Concurrent baselines write to `-c{N}`
+a complete live run, operator-initiated only, never CI. Since #156 batched the
+uncapped truth-pass oracle into one paginated visibility sweep per workload
+(replacing ~24k per-candidate federated queries), per-level runtime is
+materially lower than the earlier truth-pass-dominated ~35-minute figure — the
+dominant cost is now tier loading plus measured-workload execution; re-measure
+on your host. Concurrent baselines write to `-c{N}`
 suffixed directories and form their own trend comparability group, so they
 never pollute the sequential (`C=1`) baseline window.
 
--- a/internal/e2e_harness/federated/benchmark/runner.go
+++ b/internal/e2e_harness/federated/benchmark/runner.go
@@ -1377,6 +1377,25 @@ func collectVisibleRowIDs(ctx context.Context, h *federated.FederatedTestHarness
 	return visible, nil
 }
 
+// perCandidateVisible probes one candidate's visibility through the engine with
+// a RowID-filtered query. Capped (sampled) truth-pass mode uses this so its cost
+// stays bounded by the sample cap; the uncapped path uses collectVisibleRowIDs.
+func perCandidateVisible(ctx context.Context, h *federated.FederatedTestHarness, workload WorkloadDefinition, candidate GeneratedRecord) (bool, error) {
+	result, err := h.ExecuteFederatedQuery(ctx, &federated.QueryOptions{
+		Limit: 1,
+		Filter: &federated.Filter{
+			RowID:      candidate.RowID,
+			Conditions: workload.ResolvedFilterConditions(),
+		},
+		SortBy:   "tradeTime",
+		SortDesc: true,
+	})
+	if err != nil {
+		return false, fmt.Errorf("execute federated truth query for candidate %s: %w", candidate.RowID, err)
+	}
+	return result.TotalRecords > 0, nil
+}
+
 func buildExpectedWorkloadResultFromFederatedTruth(ctx context.Context, h *federated.FederatedTestHarness, workload WorkloadDefinition, defaultPageSize int, loadedRecords []GeneratedRecord, genCfg GeneratorConfig, sampleCap int, seed int64) (expectedWorkloadResult, truthPassSampleStats, error) {
 	if h == nil {
 		return expectedWorkloadResult{}, truthPassSampleStats{}, fmt.Errorf("harness cannot be nil")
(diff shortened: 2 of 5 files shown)


iceboundrock:
**Review: PR #164 vs issue #156**

**Findings**

1. `internal/e2e_harness/federated/benchmark/runner.go:1396` weakens the sampled truth-pass cost bound. Issue #156 explicitly notes `#100 (heavy-live) bounds the cost with a sampling cap`, but the PR now calls `collectVisibleRowIDs` before `buildTruthPassExpected` applies `sampleCap`. For capped/heavy-live workloads, this can sweep all visible hits page-by-page instead of probing only sampled candidates, so sampled mode may no longer be bounded by the cap.

2. No docs update for materially changed live-preset runtime expectations. Issue #156 AC says: `Live-preset docs updated if expected runtimes change materially.` The PR changes no docs, while `docs/federated-query/federated-query-benchmark-ci-and-ops-guide.md:166-171` still describes each small-live concurrency level as `~35 minutes on an idle machine, hours under load.`

**Assessment**

No scope creep found beyond the accepted #163 split/repoint. The batching approach appears aligned with the main #156 goal, but the sampled-cap behavior should be resolved or explicitly documented because it changes the performance envelope for sampled truth-pass mode.

**Standards Observations**

1. `internal/e2e_harness/federated/benchmark/runner.go:1344` adds new code to an already oversized source file. `coding-standard.md:7-11` says to reject changes resulting in source files over 500 lines, and `coding-standard.md:95-97` says to fix the immediate area when modifying non-conforming code. This helper likely belongs in a smaller/cohesive file.

2. `internal/e2e_harness/federated/benchmark/truthpass_batch_e2e_test.go:19` has an 80+ line test function. This is below the hard 100-line limit, but hits the refactoring trigger in `coding-standard.md:82-87`; setup/load/comparison helpers would make failures easier to localize.

**Verification**

Review evidence: fetched `origin/pr/164`, inspected `git diff origin/main...origin/pr/164`, fetched issue #156, and checked relevant standards/docs. No tests were run because this was a review-only request.

**Summary**

Spec findings: 2. Standards findings: 2. Worst issue: sampled truth-pass mode can now do an unbounded full visibility sweep before applying the sample cap.

iceboundrock:
Addressed in `a6db030`.

**1 (Spec, worst) — sampled mode no longer cap-bounded.** Fixed. The batch sweep now runs **only in the uncapped path** — the O(candidates) small/medium-live case #156 targets. Capped (heavy-live) mode restores the bounded per-candidate probe (extracted as `perCandidateVisible`), so it does ≤ `sampleCap` cheap `Limit:1` probes and never materializes the full hit set. Heavy-live's cost/memory envelope is unchanged. A shared `truthPassCapped(total, cap)` predicate is the single source of the capped/uncapped decision (used by both the oracle strategy and `selectTruthPassSampleIndices`), so the invariant can't drift. Verified: the capped guard test (`TestBenchmarkTruthPassSampledSpotCheck_RunWithHarness`, `cap=5`) and the uncapped one both pass.

**2 (Spec) — live-preset docs.** Fixed. Updated the small-live (uncapped) concurrency-sweep runtime note in `federated-query-benchmark-ci-and-ops-guide.md`: the `~35-minute` figure was truth-pass-dominated and is materially lower after batching. The heavy-live/capped runtime notes are intentionally left as-is — my finding-1 fix keeps capped mode on the per-candidate probe, so those numbers are unchanged.

**3 (Standards) — adding to oversized runner.go.** Not doing a partial extraction here. Splitting `runner.go` (2069 lines) is the dedicated purpose of **#158**, the very next PR, whose plan relocates exactly these functions (`buildExpectedWorkloadResultFromFederatedTruth`, `buildTruthPassExpected`, `collectVisibleRowIDs`, `perCandidateVisible`) into `oracle.go`. Doing a partial split in this perf PR would fragment that refactor and force a three-way rebase. This is the "suggest follow-up refactoring for the whole file" path in coding-standard.md's Incremental Improvement Mode — and the follow-up is already filed and scheduled next.

**4 (Standards) — 80-line test fn.** Addressed. The test now reuses the production `perCandidateVisible` (dropping the duplicate probe closure) and extracts the generate→split→load→snapshot sequence into `loadHarnessDataset`, so the test body is shorter and failures localize to setup vs comparison.

`make test` / `make lint` (v1.64.8) / `go vet -tags e2e` green; characterization + both guard tests pass.

PR #165, opened 2026-07-07
refactor(benchmark): split oversized runner.go and report.go under the 500-line limit (#158)

iceboundrock:
## Summary

`runner.go` (2131 lines) and `report.go` (1405) far exceeded `coding-standard.md` §1's 500-line-per-file limit. This is a **pure behavior-preserving** reorganization within `package benchmark` — functions relocated verbatim, no signature or logic changes. Same-package moves can't break references; only per-file imports changed (via `goimports`).

## runner.go → 442 lines

Keeps the core `Runner` type and lifecycle (New/Register/Run/RunWithHarness, retry, `executeWorkload` dispatch). Extracted:

| File | Lines | Contents |
|---|---|---|
| `oracle.go` | 441 | expected-result construction + truth-pass + candidate filtering |
| `execute.go` | 471 | service query execution (`executeService*`, route info) |
| `execute_query.go` | 143 | query request/condition/options builders + result helpers |
| `loaded_state.go` | 211 | loaded-state snapshot + hot-record scan |
| `assertions.go` | 453 | the `validate*` family + record matching / pushdown / per-tier |

## report.go → 309 lines

Keeps writers + formatters + metadata. Extracted:

| File | Lines | Contents |
|---|---|---|
| `report_types.go` | 237 | metadata / summary / diff / trend type declarations |
| `summary.go` | 381 | `SummarizeRunResult`, `summarize*`, `Compare*`, `compareWorkloadSummaries` |
| `trend.go` | 495 | trend history/analysis (`ReadTrendHistory` .. `AnalyzeTrend`) |

## Correctness

- **No non-generated source file in the package exceeds 500 lines.**
- **Top-level declaration counts are identical before vs after**: 108 func / 24 type / 4 const / 4 var = 140 — nothing dropped.
- `go build ./...`, `go vet -tags e2e ./internal/e2e_harness/federated/...`, `make test`, `make lint` (v1.64.8) all green.
- The e2e guard tests (`TestBenchmarkWorkloadExecution_RunWithHarness`, `TestBenchmarkTruthPassSampledSpotCheck_RunWithHarness`) and the batching characterization (`TestTruthPassBatchEqualsPerCandidate`) all pass — proving the reorg is behavior-preserving.

Refs #100, #156 (this relocates the truth-pass oracle cluster added there). **Closes #158.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #166, opened 2026-07-08
fix(sqlgen): emit ? not $n for main-column pushdown in DuckDB federated query (#161)

### LINKED ISSUE #161: Federated service query returns zero rows for a main-column AND EAV composite filter
> Surfaced while tightening the benchmark e2e assertion (#150).

## Summary

A federated **service-path** query with a composite `AND` filter that mixes a **main-column-bound** attribute and a **pure-EAV** attribute returns **zero rows**, even though each condition returns the correct set on its own.

Two benchmark workloads reproduce it, both with `FilterConditions = {symbol: "SYM00001", exchange: "NYSE"}`:

- `mixed-hot-eav-page`
- `tier-pushdown-mixed`

In the trade benchmark schema (id 102), `symbol` is main-column-bound (`text_01`; cf. `targetedHotFilterExpression("symbol") = COALESCE(hot_vals.symbol, em.text_01)`) and `exchange` is pure EAV (no `em.*` column). So the failing shape is `main(text_01 = symbol) AND eav(exchange = NYSE)`.

## Evidence

Via `TestBenchmarkWorkloadExecution_RunWithHarness` (plan mode, hotspot, TradeCount=120) with the #150 tightened assertion:

```
workload "mixed-hot-eav-page"  : total-records-match-expected actual=0 expected=20
workload "tier-pushdown-mixed" : total-records-match-expected actual=0 expected=20  (identical row-id set)
```

At live mode / uniform / TradeCount=300 the same workloads fail with `actual=0 expected=11`.

Single-condition controls **pass**:
- `eav-selective-page` (`exchange=NYSE` alone) → 80 rows ✅
- `tier-pushdown-eav` (`exchange=NYSE` alone) → ✅
- `security-symbol-page` (`symbol` alone, security schema) → ✅

So the defect is specifically in combining a main-column predicate with an EAV predicate under `AND` on the service/federated path — the composite returns the empty set instead of the intersection.

## How the composite is built

`conditionForWorkload` (`internal/e2e_harness/federated/benchmark/runner.go`) produces:

```go
&forma.CompositeCondition{Logic: forma.LogicAnd, Conditions: []forma.Condition{
    &forma.KvCondition{Attr: "symbol",   Value: "equals:SYM00001"},
    &forma.KvCondition{Attr: "exchange", Value: "equals:NYSE"},
}}
```

routed through `executeServiceWorkload` → `entity_manager_query_service` (the production `forma.QueryRequest` path).

## Suspected area (to be confirmed)

The dual-clause / hybrid composite translation where a main-table pushdown predicate (`m.text_01 = ...`) is combined with an EAV `EXISTS` predicate — likely the DuckDB `DuckClause` translation or the EAV pivot join producing an unsatisfiable combined predicate. Note this pre-dates and is independent of the #154 hybrid-leaf-emission refactor (which is byte-identical).

## Acceptance Criteria

- [ ] `mixed-hot-eav-page` and `tier-pushdown-mixed` return the correct intersection (matching the truth oracle) on the service path.
- [ ] A regression test pins a main-column `AND` EAV composite at small scale (the #147 `RunWithHarness` + small `TradeCount` + workload-subset method reproduces in ~40s).
- [ ] Remove both workloads from `knownFailingWorkloads` in `benchmark_workload_execution_test.go` (#150) once fixed.

## References

- #150 (assertion tightening that surfaced this; excludes these two workloads with a reference here)
- #147, #149 (positional-scan contract, benchmark green restoration)

iceboundrock:
## Summary

A federated **service-path** query with a composite `AND` mixing a **main-column-bound** attribute and a **pure-EAV** attribute returned **zero rows** instead of the intersection, though each condition returned correctly alone (`mixed-hot-eav-page`, `tier-pushdown-mixed`, both trade schema 102 with `{symbol, exchange}`).

## Root cause (confirmed empirically)

`PgMainClause` (the `entity_main` pushdown predicate from `ToDualClauses`) is embedded **only** in the DuckDB federated template (`pg_source` WHERE) — it is never sent to Postgres — yet it emitted Postgres-style `$n` placeholders while the co-located `DuckClause` (s3_source / visible WHERE) uses DuckDB's positional `?`.

**DuckDB cannot mix `?` and `$n` in one statement.** The `$1` aliases positional param 1, colliding with the first `?` and shifting every `?` after it. For `symbol(main) AND exchange(EAV)`, the generated SQL was:

```
s3_source  WHERE (exchange = ?) AND (symbol = ?)      -- ?, ?
pg_source  WHERE (m.text_01 = $1)                     -- $1  <-- collides
visible    WHERE (exchange = ?) AND (symbol = ?)      -- ?, ?
ARGS = [NYSE, SYM00001, SYM00001, NYSE, SYM00001]
```

`$1` bound `m.text_01` (symbol column) to `NYSE`, and the visible re-filter bound the wrong operands → empty set. Each condition alone masked it: EAV-only has no `$n`; a single main-column filter binds identical values so the collision is invisible. Confirmed directly against the DuckDB CLI: the `?,?,$1,?,?` pattern errors/mis-binds; uniform `?` binds every value to the correct position.

> Note: this differs from the originating issue's suspected area. The benchmark shim (`translateDuckClauseToBenchmark`) actually produces a **correct** attribute-named clause; the operative defect is the placeholder-style mixing.

## Fix

Emit `?` from `pgMainTypedEmitter` so the whole DuckDB statement uses a single placeholder style. The advanced-template arg interleave (`DuckArgs, PgMainArgs, DuckArgs`) already matches left-to-right `?` order. `paramIndex` is still advanced so the sibling EAV EXISTS clause (`PgClause`, genuinely Postgres-bound) keeps its `$n` numbering.

## Tests

- **New regression** `TestDuckDBMixedComposite_NoPlaceholderStyleMixing`: renders the full DuckDB federated SQL for a main+EAV composite (non-benchmark schema) and asserts no `$` appears and placeholder count equals arg count. Verified it **fails** against the old `$n` emission.
- Updated characterization goldens to the corrected `?` (PgMainClause only; EAV `PgClause` `$n` numbering unchanged).
- Removed `mixed-hot-eav-page` / `tier-pushdown-mixed` from `knownFailingAssertions`; the full e2e harness (`TestBenchmarkWorkloadExecution_RunWithHarness`) passes with them re-enabled (both now return 20 rows).

## Follow-up (separate issue)

While rendering the SQL I confirmed a **separate** production bug: for non-benchmark schemas the DuckClause references the physical column (`integer_01`) while the CTE column is attribute-aliased (`age`) — masked in the benchmark only by `translateDuckClauseToBenchmark`. Filed as #167 (production namespace mismatch + shim retirement); orthogonal to this placeholder fix.

Closes #161.

### DIFF
--- a/internal/e2e_harness/federated/benchmark_workload_execution_test.go
+++ b/internal/e2e_harness/federated/benchmark_workload_execution_test.go
@@ -29,21 +29,10 @@ import (
 //     batching proved (via TestTruthPassBatchEqualsPerCandidate) that this is a
 //     harness truth-query bug independent of the per-candidate/batch split, so
 //     it is tracked on its own (#163).
-//   - mixed-hot-eav-page / tier-pushdown-mixed: the service path returns zero
-//     rows for the composite main+EAV filter (symbol AND exchange) though each
-//     condition works alone (#161).
 var knownFailingAssertions = map[string]map[string]string{
 	"eav-low-selectivity-page": {
 		"total-records-match-expected": "#163",
 		"page-row-ids-match-expected":  "#163",
-	},
-	"mixed-hot-eav-page": {
-		"total-records-match-expected": "#161",
-		"page-row-ids-match-expected":  "#161",
-	},
-	"tier-pushdown-mixed": {
-		"total-records-match-expected": "#161",
-		"page-row-ids-match-expected":  "#161",
 	},
 }
 
@@ -137,7 +126,7 @@ type workloadCoverage struct {
 	customer, security, filtered, lowSelective, eav, mixedFilter bool
 	mixedTier, hotOnly, coldOnly                                 bool
 	expectedOracleAssertions, tradeTimeWindow, repeatedStability bool
-	pushdownHot, pushdownEAV, pushdownMixed, pushdownColdOnly     bool
+	pushdownHot, pushdownEAV, pushdownMixed, pushdownColdOnly    bool
 	pushdownAssertions                                           bool
 }
 
--- a/internal/sqlgen/dualpath_sql_generator.go
+++ b/internal/sqlgen/dualpath_sql_generator.go
@@ -94,9 +94,20 @@ func (e *pgMainTypedEmitter) EmitTypedLeaf(leaf *PredicateLeaf) (string, []any, 
 		return "", nil, nil
 	}
 
+	// PgMainClause is only ever embedded in the DuckDB federated query template
+	// (as PG_WHERE_CLAUSE for the pg_source CTE) — it is never sent to Postgres.
+	// It therefore must use DuckDB's positional "?" placeholder, not "$n":
+	// DuckDB cannot mix "?" (auto-numbered) with "$1" (numbered) in one statement,
+	// and the DuckClause (also embedded here) uses "?". Mixing the two mis-binds
+	// arguments — the "$1" aliases DuckDB positional param 1 and shifts every "?"
+	// after it — so a main-column AND EAV composite silently returned zero rows
+	// (#161). The advanced-template arg interleave (DuckArgs, PgMainArgs, DuckArgs)
+	// already matches the left-to-right "?" order once every placeholder is "?".
+	//
+	// paramIndex is still advanced so the sibling EAV EXISTS clause (PgClause),
+	// which shares this counter and does use "$n", keeps its numbering stable.
 	*e.paramIndex++
-	ph := fmt.Sprintf("$%d", *e.paramIndex)
-	sql := fmt.Sprintf("%s %s %s", p.Column, p.SQLOp, ph)
+	sql := fmt.Sprintf("%s %s ?", p.Column, p.SQLOp)
 	return sql, []any{p.Value}, nil
 }
 
--- a/internal/sqlgen/dualpath_sql_generator.go
+++ b/internal/sqlgen/dualpath_sql_generator.go
@@ -21,7 +21,10 @@ type DualClauses struct {
 // - PgClause reuses existing SQLGenerator (EAV-based EXISTS expressions).
 // - PgMainClause contains predicates suitable for entity_main pushdown.
 // - DuckClause maps attributes to column names when available and emits a simple DuckDB-style clause.
-// Note: DuckDB placeholders are "?" and args are returned in order. Postgres uses $n placeholders.
+// Placeholder styles: PgClause (the EAV EXISTS clause, genuinely Postgres-bound)
+// uses "$n"; DuckClause and PgMainClause both use DuckDB positional "?", because
+// PgMainClause is only ever embedded in the DuckDB federated template (see
+// pgMainTypedEmitter and #161). Args are returned in order for all three.
 //
 // The condition tree is normalized into the typed predicate IR exactly once
 // (#143); the three emitters below consume that shared tree.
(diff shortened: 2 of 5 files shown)


iceboundrock:
## Review Comment

### Standards

Findings:

1. `internal/sqlgen/dualpath_sql_generator.go:24`, `:37`, `:122` still document `PgMainClause` / main pushdown placeholders as `$n`, but the PR changes them to `?`. This conflicts with `docs/ai_review_guideline.md:245`: “Public API behavior differs from comment/contract.” Update the comments to match the new DuckDB-only placeholder behavior.

2. `internal/sqlgen/dualpath_sql_generator_test.go` is modified while already over 500 lines. `coding-standard.md:8-11` says to reject source files over 500 lines, and `coding-standard.md:95-98` says to fix the immediate area or suggest follow-up refactoring when touching non-compliant code.

3. `internal/sqlgen/predicate_characterization_test.go` modifies `TestToDualClauses_Characterization`, which is already over the 100-line function limit from `coding-standard.md:8-11`. This is pre-existing, but the touched area remains non-compliant.

### Spec

No Spec review findings.

PR #166 satisfies issue #161 acceptance criteria: it fixes the main-column + EAV `AND` placeholder binding path, adds a regression test, and removes the two benchmark workloads from the known-failing list.

### Assessment

The implementation appears directionally correct for issue #161. The main behavioral fix is narrowly scoped: `PgMainClause` now emits DuckDB positional `?` placeholders for the federated DuckDB template while still advancing `paramIndex` so the sibling Postgres EAV clause keeps stable `$n` numbering.

### Observations

The stale comments are the only behavior-adjacent concern: future maintainers could read the exported/public helper contract as still returning `$n` placeholders even though the PR intentionally changes that contract for the DuckDB-embedded main clause.

The file/function size findings are standards compliance issues against existing over-limit tests touched by this PR; they do not appear to invalidate the bug fix itself.

### Verification

GitHub PR checks are passing:

- `Benchmark Smoke`
- `E2E Smoke Tests`
- `Lint`
- `Test`

### Summary

Standards has 3 findings; Spec has 0 findings. Worst issue: stale `$n` comments in `dualpath_sql_generator.go` now contradict the actual placeholder contract.

iceboundrock:
Thanks for the review. Addressing the 3 Standards findings:

**1. Stale `$n` comments (`dualpath_sql_generator.go:24,37,122`) — fixed** in e72e999. All three now state that only the EAV `PgClause` (genuinely Postgres-bound) uses `$n`, while `DuckClause` and `PgMainClause` use DuckDB positional `?` (with the reason `paramIndex` is still advanced). No stale `$n` references remain — the only `$n` mentions left describe the EAV clause correctly.

**2 & 3. Over-limit test file / function — pre-existing, filed follow-up #168.** Both predate this PR:
- `dualpath_sql_generator_test.go` was already 541 lines; this PR only edited the `NestedAndOr` assertion (~+3 lines).
- `TestToDualClauses_Characterization` was already a 223-line table; this PR only updated golden strings within it.

Per `coding-standard.md` Incremental Improvement Mode (§95–98) — *"Suggest follow-up refactoring… Don't rewrite unrelated code without asking"* — splitting these test scaffolds is out of scope for a narrow bugfix (same rationale as the #158→#165 runner/report split). Tracked in **#168** (split the test file by emitter concern; extract the characterization table into a helper), with the constraint that it stays a pure restructure preserving byte-for-byte golden coverage. Happy to bundle it here instead if you'd prefer.

Issue #189, opened 2026-07-10
[E2E] Schema evolution across Parquet generations


iceboundrock:
Part of epic #172.

## Task

Verify that federated queries correctly handle Parquet files written with different schema versions (added/removed/renamed columns, changed types).

### 1. Added column
- Schema v1: 3 attributes → generate base Parquet with v1 schema
- Schema v2: 4 attributes (new column added) → generate delta Parquet with v2 schema
- Federated query → assert all records return, new column NULL/default for old rows

### 2. Removed column
- Schema v1: 4 attributes → base Parquet
- Schema v2: 3 attributes (column removed) → delta Parquet
- Federated query → assert query succeeds for v2-projected columns

### 3. Changed column type
- Schema v1: `score` as int → base Parquet
- Schema v2: `score` as float → delta Parquet
- Federated query → assert type coercion or explicit error

### 4. Attribute promoted from EAV to bound
- Schema v1: `status` as EAV → base Parquet (pivoted column)
- Schema v2: `status` as bound main column → delta Parquet (direct column)
- Assert projection resolves correctly

### 5. Mixed old/new Parquet in same query
- Query with both v1 and v2 Parquet files in the same glob
- Assert column resolution handles union of schemas

## Success criteria

- Schema evolution does not cause query failures or crashes
- NULL/default values correctly filled for missing columns
- Type coercion is predictable or explicitly rejected

iceboundrock:
Adjudications for the acceptance criteria, recorded before review (PR #254):

**Scenario 2 (removed column) — already green, pinned as characterization.** Both halves hold on pre-fix main: a lone old-generation file under the v2 projection succeeds because the explicit projection never references the dropped column, and the *mixed* base+delta read also succeeds because DuckDB's projection pushdown never requests that column from any file. The scenario's value is the pin, not a fix.

**Scenario 3 (changed type) — pre-fix behavior was silent data corruption, not a query failure.** Without `union_by_name`, DuckDB coerces all files to the first file's schema: v2 DOUBLE values were cast to the v1 base's INTEGER, destroying fractional values with totals intact; only the oracle caught it. The issue's "predictable coercion or explicit error" is therefore adjudicated to **predictable widening** (INTEGER∪DOUBLE → DOUBLE, lossless, matching the oracle's numeric-family normalization); genuinely incompatible type pairs still fail as classified read errors.

[...]

Issue #174, opened 2026-07-10
[E2E] CDC full-type producer-consumer round-trip


iceboundrock:
Part of epic #172.

## Task

Verify that every supported Forma type survives a full CDC export → Parquet → federated read round-trip without loss or corruption.

## Scenario

1. Create entities with **all attribute types**: text, numeric (int/bigint/float), bool, date, datetime, UUID — both bound (main-column) and EAV
2. Include NULL values and boundary values (max int64, empty string, zero UUID, epoch dates)
3. Run production `cdc-flush`
4. Inspect Parquet: schema column names, physical types, row values
5. Run production federated query against the flushed Parquet
6. Assert every field value matches the original write

## High-risk hypotheses to confirm or reject

- CDC export writes `time_slot` but federated read expects `changed_at` → column mismatch may break all queries
- EAV numeric values stored in `value_numeric`, but CDC export reads from `value_text` → numbers/bools/dates may be null
- Array/object attributes may not round-trip

## Success criteria

- All supported types round-trip correctly through real CDC → Parquet → real federated engine
- NULL and boundary values preserved
- Bound and EAV attributes both verified

Issue #172, opened 2026-07-10
Epic: Production-Grade E2E Test Suite for CDC & Federated Query


iceboundrock:
## Goal

Build a production-tracing E2E test suite that closes the loop through the real CDC flusher and federated query engine.

## Status (2026-07-11)

Phase 1 is implemented: PR #191 (open, review round 1 complete, CI green) wires the production harness and fixed six cross-component production bugs it surfaced. Those findings added seven follow-up issues (#192–#198), now woven into the execution order below — most notably #192/#194, which block full-type coverage in #174.

### Phase 1: Harness foundation — ✅ implemented (PR #191, in review)
- [x] [#173](https://github.com/Lychee-Technology/forma/issues/173) Wire production harness & diagnostic infrastructure → PR #191

### Phase 1.5: Harness-surfaced production fixes (blocking P0 coverage)
- [x] [#192](https://github.com/Lychee-Technology/forma/issues/192) Remove synthetic `name`/`version` injection from `BuildSchemaProjection` — **do first**: every fixture schema currently carries workaround attributes; any production schema without them fails warm/cold queries
- [x] [#194](https://github.com/Lychee-Technology/forma/issues/194) Main-column-bound date/datetime (`unix_ms`) exporter/reader mismatch — **blocks #174**: fixtures deliberately avoid bound dates until this lands
- [x] [#197](https://github.com/Lychee-Technology/forma/issues/197) Delta flusher swallows manifest update failures — same pattern as the init fix in PR #191; land before #179 so the failure matrix tests the intended contract, not the swallow

### Phase 2: Core correctness (P0)
- [x] [#174](https://github.com/Lychee-Technology/forma/issues/174) CDC full-type producer-consumer round-trip → PR #206 (merged 2026-07-10). Surfaced two follow-ups: [#204](https://github.com/Lychee-Technology/forma/issues/204) (`list` unusable end to end — write path rejects, CDC would MAX-collapse) and [#205](https://github.com/Lychee-Technology/forma/issues/205) (bound bigint float64 hops, faithful ceiling ±2^62; relevant to #195/#178)
[...]

PR #191, opened 2026-07-10
feat(e2e): wire production test harness & diagnostic infrastructure (#173)

### LINKED ISSUE #173: [E2E] Wire production harness & diagnostic infrastructure
Part of epic #172.

## Task

Build the reusable production test harness that wires:
- **Real containers**: Postgres, DuckDB, MinIO/S3 via testcontainers
- **Real CDC flusher**: `internal/cdc` package, not harness-simulated export
- **Real federated engine**: `internal/federated` package, not custom SQL
- **Real schema metadata**: load from schema JSON, register in schema registry
- **Real manifest store**: S3-backed manifest, not in-memory

## Diagnostic output

Every test failure must produce a run-specific artifact directory with:
- Fixed random seed + run ID
- Test fixture / event log
- `change_log` snapshot
- S3 object listing
- Manifest raw JSON
- Parquet schema + sampled rows
- Generated production SQL, parameters, and execution plan
- Expected vs actual diff
- `KEEP_E2E_ENV=1` keeps containers alive for manual inspection

## Oracle model

Build an **independent expected-state model**:
1. Collect events by `row_id`
2. Pick latest version
3. Apply deletions
4. Apply filters
5. Sort
6. Paginate

Do NOT reuse the federated engine to compute expected results.

## Constraints

- Each test owns isolated schema ID, tenant, bucket prefix
- Must not depend on prior run state (no history pollution)
- Deterministic fixtures with seeded randomness

iceboundrock:
## Summary

Builds the reusable production E2E test harness (`internal/e2e_harness/production`) that wires **real containers** (Postgres 16 + MinIO via testcontainers, in-process DuckDB), **real writes** (`internal.NewEntityManager`), the **real CDC flusher** (`internal/cdc`), the **real federated engine** (`internal/federated`), real schema metadata from fixture JSON, and the S3-backed manifest — with every query checked against an **independent oracle** and full diagnostic artifacts on failure.

- Isolation: containers shared per test binary (`SharedCluster`); each test gets its own PostgreSQL database (standard production table names → advisory-lock + change_log-scan isolation, `t.Parallel()`-safe), S3 prefix, and in-memory DuckDB client; `KEEP_E2E_ENV=1` keeps everything for inspection.
- Determinism: seeded event-script generator (`E2E_SEED`), wall-clock `changed_at` read back from `change_log` into the event log.
- Oracle: pure expected-state model (group by row_id → latest by ChangedAt+Seq → drop deletes → typed filters → sort with row_id tiebreak → paginate); numeric normalization verified against both mainline EAV write and read paths.
- Diagnostics: `.artifacts/e2e/<runID>/<test>/` with run.json, events.json, change_log snapshot, S3 listing, raw manifests, parquet schema+samples, per-query SQL+params+plan dumps, and diff.json; path printed in the failure message.
- Smoke (`TestProductionSmoke`): seeded creates → `cdc.RunInit` (cold) → updates/deletes → real flush (warm) → post-flush writes (hot) → oracle-checked battery (unfiltered, equality, range, sorted, page 2, PreferHot) with **explicit nonzero cold AND warm tier-hit assertions** (guards the glob-mismatch "both sides empty" false green). Wired as `make test-e2e-production` plus a CI E2E-gate step.

## Production-code changes

The two planned changes, each its own commit:

1. **`refactor(cdc)`: extract `cdc.RunInit`** (8e90de0) — mechanical move of the init driver from `cmd/tools/cdc_init.go` into `internal/cdc/init.go` + `init_sizing.go`; the CLI keeps flags and delegates. No behavior change.
2. **`feat(model)`: `DataSourcePlan.Params`** (c8b564f) — bind parameters captured in execution-plan sources, only when `IncludeExecutionPlan` is set (DuckDB rendered SQL args, dirty-id schema_id, hot-path hybrid clause args).

**Unplanned but necessary fixes.** The production DuckDB federated path had never executed against parquet produced by the real CDC exporter (all prior integration tests for it are skipped; the benchmark uses its own simulated layout). Wiring the real components together surfaced six genuine cross-component bugs, fixed minimally in two commits (cb0b1bf, bd1c829):

- exporter wrote `time_slot`, reader expects `changed_at` (nothing in-repo reads `time_slot`) → renamed in delta+base export;
- exporter read EAV values only from `value_text`, but mainline stores the numeric family (bool as 1/0, dates as epoch ms) in `value_numeric` → export from the correct storage column with reader-matching types;
- the no-EAV-pivot shortcut rendered a `UNION ALL` with mismatched column counts for any schema with EAV-only attributes → gated on the schema having none;
- `pg_source` selected `hot_vals.<attr>` ungrouped (DuckDB rejects; the benchmark projection already wraps in `ANY_VALUE`) → same aggregation, plus the production `attributes_json` now emits the array-of-EAV-objects contract `model.ParseAttributesJSON` expects, and UUID main columns are cast in `COALESCE`;
- **CDC flush dropped delete tombstones** (delete removes the entity row; the export INNER JOINed change_log→entity_main, so tombstones were marked flushed but never exported — the cold tier resurrected deleted rows) → LEFT JOIN;
- **the template filtered `s3_source` before LWW dedup**, resurrecting stale base versions whose old values still matched when the newer delta version didn't → the s3 filter is now a row_id semijoin (any version matches → all versions enter dedup; the `visible` filter applies the predicate to the winner). Placeholder count/order unchanged, so arg binding and the plan cache are unaffected.

Pinned SQL unit tests were updated to the corrected output. `internal/e2e_harness/federated/` is untouched.

## Verification

- `make test`: all 26 packages ok (oracle/generator tests run docker-free).
- `make lint` (golangci-lint v1.64.8): clean.
- `make test-e2e-production` (docker): 13/13 pass, incl. `TestProductionSmoke` — tier hits `cold=10 warm=16 hot=3`, base parquet 30 rows, delta 25 rows; dry-run immutability; init/base export + manifest entries; two parallel Envs isolation.
- Deliberate-mismatch check: rogue row injected behind the event log → `AssertQueryMatches` failed with `expected_total=5 actual_total=6 extra=1`, artifact path printed, all 8 artifact kinds present and parseable.
- `KEEP_E2E_ENV=1 TESTCONTAINERS_RYUK_DISABLED=true`: containers survived the failing run; PG DSN, S3 endpoint/credentials, and per-env prefix printed.

## Deviations from the plan

- `engine.go` landed with the Task-5 commit (the EntityManager needs the engine); `assert.go` landed with Task 9 (it calls `DumpArtifacts`). Same files/APIs as planned.
- `FlushReport.Manifests` is `map[int16]*manifest.Manifest` (per schema) instead of a single manifest — multi-schema (#189) needs it.
- Fixture schemas additionally define `name`/`version` text attributes (see follow-up 1) and cover 9 of 10 value types — `forma.ValueTypeList` is deferred to #174.
- The smoke uses `ExecSQL` to truncate `change_log` for the base-only rows after `RunInit`, modeling the production onboarding contract (base bootstrap + log cleanup); without it every row would also exist in a delta and no row would be cold-exclusive.
- Federated (DuckDB) queries require ≥1 parquet file per schema (`read_parquet` errors on an empty glob) — matching production, documented in the README; use `PreferHot` before the first flush/init.

- Size limits: every new file/function is within the 500/100 limits. Two pre-existing files already over 500 lines on main received small deltas (`internal/federated/duckdb_query.go` 618→624, `internal/sqlgen/duckdb_schema_projection.go` 701→730); splitting them is left to a dedicated refactor (cf. #158/#168 precedent). `cmd/tools/cdc_init.go` shrank 691→181.

## Follow-ups (now filed as issues)

1. #192 — **`sqlgen.BuildSchemaProjection` injects synthetic `name`/`version` attributes** into every schema; real CDC exports cannot satisfy the injected parquet columns, so any production schema *not* defining them fails warm/cold queries. The injection looks like toy-schema legacy (#86) and should be removed (with test fixture updates).
2. #193 — The generic (no attribute cache) CDC export path still writes `value_text`-only `attributes` structs; the production reader can't consume no-cache schemas anyway, but the path should be aligned or retired.
3. #194 — Main-column-bound date/datetime with `unix_ms` encoding: the exporter converts to DATE/TIMESTAMP while the reader expects raw epoch-ms BIGINT; not exercised by the fixtures (they avoid bound dates), needs the same alignment treatment.
4. #195 — The s3_source row_id semijoin scans the parquet glob twice; if profiling shows it matters, a smarter pushdown (version-aware predicate) can restore single-scan performance.
5. #196 — `internal/federated/duckdb_federated_production_e2e_test.go` and the skipped integration tests reference "root integration tests" that no longer exist; the production package now provides that coverage and the skipped stubs could be retired.

Review-round follow-ups also filed: #197 (delta flusher swallows manifest failures), #198 (migrate federated harness off the archived MinIO image to RustFS).

Closes #173
Part of epic #172


internal/cdc/init.go
```diff
@@ -0,0 +1,439 @@
...
+// RunInit performs the initialization export for all or a specific schema:
+// existing non-deleted entity_main rows are batched into base parquet files on
+// S3 and (when a manifest template is configured) recorded in the schema
+// manifest. Extracted from cmd/tools/cdc_init.go (mechanical move, #173).
+func RunInit(ctx context.Context, opts InitOptions) (InitSummary, error) {
+	if opts.Logger == nil {
+		opts.Logger = zap.NewNop()
+	}
+
+	runCtx, err := newInitRunContext(ctx, opts)
```
iceboundrock, line 154: P1: The new exported `RunInit` does not apply `CDCConfig.WithDefaults()`, unlike `Runner.RunOnce`. With `TargetFileSizeMB` and `BatchSize` unset, `resolveInitBatchSize` returns zero and the batch query uses `LIMIT 0`, so init reports success after exporting nothing. Apply defaults before constructing the run context and test the zero-value configuration.

internal/e2e_harness/production/schema.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/lychee-technology/forma/internal/sqlgen"
+)
+
+// Fixture schema IDs live in 20-40: sqlgen.IsBenchmarkSchemaID reserves
+// 100-102 for hardcoded benchmark parquet projections, and the production
+// harness must never collide with that range.
+const (
+	SchemaIDSimple int16 = 20
```
iceboundrock, line 17: P2 (spec): #173 explicitly requires each test to own an isolated schema ID, tenant, and bucket prefix. Fixture IDs are fixed at 20-22, and this harness has no tenant identity or tenant-aware DDL. Per-test databases prevent direct collisions, but do not implement the requested schema-ID/tenant isolation model.

internal/e2e_harness/production/cluster.go
```diff
@@ -0,0 +1,260 @@
...
+	return nil
+}
+
+// Shutdown stops the shared containers. Under KEEP_E2E_ENV=1 it prints
+// connection info instead and leaves everything running.
+func (c *Cluster) Shutdown(ctx context.Context) error {
+	if c == nil {
+		return nil
+	}
+	if KeepEnv() {
```
iceboundrock, line 200: P2 (spec): `KEEP_E2E_ENV=1` is meant to keep the environment alive for inspection, but it only skips this explicit shutdown. Ryuk still removes the containers unless the caller also knows to set `TESTCONTAINERS_RYUK_DISABLED=true`. Make the harness enforce the required behavior or revise the stated contract.

internal/cdc/init.go
```diff
@@ -0,0 +1,439 @@
...
+
+	var schemaIDs []int64
+	for rows.Next() {
+		var sid int64
+		if err := rows.Scan(&sid); err != nil {
+			return nil, fmt.Errorf("scan schema id: %w", err)
+		}
+		schemaIDs = append(schemaIDs, sid)
+	}
+	return schemaIDs, nil
```
iceboundrock, line 230: P2: This cursor loop returns without checking `rows.Err()`. A mid-stream database error can silently truncate the schema list while `RunInit` reports success. Check and wrap `rows.Err()` after this loop and after the equivalent batch-ID loop.

internal/e2e_harness/production/artifacts.go
```diff
@@ -0,0 +1,365 @@
...
+			"row_ids":         rowIDs,
+		}
+		if err := writeJSONArtifact(dir, fmt.Sprintf("query_%d.json", i+1), payload); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (e *Env) dumpDiff(_ context.Context, dir string) error {
```
iceboundrock, line 380: P2 (spec): #173 requires an expected-vs-actual diff for every failure. `diff.json` is skipped unless a query assertion already set `lastDiff`; failures from init/flush/provisioning therefore omit the required artifact. Write a structured empty/error diff so the artifact contract is uniform.

internal/e2e_harness/production/artifacts.go
```diff
@@ -0,0 +1,365 @@
...
+		e.T.Logf("diagnostic artifacts: %s", dir)
+	})
+}
+
+func writeJSONArtifact(dir, name string, payload any) error {
+	data, err := json.MarshalIndent(payload, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshal %s: %w", name, err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, name), data, 0o644); err != nil {
```
iceboundrock: P1: Artifacts include `e.PGDSN()` (and therefore an external PostgreSQL password) but are written with mode `0644`. Diagnostic data should not expose supplied production-like credentials to other local users. Redact the password or use restricted artifact permissions (`0600`), with a regression test.

internal/e2e_harness/production/env.go
```diff
@@ -0,0 +1,369 @@
...
+			ChangeLog:      "change_log",
+			SchemaRegistry: "schema_registry",
+		},
+	}
+
+	seq := c.envSeq.Add(1)
+	e.DBName = fmt.Sprintf("e2e_%d", seq)
+	e.S3Prefix = fmt.Sprintf("e2e/%s/env%d", c.RunID, seq)
+
+	if err := e.provision(ctx); err != nil {
```
iceboundrock, line 159: P1 (spec): #173 requires every test failure to produce run-specific artifacts. This `t.Fatalf` path happens before cleanup/artifact registration, so provisioning failures such as missing `CREATE DATABASE` privileges, schema-load errors, or DuckDB startup failures leave no diagnostics. Register a minimal dump before provisioning or dump explicitly here.

internal/cdc/init.go
```diff
@@ -0,0 +1,439 @@
...
+	if runCtx.manifestStore == nil || len(state.fileEntries) == 0 || runCtx.dryRun {
+		return
+	}
+
+	manifestPath, err := runCtx.manifestResolver.Resolve(state.schemaID)
+	if err != nil {
+		runCtx.logger.Error("failed to resolve manifest path", zap.Int16("schema_id", state.schemaID), zap.Error(err))
+		return
+	}
+	if err := manifest.AppendFiles(ctx, runCtx.manifestStore, manifestPath, state.schemaID, state.fileEntries); err != nil {
```
iceboundrock, line 402: P1: `updateSchemaManifest` logs resolver and `AppendFiles` failures, then returns successfully. The created base files are therefore absent from the manifest and cannot be discovered by consumers; a retry can duplicate exports. Please propagate this error from `initSchema` and add resolver/S3 manifest-failure coverage.

iceboundrock:
## Assessment

Request changes. The review found four P1 and four P2 issues. The P1 issues can make CDC init report success without a usable export, omit mandatory failure diagnostics, or expose external database credentials.

## Standards

Against `docs/ai_review_guideline.md`:

- Error handling: manifest-write and cursor-iteration failures must be propagated.
- API/zero-value behavior: the new exported `RunInit` must apply the same CDC defaults as `Runner.RunOnce`.
- Security: artifacts containing connection information must not be world-readable.

## Spec

Against #173:

- Provisioning failures need a run-specific artifact directory.
- Every failure needs a `diff.json`, including non-query failures.
- The harness does not implement the requested per-test tenant/schema-ID isolation model.
- `KEEP_E2E_ENV=1` alone does not preserve containers because Ryuk still cleans them up.

## Observations

The new harness correctly isolates databases and S3 prefixes per environment, but those measures do not cover the missing tenant contract or early provisioning-failure diagnostics.

## Verification

- `go test ./internal/cdc ./internal/e2e_harness/production ./internal/federated ./internal/sqlgen`
- `go vet ./internal/cdc ./internal/e2e_harness/production ./internal/federated ./internal/sqlgen`
- `git diff --check c61d9d0a39632c84e459671a7a47b1190e9c8cf1...origin/pr-191`

All three completed successfully.

## Summary

8 findings total: 4 P1, 4 P2. The highest-priority fix is to return manifest persistence failures from `RunInit`.

internal/e2e_harness/production/cluster.go
```diff
@@ -0,0 +1,260 @@
...
+	externalPGDSNVar       = "PRODUCTION_E2E_EXTERNAL_PG_DSN"
+	externalS3EndpointVar  = "PRODUCTION_E2E_EXTERNAL_S3_ENDPOINT"
+	externalS3BucketVar    = "PRODUCTION_E2E_EXTERNAL_S3_BUCKET"
+	externalS3RegionVar    = "PRODUCTION_E2E_EXTERNAL_S3_REGION"
+	externalS3AccessKeyVar = "PRODUCTION_E2E_EXTERNAL_S3_ACCESS_KEY"
+	externalS3SecretKeyVar = "PRODUCTION_E2E_EXTERNAL_S3_SECRET_KEY"
+)
+
+// Cluster owns the container set shared by every test in one test binary:
+// Postgres (admin database used only for CREATE/DROP DATABASE) and MinIO
```
iceboundrock: Let's use RustFS instead of MinIO since official docker of MinIO is Archived(no updates in the feature)

internal/cdc/init.go
iceboundrock: Fixed in 2591bd2. `RunInit` now normalizes options via `normalizeInitOptions` (applies `CDCConfig.WithDefaults()`, same as `Runner.RunOnce`, plus the nop-logger fallback). Added `TestNormalizeInitOptions_ZeroValueConfigGetsUsableBatchSize` and `TestResolveInitBatchSize_RawZeroValueConfigIsZero`, which documents the LIMIT 0 trap this prevents.

internal/cdc/init.go
iceboundrock: Fixed in 2591bd2. Both cursor loops (`getSchemaIDsToInit`, `selectEntityMainBatch`) now check and wrap `rows.Err()` after iteration.

internal/cdc/init.go
iceboundrock: Fixed in 2591bd2. `updateSchemaManifest` returns errors (resolver and `AppendFiles`) and `initSchema` propagates them; added unit coverage for both failure modes plus the no-store/dry-run no-op cases.

One observation for a separate decision: the mainline delta flusher has the same swallow pattern (`internal/cdc/flusher.go` ~:394, "manifest update failed ... Don't return - the flush succeeded, manifest is non-critical"). I left it untouched here because changing flush semantics is out of scope for this PR, but if the init rationale applies there too it deserves its own issue.

internal/e2e_harness/production/artifacts.go
iceboundrock: Fixed in 347f35b. `run.json` now writes a redacted DSN (`redactedPGDSN`, password replaced with `REDACTED`); all artifacts are written 0600 under 0700 directories. Regression coverage: `TestRedactedPGDSN_OmitsPassword` + `TestWriteJSONArtifact_OwnerOnlyPermissions` (unit), and `TestDumpArtifacts` now asserts the cluster password never appears in run.json (e2e).

internal/e2e_harness/production/env.go
iceboundrock: Fixed in 347f35b. `NewEnv` registers its cleanups (teardown + artifact dump, LIFO-ordered) before provisioning, so a provisioning failure still dumps run-specific artifacts: the dump steps tolerate unprovisioned resources (`change_log.json` gets an `unavailable` marker, parquet dumps are skipped without DuckDB) and `run.json` records the provisioning error. New e2e test `TestDumpArtifacts_UnprovisionedEnv` covers exactly the missing-database case.

internal/e2e_harness/production/artifacts.go
iceboundrock: Fixed in 347f35b. `diff.json` is now always written: query mismatches produce the full diff as before, and failures outside `AssertQueryMatches` (provisioning/init/flush) produce a structured `{"status": "no_query_mismatch_recorded", "queries_executed": N}` document. Covered by `TestDumpDiff_AlwaysWritesStructuredDocument` and the unprovisioned-env e2e test.

internal/e2e_harness/production/cluster.go
iceboundrock: Fixed in 347f35b — enforced in the harness. `StartCluster` calls `enforceContainerRetention()` before the first container is created: under KEEP_E2E_ENV=1 it sets `TESTCONTAINERS_RYUK_DISABLED=true` itself (testcontainers reads its config once per process, so the ordering matters). An explicitly set `TESTCONTAINERS_RYUK_DISABLED` — either value — is respected. Three unit tests cover set/override/no-op; README updated to drop the manual instruction.

internal/e2e_harness/production/cluster.go
iceboundrock: Done in 580e644. The base harness gains `StartRustFS` (same `rustfs/rustfs:latest` image and env-var contract as deploy/docker-compose.yml, credentials kept at the MinIO defaults so client wiring is unchanged) and the production cluster now uses it. `StartS3`/MinIO is untouched for the existing federated suite — migrating those 68 cases can ride a follow-up. Verified the full production e2e suite (15 tests) against RustFS: CDC CopyObject tmp→final, manifest conditional PUT (If-Match etag), ListObjectsV2 pagination, and DuckDB httpfs parquet reads all pass.

internal/e2e_harness/production/schema.go
iceboundrock: Responding rather than changing code — forma has no first-class tenant to isolate: the DDL carries no tenant column and no storage/query API takes a tenant key (the only "tenant" strings in the repo are incidental test values). Deploy-level tenancy in this codebase is a database + S3 prefix pair, and that is exactly what the harness gives each test: a private database (which also carries the advisory-lock and change_log-scan isolation) and a private bucket prefix. The fixture schema IDs are numerically the same across tests, but each copy lives in a private database and prefix, so no state, lock, or parquet file is reachable from another test — which is the property the spec constraint is after. Tests needing different schema IDs can register their own via `RegisterSchemas` + `WithSchemaDir`. Documented this mapping in the README isolation section (3062967). If you want tenant to become a first-class concept in the harness API (e.g. anticipating future product work), that's a design decision worth its own issue rather than a review fix.

iceboundrock:
## Review-fix round 1

All 8 findings addressed (7 fixed, 1 answered), plus the RustFS request:

| Finding | Resolution |
|---|---|
| P1 `RunInit` missing defaults | fixed 2591bd2 — `normalizeInitOptions` applies `WithDefaults()`; zero-value tests added |
| P1 manifest failures swallowed | fixed 2591bd2 — propagated from `initSchema`; failure-mode tests; noted the same pattern exists in the mainline delta flusher (potential follow-up) |
| P1 artifacts leak PG password | fixed 347f35b — redacted DSN in run.json, 0600/0700 permissions, unit+e2e regression tests |
| P1 provisioning failures leave no artifacts | fixed 347f35b — cleanups registered before provisioning; nil-tolerant dump steps; `provision_error` in run.json; new e2e test |
| P2 `rows.Err()` unchecked | fixed 2591bd2 — both loops |
| P2 diff.json not uniform | fixed 347f35b — always written; structured `no_query_mismatch_recorded` document for non-query failures |
| P2 KEEP_E2E_ENV vs ryuk | fixed 347f35b — harness disables the reaper itself before container creation; explicit setting respected |
| P2 tenant/schema-ID isolation (spec) | answered in-thread — forma has no first-class tenant; per-test database + S3 prefix is the tenancy boundary; README documents the mapping (3062967) |
| RustFS instead of archived MinIO | done 580e644 — `StartRustFS` in the base harness, production cluster switched; federated suite untouched; full e2e suite verified green on RustFS |

Verification: `make test` (all packages), `make lint` clean, `make test-e2e-production` 15/15 pass against RustFS (smoke tier hits cold=10 warm=16 hot=3).

Issue #203, opened 2026-07-10
cdc: manifest reconciliation tool (compare S3 parquet keys against manifest entries)


iceboundrock:
> *This was generated by AI during triage.*

Follow-up recommended by the final review of PR #202 (fix for #197).

## Context

Since #197 (delta, PR #202) and the init-path fix (2591bd2, PR #191 review), manifest update failures propagate and fail the CDC run instead of being silently swallowed. But for the delta flush the failure is discovered *after* rows are marked flushed:

- the parquet file is already at its final S3 key,
- the rows are already `flushed_at != 0`, so a re-run will not re-export them or retry the manifest append,
- retrying the append blindly is unsafe (`manifest.AppendFile` is load-modify-save with etag; a retry after an ambiguous save failure duplicates the entry).

The error message embeds the orphaned file's final key, and **manual reconciliation is now the documented recovery path** — but there is no tooling for it. An operator today has to hand-edit the manifest JSON or write ad-hoc scripts.

## Proposal

A `cmd/tools` subcommand (alongside `cdc_flush` / `cdc_init` / `compactor`), e.g. `manifest-reconcile`, that for each schema:

1. Lists parquet objects under the CDC layout `prefix/{schemaID}/*.parquet` (excluding `prefix/{schemaID}/_tmp/`), covering both delta files (`{uuid}.parquet`, `BuildDeltaPath`) and base files (`{min}_{max}.parquet`, `BuildBasePath`).
2. Loads the schema manifest via the configured `manifest.PathResolver` and diffs both directions:
   - **Orphans:** objects on S3 with no `FileEntry.Path` in the manifest (the #197 failure mode).
   - **Dangling entries:** manifest entries whose S3 object no longer exists.
3. Reports the diff (default: read-only, non-zero exit when discrepancies are found, so it can run as a periodic check).
4. Optional `--repair`: append missing entries for orphans. Entry metadata (`RowCount`, `RowIDMin`/`RowIDMax`, `CreatedMin`/`CreatedMax`, `Tier`) can be recomputed by reading the parquet's `row_id` column via DuckDB rather than trusted from filenames. Dangling-entry removal should probably stay manual or behind a separate explicit flag — deleting manifest entries hides data from consumers.

## Open questions

- Concurrency with live flushers: reconcile under the same per-schema advisory lock the flusher uses, or accept a benign race (a file uploaded between the S3 list and the manifest load looks like an orphan)? Locking seems cheaper than explaining false positives.
[...]

PR #208, opened 2026-07-11
test(e2e): init -> incremental CDC handoff correctness (#176)

### LINKED ISSUE #176: [E2E] Init → incremental CDC handoff correctness
Part of epic #172.

## Task

Verify that `cdc-init` backfill and subsequent incremental `cdc-flush` produce a consistent, complete cold tier with correct merge-on-read behavior.

### 1. Backfill → federated query parity
- Create data → run production `cdc-init` → verify base Parquet exists
- Run federated query against base only
- Assert exact row count and values match Postgres

### 2. Backfill → incremental delta → query
- Create initial data → `cdc-init` → create more data → `cdc-flush`
- Federated query against base + delta → assert count = total, no duplicates, no gaps
- Latest version wins where overlap exists

### 3. Backfill with deletes → delta → query
- Create data with some soft-deleted → `cdc-init` → delete more → `cdc-flush`
- Assert deleted rows invisible in federated result

### 4. Backfill rerun idempotency
- `cdc-init` → no changes → `cdc-init` again
- Assert no duplicate manifest entries, no duplicate objects
- Assert federated query returns same results

### 5. Backfill under concurrent mutation
- Start `cdc-init` → concurrently insert/update rows
- Known risk: OFFSET-based pagination may skip or duplicate rows
- Assert final federated result matches expected state

## High-risk hypotheses

- `cdc-init` uses mutable `OFFSET` pagination without snapshot → concurrent writes may cause row skip/duplication
- Rerun appends duplicate manifest entries and may leave obsolete files

## Success criteria

- Base + Delta + Hot merge-on-read is exact (count, values, versions)
- No duplicate files or manifest entries after rerun
- Backfill and CDC hand off cleanly

iceboundrock:
Closes #176 (epic #172, Phase 2).

## What

Five E2E scenarios pinning the cdc-init backfill -> incremental cdc-flush handoff, plus one production fix the rerun scenario exposed.

- **S1** `TestInitBaseOnlyParity`: multi-batch backfill -> base-only federated parity (onboarding contract: change_log cleared post-init).
- **S2** `TestInitThenIncrementalFlush` / `TestInitFlushOverlapWithoutLogCleanup`: exact counts across base+delta, delta version wins on overlap; the no-cleanup variant pins that a full first-flush re-export dedupes to exact counts.
- **S3** `TestInitDeleteHandoff` + `TestInitSoftDeletedRowExcluded`: deletes on both sides of the init boundary stay invisible (hard-delete + tombstone path), and a row already soft-deleted in `entity_main` at init time is excluded by init's `ltbase_deleted_at IS NULL` filter (soft-deleted state injected via the SQL escape hatch after an API delete keeps the oracle in sync — the API itself hard-deletes).
- **S4** `TestInitRerunIdempotency` + `TestInitRerunAfterChangesReconcilesManifest`: no-change rerun -> zero new objects (deterministic min/max row-id keys overwrite) and an unchanged manifest; rerun after inserts+deletes -> the manifest base tier becomes exactly the new run's file set (stale ranges and historical duplicates removed), and the final parity check proves glob+LWW keeps queries exact while the stale objects remain.
- **S5** `TestInitUnderConcurrentMutation`: OFFSET pagination races (issue hypothesis 1 is real — no snapshot in `selectEntityMainBatch`) are absorbed by the handoff: skipped rows stay hot then flush to delta, duplicates dedupe via LWW; end state is exact. A started-barrier plus atomic counter snapshots around `RunInit` prove at least one create and one update overlapped init on every run (observed 120-139 during-init mutations across three runs). (One harness nuance surfaced here: a single flush pass drains at most `BatchSize` rows, so the test restores the default batch size after the race window before the handoff flush.)

## Production fix

Issue hypothesis 2 confirmed: `manifest.AppendFiles` appended blindly, so every init rerun doubled the base entries. Since cdc-init is a full re-export of a schema's live rows, `updateSchemaManifest` now calls `manifest.ReplaceTierFiles`: the manifest's base tier is replaced wholesale with each run's output (delta entries preserved in order). Reruns neither duplicate entries nor leave stale ranges behind, and manifests already carrying historical duplicates are healed on their next init run.

Obsolete S3 *objects* from superseded runs are intentionally not deleted here: glob-based federated reads stay exact via LWW/tombstones (asserted end-to-end by `TestInitRerunAfterChangesReconcilesManifest`), and safe object GC needs the global manifest-vs-S3 view of the #203 reconciliation tool (orphans can exist outside the manifest too).

### DIFF
--- a/internal/e2e_harness/production/init_concurrent_e2e_test.go
+++ b/internal/e2e_harness/production/init_concurrent_e2e_test.go
@@ -1,0 +1,115 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"fmt"
+	"testing"
+)
+
+// TestInitUnderConcurrentMutation (#176 scenario 5): cdc-init pages with
+// OFFSET over live data (internal/cdc/init.go selectEntityMainBatch — no
+// snapshot), so concurrent writes can skip or duplicate rows in the base
+// tier. The system-level contract asserted here: whatever the base tier
+// captured, unflushed change_log entries shadow it (hot) and the follow-up
+// flush lands the rest in delta, so the final federated state is exact.
+func TestInitUnderConcurrentMutation(t *testing.T) {
+	cluster := SharedCluster(t)
+	env := NewEnv(t, cluster)
+	ctx := context.Background()
+	wide := DefaultSchemaFixtures()[1]
+
+	env.CDC.BatchSize = 5 // 60 seed rows -> >=12 OFFSET pages of race window
+
+	seed := env.GenerateScript(ScriptSpec{Schema: wide, Creates: 60})
+	if err := env.ApplyEvents(ctx, seed...); err != nil {
+		t.Fatalf("apply seed creates: %v", err)
+	}
+
+	stop := make(chan struct{})
+	type mutations struct {
+		created int
+		deleted int
+		err     error
+	}
+	done := make(chan mutations, 1)
+	go func() {
+		var m mutations
+		defer func() { done <- m }()
+		for i := 0; ; i++ {
+			select {
+			case <-stop:
+				return
+			default:
+			}
+			var batch []*Event
+			// One new create per round (created rows are hot until the
+			// post-init flush) — capped so the query limit below stays safe.
+			if m.created < 200 {
+				news := env.GenerateScript(ScriptSpec{Schema: wide, Creates: 1})
+				batch = append(batch, news...)
+				m.created++
+			}
+			// Rolling updates over seed[0:40] (disjoint from delete range).
+			batch = append(batch, UpdateEvent(wide, seed[i%40].RowID, map[string]any{
+				"title": fmt.Sprintf("race-%03d", i),
+				"count": float64(400000 + i),
+			}))
+			// Ten deletes from seed[40:50]: shifts OFFSET pages leftward,
+			// the classic skip trigger.
+			if m.deleted < 10 {
+				batch = append(batch, DeleteEvent(wide, seed[40+m.deleted].RowID))
+				m.deleted++
+			}
+			if err := env.ApplyEvents(ctx, batch...); err != nil {
+				m.err = fmt.Errorf("mutator round %d: %w", i, err)
+				return
+			}
+		}
+	}()
+
+	report, initErr := env.RunInit(ctx, wide)
+	close(stop)
+	m := <-done
+	if m.err != nil {
+		t.Fatalf("concurrent mutator failed: %v", m.err)
+	}
+	if initErr != nil {
+		t.Fatalf("run init under concurrent mutation: %v", initErr)
+	}
+	if report.RowsExported == 0 {
+		t.Fatal("init exported no rows")
+	}
+	t.Logf("init exported %d rows in %d files; mutator created %d, deleted %d",
+		report.RowsExported, report.FilesCreated, m.created, m.deleted)
+
+	// The small BatchSize above only served to widen the init OFFSET race
+	// window; it also caps rows-per-flush, so restore the harness default
+	// (env.go) before the handoff flush so a single RunOnce drains the whole
+	// hot tier in one snapshot, exactly as the sibling handoff tests do. The
+	// mutator has already stopped (m received above), so this is race-free.
+	env.CDC.BatchSize = 10000
+
+	// Handoff: flush everything the mutator produced (plus any rows init's
+	// pagination missed — their change_log entries are still unflushed).
+	flush, err := env.RunFlush(ctx)
+	if err != nil {
+		t.Fatalf("post-init flush: %v", err)
+	}
+	if flush.UnflushedAfter != 0 {
+		t.Fatalf("flush left %d unflushed rows", flush.UnflushedAfter)
+	}
+
+	// Exact end state: 60 seed + created - deleted, every value the latest.
+	expected := 60 + m.created - m.deleted
+	result := env.AssertQueryMatches(ctx, Query{Schema: wide, Limit: expected + 50})
+	if result != nil && len(result.Records) != expected {
+		t.Fatalf("federated result has %d rows, want %d (no skips, no duplicates)", len(result.Records), expected)
+	}
+	env.AssertQueryMatches(ctx, Query{
+		Schema:  wide,
+		Filters: []Filter{{Attr: "count", Op: "gte", Value: "400000"}},
+		Limit:   expected + 50,
+	})
+}
--- a/internal/e2e_harness/production/init_concurrent_e2e_test.go
+++ b/internal/e2e_harness/production/init_concurrent_e2e_test.go
@@ -5,8 +5,18 @@ package production
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"testing"
 )
+
+// raceCounters tracks mutator progress. Written only by the mutator
+// goroutine; the main goroutine reads them atomically to prove mutations
+// overlapped RunInit (snapshot before the call vs after it returns).
+type raceCounters struct {
+	created atomic.Int64
+	updated atomic.Int64
+	deleted atomic.Int64
+}
 
 // TestInitUnderConcurrentMutation (#176 scenario 5): cdc-init pages with
 // OFFSET over live data (internal/cdc/init.go selectEntityMainBatch — no
@@ -14,6 +24,9 @@ import (
 // tier. The system-level contract asserted here: whatever the base tier
 // captured, unflushed change_log entries shadow it (hot) and the follow-up
 // flush lands the rest in delta, so the final federated state is exact.
+// The started barrier plus counter snapshots around RunInit prove at least
+// one create and one update genuinely overlapped the init pagination
+// (review round 1: without them the scheduler could serialize the test).
 func TestInitUnderConcurrentMutation(t *testing.T) {
 	cluster := SharedCluster(t)
 	env := NewEnv(t, cluster)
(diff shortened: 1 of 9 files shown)


internal/e2e_harness/production/init_delete_e2e_test.go
```diff
@@ -0,0 +1,89 @@
...
+package production
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/uuid"
+)
+
+// TestInitDeleteHandoff (#176 scenario 3): rows deleted before the backfill
```
iceboundrock, line 13: **Spec · P2** #176 explicitly calls for rows that are already soft-deleted at init time. This test documents and exercises hard deletes, so no row remains in `entity_main` with `ltbase_deleted_at` set. It therefore does not verify the `cdc-init` soft-delete filter. Inject a soft-deleted row through the harness SQL escape hatch, with a matching change-log tombstone, and assert it is absent from base and the final federated result.

internal/e2e_harness/production/init_concurrent_e2e_test.go
```diff
@@ -0,0 +1,115 @@
...
+				m.deleted++
+			}
+			if err := env.ApplyEvents(ctx, batch...); err != nil {
+				m.err = fmt.Errorf("mutator round %d: %w", i, err)
+				return
+			}
+		}
+	}()
+
+	report, initErr := env.RunInit(ctx, wide)
```
iceboundrock, line 58: **Spec · P1** This starts the mutator but does not wait for a successful mutation before init begins. The scheduler can therefore run all of `RunInit` before the goroutine writes, making the race test pass without testing the OFFSET-pagination handoff. Add a started/first-mutation barrier and assert that at least one create and update completed during init.

internal/e2e_harness/production/init_handoff_e2e_test.go
```diff
@@ -0,0 +1,210 @@
...
+	env.AssertQueryMatches(ctx, Query{
+		Schema: wide,
+		Sorts:  []Sort{{Attr: "count", Desc: true}},
+		Limit:  10,
+	})
+}
+
+// parquetKeys filters an object-key diff down to parquet files: the first
+// manifest write also shows up in NewObjects.
+func parquetKeys(keys []string) []string {
```
iceboundrock: **Standards · Hard** `coding-standard.md` §4 requires private helper names to be verb-first. Rename `parquetKeys` and the nearby `basePaths` to action-oriented names such as `filterParquetKeys` and `buildBasePaths`.

internal/e2e_harness/production/init_concurrent_e2e_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"testing"
+)
+
+// TestInitUnderConcurrentMutation (#176 scenario 5): cdc-init pages with
+// OFFSET over live data (internal/cdc/init.go selectEntityMainBatch — no
+// snapshot), so concurrent writes can skip or duplicate rows in the base
+// tier. The system-level contract asserted here: whatever the base tier
+// captured, unflushed change_log entries shadow it (hot) and the follow-up
+// flush lands the rest in delta, so the final federated state is exact.
+func TestInitUnderConcurrentMutation(t *testing.T) {
```
iceboundrock, line 30: **Standards · Observation** This test is close to the 100-line function limit and exceeds the >80-line refactoring trigger in `coding-standard.md` §7. Consider extracting the mutator lifecycle or the post-init assertions to keep the scenario easy to audit.

internal/manifest/manifest.go
```diff
@@ -176,3 +176,34 @@ func AppendFiles(ctx context.Context, st Store, path string, schemaID int16, ent
...
+// optimistic locking via etag to handle concurrent updates.
+func UpsertFiles(ctx context.Context, st Store, path string, schemaID int16, entries []FileEntry) error {
+	if len(entries) == 0 {
+		return nil
+	}
+	m, etag, err := LoadOrCreate(ctx, st, path, schemaID)
+	if err != nil {
+		return fmt.Errorf("load manifest: %w", err)
+	}
+	index := make(map[string]int, len(m.Files))
```
iceboundrock: **Spec · P1** `UpsertFiles` indexes each existing path to only its last occurrence and only replaces matching paths. A rerun after inserts or deletes changes deterministic min/max batch paths, so old manifest entries and objects remain while new paths are appended; existing duplicate paths also remain. This does not meet issue #176 obsolete-file and no-duplicate-manifest success criteria. Please reconcile the base-tier manifest to the full set produced by the current init run, or explicitly split that required behavior from this PR.

iceboundrock:
## Review Assessment
This PR adds valuable end-to-end coverage for the init-to-flush handoff, but it does not fully satisfy issue #176 yet.

## Findings
- **Spec P1:** Reruns after data changes retain obsolete base manifest entries and objects; historical duplicates are also preserved.
- **Spec P1:** The concurrent-mutation test has no barrier proving a mutation overlaps `RunInit`.
- **Spec P2:** The delete scenario tests hard-delete tombstones rather than rows already soft-deleted at init time.
- **Standards hard:** `parquetKeys` and `basePaths` do not follow the verb-first private-helper rule in `coding-standard.md` §4.
- **Standards observation:** `TestInitUnderConcurrentMutation` exceeds the >80-line refactoring trigger in `coding-standard.md` §7.

The detailed findings are attached as inline comments.

## Observations
The no-change rerun behavior is covered and the targeted E2E run observed concurrent mutations. Neither result proves reruns reconcile changed batch ranges or that every run of the concurrency test overlaps a mutation.

## Verification
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/manifest ./internal/cdc` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -tags=e2e -v ./internal/e2e_harness/production -run "^TestInit(BaseOnlyParity|ThenIncrementalFlush|FlushOverlapWithoutLogCleanup|DeleteHandoff|RerunIdempotency|UnderConcurrentMutation)$" -count=1` passed: 6 tests.
- `git diff --check 972e0be...4e13cf4` was clean.

## Summary
Three specification findings and two standards findings. The stale-manifest rerun behavior is the highest-impact issue.

internal/manifest/manifest.go
iceboundrock: Fixed in 6286e0f + f7c0da4. `UpsertFiles` is removed; since cdc-init is a full re-export, `updateSchemaManifest` now calls `manifest.ReplaceTierFiles` — the base tier is replaced wholesale with the current run's output (delta entries preserved in order), so stale ranges from earlier runs and pre-existing historical duplicates are gone after the next init. Pinned at three levels: `TestReplaceTierFiles_*` (manifest), `TestUpdateSchemaManifest_RerunReconcilesBaseTier` (cdc, seeded with a duplicate + a stale range), and E2E `TestInitRerunAfterChangesReconcilesManifest` (inserts+deletes -> exactly ceil(23/7)=4 entries, >=1 run-1 path evicted, row sum 23).

Obsolete S3 objects: split out, staying with #203. Two technical reasons: (a) the new E2E asserts end-to-end that stale objects are query-invisible under glob+LWW once tombstones flush, so they are storage garbage, not a correctness gap; (b) deleting only manifest-known obsoletes is false safety — orphans can exist outside the manifest (e.g. past partial failures), so object GC needs #203's global manifest-vs-S3 diff. Noted in the PR body and the `updateSchemaManifest` comment.

internal/e2e_harness/production/init_concurrent_e2e_test.go
iceboundrock: Fixed in b47af33. The mutator now closes a `started` channel after its first fully-applied round, and the main goroutine blocks on it before calling `RunInit`; progress is tracked in atomic counters snapshotted immediately before the call and immediately after it returns, and the test fails if either `createdDuring` or `updatedDuring` is zero. Observed during-init mutations across the three verification runs: 139/134/120. (The barrier wait also selects on the error channel so a mutator failure before round 1 cannot deadlock the test.)

internal/e2e_harness/production/init_delete_e2e_test.go
iceboundrock: Fixed in cfbf404, `TestInitSoftDeletedRowExcluded`. Injection shape: the row is first deleted through the API (real change_log tombstone, oracle stays in sync), then its `entity_main` row is re-materialized with `ltbase_deleted_at` set via `ExecSQL` — a legal storage state the API cannot produce (it hard-deletes). Asserts init's `ltbase_deleted_at IS NULL` filter live: `RowsExported == 5` of 6, base parquet rows == 5, and the row invisible in the federated result both before and after the flush (with the 5 survivors as the positive guard).

internal/e2e_harness/production/init_handoff_e2e_test.go
iceboundrock: Fixed in 176d203: `parquetKeys` -> `filterParquetKeys`, `basePaths` -> `buildBasePaths`. (`assertBaseRows` was already verb-first.)

internal/e2e_harness/production/init_concurrent_e2e_test.go
iceboundrock: Addressed in b47af33 together with the overlap barrier: the mutator lifecycle is extracted to `runInitRaceMutator` (with the sole-caller/disjoint-ranges constraints documented on it), leaving both functions under 80 lines.

Issue #196, opened 2026-07-10
federated: retire skipped integration test stubs superseded by the production E2E harness


iceboundrock:
Follow-up from PR #191 (#173, epic #172).

## Problem

`internal/federated/duckdb_federated_production_e2e_test.go` and other skipped integration tests reference "root integration tests" / setups that no longer exist, so they skip unconditionally and provide no coverage. The production harness (`internal/e2e_harness/production`) now covers the real CDC → parquet → federated read path end to end.

## Suggested fix

Audit the skipped stubs in `internal/federated`; for each, either (a) delete it and note which production-harness test supersedes it, or (b) port it to the production harness if it covers something the harness doesn't yet.

## Acceptance

- No permanently-skipped test remains in `internal/federated` whose coverage is already provided by the production harness.
- Anything genuinely uncovered is ported, not deleted.

iceboundrock:
Partial progress via PR #209 (review round 1 of #177): the seven three-tier stubs in `duckdb_federated_production_e2e_test.go` are deleted — their scenarios now run oracle-checked in `internal/e2e_harness/production` (`TestThreeTierIsolation`, `TestThreeTierMerge/{union,triple_overlap_hot_wins,dirty_hot_shadows_cold,changed_at_beats_tier_layout}`, `TestThreeTierSoftDelete`).

Remaining in this issue's audit: three permanently-skipped stubs in `duckdb_federated_integration_test.go` — `TestExecuteDuckDBFederatedQuery_ClientUnavailable` (:87), `TestStreamDuckDBFederatedQuery_BasicExecution` (:101), `TestBuildDuckDBQuery_InvalidTemplateSyntax` (:111). These are engine-level concerns (client-unavailable failure mode, basic execution, template-syntax errors); per the acceptance criteria they need per-stub adjudication — delete if the production harness (or #185/#187 work) demonstrably covers them, port if not.

iceboundrock:
Completed the remaining audit for this issue — the three permanently-skipped stubs in `internal/federated/duckdb_federated_integration_test.go`, whose "covered by root integration tests" claim was stale (those root monolith tests no longer exist):

[...]

PR #209, opened 2026-07-11
test(e2e): production three-tier federated query correctness (#177)

### LINKED ISSUE #177: [E2E] Production three-tier federated query correctness
Part of epic #172.

## Task

Verify that the real production federated engine correctly merges Base (Parquet), Delta (Parquet), and Hot (Postgres) tiers using exact, deterministic oracles. This replaces the 7 currently-skipped tests in `internal/federated/duckdb_federated_production_e2e_test.go`.

### 1. Base-only query
- Seed data → `cdc-init` → federated query with base glob only

### 2. Delta-only query
- Seed data → `cdc-flush` → federated query with delta glob only

### 3. Hot-only query
- Seed data (no flush) → federated query (Postgres only)

### 4. All three tiers
- Seed old data → `cdc-init` → seed mid data → `cdc-flush` → seed new data
- Federated query across all tiers → assert non-overlapping rows all present

### 5. Overlapping rows: LWW
- Same `row_id` in base, delta, and hot with different versions
- Assert latest hot version wins

### 6. Overlapping rows: dirty shadow
- Same `row_id` in cold tier AND in hot (unflushed)
- Assert hot row shadows cold row; cold version excluded via anti-join

### 7. Soft-deleted across mixed tiers
- Delete in one tier, active in another
- Query: assert deleted row invisible

### 8. Generic (non-benchmark) schema
- Use a production-like schema with real attribute types, not benchmark schema IDs 100-102

## Success criteria

- Exact row IDs, values, ordering, and total count match the independent oracle
- Anti-join correctly excludes dirty rows from Parquet sources
- All 7 previously-skipped test scenarios are now executed

iceboundrock:
Closes #177 (part of epic #172).

Adds oracle-checked production-harness coverage for three-tier federated merge-on-read:

- **Tier isolation** (`TestThreeTierIsolation`): base-only (init + onboarding cleanup), delta-only (drained flush), hot-only (PreferHot pre-init window). Isolation is physical, not glob-scoped — production has a single flat parquet glob per schema and a manifest-blind reader, so per-tier globs are not a production input (acceptance note posted on #177).
- **Union** (`TestThreeTierMerge/union`): 10 base + 8 delta + 6 hot disjoint rows — exact set, total order, pagination across tier boundaries.
- **Triple-overlap LWW** (`triple_overlap_hot_wins`): same row_ids in base(v1)/delta(v2)/hot(v3); newest hot version wins, v1-only attributes survive partial updates.
- **Dirty shadow** (`dirty_hot_shadows_cold`): unflushed hot version evicts the cold base version via the dirty-set anti-join — including under a filter only the stale version matches.
- **changed_at beats created_at and layout** (`changed_at_beats_tier_layout`): two delta generations (v1 @ create ver_ts, v2 @ update ver_ts) plus the base duplicate of v2 @ create ver_ts — the winner is strictly newest by changed_at, with per-row strict-ordering assertions. Supersedes the ConflictingCreatedAndUpdatedAt stub.
- **Soft-delete mixed tiers** (`TestThreeTierSoftDelete`): hot tombstone vs live base row (anti-join) and flushed delta tombstone vs live base row (parquet LWW), with live controls in every tier.

All scenarios run on the generic full-type `e2e_wide` fixture (schema ID 21 — non-benchmark, issue scenario 8) and assert exact row IDs, values, ordering, and totals against the independent oracle.

The seven superseded stubs in `internal/federated/duckdb_federated_production_e2e_test.go` are **deleted** (folded from #196's scope at review round 1; the three engine-level skips in `duckdb_federated_integration_test.go` stay with #196's audit).

**Review round 1 findings** (https://github.com/Lychee-Technology/forma/pull/209#pullrequestreview-4676717407): the original changed_at scenario rested on an undefined equal-ver_ts tie — cdc-init stamps base rows with `ltbase_created_at` as the version timestamp, which also produces a deterministic stale-read window through the onboarding flow. Scenario reworked onto well-defined timestamps (19a5728); the underlying init export contract is **#210**.

Verification: full production E2E suite green (includes all new tests), `make test` green, `make lint` clean.


internal/e2e_harness/production/three_tier_merge_e2e_test.go
```diff
@@ -0,0 +1,275 @@
...
+	mustFlush(ctx, t, env) // delta = v1
+
+	v2 := make([]*Event, 0, len(creates))
+	for i, c := range creates {
+		v2 = append(v2, UpdateEvent(wide, c.RowID, map[string]any{
+			"title": fmt.Sprintf("newer-%02d", i),
+			"count": float64(400000 + i),
+		}))
+	}
+	if err := env.ApplyEvents(ctx, v2...); err != nil {
```
iceboundrock, line 240: **Spec: deterministic oracle / LWW coverage**

This scenario assumes v2 has a later `changed_at`, but it neither waits for nor asserts `v2.ChangedAt > v1.ChangedAt`. The harness records wall-clock timestamps at millisecond resolution, and its oracle resolves ties with the harness-only `Seq`; DuckDB has no equivalent version tie-breaker. Equal-millisecond writes can therefore make this test flaky or validate an undefined tie rather than LWW by `changed_at`. Please force or assert a strictly later timestamp before exporting v2.

internal/e2e_harness/production/three_tier_isolation_e2e_test.go
```diff
@@ -0,0 +1,125 @@
...
+	report, err := env.RunInit(ctx, wide)
+	if err != nil {
+		t.Fatalf("run init: %v", err)
+	}
+	if report.RowsExported != 12 {
+		t.Fatalf("init exported %d rows, want 12", report.RowsExported)
+	}
+	env.ExecSQL(ctx, "DELETE FROM change_log WHERE schema_id = $1", wide.ID)
+
+	result := env.AssertQueryMatches(ctx, Query{Schema: wide, Limit: 100})
```
iceboundrock, line 61: **Spec: base-only / delta-only query inputs**

Both this base-only query and the delta-only query below use the shared `*.parquet` glob and default Hot/Warm/Cold tiers. Physical absence of the other tier is useful isolation, but issue #177 explicitly calls for a base-glob-only and delta-glob-only federated query. Please use tier-specific query inputs, or revise the issue/acceptance criteria if the production layout intentionally cannot express them.

internal/federated/duckdb_federated_production_e2e_test.go
```diff
@@ -2,30 +2,34 @@ package federated
 
 import "testing"
 
+// The production three-tier suite (#177) in internal/e2e_harness/production
+// supersedes these stubs with oracle-checked coverage; their removal is
+// tracked by #196.
+
 func TestStreamDuckDBFederatedQuery_GivenBaseAndHotVersions_WhenQueried_ThenLatestHotVersionWins(t *testing.T) {
-	t.Skip("requires root symbols (NewDBPersistentRecordRepository, loadProductionFederatedMetadata); covered by root integration tests")
+	t.Skip("superseded by TestThreeTierMerge/dirty_hot_shadows_cold (internal/e2e_harness/production, #177); removal tracked by #196")
```
iceboundrock: **Spec and standards**

Issue #177 says all seven previously skipped scenarios should execute, but this and the other six functions remain unconditional `t.Skip` calls; changing the messages to point at replacement tests does not execute these tests.

Also, the seven touched exported test functions lack individual Go doc comments. The file-level comment does not document each declaration, contrary to `coding-standard.md:80`.

iceboundrock (commented):
## Findings

Three specification gaps and one coding-standard violation were identified.

## Standards

- `internal/federated/duckdb_federated_production_e2e_test.go`: the seven touched exported test functions lack individual Go doc comments. The file-level comment does not document each declaration, which conflicts with `coding-standard.md:80`.

## Spec

- The changed-at conflict scenario does not ensure v2 has a strictly later `changed_at` than v1.
- The seven original test functions remain unconditional skips rather than executing.
- Base-only and delta-only use the shared `*.parquet` path and default tiers instead of isolated query inputs.

## Assessment

The timestamp issue is the highest risk: it makes the LWW regression test nondeterministic and can validate an undefined equal-timestamp tie instead of the intended `changed_at` ordering.

## Observations

Physical tier isolation still provides useful coverage, but it is not equivalent to exercising base-only and delta-only glob inputs. The PR adds the intended production-harness scenarios, but leaves the original skipped functions in place.

## Verification

- Reviewed the PR diff against `main` at `d7bfe0e5e1edec7d15891074ee55df2093d1aa51`.
- Inspected the affected harness, oracle, query construction, and DuckDB ranking paths.
- `git diff --check d7bfe0e5e1edec7d15891074ee55df2093d1aa51...bd40df377c3002885d3d94dee4d1b52f3b89a980` completed with no whitespace errors.
- No PR test suite was run.

## Summary

1 standards finding and 3 specification findings. The nondeterministic `changed_at` assertion is the most significant issue.

internal/e2e_harness/production/three_tier_merge_e2e_test.go
iceboundrock: Confirmed — and verified to be worse than a millisecond race. Two probe runs against the harness established:

1. **The scenario was structurally in the tie state on every run, not occasionally.** cdc-init stamps base rows with `m.ltbase_created_at AS changed_at` (`internal/cdc/init_exporter.go:35`), while the delta flushed before the update carried the create-time `cl.changed_at` — so both versions carried **identical** ver_ts with different attrs. Probe dump for one row: `tier=delta changed_at=1783749882821 title="title-0000-8764"` vs `tier=base changed_at=1783749882821 title="newer-00"`. With `ORDER BY ver_ts DESC, source_tier_priority DESC, deleted_ts DESC, row_id ASC` (`advanced_query_template_duckdb.go:80`) and all parquet rows sharing `source_tier_priority = 1`, every rank key compared equal — the pass rested on DuckDB scan order. Forcing/asserting event timestamps alone could not have fixed this, since init exports created_at regardless of when v2 was written.

2. **The same contract produces a deterministic stale read** in the sequence create → update v2 → flush (delta: v2@T2) → update v3 → init (base: v3@T1-created) → change_log cleanup: delta's stale v2 beats base's fresh v3 and the oracle diff fails (`expected "fresh-v3", actual "stale-v2"`). That is a product bug in the init export contract, reachable through the documented onboarding steps; issue draft with full evidence is ready to file (see PR conversation).

Fix in 19a5728: the scenario now builds well-defined version timestamps — two delta generations (v1 @ create ver_ts, v2 @ update ver_ts) plus the base duplicate of v2 @ create ver_ts; the winner is strictly newest by changed_at, physical preconditions assert all 15 copies exist, and a per-row strict `v2.ChangedAt > create.ChangedAt` assertion pins the ordering the scenario depends on (your suggestion, now landing on the load-bearing timestamps). A ranking keyed on created_at still sees a three-way tie and can surface v1, so the ConflictingCreatedAndUpdatedAt discrimination is preserved. The undefined equal-ver_ts tie is deliberately not exercised and is documented as belonging to the follow-up issue.

internal/e2e_harness/production/three_tier_isolation_e2e_test.go
iceboundrock: Per-tier glob inputs do not exist on the production read path, so the test cannot exercise them without inventing a non-production query shape: the engine takes a single `S3ParquetPathTemplate` per schema (`DuckDBHints`, wired by `Env.ParquetGlob()` to the real CDC layout `<prefix>/<schemaID>/*.parquet`), base `<min>_<max>.parquet` and delta `<uuid>.parquet` share that one flat directory, and the reader is manifest-blind by design — the benchmark harness's `base/`/`delta/` subdirectory globs that #177's wording descends from were simulation-only layout (#173 decision). A filename-shape glob (`*_*.parquet` vs UUID hyphens) could technically split them, but that would validate an input production never issues.

Physical isolation is the production-faithful expression of scenarios 1–2's intent (each tier alone serves the query, checked against the oracle), and the delta-only subtest additionally pins the "no base files" precondition via the manifest. Took the second branch of your suggestion: 5e2d4fd documents this in the test's doc comment, and I'd annotate #177's scenarios 1–2 with: "Production uses a single flat glob per schema (no per-tier globs; reader is manifest-blind). Tier isolation is asserted physically: the query runs while the other tiers are verifiably empty." Want me to post that on the issue?

internal/federated/duckdb_federated_production_e2e_test.go
iceboundrock: Two parts:

**Godoc:** fixed in 3dabda6 — each of the seven exported stubs now carries its own doc comment naming the superseding scenario, per coding-standard.md ("Add godoc comments for exported items").

**Stubs still skipping:** the issue's "all 7 previously-skipped test scenarios are now executed" is read at scenario level — each stub's Given/When/Then now executes in the production harness against the oracle (mapping in the file header and per-stub docs). The stub *functions* themselves were never executable: they reference root symbols (`NewDBPersistentRecordRepository`, `loadProductionFederatedMetadata`) that predate #173 and no longer exist, so "executing" them can only mean deletion — which epic #172 deliberately sequences into #196 ("retire skipped stubs superseded by the production harness — best after Phase 2 lands, when the superseding coverage exists") so the supersession is reviewable before the names disappear. Kept them as sign-posted skips per that sequencing. If you'd rather fold the deletion into this PR, I'll drop the file here and note it on #196.

iceboundrock: Deletion folded in per your call: 1787b0a removes the seven-stub file entirely (their scenarios run oracle-checked in the production harness; mapping preserved in the commit message and PR body). Scope note for #196: `internal/federated/duckdb_federated_integration_test.go` still holds three permanently-skipped engine-level stubs (`TestExecuteDuckDBFederatedQuery_ClientUnavailable`, `TestStreamDuckDBFederatedQuery_BasicExecution`, `TestBuildDuckDBQuery_InvalidTemplateSyntax`) whose supersession is not established by #177 — those stay with #196's audit; noted on the issue.

Issue #178, opened 2026-07-10
[E2E] Filter-before-LWW semantic correctness


iceboundrock:
Part of epic #172.

## Task

Verify that filter predicates are applied **after** last-write-wins deduplication, not before. The current production SQL applies filters in both S3/PG sources before ranking, then again after ranking — which can allow an older matching version to reappear when the newer version does not match the filter.

### 1. Older matches, newer does not match
- Write entity v1 with `status = "open"` → flush
- Update entity v2 with `status = "closed"` → flush
- Query: `status = "open"`
- **Assert**: row must NOT appear (latest version is "closed")

### 2. Older matches, newer is a tombstone with no filter attribute
- Write entity with `status = "open"` → flush
- Delete entity (tombstone has no `status` attribute)
- Query: `status = "open"`
- **Assert**: deleted row must NOT appear

### 3. Hot shadow equivalent
- Same scenarios but with dirty (unflushed) hot row as the newer version

### 4. Multi-attribute filter
- Filter on multiple attributes, mix of MAIN and EAV
- Text, numeric, bool, and date operators

## Success criteria

- The newest version always determines visibility regardless of filter match
- A newer non-matching version never exposes an older matching version
- Tombstones correctly suppress filtered rows

PR #221, opened 2026-07-12
docs(federated): make §5 LWW SQL template executable and match the runtime source schema (#214)

### LINKED ISSUE #214: docs(federated): make the §5 LWW SQL template executable and match the runtime source schema
Follow-up from merged PR #211 / issue #178.

## Problem

PR #211 rewrote `docs/federated-query/design.md` §5 to document filter-after-LWW, but the resulting SQL sketch still cannot execute against the runtime source contract. The section calls itself the “core logic of the Federated Query Engine” and PR #211 says the sketch is self-consistent, so it must be safe to use as a reference.

### Invalid `postgres_scan` calls

The runtime template calls `postgres_scan(connection, schema, table)` (for example `internal/sqlgen/advanced_query_template_duckdb.go:58`). The documented SQL instead:

- passes `'flushed_at = 0'` as the table argument at `docs/federated-query/design.md:152`;
- passes only two arguments and treats a dynamic `SELECT` as the table name at `:206-213`;
- passes only two arguments for the EAV scan at `:216`.

These calls fail before the LWW/filter ordering can be evaluated.

### Stale parquet system-column aliases

The sketch projects `ltbase_created_at`, `ltbase_updated_at`, and `ltbase_deleted_at` from Parquet at `docs/federated-query/design.md:163-165`. The production projection is `changed_at AS created_at`, `changed_at AS ver_ts`, and `deleted_at AS deleted_ts` (`internal/sqlgen/duckdb_schema_projection.go:143-151`). Thus the example also references parquet columns that do not exist in the runtime schema.

## Suggested fix

Base §5 on the actual DuckDB template’s scan shape and source aliases, while retaining the row-id semijoin and final post-LWW logical filter added by #211. Add a rendered-SQL smoke or documentation example test that validates the sketch against DuckDB so its table-function arity and projected columns cannot drift again.

## Acceptance criteria

- The documented query uses valid runtime `postgres_scan` invocation forms.
- Its S3 source columns and aliases match the runtime projection.
- The example preserves filter-after-LWW semantics from #173/#178.
- A test or executable documentation check detects future SQL/schema drift.

Refs #178, #173, #211.

Part of epic #172.

iceboundrock:
## Summary

Follow-up from #211: the §5 "SQL Execution Template" in `docs/federated-query/design.md` could not execute against the runtime source contract. This PR aligns the sketch with the actual DuckDB template and adds two anti-drift guards so its table-function arity and projected columns cannot drift again.

### Doc changes (`docs/federated-query/design.md`)

- Every `postgres_scan` now uses the runtime 3-arg `(connection, schema, table)` form (`advanced_query_template_duckdb.go`); `flushed_at = 0` and `$PG_WHERE_CLAUSE` moved into `WHERE` predicates. Previously `:152`/`:206` passed `'flushed_at = 0'` as the table argument, `:209-213` used a 2-arg dynamic-`SELECT` form, and `:216` scanned EAV with 2 args.
- S3 projection aliases replaced with the runtime parquet columns from `buildS3Projection` (`duckdb_schema_projection.go:141-152`): `changed_at AS created_at`, `changed_at AS ver_ts`, `deleted_at AS deleted_ts` — the `ltbase_*` columns do not exist in parquet.
- §4.2 and §6.1 carried the same stale "inject SQL into postgres_scan's argument" wording; both now describe the predicate-in-WHERE pushdown the runtime performs.
- The #211 filter-after-LWW semantics are untouched: row_id semijoin, `source_tier_priority`, the `rn = 1` post-dedup logical filter, and tombstone-wins-then-drop. The section remains a deliberately simplified sketch (no `visible` CTE / keyset / pagination metadata).

### Anti-drift guards

- **`internal/federated/design_doc_sql_test.go`** (executable documentation): extracts the §5 fenced block by heading anchor, asserts the 3-arg scan form and the S3 aliases, then executes the sketch on in-memory DuckDB — scan calls map to same-named seed relations, `$S3_PATHS` points at a real parquet file — and asserts the #173/#178 scenarios: a newer non-matching version must not resurrect an older matching one, a tombstone must win dedup and then be dropped, dirty rows are served from the hot tier, and an already-flushed change_log row stays out of the dirty set.
- **`internal/sqlgen/duckdb_scan_contract_test.go`** (runtime side): renders `AdvancedQueryTemplateDuckDB` with production-shaped params and pins the 3-arg scan form and `flushed_at = 0` as a predicate — catching runtime drift the doc-side test cannot see.

Red-green verified: all three doc-side tests fail against the pre-fix document (arity, aliases, and relation mapping), and a mutation of the LWW `ORDER BY` makes the execution test fail, confirming the semantic assertions bite.

### Out of scope (pre-existing, noted during review)

- `internal/sqlgen/duckdb_template_renderer.go:170` — the legacy default `S3SourceSelect` fallback still injects the phantom `ltbase_*` aliases; the real federated path always overrides it from `SchemaProjection`, so it is legacy-only and untouched here.

## Acceptance criteria (#214)

- [x] The documented query uses valid runtime `postgres_scan` invocation forms.
- [x] Its S3 source columns and aliases match the runtime projection.
- [x] The example preserves filter-after-LWW semantics from #173/#178.
- [x] A test / executable documentation check detects future SQL/schema drift.

## Test plan

- `go test ./internal/federated ./internal/sqlgen -run 'TestDesignDocSQL|TestAdvancedTemplate_PostgresScanContract' -v` — all pass
- `make test` — full unit suite passes
- `make lint` — clean

Fixes #214. Refs #173, #178, #211. Part of epic #172.

### DIFF
--- a/internal/federated/design_doc_sql_test.go
+++ b/internal/federated/design_doc_sql_test.go
@@ -1,0 +1,257 @@
+package federated
+
+import (
+	"database/sql"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	_ "github.com/duckdb/duckdb-go/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// These tests are the #214 executable-documentation guard for the §5 "SQL
+// Execution Template" sketch in docs/federated-query/design.md. The sketch
+// calls itself the core logic of the Federated Query Engine, so it must stay
+// executable against the runtime source contract:
+//
+//   - every postgres_scan is the 3-arg (connection, schema, table) form used
+//     by AdvancedQueryTemplateDuckDB — never a 2-arg form or a predicate/query
+//     smuggled into the table argument;
+//   - the S3 projection uses the runtime parquet columns from
+//     buildS3Projection (changed_at AS created_at / ver_ts, deleted_at AS
+//     deleted_ts), not phantom ltbase_* columns;
+//   - the documented query preserves the #173/#178 filter-after-LWW
+//     semantics that PR #211 encoded.
+//
+// TestDesignDocSQL_ExecutesWithLWWFilterSemantics enforces the above
+// behaviorally: postgres_scan calls are mapped to same-named local relations
+// (a malformed call maps to nothing and fails the parse assertion), the
+// $S3_PATHS placeholder points at a real parquet file, and the whole block
+// runs on an in-memory DuckDB against seed data built for the #173 scenarios.
+
+const designDocRelPath = "../../docs/federated-query/design.md"
+
+// extractDesignDocSection5SQL returns the fenced SQL block of design.md §5.
+// It anchors on the section heading (not line numbers, which drift) and
+// captures the first ```sql fence after it.
+func extractDesignDocSection5SQL(t *testing.T) string {
+	t.Helper()
+
+	raw, err := os.ReadFile(filepath.Clean(designDocRelPath))
+	require.NoError(t, err, "design.md must be readable from internal/federated")
+	doc := string(raw)
+
+	heading := regexp.MustCompile(`(?m)^##\s+\*\*5\..*$`)
+	loc := heading.FindStringIndex(doc)
+	require.NotNil(t, loc, "design.md must contain the §5 heading (## **5. ...**)")
+
+	fence := regexp.MustCompile("(?is)```sql\\s*\n(.*?)```")
+	m := fence.FindStringSubmatch(doc[loc[1]:])
+	require.NotNil(t, m, "design.md §5 must contain a fenced SQL block")
+	return m[1]
+}
+
+// findPostgresScanCalls returns the raw argument text of every postgres_scan
+// invocation in sqlText, matching parentheses so multi-line and nested forms
+// are captured whole.
+func findPostgresScanCalls(sqlText string) []string {
+	const marker = "postgres_scan("
+	var calls []string
+	for i := 0; ; {
+		idx := strings.Index(sqlText[i:], marker)
+		if idx < 0 {
+			break
+		}
+		start := i + idx + len(marker)
+		depth := 1
+		j := start
+		for ; j < len(sqlText) && depth > 0; j++ {
+			switch sqlText[j] {
+			case '(':
+				depth++
+			case ')':
+				depth--
+			}
+		}
+		calls = append(calls, strings.TrimSpace(sqlText[start:j-1]))
+		i = j
+	}
+	return calls
+}
+
+// threeArgScanRe is the only valid documented postgres_scan shape: the
+// $PG_CONN placeholder plus two single-quoted identifiers (schema, table).
+// A predicate or a dynamic SELECT in any argument does not match.
+var threeArgScanRe = regexp.MustCompile(
+	`^\$PG_CONN\s*,\s*'([A-Za-z_][A-Za-z0-9_]*)'\s*,\s*'([A-Za-z_][A-Za-z0-9_]*)'$`)
+
+// TestDesignDocSQL_PostgresScanIsThreeArg pins the runtime scan arity: every
+// documented postgres_scan must be the 3-arg (connection, schema, table) form
+// of AdvancedQueryTemplateDuckDB.
+func TestDesignDocSQL_PostgresScanIsThreeArg(t *testing.T) {
+	sqlText := extractDesignDocSection5SQL(t)
+
+	calls := findPostgresScanCalls(sqlText)
+	require.NotEmpty(t, calls, "§5 must document postgres_scan usage")
+
+	for _, args := range calls {
+		require.Regexp(t, threeArgScanRe, args,
+			"every §5 postgres_scan must be the 3-arg ($PG_CONN, 'schema', 'table') runtime form; "+
+				"predicates such as flushed_at = 0 belong in WHERE, not in a scan argument")
+	}
+}
+
+// TestDesignDocSQL_S3ProjectionAliasesMatchRuntime pins the §5 S3 source
+// projection to the runtime parquet schema from buildS3Projection
+// (duckdb_schema_projection.go): one physical changed_at column feeds both
+// created_at and ver_ts, and deleted_at feeds deleted_ts.
+func TestDesignDocSQL_S3ProjectionAliasesMatchRuntime(t *testing.T) {
+	sqlText := extractDesignDocSection5SQL(t)
+
+	for _, alias := range []string{
+		"changed_at AS created_at",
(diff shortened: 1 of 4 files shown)


internal/federated/design_doc_sql_test.go
```diff
@@ -0,0 +1,257 @@
...
+// created_at and ver_ts, and deleted_at feeds deleted_ts.
+func TestDesignDocSQL_S3ProjectionAliasesMatchRuntime(t *testing.T) {
+	sqlText := extractDesignDocSection5SQL(t)
+
+	for _, alias := range []string{
+		"changed_at AS created_at",
+		"changed_at AS ver_ts",
+		"deleted_at AS deleted_ts",
+	} {
+		require.Contains(t, sqlText, alias,
```
iceboundrock: **P2 - S3 anti-drift guard is not linked to the runtime projection.** Issue #214 requires a test that detects future SQL/schema drift. This test hard-codes aliases, while the DuckDB smoke also creates Parquet with that assumed schema and the runtime contract test only checks `postgres_scan` arity. If `buildS3Projection` changes, both tests can remain green while section 5 is stale. Derive the expected projection from `BuildSchemaProjection(...).S3SourceSelect` (or compare a normalized projection) so a runtime change fails this guard.

internal/federated/design_doc_sql_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+	heading := regexp.MustCompile(`(?m)^##\s+\*\*5\..*$`)
+	loc := heading.FindStringIndex(doc)
+	require.NotNil(t, loc, "design.md must contain the §5 heading (## **5. ...**)")
+
+	fence := regexp.MustCompile("(?is)```sql\\s*\n(.*?)```")
+	m := fence.FindStringSubmatch(doc[loc[1]:])
+	require.NotNil(t, m, "design.md §5 must contain a fenced SQL block")
+	return m[1]
+}
```
iceboundrock, line 58: **P3 - Duplicated `postgres_scan` parser.** This parser is materially identical to `findRenderedScanCalls` in `internal/sqlgen/duckdb_scan_contract_test.go`. The duplicated 26-line helpers violate the `coding-standard.md` section 2 duplication rule and can drift when the scan syntax changes. Consolidate the helper if practical.

internal/federated/design_doc_sql_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+// TestDesignDocSQL_ExecutesWithLWWFilterSemantics executes the §5 sketch on a
+// real DuckDB engine and asserts the #173/#178 filter-after-LWW semantics it
+// documents. See seedDesignDocScenario for the per-row expectations.
+func TestDesignDocSQL_ExecutesWithLWWFilterSemantics(t *testing.T) {
+	docSQL := extractDesignDocSection5SQL(t)
+	parquetPath := filepath.Join(t.TempDir(), "base.parquet")
+
+	db, err := sql.Open("duckdb", "")
```
iceboundrock, line 241: **P3 - Ignored database cleanup errors.** `db.Close()` and `rows.Close()` return errors that this test discards, contrary to `docs/ai_review_guideline.md` section 3.1. Prefer `t.Cleanup` callbacks that assert cleanup errors with context.

iceboundrock:
## Assessment
The documented query currently matches the runtime `postgres_scan` form and S3 aliases, and the DuckDB example preserves post-LWW logical filtering.

## Spec (#214)
- Acceptance criteria 1-3 are implemented.
- Acceptance criterion 4 is partial: the S3 projection guard is not derived from the runtime projection, so it cannot detect all runtime-to-documentation schema drift. See the inline P2 comment.
- No material scope creep found.

## Standards
- P3: duplicated `postgres_scan` parser across the two new test files.
- P3: database cleanup errors are ignored in the executable-documentation test.
See the inline comments for locations and suggested fixes.

## Observations
The executable documentation test substitutes local DuckDB relations for `postgres_scan`; it therefore validates SQL shape and LWW behavior, but not actual PostgreSQL scanner predicate-pushdown behavior.

## Verification
Ran on PR commit `4cf29c7`:
`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/federated ./internal/sqlgen -run 'TestDesignDocSQL|TestAdvancedTemplate_PostgresScanContract' -v`

Result: all four targeted tests passed. The full unit suite and lint were not run in this review.

## Summary
1 P2 specification/coverage finding and 2 P3 standards findings. The P2 anti-drift coverage gap is the item to address before relying on the new guard.

internal/federated/design_doc_sql_test.go
iceboundrock: Fixed in 25131f8. The guard now extracts the §5 s3_source SELECT items and compares them (order-insensitively, minus the template-appended `source_tier_priority`) against `BuildSchemaProjection(...).S3SourceSelect` derived for the documented name/age/tag schema. Mutation-verified: renaming the `ver_ts` alias inside `buildS3Projection` turns this test red.

internal/federated/design_doc_sql_test.go
iceboundrock: Fixed in 25131f8. The parser is consolidated into `internal/sqlgen/sqlgentest.FindPostgresScanCalls` (a small test-support package, named after the `httptest` convention); both contract tests import it. The two `threeArgScanRe` regexes stay local because they pin different shapes — the doc side matches the `$PG_CONN` placeholder, the runtime side matches a rendered quoted connection string.

internal/federated/design_doc_sql_test.go
iceboundrock: Fixed in 25131f8. `db.Close()` is asserted via `t.Cleanup` and `rows.Close()` via a deferred `require.NoError`, both with context. Note the pre-existing tests in this package (e.g. `duckdb_namespace_regression_test.go`) still use bare `defer db.Close()`; left untouched as out of scope.

iceboundrock:
Review round 1 addressed in 25131f8 — all three findings fixed (see inline replies): the S3 projection guard is now derived from `BuildSchemaProjection` (mutation-verified), the scan parser is consolidated into `internal/sqlgen/sqlgentest`, and cleanup errors are asserted.

On the observation that the executable-doc test does not validate real PostgreSQL scanner pushdown: correct, and inherent to the unit level — `postgres_scan` requires the DuckDB postgres extension plus a live PostgreSQL connection, which in-memory DuckDB cannot provide. Real scanner behavior (including pushdown) is exercised by the `-tags=e2e` federated suite and the production harness, which run the actual engine against containers. This test's scope is deliberately the documented SQL's shape, executability, and LWW/filter semantics.

Issue #197, opened 2026-07-10
cdc: delta flusher swallows manifest update failures


iceboundrock:
Follow-up from PR #191 review (same pattern as the init fix in commit 2591bd2).

## Problem

The mainline delta flush logs and swallows manifest update failures (`internal/cdc/flusher.go` ~:394):

```go
if err := updateManifest(...); err != nil {
    e.logger.Sugar().Errorw("manifest update failed", "err", err)
    // Don't return - the flush succeeded, manifest is non-critical.
}
```

PR #191's review established for the init path that silently missing manifest entries make exported files invisible to manifest consumers (compaction) while the run reports success; `cdc.RunInit` now propagates these errors. The delta flusher has the exact same swallow.

## Considerations

Delta differs from init in one way: by this point rows are already marked flushed, so failing the pass cannot undo the flush — the design question is whether to (a) propagate the error so the run reports failure (operator alert, retry can re-append), (b) retry the manifest append with backoff, or (c) keep swallowing but surface a metric/alarm. Note manifest append is idempotent-unsafe on retry (duplicate entries) — check `manifest.AppendFile` semantics before choosing.

## Acceptance

- Manifest failures during delta flush are no longer silent: they either fail the run or raise an observable signal, per the chosen design.
- Failure-mode test coverage (failing store), mirroring `internal/cdc/init_test.go`.

Issue #181, opened 2026-07-10
[E2E] Deep pagination empty-page total count


iceboundrock:
Part of epic #172.

## Task

Verify that offset pagination preserves the correct `totalRecords` even when the requested page is empty (offset >= total). The production engine reads `COUNT(*) OVER()` from returned rows — DuckDB returns zero rows when offset exceeds total, causing `totalRecords=0`.

### 1. Offset equals total
- Seed N records → query with `offset=N, limit=10`
- Assert `data=[]` but `totalRecords=N`

### 2. Offset exceeds total
- Seed N records → query with `offset=N+10, limit=10`
- Assert `data=[]` but `totalRecords=N`

### 3. Multi-tier
- Same scenarios with data spread across Base, Delta, and Hot tiers

### 4. After filter
- Filter reduces visible count below offset
- Assert total reflects filtered count, not unfiltered

## Success criteria

- `totalRecords` always reflects the true matching count
- Empty pages never report `totalRecords=0` when matches exist

iceboundrock:
Scope note (from PR #228 review): the fix covers both engines, not only the DuckDB path named in the task description. Rationale: the success criteria here are path-agnostic, and in production the hybrid routing strategy sends small offset pages (`limit < 1000 && offset < 10000`) to the Postgres-only path — so the literal reproduction above (`offset=N, limit=10`) exercises the OLTP `runOptimizedQuery` reader, which had the same read-total-off-returned-rows defect. PR #228 fixes both and pins each with e2e scenarios (`hot_only_oltp` for the OLTP route).

PR #225, opened 2026-07-12
test(e2e): CDC flush failure boundary matrix & idempotency (#179)

### LINKED ISSUE #179: [E2E] CDC flush failure boundary matrix & idempotency
Part of epic #172.

## Task

Inject failures at every step of the CDC flush pipeline and verify that retry produces a correct, deduplicated, consistent result.

## Flush pipeline steps (inject failure at each):

1. Dirty ID selection / snapshot
2. DuckDB COPY (export to temp S3 object)
3. S3 CopyObject (temp → final random key)
4. DeleteObject (temp cleanup)
5. Postgres mark-flushed (`UPDATE change_log SET flushed_at`)
6. Manifest load
7. Manifest save

### Per-step failure + retry
For each step above:
1. Run flush with injected failure at that step
2. Assert the steps side effects are not visible (no partial commit)
3. Retry flush
4. Assert: correct row count, one final object per batch, no duplicate objects, no orphan temp files, manifest matches object inventory, change_log `flushed_at` consistent

### Known risky boundaries
- **Mark-flushed failure after CopyObject**: leaves a final S3 object while row remains dirty. Retry creates a second final object. Must not cause duplicate rows in federated query.
- **Manifest failure after mark**: manifest is treated as non-fatal. Verify manifest can self-repair on next flush.
- **Temp object cleanup failure**: verify temp objects do not affect correctness.

### Multi-batch failure
- Force byte-size splitting into multiple batches
- Fail a middle batch, retry
- Assert completed batches are NOT duplicated; only remaining dirty rows are re-exported

## Success criteria

- No data loss or duplication after any single failure + retry
- change_log, S3 objects, and manifest remain consistent
- Federated query result is identical to no-failure baseline

iceboundrock:
Closes #179 (epic #172, Phase 2).

Injects a failure at every step of the CDC flush pipeline and verifies that retry produces a correct, deduplicated, consistent result — with **zero production-code changes**: every fault is injected from outside `internal/cdc`/`internal/manifest`/`internal/federated`.

## Fault matrix

| # | Pipeline step | Injection | Scenario |
|---|--------------|-----------|----------|
| 1 | Dirty ID selection | Config poison: `PGHost/PGPort = 127.0.0.1:1` | `TestFlushFaultDirtySelection` — zero side effects, clean retry converges |
| 2 | DuckDB COPY → tmp | Config poison: `S3Endpoint = 127.0.0.1:1` (httpfs unreachable, fails fast ~3s) | `TestFlushFaultDuckExport` — zero side effects, clean retry converges |
| 3 | S3 CopyObject tmp→final | `FaultInjectingS3{Op: S3OpCopy}` | `TestFlushFaultCopyObject` — tmp orphan only, rows stay dirty, retry converges |
| 4 | S3 DeleteObject (tmp cleanup) | `FaultInjectingS3{Op: S3OpDelete, KeyContains: "/_tmp/"}` | `TestFlushFaultTempCleanup` — flush **succeeds** (swallow is by design, `helpers.go:165-167`), orphan invisible to reads |
| 5 | Postgres mark-flushed | `ExecSQL` poison trigger `BEFORE UPDATE OF flushed_at ON change_log` | `TestFlushFaultMarkFlushed` — final object already landed, retry creates a second identical one, LWW dedups |
| 6 | Manifest load | `FaultInjectingS3{Op: S3OpGet, KeyContains: "manifest/"}` | `TestFlushFaultManifestLoad` — corrected #197 contract (below) |
| 7 | Manifest save | `FaultInjectingS3{Op: S3OpPut, KeyContains: "manifest/"}` | `TestFlushFaultManifestSave` — corrected #197 contract (below) |
| — | Multi-batch middle chunk | `EstimatedRowBytes=1, MaxBatchBytes=3` (9 rows → 3 chunks) + `SkipMatches: 1` on CopyObject | `TestFlushFaultMultiBatchMiddleChunk` — committed chunk 1 not re-exported; `assertNoRowExportedTwice` guards physical duplication |

Every scenario carries a positive control (`faulty.Injected() == 0 → Fatal`) and closes with the fault-independent event-fold oracle (`AssertQueryMatches`) — the issue's "identical to no-failure baseline" criterion.

New harness surface: `FaultInjectingS3` (implements `cdc.S3FullClient`; injected error text deliberately avoids the not-found phrases `manifest.LoadOrCreate` treats as "fresh manifest" — unit-pinned) and `Env.RunFlushWith(ctx, FlushOverrides{S3, Config, DryRun})`, which returns a best-effort non-nil report even when the flush fails so partial side effects can be asserted. `RunFlush`/`RunFlushDry` semantics unchanged.

## Spec adjudications (issue text vs. current contracts)

1. **"Manifest failure is non-fatal / self-repairs on next flush" is superseded by #197 (PR #202).** The corrected contract is pinned instead: the run fails with the orphaned final key embedded in the error, rows are already marked flushed, and the retry is a strict no-op that does **not** re-append the manifest entry (blind `AppendFile` retry would duplicate entries). Reconciliation is #203's scope; federated reads are unaffected because the read path never consults the manifest.
2. **"No orphan temp files" — deferred to #226 (review round 1 decision).** A step-3 failure orphans its tmp object permanently (retry uses fresh UUIDs; `CopyTmpToFinal` deletes only its own key), and step-4 delete failures are swallowed. Review adjudicated that the #197/#203 supersession does **not** extend to this criterion, so restoring the invariant is production work tracked in #226 (tmp GC design: flush holds the schema advisory lock but `RunInit` does not and shares `_tmp/`, so a naive sweep is unsafe). Until #226 lands, the tests pin the current behavior and assert correctness is unaffected: the production glob's `*` does not cross `/`, so `_tmp/` is excluded from reads (`production/query.go:52-56`), and orphans never enter the manifest.
3. **"One final object per batch" is intentionally violated in the mark-flushed scenario.** As the issue itself acknowledges, retry after a step-5 failure creates a second final object with identical `(row_id, ver_ts)` content; the assertion is at the query layer (LWW `rn=1` dedups, no duplicate rows).
4. **`production/doc.go` consumer-map drift**: the `#179 flush thresholds` label predated this issue's actual scope and is fixed here. The `#181`/`#189` rows have similar drift (actual #181 = deep pagination, #189 = schema evolution per epic #172) — left untouched in this PR.

## Review round 1 (c46a291)

- Bare error returns in `runFlush`/`RunFlushWith` pre-capture and the `FaultInjectingS3` forwarders now wrapped with operation + key context; `S3Op` constants godoc'd.
- Multi-batch scenario now asserts **row-level** `flushed_at` consistency: after the fault the flushed set equals the chunk-1 parquet contents (3/6 split), and the retry files contain exactly the previously-dirty 6 rows — pinning "only remaining dirty rows are re-exported" at row granularity.
- Orphan-tmp adjudication replaced by #226 (see above).

## Test notes

- One characterization finding en route: a federated query against a schema with zero parquet files errors on the empty `read_parquet` glob, so S1's pre-flush baseline uses `PreferHot: true` (established pattern: `boundary_roundtrip_e2e_test.go`). Fault and convergence assertions are untouched.
- Follow-up candidates (non-blocking, from final review): a `requireReport` nil-guard for `RunFlushWith` fault call sites (pre-run capture failure currently panics instead of printing a diagnostic — still a test failure, never a false green); `loadManifests` swallows `manifest.Load` errors (pre-existing; relevant to #203's reconciliation tests).

## Verification

- `make lint` (golangci-lint v1.64.8) — clean
- `make test` — all packages ok
- Full production e2e suite (`-tags=e2e`, Docker): 54 top-level tests PASS in ~30s — the 8 new `TestFlushFault*` tests co-exist with the pre-existing suite under the shared cluster + `t.Parallel()`

### DIFF
--- a/internal/e2e_harness/production/README.md
+++ b/internal/e2e_harness/production/README.md
@@ -131,6 +131,14 @@ where cdc-init bootstraps base files before federated reads. Use
   unflushed counts, new S3 objects, and parsed manifests. Note the
   mainline dry-run still writes parquet objects — it skips `flushed_at`
   marking and manifest updates (#180).
+- `RunFlushWith(FlushOverrides{S3, Config, DryRun})` runs one flush pass
+  with per-run overrides and reports observable state *even when the run
+  fails* (report is non-nil unless the pre-run capture fails), so
+  failure-boundary tests can assert partial side effects alongside the
+  error. Pair it with `FaultInjectingS3` — a decorator over the real S3
+  client that fails calls matching an `S3Fault{Op, KeyContains,
+  SkipMatches}` — to break exactly one step of the flush pipeline and pin
+  the failure matrix (#179).
 - `RunInit` wraps `cdc.RunInit` (base file export + manifest base entries).
   Base keys are deterministic (`<prefix>/<id>/<minRowID>_<maxRowID>.parquet`),
   and init replaces the manifest's base tier with each run's output, so
@@ -195,7 +203,7 @@ On failure (or `KEEP_E2E_ENV=1`) the Env writes
 | #174 full-type coverage | `e2e_wide` + `FullTypeProfile` |
 | #176 init handoff | `RunInit` + `RunFlush` (`init_handoff_e2e_test.go`, `init_delete_e2e_test.go`, `init_rerun_e2e_test.go`, `init_concurrent_e2e_test.go`) |
 | #175 lifecycle sequences | `Event` model + `ApplyEvents` + `InjectRestore` + `RestartPostgres` (`lifecycle_e2e_test.go`, `restart_e2e_test.go`) |
-| #179 flush thresholds | `WithFlushThresholds` + `FlushReport` |
+| #179 flush failure matrix | `FaultInjectingS3` + `RunFlushWith` + `ExecSQL` |
 | #180 dry-run semantics | `RunFlushDry` |
 | #181 failure injection | `ExecSQL` |
 | #185 circuit breaker | `WithBreaker` |
--- a/internal/e2e_harness/production/cdc.go
+++ b/internal/e2e_harness/production/cdc.go
@@ -52,7 +52,7 @@ type FlushOverrides struct {
 func (e *Env) runFlush(ctx context.Context, dryRun bool) (*FlushReport, error) {
 	report, err := e.RunFlushWith(ctx, FlushOverrides{DryRun: dryRun})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("run flush (dry=%t): %w", dryRun, err)
 	}
 	return report, nil
 }
@@ -64,11 +64,11 @@ func (e *Env) runFlush(ctx context.Context, dryRun bool) (*FlushReport, error) {
 func (e *Env) RunFlushWith(ctx context.Context, ov FlushOverrides) (*FlushReport, error) {
 	before, err := e.countUnflushed(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("capture pre-flush change_log state: %w", err)
 	}
 	keysBefore, err := e.listS3Keys(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("capture pre-flush s3 listing: %w", err)
 	}
 
 	cfg := e.CDC
--- a/internal/e2e_harness/production/cdc.go
+++ b/internal/e2e_harness/production/cdc.go
@@ -2,6 +2,7 @@ package production
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 
@@ -40,7 +41,27 @@ func (e *Env) RunFlushDry(ctx context.Context) (*FlushReport, error) {
 	return e.runFlush(ctx, true)
 }
 
+// FlushOverrides customizes a single flush pass for failure-injection tests
+// (#179). Zero-value fields fall back to the Env defaults RunFlush uses.
+type FlushOverrides struct {
+	S3     cdc.S3FullClient // nil: e.Cluster.S3
+	Config *cdc.CDCConfig   // nil: e.CDC
+	DryRun bool
+}
+
 func (e *Env) runFlush(ctx context.Context, dryRun bool) (*FlushReport, error) {
+	report, err := e.RunFlushWith(ctx, FlushOverrides{DryRun: dryRun})
+	if err != nil {
+		return nil, err
+	}
+	return report, nil
+}
+
+// RunFlushWith executes one real CDC flush pass with per-run overrides.
+// Unlike RunFlush it reports observable state even when the run fails: the
+// report is always non-nil unless the pre-run state capture itself fails, so
+// failure-boundary tests can assert partial side effects alongside the error.
+func (e *Env) RunFlushWith(ctx context.Context, ov FlushOverrides) (*FlushReport, error) {
 	before, err := e.countUnflushed(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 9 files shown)


iceboundrock (commented):
## Standards

1. **[P2]** `internal/e2e_harness/production/cdc.go:55,67,71` returns bare errors in newly extracted or modified code. Wrap them with operation context per `coding-standard.md` section 3.
2. **[P2]** `internal/e2e_harness/production/faults.go:76,83,90,97` propagates underlying S3 errors without operation or key context.
3. **[P3]** `internal/e2e_harness/production/faults.go:19-22` has no Godoc comments for the exported `S3Op*` constants.

**Observation:** The four forwarding methods duplicate code, but their distinct interface signatures make further abstraction a judgment call rather than a required change.

## Spec

1. **[P1]** Issue #179 requires "no orphan temp files," but this PR intentionally asserts permanent tmp-object orphans for CopyObject and temp-cleanup failures in `internal/e2e_harness/production/cdc_fault_matrix_e2e_test.go:81-118,121-158`. The production cleanup error remains swallowed in `internal/cdc/helpers.go:165-167`, so a retry cannot restore the stated invariant.
2. **[P2]** The matrix verifies aggregate unflushed counts, but not the exact `change_log` rows and `flushed_at` markers. This leaves the issue requirements "flushed_at consistent" and "only remaining dirty rows are re-exported" partially unverified, especially in `internal/e2e_harness/production/cdc_fault_multibatch_e2e_test.go:40-42,51-52`.

## Assessment

The changed manifest expectation is supported by downstream work: #197 requires delta manifest failures to be observable, and #203 explicitly makes manual reconciliation the recovery path. That supersession does not apply to #179s temporary-object requirement.

## Verification

`make test-e2e-production` passed at PR head `17607f6` in 35.6 seconds.

## Summary

Three standards violations and two specification gaps. The permanent orphan-temp behavior is the blocking issue.

iceboundrock:
Review round 1 addressed in c46a291. Per item:

**Standards 1 (cdc.go bare errors)** — Fixed. `runFlush` wraps with `run flush (dry=%t)`, the two `RunFlushWith` pre-capture returns wrap with `capture pre-flush change_log state` / `capture pre-flush s3 listing`.

**Standards 2 (faults.go pass-through errors)** — Fixed. All four forwarders now wrap real client errors with `s3 <Op> "<key>": %w`. Note the wrap is `%w`-transparent, so `manifest.LoadOrCreate`'s not-found substring detection (which the S7 save-failure path relies on to treat a genuine 404 as "fresh manifest") still sees the underlying message — verified by re-running both manifest scenarios.

**Standards 3 (S3Op godoc)** — Fixed. Each constant documents which flush-pipeline step it matches.

**Observation (forwarder duplication)** — Keeping as-is. The four methods differ in input/output types on every line that matters; a generics/reflection abstraction would cost more than the ~6 duplicated lines per method it removes.

**Spec 1 [P1] (orphan temp files)** — Resolved per your call by splitting the production work into #226 rather than expanding this PR. The invariant genuinely cannot be restored test-side: cleanup requires production design — flush holds the per-schema advisory lock but `RunInit` does not and shares the `_tmp/` layout, so the obvious flush-start sweep would delete a concurrent init's in-flight objects, and `cdc.S3ObjectClient` has no List operation. #226 carries the design constraints, three candidate approaches, and an acceptance item to flip these tests from pinning orphan persistence to asserting the restored invariant. The test comments and the PR-body adjudication now point at #226 instead of calling the orphans by-design.

**Spec 2 [P2] (row-level flushed_at)** — Fixed. The multi-batch scenario now asserts at row granularity: after the fault, the `flushed_at != 0` set must equal the chunk-1 parquet contents exactly (3/6 split); after retry, the retry files must contain exactly the previously-dirty 6 rows and the split must be 9/0 — this pins both "flushed_at consistent" and "only remaining dirty rows are re-exported" beyond aggregate counts (new helpers: `fetchChangeLogRowIDs`, `fetchParquetRowIDs`, `assertSameRowIDs`). For the single-batch scenarios the aggregates were already exact: each test runs in an isolated per-test database with a fixed seed count, so `UnflushedAfter == 0` (or `== 3`) over 3 seeded rows determines every row's marker; I left those as-is.

Verification after the fixes: all 8 `TestFlushFault*` scenarios re-run green (`-tags=e2e`, 17.7s), `make lint` and `make test` clean.

Issue #210, opened 2026-07-11
cdc-init exports created_at as LWW version timestamp: backfilled base loses to previously flushed updates (deterministic stale reads) and ties with same-timestamp versions


iceboundrock:
Found during the PR #209 review round for #177 (part of epic #172), while verifying review finding https://github.com/Lychee-Technology/forma/pull/209#pullrequestreview-4676717407.

## Mechanism

- Base export (cdc-init) stamps the LWW version timestamp from creation time: `m.ltbase_created_at AS changed_at` (`internal/cdc/init_exporter.go:35`; contract pinned by `full_type_parquet_e2e_test.go` / `export_sql_shared_test.go` in #174's review round).
- Delta export (cdc-flush) stamps the mutation time: `cl.changed_at AS changed_at` (`internal/cdc/duckdb_exporter.go:153`).
- Federated dedup ranks `ORDER BY ver_ts DESC, source_tier_priority DESC, deleted_ts DESC, row_id ASC` (`internal/sqlgen/advanced_query_template_duckdb.go:80`), and ALL parquet rows share `source_tier_priority = 1` — base and delta files are one glob scan with no per-file tier distinction.

Consequence: for any row that was **updated after creation**, the base file init exports carries a version timestamp **older than** (or equal to) every delta version of that row, even though init exported the *newest* attribute state.

## Failure mode 1 — deterministic stale read (confirmed by probe)

Sequence (all through real production drivers): create(T1) → update v2(T2) → **cdc-flush** (delta: v2 attrs @ ver_ts=T2) → update v3(T3) → **cdc-init** (base: v3 attrs @ ver_ts=T1) → onboarding change_log cleanup.

Result: delta's stale v2 (@T2) deterministically beats base's fresh v3 (@T1) in LWW. Probe evidence (production E2E harness, oracle diff):

```json
"attribute_mismatches": [
  {"attr": "count", "expected": 500000, "actual": 400000},
  {"attr": "title", "expected": "fresh-v3", "actual": "stale-v2"}
]
```

Any onboarding that runs cdc-init over rows with previously flushed post-create updates, then clears change_log per the onboarding contract (#176), serves stale attributes on the federated path until the next mutation of the affected row.

## Failure mode 2 — undefined tie (confirmed by probe)

Sequence: create(T1) → cdc-flush (delta: v1 @ T1) → update v2(T2) → cdc-init (base: v2 @ T1) → cleanup.

[...]

PR #237, opened 2026-07-13
test(e2e): concurrent mutation during flush snapshot (#182)

### LINKED ISSUE #182: [E2E] Concurrent mutation during flush snapshot
Part of epic #172.

## Task

Verify that rows updated concurrently with a CDC flush are correctly handled: updates after the snapshot must remain dirty (not prematurely marked flushed), and queries must always see the latest version.

### 1. Update after snapshot, before export
- Pause flush after dirty-ID selection
- Update a selected row in Postgres
- Resume flush
- Assert: updated rows `flushed_at` remains 0 (still dirty)
- Assert: federated query sees the NEW value (hot shadow), not the exported old value

### 2. Delete after snapshot
- Pause flush after dirty-ID selection
- Soft-delete a selected row
- Resume flush
- Assert: row stays dirty, deletion is visible in hot tier

### 3. Insert after snapshot
- Pause flush after dirty-ID selection
- Insert a new row
- Resume flush
- Assert: new rows `flushed_at` is 0, visible in hot tier

### 4. Snapshot guard correctness
- The `changed_at <= snapshot` guard in mark-flushed must be tested against a real concurrent write

## Success criteria

- Only rows with `changed_at <= snapshot` are marked flushed
- Post-snapshot mutations remain dirty and are correctly visible
- Retry flush after concurrent mutation yields exactly one visible version

iceboundrock:
Closes #182 (epic #172, Phase 3).

Pins the CDC snapshot-guard contract end to end: a real flush is held open inside each race window while the test mutates rows through the real EntityManager, then the guard's row-level effects, tier visibility, and retry convergence are asserted against the event-fold oracle. The only production-code change is the `BeforeExportHook` test seam added in review round 2 (nil in production).

## Mechanism

New `PausingS3` decorator (mirrors `FaultInjectingS3`/`RecordingS3`, injected via `Env.RunFlushWith(FlushOverrides{S3})` from #179): the first matching `CopyObject` closes a `reached` channel and blocks until `Resume()`. `CopyObject` fires after dirty-ID selection, snapshot capture, and the DuckDB export, but before `MarkFlushedIDsAtSnapshot` — exactly the window the `changed_at <= snapshot` guard (`internal/cdc/helpers.go`) protects. The paused call also unblocks on context cancellation, and the orchestration helper releases it via `t.Cleanup`, so no failure mode hangs the suite.

## Scenario map (issue → test)

| Issue scenario | Test | Pinned contract |
|---|---|---|
| 1. Update after snapshot, before export | `TestConcurrentFlushSnapshot/UpdateBeforeExport` *(review round 2)* | Mutation inside the selection→export window via `cdc.CDCConfig.BeforeExportHook`: the row is neither exported (delta holds exactly the 3 flushed siblings) nor marked flushed; `victim.changed_at <= snapshot < update.changed_at` asserted against the actual snapshot the hook exposes |
| 1. Update after export (harder window) | `TestConcurrentFlushSnapshot/UpdateAfterSnapshot` | Updated row keeps `flushed_at = 0` while its 3 siblings flush (row-level split via `fetchChangeLogRowIDs`); federated read serves the hot value; the exported stale value is unreachable even by direct filter (anti-join) |
| 2. Delete after snapshot | `.../DeleteAfterSnapshot` | Concurrent tombstone stays dirty; deletion immediately visible; the live copy in the just-written delta never resurrects the row |
| 3. Insert after snapshot | `.../InsertAfterSnapshot` | New row stays dirty (never in the selected batch — distinct mechanism from the guard, doc-commented), absent from the parquet, visible via hot tier |
| 4. Snapshot guard vs real concurrent write | row-level split assertions in scenarios 1–2 | Exactly the mutated row (which WAS in `batchIDs`) stays dirty = "only rows with `changed_at <= snapshot` are marked flushed"; `assertStrictlyNewer` guards the degenerate equal-ver_ts tie (#210); unit twin: `TestMarkFlushedIDsAtSnapshot_SkipsUpdatedRows` |

Retry convergence (issue success criterion 3): scenarios 1–2 re-run a clean flush — all dirty rows drain (`UnflushedAfter == 0`), the victim lands in a second delta, and the oracle-checked probes confirm exactly one visible version (now via LWW on ver_ts instead of the anti-join). `assertNoRowExportedTwice` is deliberately **not** used across the two flushes: the victim legitimately appears in both deltas at different ver_ts (that helper is documented create-only).

## Spec adjudications

1. **Both mutation windows are now exercised (review round 2 superseded the original adjudication).** The `*AfterSnapshot` scenarios pause at `CopyObject` — the *harder* window, where the parquet already holds the stale value and only the mark-flushed guard saves the row (the physical shape the issue's assertion describes: "sees the NEW value, not the **exported old value**"). The literal "before export" window has no external seam (the DuckDB export bypasses the injected S3 client; Postgres lock barriers are nondeterministic because `postgres_scan` reads each table with its own statement snapshot), so review round 2 added a minimal production seam: `cdc.CDCConfig.BeforeExportHook`, nil in production, unit-pinned both ways (ordering; a hook error aborts before any side effect). `UpdateBeforeExport` drives it through the existing `FlushOverrides{Config}` injection.
2. **The paused flush's manifest tracks a delta whose parquet physically contains the victim's stale value (row-id bounds exclude it).** Expected, doc-commented in the test: reads never consult the manifest, the dirty row_id anti-joins the stale copy out, and after retry LWW suppresses it.
3. **The federated-harness flush reimplementation (`internal/e2e_harness/federated/cdc.go`) is not used** — its hand-rolled mark-flushed has no snapshot guard; only the production harness drives the real `cdc.Runner.RunOnce`.

All scenarios passed on first run — the dual guard (export CTE + mark-flushed) behaves as designed, so this lands as characterization with no production bug fix.

## Verification

- `go test -v ./internal/e2e_harness/production/ -tags=e2e -run TestConcurrentFlushSnapshot -timeout=10m` — 3/3 PASS (~6s each, shared cluster)
- same with `-race` — clean
- `make lint` (golangci-lint v1.64.8) — clean
- `make test` — all packages ok
- full production e2e suite (`-tags=e2e`, Docker) — PASS

## Review round 1 (49a6187)

Two-axis review (standards + spec) surfaced three items, all applied:

- `PausingS3` delegate methods now wrap `Inner` errors with op+key context (coding-standard §3, matching `FaultInjectingS3`); the unused `KeyContains` field is removed.
- `DeleteAfterSnapshot` gained the same-filter positive control its zero-row probes lean on (mirrors `UpdateAfterSnapshot`).
- The post-flush/converged count probes now also assert `Plan.Routing.UseDuckDB` (`assertFederatedRowCount`): the flushed siblings are only served by the parquet side (the hot scan reads `flushed_at = 0` rows), so the counts now prove the cold delta was genuinely consulted rather than the whole page coming from Postgres.

## Review round 2 (81d4dcb, 6b8ccd4)

All four findings applied:

- **High (spec):** new `cdc.CDCConfig.BeforeExportHook` seam + `UpdateBeforeExport` scenario covering the literal selection→export window — the mutated row is neither exported nor marked flushed, with `changed_at` ordering asserted against the actual snapshot. Adjudication 1 above rewritten accordingly.
- **Medium (assertion strength):** parquet checks upgraded to record level — exactly one stale live v1 record in the paused flush's delta (`assertLiveParquetRow`), exactly one v2 record in the update retry delta, exactly one all-NULL tombstone in the delete retry delta (`assertTombstoneParquet`).
- **Standards (×2):** `PausingS3` delegates wrap the interceptor error with op+key context (single-layered — `intercept` returns bare `ctx.Err()`); `runPausedFlush` wraps the flush error.

Verification: unit `TestExecuteBatch_BeforeExportHook*` green; 4/4 e2e scenarios green incl. `-race`; `make lint` / `make test` clean; full production e2e suite 34s PASS.

### DIFF
--- a/internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
+++ b/internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
@@ -124,6 +124,24 @@ func assertPausedFlushSplit(ctx context.Context, t *testing.T, env *Env, wide Sc
 	return finals
 }
 
+// assertFederatedRowCount is assertRowCount plus a routing check: the flushed
+// sibling rows are only served by the parquet side (the hot scan reads
+// flushed_at = 0 rows), so the count is meaningful evidence that the cold
+// delta was consulted only if the query actually routed to DuckDB.
+func assertFederatedRowCount(ctx context.Context, t *testing.T, env *Env, name string, q Query, want int) {
+	t.Helper()
+	res := env.AssertQueryMatches(ctx, q)
+	if res == nil {
+		return
+	}
+	if len(res.Records) != want {
+		t.Fatalf("%s returned %d rows, want %d", name, len(res.Records), want)
+	}
+	if !res.Plan.Routing.UseDuckDB {
+		t.Errorf("%s did not route to duckdb: %+v", name, res.Plan.Routing)
+	}
+}
+
 // testUpdateAfterSnapshot is issue #182 scenarios 1 and 4: a row updated
 // between snapshot capture and mark-flushed keeps flushed_at = 0, the
 // federated read serves the hot (new) value, and the exported stale value is
@@ -170,7 +188,7 @@ func testUpdateAfterSnapshot(ctx context.Context, t *testing.T) {
 	// Visibility: the dirty hot version shadows the exported copy. The oracle
 	// expects the victim to carry v2; the stale v1 must be unreachable even by
 	// direct filter (anti-join evicts it before the filter runs).
-	assertRowCount(ctx, t, env, "post-flush rows", Query{Schema: wide, Limit: 10}, 4)
+	assertFederatedRowCount(ctx, t, env, "post-flush rows", Query{Schema: wide, Limit: 10}, 4)
 	assertRowCount(ctx, t, env, "hot value reachable",
 		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "hot-00"}}, Limit: 10}, 1)
 	assertZeroRows(ctx, t, env, "exported stale value unreachable",
@@ -186,7 +204,7 @@ func testUpdateAfterSnapshot(ctx context.Context, t *testing.T) {
 	assertSameRowIDs(t, "retry parquet vs victim",
 		fetchParquetRowIDs(ctx, t, env, retryFinals), map[string]bool{victim.RowID.String(): true})
 	assertManifestDeltaPaths(t, retry.Manifests, wide, append(finals, retryFinals...))
-	assertRowCount(ctx, t, env, "converged rows", Query{Schema: wide, Limit: 10}, 4)
+	assertFederatedRowCount(ctx, t, env, "converged rows", Query{Schema: wide, Limit: 10}, 4)
 	assertRowCount(ctx, t, env, "converged hot value",
 		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "hot-00"}}, Limit: 10}, 1)
 	assertZeroRows(ctx, t, env, "converged stale value",
@@ -206,6 +224,10 @@ func testDeleteAfterSnapshot(ctx context.Context, t *testing.T) {
 	mustApplyEvents(ctx, t, env, "seed creates", creates...)
 	victim := creates[0]
 	assertRowCount(ctx, t, env, "pre-flush rows", Query{Schema: wide, Limit: 10, PreferHot: true}, 4)
+	// Same-filter positive control for the zero-row probes below: the victim
+	// is reachable by this value until the concurrent delete lands.
+	assertRowCount(ctx, t, env, "pre-flush victim by v1",
+		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "open-00"}}, Limit: 10, PreferHot: true}, 1)
 
 	report, err := runPausedFlush(ctx, t, env, func() {
 		del := DeleteEvent(wide, victim.RowID)
@@ -220,7 +242,7 @@ func testDeleteAfterSnapshot(ctx context.Context, t *testing.T) {
 
 	// Deletion visible now: 3 rows; the exported live copy must not resurrect
 	// the victim (dirty tombstone anti-joins it out before the filter).
-	assertRowCount(ctx, t, env, "post-flush rows", Query{Schema: wide, Limit: 10}, 3)
+	assertFederatedRowCount(ctx, t, env, "post-flush rows", Query{Schema: wide, Limit: 10}, 3)
 	assertRowCount(ctx, t, env, "survivor reachable",
 		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "open-01"}}, Limit: 10}, 1)
 	assertZeroRows(ctx, t, env, "deleted row unreachable by exported value",
@@ -229,7 +251,7 @@ func testDeleteAfterSnapshot(ctx context.Context, t *testing.T) {
 	// Retry flushes the tombstone; LWW + deleted_ts must keep both cold copies
 	// of the victim invisible.
 	runCleanRetry(ctx, t, env)
-	assertRowCount(ctx, t, env, "converged rows", Query{Schema: wide, Limit: 10}, 3)
+	assertFederatedRowCount(ctx, t, env, "converged rows", Query{Schema: wide, Limit: 10}, 3)
 	assertZeroRows(ctx, t, env, "deleted row stays unreachable",
 		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "open-00"}}, Limit: 10})
 }
@@ -276,12 +298,12 @@ func testInsertAfterSnapshot(ctx context.Context, t *testing.T) {
 	assertManifestDeltaPaths(t, report.Manifests, wide, finals)
 
 	// The new row is visible through the hot tier right away.
-	assertRowCount(ctx, t, env, "post-flush rows", Query{Schema: wide, Limit: 10}, 5)
+	assertFederatedRowCount(ctx, t, env, "post-flush rows", Query{Schema: wide, Limit: 10}, 5)
 	assertRowCount(ctx, t, env, "inserted row reachable",
 		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "new-99"}}, Limit: 10}, 1)
 
 	runCleanRetry(ctx, t, env)
-	assertRowCount(ctx, t, env, "converged rows", Query{Schema: wide, Limit: 10}, 5)
+	assertFederatedRowCount(ctx, t, env, "converged rows", Query{Schema: wide, Limit: 10}, 5)
 	assertRowCount(ctx, t, env, "converged inserted row",
 		Query{Schema: wide, Filters: []Filter{{Attr: "title", Value: "new-99"}}, Limit: 10}, 1)
 }
--- a/internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
+++ b/internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
@@ -1,0 +1,287 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+// TestConcurrentFlushSnapshot pins issue #182: rows mutated concurrently with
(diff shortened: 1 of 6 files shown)


internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
```diff
@@ -0,0 +1,309 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+// TestConcurrentFlushSnapshot pins issue #182: rows mutated concurrently with
+// a CDC flush must stay dirty and queries must always see the latest version.
+//
+// Mechanism: PausingS3 holds the flush open at CopyObject — after dirty-ID
```
iceboundrock: **Spec gap (high):** #182 requires a real pause immediately after dirty-ID selection for the "Update after snapshot, before export" scenario. `CopyObject` occurs after `ExportSnapshotToTmp`, so this only exercises the post-export `MarkFlushedIDsAtSnapshot` guard. Export SQL unit coverage does not meet that explicit E2E requirement. Please add a pre-export barrier and assert the mutation is neither exported nor marked flushed.

internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	finals := assertPausedFlushSplit(ctx, t, env, wide, report, 3, victim.RowID.String())
+	// The exported parquet physically contains all 4 selected rows — the
+	// victim with its stale v1 value.
+	wantExported := map[string]bool{}
+	for _, c := range creates {
+		wantExported[c.RowID.String()] = true
+	}
+	assertSameRowIDs(t, "paused-flush parquet vs selected batch",
```
iceboundrock, line 301: **Test-strength gap (medium):** `fetchParquetRowIDs` reduces the parquet contents to `SELECT DISTINCT row_id`, so this assertion proves only ID membership. It cannot prove that the paused flush wrote exactly one stale version or that the retry wrote the new version. Query the payload/version timestamp too, and assert one stale v1 record in the first delta and one v2 record in the retry delta.

internal/e2e_harness/production/pausing.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("paused s3 %s %q: %w", op, key, ctx.Err())
+	}
+}
+
+func (p *PausingS3) CopyObject(ctx context.Context, in *s3.CopyObjectInput, optFns ...func(*s3.Options)) (*s3.CopyObjectOutput, error) {
+	key := aws.ToString(in.Key)
+	if err := p.intercept(ctx, S3OpCopy, key); err != nil {
+		return nil, err
```
iceboundrock: **Standards:** this returns the interceptor error bare (also lines 92, 104, and 116). `coding-standard.md` §3 and `AGENTS.md:69` require errors to be wrapped with operation context. Wrap this boundary, e.g. `fmt.Errorf("pause CopyObject %q: %w", key, err)`.

iceboundrock:
## Assessment

Changes requested before merge: the suite provides useful coverage of mutations after export but before `MarkFlushedIDsAtSnapshot`, yet it does not meet issue #182's explicitly required real pre-export race.

## Findings

1. **High:** The `CopyObject` pause is after `ExportSnapshotToTmp`, not immediately after dirty-ID selection. Add a real pre-export barrier and prove a concurrent update is neither exported nor marked flushed.
2. **Medium:** The parquet checks collapse data to distinct row IDs. Assert each delta's payload and `ver_ts`, including exactly one stale v1 record before retry and the v2 record after retry.

## Spec

Issue #182 requires: “Pause flush after dirty-ID selection” and, for scenario 1, “Update after snapshot, before export.” The PR deliberately substitutes a post-export pause plus export-SQL unit coverage. That validates a harder mark-flushed/anti-join case, but does not substitute for the required E2E scenario. No unrelated production-code scope creep was found.

## Standards

`coding-standard.md` §3 and `AGENTS.md:69` require contextual error wrapping. `PausingS3` returns interceptor errors bare at lines 80, 92, 104, and 116; `runPausedFlush` also returns `out.err` bare at line 87. Wrap these helper-boundary errors with their operation context.

## Observations

- The row-level dirty/flushed split and DuckDB-routing checks are valuable positive controls for the post-export contract.
- The current ID-set parquet helper cannot substantiate the comment that the first delta physically carries the victim's stale v1 value, nor distinguish duplicated or mixed-version records.

## Verification

- `git diff --check df0e45eeaf166872694179529ed072bef869dff0...49a61876b1d90faebd4f55b281057f642cc2dc15` completed without whitespace errors.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -v ./internal/e2e_harness/production/ -tags=e2e -run TestConcurrentFlushSnapshot -timeout=10m` passed all three scenarios in 7.665s.

## Summary

Two standards findings, one specification finding, and one additional assertion-strength finding. The blocking concern is the missing real pre-export concurrency test.

internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
iceboundrock: Implemented in 81d4dcb + 6b8ccd4. There is no external seam between selection and export (the DuckDB export bypasses the injected S3 client, and Postgres lock barriers are nondeterministic because postgres_scan reads each table with its own statement snapshot), so `CDCConfig` gained an explicit `BeforeExportHook` test seam — runs per batch after the snapshot is captured and before the export, rides the existing `FlushOverrides{Config}` injection, nil in production, unit-pinned in both directions (ordering + error-aborts-before-side-effects).

The new `UpdateBeforeExport` scenario mutates inside the selection→export window and asserts the requirement exactly: the mutation is neither exported (the delta holds exactly the three flushed siblings) nor marked flushed, and — since the hook exposes the actual snapshot — `victim.changed_at <= snapshot < update.changed_at` is asserted directly against the real value.

internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go
iceboundrock: Fixed in 6b8ccd4. The parquet checks are now record-level, not ID-membership: the paused update/delete flush's first delta is asserted to hold exactly one stale live v1 record of the victim (payload + ver_ts via `assertLiveParquetRow`); the update retry delta exactly one v2 record (shared `assertUpdateRetryConverges`); the delete retry delta exactly one all-NULL tombstone carrying the delete's timestamps (`assertTombstoneParquet`). The ID-set comparisons stay for whole-file membership coverage.

internal/e2e_harness/production/pausing.go
iceboundrock: Fixed in 6b8ccd4. Each delegate now wraps the interceptor error with op+key context (`pause CopyObject %q: %w` etc.); `intercept` returns the bare `ctx.Err()` so the wrap stays single-layered rather than double-wrapping the same op+key. `runPausedFlush` also wraps the flush error (`flush under CopyObject pause: %w`) per the summary's line-87 note.

iceboundrock:
Review round 2 applied in 81d4dcb + 6b8ccd4 — all four findings addressed:

1. **High:** the literal selection→export race is now e2e-covered. Since no external seam exists there, `cdc.CDCConfig.BeforeExportHook` was added as an explicit test seam (nil in production; unit-pinned: runs before the export, and a hook error aborts before any side effect). `TestConcurrentFlushSnapshot/UpdateBeforeExport` asserts the mutation is neither exported nor marked flushed, and pins scenario 4 directly against the actual snapshot value: `victim.changed_at <= snapshot < update.changed_at`.
2. **Medium:** parquet assertions are record-level now — exactly one stale v1 record in the paused flush delta, one v2 record in the update retry delta (`assertLiveParquetRow`), one all-NULL tombstone in the delete retry delta (`assertTombstoneParquet`).
3. **Standards:** `PausingS3` delegates wrap interceptor errors with op+key context (`intercept` returns bare `ctx.Err()` to keep the wrap single-layered), and `runPausedFlush` wraps the flush error.

Verification: 4/4 e2e scenarios green incl. `-race`; new unit pins green; `make lint`/`make test` clean; full production e2e suite PASS (34s).

Note: this PR is still **open** — `mergedAt` is null and `origin/main` is unchanged, so the earlier merge attempt did not land. The round-2 commits are on the PR branch and CI is running against them.

PR #239, opened 2026-07-13
test(e2e): warm/cold tiebreaking & sort/pagination stability (#183)

### LINKED ISSUE #183: [E2E] Warm/cold tiebreaking & sort/pagination stability
Part of epic #172.

## Task

Verify deterministic behavior when rows have equal sort keys, same timestamps, or NULL sort values. The production advanced DuckDB template assigns all Parquet sources priority=1 (no warm/cold distinction), and pagination lacks a stable `row_id` tiebreaker.

### Warm/cold tiebreaking
1. Same `row_id` and same `changed_at` timestamp in base AND delta
2. Different values or different deletion states
3. Assert: deterministic winner (or at minimum, exactly one row returned)
4. Repeat N times to confirm no flapping

### Sort stability
1. Multiple rows with same `created_at` value
2. Paginate through them; assert no row duplication or omission across pages
3. Repeat query N times; assert same order each time

### NULL sort values
1. Rows with NULL in the sort column
2. Assert consistent placement (first or last), no flapping between runs

### Multi-key sort
1. Sort by `(status, created_at DESC)`
2. Assert correct grouping and ordering within groups

### Insert between pages
1. Start paginating → insert new row that sorts into already-visited pages
2. Assert: row appears in current or future page, never duplicated

## Success criteria

- Equal-sort-value rows are returned in a stable, repeatable order
- No row appears twice or is skipped during pagination
- NULLs handled consistently

iceboundrock:
Closes #183 (epic #172, Phase 3).

## What this delivers

Three production determinism fixes (landed red-first) plus four new oracle-checked e2e suites in `internal/e2e_harness/production` pinning warm/cold LWW tiebreaking and sort/pagination stability.

### Production fixes

**F1 — DuckDB federated ORDER BY gains a stable `row_id ASC` tiebreak** (`buildNonKeysetOrderBy`, `internal/sqlgen/duckdb_template_renderer.go`). The issue body's "pagination lacks a stable `row_id` tiebreaker" was real on the DuckDB path only: the PG optimized template already ends both ORDER BYs with `row_id`, and the harness oracle tiebreaks `row_id ASC`. Equal-sort-key rows previously fell to DuckDB scan order, so LIMIT/OFFSET windows could duplicate/skip rows across page boundaries. `design.md` §5 updated in lockstep (#214 executable-doc guard green).

**F2 — PG optimized template: explicit `NULLS LAST` on DESC sort keys** (`internal/advanced_query_template.go`, both the `ordered` CTE — whose LIMIT window depends on NULL placement — and the final ORDER BY).

> **Scope note (user-visible behavior change, accepted at plan adjudication A2, recorded on #183):** on the Postgres-only/OLTP path, a DESC sort over a nullable attribute previously placed NULL rows FIRST (PG default); they now place LAST, matching the DuckDB federated path, giving one NULL-placement contract regardless of routing. The red commit (03cf2b9) demonstrates the pre-fix divergence via the oracle.

**F3 — keyset cursors lacking a trailing `row_id` tiebreak are rejected** (`validateKeysetTiebreak`, `internal/federated/keyset.go`; added in review round 1). A cursor whose final column is a non-unique key applies a strict inequality at the page boundary and silently skips every row tied there. The validator runs at both federated seams that consume a cursor: `DBFederatedQueryEngine.Query` entry (guarding the live DuckDB renderer path, which read the cursor unvalidated) and `executeFederatedKeysetQuery` (alongside `validateKeysetColumns`, whose plain read-path error convention it mirrors). All in-repo cursor constructors already ended (or now end) with `row_id`.

### New e2e suites

- **`TestSortNulls`** — NULLS LAST pinned on both routing paths, both directions, with repeat-N order stability.
- **`TestSortStability`** — duplicate-key page-union (no dup/omission across windows that split tie groups, adversarial reverse-order-update fixture), repeat-N determinism, hot-OLTP parity, multi-key sort on both surfaces: `multi_key_public_api_desc` drives the real `EntityManager.Query` with a uniform-DESC two-key sort over the federated path (the publicly expressible slice — `QueryRequest` threads one shared `SortOrder` across all `SortBy` keys), and `multi_key_mixed_directions` pins the internal per-key-direction `AttributeOrders` contract. The public per-key-direction API gap is tracked in a follow-up issue.
- **`TestWarmColdTiebreak`** — equal-ver_ts constructions through pure production verbs (create→flush→init): identical-copies hard invariant, tombstone-wins-on-tie (read-side pin of `InjectRestore`'s same-millisecond assumption, with a select-back guard proving the restamp took), and the **#210 failure-mode-2 probe**.
- **`TestPaginationUnderInsert`** — OFFSET shift characterization (inherent window re-basing, exact shifted pages pinned; adjudication A5: the issue's "never duplicated" criterion is structurally a keyset property), and the keyset liveness contract split by direction: `keyset_insert_into_future_page` proves an ascending scan surfaces a mid-pagination insert on a later page exactly once, while `keyset_insert_front_desc_unreachable` characterizes the DESC half (a front-inserted row sorts into already-visited territory — unreachable without snapshot isolation). `keyset_incomplete_cursor_rejected` pins F3 with a positive control (the identical query with a trailing `row_id` succeeds).

### Empirical finding for #210

The plan's reconnaissance hypothesized — and the 20-run probe confirmed (**base won 20/20**) — that #210 failure mode 2's "undefined" equal-ver_ts tie is deterministic today: live base rows export `deleted_at = 0` (init COALESCE) while live delta rows export `NULL`, and under `deleted_ts DESC` + DuckDB NULLS LAST the base copy wins. The winner pin is explicitly commented as a **characterization of the #174-pinned export asymmetry, not a designed contract**, flagged for revisit when #210 lands. A comment draft for #210 is prepared (pending approval to post).

## Verification

`make test`, `make lint`, all four new suites, regression gates (`TestDeepPagination`/`TestKeysetLWW`/`TestFilterLWW`/`TestThreeTierMerge`/`TestThreeTierIsolation`/`TestQueryAcrossTiers`), and `go test -race` over `internal/sqlgen` + `internal/federated` — all green, re-run in full after the review-round commits.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -264,6 +264,15 @@ ORDER BY created_at DESC, row_id ASC
 LIMIT $PAGE_SIZE OFFSET $OFFSET;
 ```
 
+Keyset (cursor) pagination carries the same total-order requirement, and it is
+now **enforced**, not merely documented: the engine rejects any cursor whose
+final column is not `row_id` (`validateKeysetTiebreak`, guarding both the live
+renderer path in `DBFederatedQueryEngine.Query` and the `KeysetEnabled`
+`executeFederatedKeysetQuery` seam). A cursor ending on a non-unique key applies
+a strict inequality on that key at the boundary, which silently skips every row
+tied there; the trailing `row_id` gives the composite key a unique tiebreak so
+each boundary tie is resolvable (#183).
+
 ## **6. Optimization Strategies**
 
 ### **6.1 Predicate Pushdown (Critical)**
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -255,8 +255,12 @@ WHERE rn = 1
     -- a newer non-matching version must never expose an older matching one.
     AND (age > 18 AND name LIKE 'John%' AND tag = 'developer')
 
--- Sorting & Pagination  
-ORDER BY created_at DESC  
+-- Sorting & Pagination
+-- A trailing row_id ASC tiebreak gives equal-key rows a total order, so
+-- LIMIT/OFFSET page windows stay stable across requests (#183). This mirrors
+-- buildNonKeysetOrderBy, the PG optimized template's trailing m.ltbase_row_id,
+-- and the production-harness oracle.
+ORDER BY created_at DESC, row_id ASC
 LIMIT $PAGE_SIZE OFFSET $OFFSET;
 ```
 
--- a/internal/advanced_query_template.go
+++ b/internal/advanced_query_template.go
@@ -85,7 +85,7 @@ var optimizedQuerySQLTemplate = template.Must(template.New("optimizedQuery").Fun
             ORDER BY
                 {{- if gt (len .SortKeys) 0 }}
                 {{- range $i, $k := .SortKeys }}
-                k{{$i}} {{ if $k.Desc }}DESC{{ else }}ASC{{ end }}{{ if lt (add $i 1) (len $.SortKeys) }},{{ end }}
+                k{{$i}} {{ if $k.Desc }}DESC NULLS LAST{{ else }}ASC{{ end }}{{ if lt (add $i 1) (len $.SortKeys) }},{{ end }}
                 {{- end }}
                 {{- if gt (len .SortKeys) 0 }},{{ end }}
                 {{- end }}
@@ -136,7 +136,7 @@ var optimizedQuerySQLTemplate = template.Must(template.New("optimizedQuery").Fun
         ORDER BY
             {{- if gt (len .SortKeys) 0 }}
             {{- range $i, $k := .SortKeys }}
-            m.k{{$i}} {{ if $k.Desc }}DESC{{ else }}ASC{{ end }}{{ if lt (add $i 1) (len $.SortKeys) }},{{ end }}
+            m.k{{$i}} {{ if $k.Desc }}DESC NULLS LAST{{ else }}ASC{{ end }}{{ if lt (add $i 1) (len $.SortKeys) }},{{ end }}
             {{- end }}
             {{- if gt (len .SortKeys) 0 }},{{ end }}
             {{- end }}
--- a/internal/e2e_harness/production/keyset_lww_e2e_test.go
+++ b/internal/e2e_harness/production/keyset_lww_e2e_test.go
@@ -11,6 +11,14 @@ import (
 	"github.com/lychee-technology/forma/internal/model"
 )
 
+// keysetZeroRowID is the canonical zero UUID, used as the row_id arg for a
+// keyset cursor pinned to a synthetic boundary value: the row_id disjunct only
+// fires on an exact tie with the preceding key, which a synthetic boundary
+// (a value no seeded row carries) never hits, so the arm is inert. Every cursor
+// must still END on row_id — the engine now rejects cursors lacking that
+// trailing tiebreak (validateKeysetTiebreak, #183).
+const keysetZeroRowID = "00000000-0000-0000-0000-000000000000"
+
 // TestKeysetLWW (#212): the keyset cursor must be applied AFTER last-write-
 // wins deduplication. Pre-fix, the DuckDB template injected the cursor into
 // the ranked CTE — before ROW_NUMBER picked rn = 1 — so a superseded version
@@ -79,9 +87,12 @@ func testKeysetAttributeCursorResurrection(ctx context.Context, t *testing.T, en
 		Schema: wide,
 		Sorts:  []Sort{{Attr: "count"}},
 		Keyset: &model.KeysetCursor{
-			Columns: []model.KeysetColumn{{Attribute: "count", Direction: forma.SortOrderAsc}},
-			Values:  []any{float64(500)},
-			Mode:    model.KeysetCursorModeAfter,
+			Columns: []model.KeysetColumn{
+				{Attribute: "count", Direction: forma.SortOrderAsc},
+				{Attribute: "row_id", Direction: forma.SortOrderAsc}, // required trailing tiebreak (#183)
+			},
+			Values: []any{float64(500), keysetZeroRowID}, // 500 matches no row ⇒ row_id arm inert
+			Mode:   model.KeysetCursorModeAfter,
 		},
 		Limit: 10,
 	})
@@ -134,9 +145,15 @@ func testKeysetCreatedAtCursorResurrection(ctx context.Context, t *testing.T, en
 	page, err := env.Query(ctx, Query{
 		Schema: wide,
 		Keyset: &model.KeysetCursor{
-			Columns: []model.KeysetColumn{{Attribute: "created_at", Direction: forma.SortOrderDesc}},
-			Values:  []any{c.ChangedAt},
-			Mode:    model.KeysetCursorModeAfter,
+			Columns: []model.KeysetColumn{
+				{Attribute: "created_at", Direction: forma.SortOrderDesc},
+				{Attribute: "row_id", Direction: forma.SortOrderAsc}, // required trailing tiebreak (#183)
+			},
+			// Record-derived boundary at C: the row_id arm (created_at = t3 AND
+			// row_id > c.RowID) excludes C itself while B(t2) qualifies via the
+			// leading created_at < t3 disjunct.
+			Values: []any{c.ChangedAt, c.RowID.String()},
+			Mode:   model.KeysetCursorModeAfter,
 		},
 		Limit: 10,
 	})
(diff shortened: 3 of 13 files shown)


iceboundrock:
## Findings

1. **Important - the keyset insert scenario accepts a skipped new row.** `pagination_mutation_e2e_test.go:219` requires `seen[newID] == 0`, and its comment calls the issue requirement "vacuously satisfied." Issue #183 instead requires: "row appears in current or future page, never duplicated." The fixture therefore accepts precisely the omission the scenario is intended to detect. Either define the cursor/snapshot contract so this insertion is represented in a later window, or record an explicit spec change before treating this scenario as covered.

2. **Important - incomplete keysets still silently omit tied rows.** `pagination_mutation_e2e_test.go:227-277` pins the current empty continuation page for a `created_at`-only keyset cursor, then references unresolved `#NNN` follow-ups. #183 success criteria state: "No row appears twice or is skipped during pagination." The PR neither appends/enforces `row_id` nor rejects the unsafe cursor, so the public behavior remains noncompliant.

3. **Moderate - multi-key requirement is only covered below the public API.** `sort_stability_e2e_test.go:270-296` documents that `QueryRequest` cannot express `(status, created_at DESC)`. #183 explicitly calls for that sort. Internal `AttributeOrders` coverage is useful but does not deliver the requested public query behavior.

4. **Moderate - OLTP DESC NULL semantics change exceeds #183.** `advanced_query_template.go:88,139` changes PostgreSQL from its default `NULLS FIRST` to `NULLS LAST`. #183 asks only for consistent placement, "first or last," and does not specify an OLTP behavior change. The PR description records an adjudication, but it is absent from the authoritative issue; confirm the intended compatibility decision separately.

iceboundrock:
## Spec

Authoritative source: #183, with #172 Phase 3 as execution context.

- Warm/cold equal-timestamp cases are covered for identical copies, divergent values, deletion state, exactly-one results, and repeat queries.
- Equal-sort-key stability is covered with page-union and repeat-order checks, and the DuckDB non-keyset path now has a `row_id ASC` tiebreak.
- NULL placement is covered for both routing paths and directions.
- The insert-between-pages requirement is not met by the test as written: the new row is explicitly expected to be absent.
- The no-skip pagination success criterion is contradicted by the characterized, still-unfixed keyset tie omission.
- The specified `(status, created_at DESC)` public-query scenario is partial: mixed directions are tested only through internal query state.
- Explicit `NULLS LAST` on the OLTP path is additional scope relative to #183, whose wording permits either stable placement.

iceboundrock:
## Standards

No hard violation of the documented Go standards was found.

- New/changed Go files remain below the 500-line limit.
- New/changed functions remain below the 100-line limit.
- No changed production error return lacks contextual wrapping.

Judgement call: the parallel scenario-runner scaffolds in the four new production E2E files repeat more than the 15-line duplication threshold described in `coding-standard.md:13-19,82-88`. A shared test helper could remove it, although keeping setup local is defensible for test readability.

iceboundrock:
## Assessment

**Needs changes before this can be considered complete for #183.**

The core DuckDB non-keyset ordering fix and the E2E coverage are valuable, but two stated pagination acceptance criteria remain knowingly unsatisfied: the front-insert row can be skipped, and an incomplete keyset cursor can skip an equal-key group. The public mixed-direction API gap and unrecorded OLTP NULL-order compatibility change should also be resolved or moved into explicitly approved follow-up work.

iceboundrock:
## Observations

- `pagination_mutation_e2e_test.go` contains two unresolved `#NNN` references. They make the intended keyset follow-up undiscoverable from the committed test.
- The warm/cold divergent-value test deliberately characterizes the current base-wins result rather than a designed LWW contract, appropriately noting #210. This is useful diagnostic coverage, but it must not be interpreted as a durable product guarantee.
- `docs/federated-query/design.md` now matches the non-keyset DuckDB tiebreak implementation.
- The PR diff is clean under `git diff --check`.

iceboundrock:
## Verification

Run against PR head `d00bd4c99590326a99b4bd745a1c8274247e194c` in an isolated worktree:

```text
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/sqlgen -run TestBuildNonKeysetOrderBy -count=1
PASS (0.220s)

GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -tags=e2e ./internal/e2e_harness/production -run ^$ -count=1
PASS (0.621s; compilation only)

GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -tags=e2e ./internal/e2e_harness/production -run ^(TestSortNulls|TestSortStability|TestWarmColdTiebreak|TestPaginationUnderInsert)$ -count=1 -timeout=15m
PASS (13.684s)

git diff --check 2c8fe50...d00bd4c
PASS
```

Full `make test`, `make lint`, race, and the complete production E2E suite were not run in this review.

iceboundrock:
## Summary

Review against `main` merge-base `2c8fe50` and issue #183: **2 important findings, 2 moderate spec concerns, 0 hard standards violations, and 1 maintainability suggestion.**

Worst issue: the new insert-between-pages test asserts that the inserted row is absent, contrary to #183 requirement that it appear in the current or a future page without duplication.

iceboundrock:
Review round 1 addressed at bc2e68c (commits 82d50f1 + bc2e68c; PR body updated). Per finding:

**1 — keyset insert scenario accepted a skipped row.** The requirement splits by scan direction, and the suite now covers both halves. New scenario `keyset_insert_into_future_page`: an ascending `[created_at ASC, row_id ASC]` scan over 30 cold rows; a row inserted after page 1 carries the newest `created_at`, sorts into an unvisited window, and is asserted to appear **exactly once** on a later page, with all 30 originals exactly once and no duplicates — the "current or future page, never duplicated" criterion, satisfied structurally rather than vacuously. The DESC front-insert scenario is renamed `keyset_insert_front_desc_unreachable` and its comment now states the spec tension explicitly: under DESC the insert sorts into already-visited territory, so "appears in a future page" is not satisfiable without snapshot isolation; `seen[newID] == 0` remains as the characterization of that half, cross-referencing the ascending scenario for liveness.

**2 — incomplete keyset cursors silently omitted tied rows.** Enforcement landed in 82d50f1: `validateKeysetTiebreak` rejects any cursor whose final column is not `row_id`, called at both federated seams that consume a cursor — `DBFederatedQueryEngine.Query` entry (the live DuckDB renderer path previously read the cursor unvalidated) and `executeFederatedKeysetQuery` (alongside `validateKeysetColumns`, whose plain read-path error convention the new validator mirrors; the message names the offending column and the expected `row_id` state). Rejection rather than silent append: cursor `Values` pair positionally with `Columns`, and the engine cannot synthesize a `row_id` boundary value for an in-flight scan without changing which page the cursor denotes. The empty-page pin flipped to `keyset_incomplete_cursor_rejected`, with a positive control proving the identical query with a trailing `row_id` succeeds. Both `#NNN` placeholders are gone — the drafted follow-up issue is superseded. All in-repo cursor constructors audited: `keyset_lww`'s three cursors gained the trailing `row_id` (record-derived cursors use the record's RowID; synthetic boundaries use the zero UUID, inert because the `row_id` arm only fires on an exact composite-key tie the synthetic boundary never hits); the benchmark cursor already ended with `row_id`; sqlgen/queryplan unit tests construct cursors below the validated seams and are unaffected.

**3 — multi-key sort only covered below the public API.** New scenario `multi_key_public_api_desc` (bc2e68c) drives the real `EntityManager.Query` with `SortBy: ["rank","count"]` + shared `SortOrder: DESC` over cold-only data, so a correctly ordered result can only come back through the federated DuckDB path. Two deliberate substitutions, stated in the scenario comment: (a) `e2e_wide` exposes no `status`/`created_at` public attributes, so `rank` (categorical, repeating — makes the secondary key load-bearing) + `count` (distinct — total order) stand in as the faithful analog of the issue's key pair; (b) the literal mixed-direction form of `(status, created_at DESC)` remains publicly inexpressible because `QueryRequest` threads one shared `SortOrder` across all keys (`entity_query_service.go:79-104`) — the uniform-DESC slice is now pinned publicly, the per-key-direction contract stays pinned internally by `multi_key_mixed_directions`. Extending the public API to per-key directions is a product API change beyond this e2e issue; a follow-up issue draft is ready (`.superpowers/issue-public-mixed-direction-sort.md` in the working tree) — issue creation is blocked by session permissions, so it needs your go-ahead or a paste.

**4 — OLTP `NULLS LAST` exceeds #183's wording.** Confirming the intended compatibility decision: `NULLS LAST` was chosen over preserving PG-default `NULLS FIRST` because the DuckDB federated path already places NULLs last and routing (hot-only vs federated) is not observable by the caller — keeping per-path defaults would leave NULL placement dependent on routing, which is the instability class #183 targets, so one contract requires the OLTP side to move. Impact: consumers relying on PG-default NULLS FIRST under DESC sorts of nullable attributes will see those rows move to the end; commit 03cf2b9 demonstrates the pre-fix divergence red-first and `TestSortNulls` pins the new contract on both paths. A scope-note comment for the authoritative issue is drafted, but posting to #183 was denied by session permissions — it needs your approval or the text can be lifted from the PR body's scope note.

**Standards judgment (scenario-runner duplication).** Leaving as-is: the dispatcher scaffold matches the idiom of the six pre-existing production suites (`deep_pagination`, `keyset_lww`, `filter_lww`, `three_tier_*`); extracting a shared runner for only the four new files would fork the suite into two idioms, and a suite-wide refactor is better taken as its own change if wanted.

**Verification (re-run in full after the round):** `make test`, `make lint` (incl. `--build-tags=e2e` over the touched packages), the five keyset/sort/tiebreak/pagination suites, the six regression gates, and `go test -race ./internal/sqlgen ./internal/federated` — all green.

iceboundrock:
External writes from the review round are now posted: the A2 NULLS-LAST compatibility decision is recorded on #183 (https://github.com/Lychee-Technology/forma/issues/183#issuecomment-4963138786), the equal-ver_ts probe finding is on #210 (https://github.com/Lychee-Technology/forma/issues/210#issuecomment-4963134749), and the public per-key sort-direction API gap is tracked as #240.

Issue #184, opened 2026-07-10
[E2E] Tier hints & preference contracts


iceboundrock:
Part of epic #172.

## Task

Verify that every `PreferredTiers` combination actually restricts the physical data sources used, not just routing metadata. The current advanced DuckDB template always includes both S3 and Postgres sources regardless of tier hints.

### Per-tier-combination assertions
Test each subset: `[hot]`, `[warm]`, `[cold]`, `[hot,warm]`, `[hot,cold]`, `[warm,cold]`, `[hot,warm,cold]`, `[]`

For each combination:
1. Seed data in all three tiers
2. Query with that tier list
3. Request `include_execution_plan`
4. Assert execution plan lists ONLY the requested tiers as data sources

### Preference behaviors
1. `PreferHot`: verify hot-only query path
2. `UseMainAsAnchor`: verify plan includes the hint (even if not yet fully implemented)
3. Empty/default tiers: verify all three tiers used

### S3 path template
1. Supply explicit `S3ParquetPathTemplate`
2. Verify DuckDB reads from the specified glob, not the default path

## Success criteria

- Each tier subset produces a query that only accesses the specified data sources
- Execution plan accurately reflects actual sources (not just requested tiers)
- `cold` only: no Postgres hot source in the generated SQL

iceboundrock:
Acceptance-criteria notes from the implementation round (PR #242), three adjudications:

**(a) "cold only: no Postgres hot source in the generated SQL" — relaxed to "no `pg_source` data CTE; the `dirty_ids` consistency barrier is retained".** Hot-excluded `PreferredTiers` now render the advanced template without the `pg_source` data CTE (no `entity_main`/EAV `postgres_scan`), but the `dirty_ids` scan of `change_log` deliberately survives every DuckDB tier form: it is the consistency barrier, not a hot data source. Semantics of excluding hot: unflushed rows are **consistently invisible** — their stale parquet versions are discarded and not replaced — rather than resurfacing as stale reads. Pinned by the `TestTierHintPruningMatrix` adversary scenario (cold v1 + unflushed v2 update ⇒ the row vanishes entirely under `[warm,cold]`, and a filter on the v1 value returns zero rows with a same-shape positive control).

[...]

Issue #243, opened 2026-07-14
include_execution_plan never reaches the HTTP response: QueryResult has no execution-plan field


iceboundrock:
Found during #184 (PR #242 review round).

## Problem

`forma.FederatedQueryRequest.IncludeExecutionPlan` (`include_execution_plan`, types.go) is plumbed all the way through the engine — `model.FederatedQueryOptions.IncludeExecutionPlan` is honored, the plan is built, and `engine.Query` attaches it to `model.PersistentRecordPage.ExecutionPlan` — but the public `forma.QueryResult` has **no execution-plan field**, so the conversion in `entityQueryService.queryRecords` (internal/entity_query_service.go, `PersistentRecordPage` → `QueryResult`) silently drops it. `internal/httpapi/` has zero references to the plan.

Net: the API flag is accepted and does work, but its output is unreachable over HTTP. The #184 e2e suite observes plans via the in-process engine (`internal/e2e_harness/production/query.go`), which is why coverage exists despite the gap.

## Options

1. Add an execution-plan field to `forma.QueryResult` (+ JSON shape). Note `DataSourcePlan.SQL`/`Params` carry the full rendered SQL and bind parameters (#173 diagnostic contract) — decide whether the HTTP surface should expose, truncate, or redact them.
2. Retire the public flag if the plan is intended to stay an internal/diagnostic surface (then `include_execution_plan` should be rejected or documented as no-op).

## Acceptance

- Either the flag round-trips end to end over HTTP (with an e2e asserting the response carries routing + sources), or the flag is removed/documented and the request field no longer implies a response field.

iceboundrock:
_From the PR #244 / issue #185 (degraded mode & circuit breaker) review round._

While wiring the degraded-mode fallback we hit the plan-delivery gap this issue tracks, from a second angle. Status after PR #244:

- The two **degraded-fallback sites** in `internal/federated/engine.go` now stitch the recorded plan onto the returned page (`attachExecutionPlan`, review finding on PR #244) — engine-level callers that only see the page observe the fallback.
- **Still open, and squarely in #243's scope:**
  1. The **hot-only gate** (`recordHotOnlyGatePlan`) still returns pages with nil `page.ExecutionPlan` — the plan lives only on `opts.ExecutionPlan`. Same one-line stitch applies.
[...]

PR #244, opened 2026-07-14
test(e2e): degraded mode & circuit breaker under real failures (#185)

### LINKED ISSUE #185: [E2E] Degraded mode & circuit breaker behavior
Part of epic #172.

## Task

Verify degraded mode fallback and circuit breaker behavior under real infrastructure failures (not harness-simulated boolean toggles).

### 1. S3 unavailable, degraded ON
- Seed all tiers → make S3 unreachable → query with `AllowPartialDegradedMode=true`
- Assert: query falls back to Postgres-only, returns correct results, no panic

### 2. S3 unavailable, degraded OFF
- Same setup → query with `AllowPartialDegradedMode=false`
- Assert: query returns error, does not silently return partial results

### 3. DuckDB unavailable, degraded ON
- Stop/break DuckDB → query with degraded mode
- Assert: fallback to Postgres

### 4. Dirty fetch failure
- Break the dirty-ID query path → verify error propagation

### 5. Missing schema metadata
- Remove schema metadata → assert query fails (documented as non-degradable)

### 6. Fallback page metadata
- After degraded fallback, assert page metadata (total, counts) is correct
- Verify execution plan reflects the fallback path

### 1. Threshold & open
- Inject N consecutive DuckDB failures (equal to threshold)
- Assert: after threshold, queries are rejected (breaker open) without attempting DuckDB

### 2. Auto-recovery
- Let breaker open → wait for `openDuration` → next query succeeds
- Assert: breaker transitions to closed on first success

### 3. Half-open behavior
- Verify the single-probe recovery pattern (no half-open state accumulation)

### 4. Concurrent breaker
- Multiple goroutines hit the breaker simultaneously during open/close transitions
- Assert: no race conditions, consistent rejection/acceptance

### 5. Execution failure vs stream failure
- Breaker should record both query execution errors and row-stream errors
- Verify both paths increment the failure count

## Success criteria

- Degraded mode correctly falls back to Postgres and returns complete results
- Disabled degraded mode returns errors, not partial/silent results
- Circuit breaker opens at threshold, recovers after open duration
- No data inconsistency during failover

iceboundrock:
Closes #185. Part of epic #172.

## What

- **Production fix**: degraded fallback now records the execution plan (`recordDegradedFallbackPlan`, both fallback sites) — previously the plan kept the stale `UseDuckDB=true` routing decision with no fallback marker (scenario 6 gap, counterpart of the #184 hot-only gate plan).
- **Degraded-mode e2e** (`degraded_mode_e2e_test.go`): real S3 outage via `Cluster.HaltS3` (docker stop; DuckDB httpfs cannot be faulted at the Go-client layer), closed embedded DuckDB client, renamed `change_log` (dirty-fetch), ghost schema outside the metadata snapshot (`ErrSchemaMetadataCacheRequired` non-degradable, degraded-on included).
- **Circuit breaker e2e** (`circuit_breaker_e2e_test.go`): threshold→open→reject-before-DuckDB, auto-recovery after openDuration, no-half-open characterization, concurrent open/close transitions.
- **Engine integration** (`circuit_breaker_integration_test.go`): execution failures AND row-stream failures both increment the breaker through the real query path; success resets history. Deterministic mid-stream infra failure cannot be staged from containers, hence the seam-level tests.
- Harness verbs: `HaltS3`, `ReopenDuckDB` (breaker state survives rebuilds), `WithDuckMaxConnections`, `Query.AllowPartialDegradedMode`.

## Adjudications for review

1. **Scenario 3 (half-open)**: the breaker has no half-open state by design (circuit_breaker.go design note) — open expiry releases all queries and the first success clears history. Tests characterize this; strict single-probe semantics would be a follow-up.
2. **Dirty-fetch outage is effectively non-degradable (characterized)**: the plan assumed a broken change_log degrades to a complete postgres-only result, masking the outage. Implementation showed otherwise: the PG optimized template UNIONs the change_log real-time buffer into its anchor CTE (`internal/advanced_query_template.go`), so the fallback query fails too — with degraded mode ON, a change_log outage surfaces as an error rather than being masked. The e2e characterizes this (deviation comment in `TestDegradedMode_DirtyFetchFailure`); no masking risk exists.

## Follow-ups

- DuckDB session-scoped S3 config lands on only one pooled connection (`duckdb_conn.go`); `MaxConnections>1` over `:memory:` yields S3-unconfigured connections that 404 under concurrency. Latent in production (default MaxConnections=1); surfaced by the concurrent breaker e2e, which pins `WithDuckMaxConnections(1)`. Filed as #245.

### DIFF
--- a/internal/e2e_harness/harness.go
+++ b/internal/e2e_harness/harness.go
@@ -118,6 +118,22 @@ func (h *TestHarness) RestartPostgres(ctx context.Context) error {
 	return nil
 }
 
+// HaltS3 stops the S3 container in place (docker stop — the container and
+// its data survive, unlike StopS3, which terminates it). The endpoint
+// becomes really unreachable for both the Go S3 client and DuckDB httpfs,
+// which is the only fault vector that reaches DuckDB reads: client-level
+// decorators cannot, because httpfs bypasses the Go S3 client entirely.
+func (h *TestHarness) HaltS3(ctx context.Context) error {
+	if h.S3Container == nil {
+		return fmt.Errorf("halt s3: no container (external infrastructure or already terminated)")
+	}
+	stopTimeout := 30 * time.Second
+	if err := h.S3Container.Stop(ctx, &stopTimeout); err != nil {
+		return fmt.Errorf("stop s3 container: %w", err)
+	}
+	return nil
+}
+
 // StopPostgres stops the Postgres container and closes DB handle.
 func (h *TestHarness) StopPostgres(ctx context.Context) error {
 	if h.PGDB != nil {
--- a/internal/e2e_harness/production/README.md
+++ b/internal/e2e_harness/production/README.md
@@ -79,8 +79,8 @@ needs schema IDs beyond the fixtures can register its own via
 `RegisterSchemas` + `WithSchemaDir`.
 
 Options: `WithSeed`, `WithSchemaDir`, `WithFlushThresholds` (#179),
-`WithDuckMemoryMB`, `WithBreaker` (#185), `WithRoutingStrategy`,
-`WithoutManifest`.
+`WithDuckMemoryMB`, `WithBreaker` (#185), `WithDuckMaxConnections` (#185),
+`WithRoutingStrategy`, `WithoutManifest`.
 
 ## Fixture schemas (`schemas/`)
 
@@ -168,6 +168,26 @@ where cdc-init bootstraps base files before federated reads. Use
   DuckDB client, and lazy engine/manager. Restart tests must own a
   `DedicatedCluster` — restarting the shared cluster would break every
   parallel Env — and are skipped on external infrastructure.
+- `Cluster.HaltS3` (#185) stops the S3 container in place (docker stop) so
+  the endpoint becomes genuinely unreachable — the honest way to fault the
+  DuckDB httpfs read path, which cannot be broken at the Go-client layer.
+  Like `RestartPostgres` it mutates cluster-wide state, so only tests owning
+  a `DedicatedCluster` may call it (halting the shared cluster's S3 would
+  break every parallel Env); external infrastructure cannot be halted. The
+  halted container is reaped by the dedicated cluster's normal shutdown.
+- `Env.ReopenDuckDB` (#185) replaces the Env's (possibly closed) DuckDB
+  client with a fresh one and drops the lazy engine/manager so the next use
+  rebinds them. The cached circuit breaker **deliberately survives** the
+  rebuild — breaker state must span client rebuilds so recovery scenarios can
+  observe the open-to-closed transition against a healthy DuckDB.
+- `WithBreaker(maxFailures, cooldown)` (#185) enables the federated engine's
+  circuit breaker; `WithDuckMaxConnections(n)` overrides the per-test DuckDB
+  pool size (default 2). Pin it to `1` for tests that issue **concurrent**
+  DuckDB queries: `NewDuckDBClient` applies the session-scoped S3 `SET`
+  statements on a single pooled connection (`duckdb_conn.go`), so over
+  `:memory:` a second connection opened under load lacks S3 config and 404s.
+  Set `Query.AllowPartialDegradedMode` to forward the public degraded-mode
+  flag so a tier outage returns a partial result instead of erroring.
 
 ## Diagnostic artifacts
 
@@ -206,5 +226,5 @@ On failure (or `KEEP_E2E_ENV=1`) the Env writes
 | #179 flush failure matrix | `FaultInjectingS3` + `RunFlushWith` + `ExecSQL` |
 | #180 dry-run semantics | `RunFlushDry` |
 | #181 failure injection | `ExecSQL` |
-| #185 circuit breaker | `WithBreaker` |
+| #185 degraded mode & circuit breaker | `WithBreaker`, `Query.AllowPartialDegradedMode`, `Cluster.HaltS3`, `Env.ReopenDuckDB`, `WithDuckMaxConnections` |
 | #189 multi-schema | `e2e_simple`/`e2e_second` + per-test DB |
--- a/internal/e2e_harness/production/README.md
+++ b/internal/e2e_harness/production/README.md
@@ -187,7 +187,8 @@ where cdc-init bootstraps base files before federated reads. Use
   statements on a single pooled connection (`duckdb_conn.go`), so over
   `:memory:` a second connection opened under load lacks S3 config and 404s.
   Set `Query.AllowPartialDegradedMode` to forward the public degraded-mode
-  flag so a tier outage returns a partial result instead of erroring.
+  flag so a tier outage falls back to a postgres-only result (complete in
+  today's PG-retains-all model) instead of erroring.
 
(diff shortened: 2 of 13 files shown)


iceboundrock:
## Findings

1. **[Medium] Returned fallback pages do not carry the recorded execution plan.** `internal/federated/engine.go:150-151` and `:168-169` call `recordDegradedFallbackPlan(opts, ...)` then return the repository page from `queryPostgresOnly`. That page has no `ExecutionPlan`; only the mutable input options do. Direct `DBFederatedQueryEngine.Query` callers requesting a plan therefore cannot observe that a successful page degraded to PostgreSQL. The new unit tests only inspect `opts.ExecutionPlan`, which masks this. Attach the plan to the fallback `PersistentRecordPage` before returning it and assert `page.ExecutionPlan` at both fallback sites.

2. **[Spec mismatch] Recovery is not a single-probe pattern.** `internal/federated/circuit_breaker.go:58-83` has no atomic half-open/probe reservation. Once `openUntil` expires, concurrent callers all pass `IsOpen()` and execute DuckDB. `internal/e2e_harness/production/circuit_breaker_e2e_test.go:159-180` explicitly expects every worker to pass after expiry. This conflicts with #185: "Verify the single-probe recovery pattern (no half-open state accumulation)." Either implement a guarded recovery probe or amend/adjudicate the issue before treating the scenario as complete.

## Spec

- **All tiers are not seeded.** #185 says "Seed all tiers" for the S3 outage scenario. `seedTwoTiers` in `internal/e2e_harness/production/degraded_mode_e2e_test.go:18-32` creates hot rows plus flushed delta, but does not compact/init a cold/base Parquet tier.
- **Fallback page metadata is only partially asserted.** #185 requires "page metadata (total, counts) is correct." `QueryResult` retains records and `Total` but not `TotalPages` or `CurrentPage` (`internal/e2e_harness/production/query.go:57-63,92-98`), so the outage test cannot assert those fields.
- **Row-stream failure is seam-simulated, not an infrastructure failure.** `internal/federated/circuit_breaker_integration_test.go:20-32,91-106` uses a fake executor and `failingScanRows`. This is useful accounting coverage, but it is a residual gap against #185 framing: "under real infrastructure failures." The PR description documents the limitation; it should remain tracked until a deterministic real stream-failure method exists.

## Standards

Low-severity documented-standard deviations:

- `internal/e2e_harness/production/env.go:330` ignores `Duck.Close()` failure in new `ReopenDuckDB`.
- `internal/e2e_harness/production/degraded_mode_e2e_test.go:156` ignores the cleanup `Exec` failure while restoring `change_log`.
- `internal/e2e_harness/production/circuit_breaker_e2e_test.go:119,122` returns delegated errors without context.
- Modified functions retain immediate bare error forwarding at `internal/e2e_harness/production/query.go:78` and `internal/federated/engine.go:115`.

These conflict with `AGENTS.md:69-71` and `coding-standard.md` sections 2, Error Handling, and Incremental Improvement Mode. No file/function-size, nesting, naming, exported-comment, or handler-size violations found.

## Assessment

**Changes requested.** The missing `page.ExecutionPlan` is a user-visible correctness gap in the feature this PR adds. The single-probe recovery requirement also needs an explicit implementation decision against #185. The remaining spec and standards items are coverage/quality follow-ups.

## Observations

- No material scope creep: the plan mutation and new harness controls directly support #185.
- The PR correctly characterizes a `change_log` outage as non-degradable because the PostgreSQL fallback also reads the real-time buffer.
- The PR documents an unrelated DuckDB pooled-connection/S3 session-config gap and pins concurrent E2E to one connection. That limitation is not introduced by this change, but it remains a production risk if `MaxConnections > 1`.

## Verification

Reviewed `284479db119754e86602be9543dd59e02bcd76df...065cd5505eef20788b4c3f1f58183bb41d017d49` (`main` merge base to PR head).

- Passed: `git diff --check origin/main...origin/pr-244`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/federated`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -v ./internal/e2e_harness/production -tags=e2e -run "Test(DegradedMode|CircuitBreaker)" -timeout=10m` (6 tests, 24.158s)
- Not run: full `make test`, `make lint`, and race suite.

## Summary

One medium correctness finding blocks approval: returned degraded pages omit the fallback execution plan. Four #185 coverage/spec gaps and four low-severity standards deviations are recorded above.

iceboundrock:
All findings addressed — fixes in `f8743e7` + `584f3b8`, adjudications recorded on the issues. Point-by-point:

## Findings

**1. [Medium] Fallback pages don't carry the execution plan → fixed in `f8743e7`.**
Both degraded-fallback sites now stitch the recorded plan onto the returned page (`attachExecutionPlan`), and both unit tests assert `page.ExecutionPlan` (not just `opts`) — routing flipped, plan present on the page a caller actually receives. The **hot-only gate** is intentionally left as-is: it has the same pre-existing gap, and plan-into-response delivery (service layer never reads `opts.ExecutionPlan` back) is #243's scope — noted there with specifics ([#243 comment](https://github.com/Lychee-Technology/forma/issues/243#issuecomment-4965109199)).

**2. [Spec mismatch] Single-probe recovery → adjudicated on the issue ([#185 comment](https://github.com/Lychee-Technology/forma/issues/185#issuecomment-4965109283)).**
The breaker's immediate-forgiveness/no-half-open design is a documented pre-existing choice (circuit_breaker.go:58-65); this PR characterizes it end to end (including "no half-open state accumulation": a single post-recovery failure cannot reopen the breaker). Implementing a guarded single-probe is a behavioral change with its own test matrix — if you want that semantics instead, we'll open a follow-up to implement it; the adjudication comment makes that trigger explicit.

## Spec

**Seed all tiers → fixed in `584f3b8`.** New `seedAllTiers` mirrors the proven `testThreeTierUnion` recipe (RunInit base → delta flush → hot rows); `TestDegradedMode_S3Unavailable` now runs against base + delta + hot with the DuckDB-routing precondition intact.

**Page metadata → fixed in `584f3b8`.** `QueryResult` now carries `TotalPages`/`CurrentPage`; the outage test asserts `TotalPages == ceil(Total/limit)` and `CurrentPage == 1` on the degraded result, alongside the oracle-checked `Total`.

**Row-stream failure seam-simulated → tracked.** Recorded as a residual gap in the [#185 adjudication comment](https://github.com/Lychee-Technology/forma/issues/185#issuecomment-4965109283) with candidate methods (byte-cutting proxy, late-row-decode parquet) — stays open until a deterministic real mid-stream fault exists.

## Standards

All four fixed: `ReopenDuckDB` logs the ignored close error; the change_log restore cleanup logs its error; `runOne` wraps both delegated errors with context; `Env.Query` and the keyset-validation return in `engine.go` wrap with context (`584f3b8` / `f8743e7`).

## Observations

The pooled-connection S3-config gap is now filed as #245 (was "awaiting filing" in the PR body).

## Verification

All gates re-run green after the fix wave: `internal/federated` unit tests, degraded+breaker e2e 6/6 (27.4s, S3 outage passing with the three-tier seed), full `make test`, `make lint`.

Issue #226, opened 2026-07-12
cdc: orphaned flush tmp objects are never cleaned up (step-3 failure + swallowed DeleteObject)


iceboundrock:
> *Split out of PR #225 review (issue #179's "no orphan temp files" criterion).*

## Problem

Flush tmp objects under `{prefix}/{schemaID}/_tmp/` can be orphaned **permanently** — no code path ever deletes them:

1. **CopyObject failure (pipeline step 3):** the exported tmp object stays behind. A retry generates fresh UUIDs (`buildFlushS3Keys`) and `CopyTmpToFinal` only deletes its *own* tmp key (`internal/cdc/helpers.go:153-169`), so the failed attempt's tmp is unreachable garbage.
2. **DeleteObject failure (step 4):** `CopyTmpToFinal` deliberately swallows the cleanup error (`helpers.go:165-167`, log-only), so the flush succeeds and the tmp object survives.

Issue #179 requires "no orphan temp files" after failure + retry. PR #225's failure-matrix tests currently pin the *actual* behavior instead: orphans persist but are correctness-invisible (the production read glob's `*` does not cross `/`, so `_tmp/` is excluded — `internal/e2e_harness/production/query.go:52-56` — and orphans never enter the manifest). Correctness holds; the storage-hygiene invariant does not. Review of PR #225 adjudicated that #197/#203 supersession does **not** extend to this criterion, so the invariant needs to be restored by production work, tracked here.

## Why the naive fix is unsafe

"Sweep `{prefix}/{schemaID}/_tmp/` at flush start" looks obvious — the flusher holds the per-schema advisory lock (`AcquireSchemaLock`, `internal/cdc/helpers.go:31`), so no *flush* can have in-flight tmp objects for that schema. But:

- **`cdc.RunInit` does not take the advisory lock** and writes its base-export tmp objects under the same `_tmp/` layout (`internal/cdc/init_exporter.go`, e.g. `.../{schemaID}/_tmp/<uuid>.parquet`). A concurrent init's in-flight tmp would be deleted mid-export.
- The flush S3 interface (`cdc.S3ObjectClient`) has no List operation; a sweep needs `ListObjectsV2`, which is an interface-surface change for every implementer (public alias `cdc.S3ObjectClient`), or an optional capability interface + type assertion.

## Candidate designs (to be weighed)

[...]

PR #247, opened 2026-07-14
test(e2e): multi-schema isolation & backfill rerun safety (#186)

### LINKED ISSUE #186: [E2E] Multi-schema isolation & backfill rerun safety
Part of epic #172.

## Task

Verify that a failure in one schema does not corrupt another schema during CDC operations, and that backfill can be safely rerun.

### 1. One valid, one failing schema
- Schema A: valid data, valid metadata
- Schema B: missing/invalid metadata or export failure
- Run CDC flush on both
- Assert: schema A commits successfully (rows flushed, manifest updated)
- Assert: schema B fails, remaining rows stay dirty
- Assert: aggregate error returned

### 2. Retry repairs only the failed schema
- After scenario 1, fix schema Bs issue → retry flush
- Assert: only schema B rows are flushed; schema A not re-exported
- Assert: no duplicate objects or manifest entries for schema A

### 3. Concurrent flush for different schemas
- Run flush for schema A and schema B concurrently
- Assert: advisory lock scoped per-schema, no cross-contention
- Assert: both complete correctly

### 1. Idempotent rerun (no changes)
- `cdc-init` → no data changes → `cdc-init` again
- Assert: no duplicate manifest entries
- Assert: no duplicate S3 objects (or old objects cleaned up)
- Assert: federated query returns same results

### 2. Rerun with new data
- `cdc-init` → insert new rows → `cdc-init` again
- Assert: new rows are exported; old rows not duplicated
- Assert: manifest reflects all data correctly

### 3. Rerun under concurrent writes (OFFSET safety)
- Start `cdc-init` → concurrently insert/update/delete rows
- Known risk: mutable OFFSET pagination without snapshot
- Assert: no skipped rows, no duplicated rows in final Parquet
- Assert: federated query count = expected count

## Success criteria

- Schema-level failure isolation: one bad schema never blocks another
- Retry after partial failure is safe and complete
- Backfill rerun is idempotent for unchanged data
- Concurrent writes during backfill do not cause data loss or duplication

iceboundrock:
Closes #186 (epic #172, Phase 3 — final item).

## What

The multi-schema isolation half of #186, as four production-harness E2E tests; the backfill rerun-safety half is already covered by #176/PR #208 (adjudication with the scenario→test mapping is [on the issue](https://github.com/Lychee-Technology/forma/issues/186#issuecomment-4966244421)).

- **`TestMultiSchemaFlushIsolation`** — both schemas dirty in one pass, a fault scoped to one schema via its S3 key partition / manifest key:
  - *CopyFault* (pre-mark): the healthy schema commits fully (row-level flushed set == parquet row-ids == manifest entry) while the faulted schema keeps every row dirty with no final and no manifest entry (a `/_tmp/` orphan is tolerated, #226); the `errors.Join` aggregate names `schema 22` and the copy step, and does not implicate `schema 20`.
  - *ManifestSaveFault* (post-mark): the faulted schema lands in the #197 partial commit per-schema (rows flushed, final promoted, error names the orphaned key, retry is a strict no-op with no manifest self-repair) without disturbing the sibling.
- **`TestMultiSchemaRetryRepairsOnlyFailed`** — healed retry re-exports only the failed schema: no new finals under the healthy partition, its manifest keeps exactly the first pass's entry, and no row of either schema is exported twice.
- **`TestConcurrentFlushDifferentSchemas`** — deterministic per-schema advisory-lock scope proof on the real `cdc.Runner` path: runner1 pauses at schema A's CopyObject (holding A's `pg_try_advisory_lock(schemaID, schemaID)`), schema B is seeded only after the pause is reached (so runner1's enumeration provably contains A alone), and a concurrent full pass skips locked A with **no error and no side effects** while flushing B to completion; after resume both schemas converge to exactly one final each with disjoint manifests and oracle parity. This also gives same-schema exclusion its first production-path proof (the federated-harness version is hand-rolled).
- **`TestInitSchemaScopedIsolation`** — `cdc-init` runs/reruns scoped by `SchemaIDFilter` never touch a sibling schema's partition or manifest (symmetric documentation of the structurally-per-schema layout; the rerun contracts themselves are #176's tests).

All four are characterization tests and went green on first run — the believed-correct contracts (sequential per-schema loop + `errors.Join`, per-schema advisory lock, per-schema S3/manifest paths) held.

## Harness

- New per-schema helpers in `cdc_fault_helpers_e2e_test.go`: `finalsForSchema` / `assertSchemaUntouchedByFault` / `assertSchemaFullyFlushed` attribute objects by the full `{prefix}/{schemaID}/` key prefix (a bare `/22/` substring would be ambiguous against env sequence numbers), and take an explicit key set because a paused runner's report diff absorbs its sibling's writes — concurrent scenarios attribute from a fresh listing instead.
- First consumer of the `e2e_second` fixture + `SecondProfile` generator; their `#189` labels (and the README/doc.go consumer-map rows) were pre-renumber leftovers and now point at #186.

## Verification

- New tests: 3 runs (1 fresh + `-count=2`), all green, no flakes.
- `make lint` clean (golangci-lint v1.64.8), `make test` green.

### DIFF
--- a/internal/e2e_harness/production/README.md
+++ b/internal/e2e_harness/production/README.md
@@ -60,7 +60,7 @@ is unavailable). Each `NewEnv`:
   `change_log.deleted_at`.
   CDC advisory locks are keyed `(schemaID, schemaID)` *per database OID*
   and `RunOnce` scans the whole `change_log`, so a private database gives
-  natural isolation and free multi-schema support (#189);
+  natural isolation and free multi-schema support (#186);
 - gets its own S3 prefix `e2e/<runID>/env<N>` and in-memory DuckDB client
   (1 GB default memory limit, `WithDuckMemoryMB` to override);
 - registers the fixture schemas (IDs 20-22; registration rejects the
@@ -88,7 +88,7 @@ Options: `WithSeed`, `WithSchemaDir`, `WithFlushThresholds` (#179),
 |---|---|---|
 | `e2e_simple` | 20 | minimal: bound text + EAV numeric |
 | `e2e_wide` | 21 | one attribute per scalar `forma.ValueType`, mixed main-column/EAV storage (#174) |
-| `e2e_second` | 22 | second schema for multi-schema isolation (#189) |
+| `e2e_second` | 22 | second schema for multi-schema isolation (#186) |
 
 Fixtures define only the attributes their tests use: the synthetic
 `name`/`version` injection in `sqlgen.BuildSchemaProjection` was removed in
@@ -228,4 +228,4 @@ On failure (or `KEEP_E2E_ENV=1`) the Env writes
 | #180 dry-run semantics | `RunFlushDry` |
 | #181 failure injection | `ExecSQL` |
 | #185 degraded mode & circuit breaker | `WithBreaker`, `Query.AllowPartialDegradedMode`, `Cluster.HaltS3`, `Env.ReopenDuckDB`, `WithDuckMaxConnections` |
-| #189 multi-schema | `e2e_simple`/`e2e_second` + per-test DB |
+| #186 multi-schema isolation | `e2e_simple`/`e2e_second` + `FaultInjectingS3`/`PausingS3` (`multi_schema_isolation_e2e_test.go`, `concurrent_flush_multi_schema_e2e_test.go`) |
--- a/internal/e2e_harness/production/cdc_fault_helpers_e2e_test.go
+++ b/internal/e2e_harness/production/cdc_fault_helpers_e2e_test.go
@@ -195,3 +195,68 @@ func assertSameRowIDs(t *testing.T, label string, got, want map[string]bool) {
 		}
 	}
 }
+
+// schemaKeyPrefix returns the Env-scoped S3 key prefix of one schema's
+// parquet partition (e.g. "e2e/<run>/env3/22/"). Matching on the full prefix
+// rather than a bare "/22/" substring keeps the filter unambiguous no matter
+// what the run ID or env sequence look like.
+func schemaKeyPrefix(env *Env, schema SchemaRef) string {
+	return fmt.Sprintf("%s/%d/", env.S3Prefix, schema.ID)
+}
+
+// finalsForSchema filters keys down to one schema's final parquet objects
+// (multi-schema passes legitimately mix partitions in one report, so the
+// global splitKeys alone cannot attribute objects to a schema).
+func finalsForSchema(env *Env, keys []string, schema SchemaRef) []string {
+	finals, _ := splitKeys(keys)
+	var got []string
+	for _, k := range finals {
+		if strings.HasPrefix(k, schemaKeyPrefix(env, schema)) {
+			got = append(got, k)
+		}
+	}
+	return got
+}
+
+// assertSchemaUntouchedByFault is the per-schema face of assertUntouched for
+// multi-schema passes (#186): the faulted schema promoted no final parquet,
+// every one of its rows stays dirty, and its manifest tracks nothing — while
+// sibling schemas may legitimately create objects in the same report. A
+// surviving /_tmp/ orphan under the schema is tolerated (cleanup is #226).
+func assertSchemaUntouchedByFault(ctx context.Context, t *testing.T, env *Env, report *FlushReport, schema SchemaRef, wantDirty int) {
+	t.Helper()
+	if finals := finalsForSchema(env, report.NewObjects, schema); len(finals) != 0 {
+		t.Errorf("schema %d must promote no final parquet after its fault, got %v", schema.ID, finals)
+	}
+	flushed, dirty := fetchChangeLogRowIDs(ctx, t, env, schema)
+	if len(flushed) != 0 || len(dirty) != wantDirty {
+		t.Errorf("schema %d rows must all stay dirty: flushed=%v dirty=%v, want %d dirty",
+			schema.ID, flushed, dirty, wantDirty)
+	}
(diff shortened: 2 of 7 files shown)


iceboundrock:
## Findings

1. **P2 - `cdc-init` sibling isolation is not fully asserted.** `internal/e2e_harness/production/multi_schema_isolation_e2e_test.go:228-236,255,268-280` treats `InitReport.NewObjects` and a base-path-set comparison as proof that a filtered init did not touch its sibling. Existing deterministic sibling base objects can be overwritten without appearing in `NewObjects`; its manifest can also be rewritten while retaining the same paths. Capture and compare the sibling manifest contents plus object versions/ETags (or inject an operation recorder) across the scoped runs so the stated no-write contract fails on overwrite as well as create/delete.

2. **P2 - the base-tier visibility assertion can pass without reading Parquet.** `internal/e2e_harness/production/multi_schema_isolation_e2e_test.go:283-292` clears `change_log` and checks the result count, but it does not assert `result.Plan.Routing.UseDuckDB`. A routing regression to PostgreSQL-only can still read `entity_main` and produce five rows. Assert DuckDB routing for both schemas before accepting this as cold-tier coverage.

## Spec

- The copy-fault isolation, healed retry, and deterministic per-schema advisory-lock scenarios cover Issue #186 multi-schema scenarios 1-3.
- The scoped `cdc-init` check is only partial: the adjudication requires scoped runs and reruns to never touch a sibling partition **or manifest**; finding 1 does not detect overwrites or same-path manifest mutations.
- The three generic backfill rerun scenarios are intentionally satisfied by the Issue #186 adjudication mapping to merged #176 / PR #208 tests. No unapproved scope expansion found.

## Standards

- `internal/e2e_harness/production/cdc_fault_helpers_e2e_test.go:203,210`: private helpers `schemaKeyPrefix` and `finalsForSchema` are noun-first, contrary to `coding-standard.md:56`, which requires verb-first private helpers.
- `internal/e2e_harness/production/multi_schema_isolation_e2e_test.go:64-68,90,143`: new assertions inspect wrapped error text with `err.Error()` / `strings.Contains`, contrary to `AGENTS.md:42` and `coding-standard.md:22-33`. Prefer structured/sentinel error matching with `errors.Is`/`errors.As`; if these failure stages lack typed errors, expose a semantic error value rather than coupling tests to context formatting.
- No file exceeds 500 lines and no changed function exceeds 100 lines.

## Assessment

**Changes requested.** The production behavior is not implicated by these findings, but this PR is a characterization-test change: the two false-negative paths above prevent the tests from proving two claims made by the issue and PR.

## Observations

- The concurrent test deliberately establishes separate PostgreSQL sessions and documents the advisory-lock reentrancy invariant.
- Per-schema S3 prefix filtering avoids ambiguous bare-ID substring matching.
- No scope creep or additional concrete concurrency/resource-cleanup regression found.

## Verification

- `git diff --check main...origin/pr/247` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -v -tags=e2e ./internal/e2e_harness/production -run '^(TestMultiSchemaFlushIsolation|TestMultiSchemaRetryRepairsOnlyFailed|TestConcurrentFlushDifferentSchemas|TestInitSchemaScopedIsolation)$' -count=1 -timeout=10m` passed.
- The same target with `-race` passed.
- I did not rerun `make lint`, `make test`, or the full production E2E suite.

## Summary

Two P2 test-coverage findings and two documented standards violations. The most important fix is making the scoped-init test detect sibling overwrites and manifest mutations, not only new object keys.

iceboundrock:
Per-finding disposition (PR already merged, so actionable items moved to a follow-up issue):

**P2-1 (scoped-init blind to overwrites) — confirmed → #248.** `InitReport.NewObjects` is a key diff (`diffKeys`), so a same-key overwrite of a sibling base object is invisible, and the `buildBasePaths` comparison only checks path sets. Fix planned: snapshot the sibling partition (key+size+ETag+LastModified via `snapshotS3Inventory`) and the sibling manifest raw bytes+ETag (`snapshotManifest`) across the scoped runs — both helpers already exist in this package from #180.

**P2-2 (count-only base-tier check) — confirmed → #248.** Will assert `result.Plan.Routing.UseDuckDB` for both schemas, mirroring `assertFederatedRowCount`. Noted in the issue that #176's `TestInitRerunIdempotency`/`TestInitRerunAfterChangesReconcilesManifest` share the same clear-and-count pattern without a routing assertion.

**Standards: noun-first helpers — confirmed → #248.** `coding-standard.md` §4 is explicit; `schemaKeyPrefix`/`finalsForSchema` will be renamed verb-first alongside the coverage fixes (same files, one PR).

**Standards: `err.Error()`/`strings.Contains` matching — declining, with reasoning.** The `errors.Is`/`errors.As` rule (AGENTS.md, coding-standard.md) governs matching the root package's sentinel errors; these assertions pin something different — the *content* of the operator-facing aggregate message:

- Schema attribution exists only in the `fmt.Errorf("schema %d: %w")` wrapping (`runner.go`); there is no typed error carrying a schema ID, so `errors.Is`/`errors.As` cannot distinguish *which* schema failed — the text is the only observable.
- For the manifest-save arm, the message content is itself the pinned contract: #197 specifies the error must name the orphaned final key for the operator, and the merged #179 tests (`cdc_fault_postmark_e2e_test.go:30,99-100`) assert the same way after review.

If we want structural matching here, the real change is a typed per-schema flush error in `internal/cdc` (production API addition) with the message contract kept as a secondary assertion — say the word and I'll file it as a separate enhancement, but I don't think the test-side pattern alone warrants it.

Issue #245, opened 2026-07-14
configureS3 applies session-scoped S3 settings to a single pooled connection; MaxConnections>1 over :memory: yields unconfigured connections


iceboundrock:
## Summary

`NewDuckDBClient` (`internal/federated/duckdb_conn.go`) configures extensions and S3
credentials by issuing `INSTALL/LOAD/SET` statements through `db.ExecContext` on the
`database/sql` pool. Each statement lands on ONE arbitrary pooled connection. DuckDB's
`SET s3_region/s3_endpoint/s3_access_key_id/...` are session-scoped, so any additional
connection the pool opens later starts without S3 configuration. With `DBPath: ":memory:"`
and `MaxConnections > 1`, concurrent queries deterministically route some work to an
unconfigured connection and fail.

## Reproduction

Discovered by `TestCircuitBreaker_ConcurrentTransitions` (#185 scenario 4) in the
production e2e harness, which at the time ran DuckDB with `DBPath: ":memory:"` and
`MaxConnections: 2` (`internal/e2e_harness/production/env.go`). Recovery phase fires 8
concurrent federated queries through the engine; with 2 pooled connections, one query
per run failed:

```
duckdb federated query: execute duckdb query: HTTP Error: HTTP GET error reading
's3://e2e-.../env1/21/?encoding-type=url&list-type=2&prefix=...%2Fenv1%2F21%2F'
in region '' (HTTP 404 Not Found)
```

Note the tell-tale `in region ''` — the failing connection never received
`SET s3_region`/`SET s3_endpoint`, so httpfs listed against the wrong (default AWS)
endpoint semantics and 404'd on the rustfs mock bucket.

A/B evidence (5+1 runs, 2026-07-13):

- `MaxConnections: 2` (`:memory:`): **5/5 runs fail** in the concurrent recovery phase,
  identical error, always exactly one of 8 queries.
- `MaxConnections: 1`: **passes**.
- Single-threaded sibling `TestCircuitBreaker_OpensAtThresholdAndRecovers` with
  `MaxConnections: 2`: passes (never exercises the second connection concurrently).

## Root cause

`internal/federated/duckdb_conn.go:105-118`: `configureExtensions`, `configureS3`, and
`applyResourcePragmas` all run via `db.ExecContext(...)` after `sql.Open` — one pooled
connection gets the session-scoped settings; connections lazily opened afterwards do not.
(For `:memory:` DSNs the situation may be worse still: each pooled connection can be an
[...]

PR #249, opened 2026-07-14
test(e2e): Parquet corruption & S3 failure modes — manifest-driven reads + classified errors (#187)

### LINKED ISSUE #187: [E2E] Parquet corruption & S3 failure modes
Part of epic #172.

## Task

Verify federated query behavior under real Parquet and S3 failures — not harness-simulated boolean toggles. The current `TestFailureMode_CorruptedParquet` tests only missing files and valid files, not actual corrupt bytes. `TestFailureMode_S3Unavailable` only toggles a harness write flag without affecting DuckDBs S3 reads.

### 1. Corrupt Parquet bytes
- Upload a valid Parquet file → overwrite some bytes → federated query
- Assert: error returned (or graceful degradation if enabled)

### 2. Missing Parquet file (one of multiple globs)
- Reference 3 Parquet files → delete one → federated query
- Assert: error, or partial results with degraded mode

### 3. Permission error
- Upload Parquet with restricted S3 permissions → federated query
- Assert: appropriate error, not a panic or hang

### 4. Truncated Parquet
- Upload a valid Parquet file → truncate to half size → federated query

### 5. Wrong schema Parquet
- Upload Parquet with different column names/types → federated query

### 6. Empty Parquet
- Upload a valid but empty (0-row) Parquet file → federated query
- Assert: no crash, results from other tiers unaffected

### 7. One good + one bad file
- Mix valid and corrupt Parquet files in the same glob
- Assert: error or degraded; verify remaining good data is surfaced

### 8. Real S3 connection failure
- Actually stop/network-isolate the S3 container (not harness boolean)
- Query with and without degraded mode
- Restore S3 → assert recovery

### 9. Real Postgres connection failure
- Stop Postgres → query with parquet-only tiers
- Assert: error (not degraded, since hot tier is unavailable)

## Success criteria

- Each failure mode produces a classified, actionable error
- Corrupt Parquet does not crash the process
- Degraded mode correctly recovers partial results where possible
- Recovery after infrastructure restoration is clean

iceboundrock:
Closes #187. Part of epic #172 (Phase 4).

## What this delivers

Real-fault e2e coverage for all nine #187 scenarios, plus two production changes the scenarios exposed (both red-test-first, scope adjudicated with the issue owner during planning):

### Production change A — manifest-driven parquet reads (`internal/federated`, `internal/manifest`)

Scenario 2 exposed a real defect: production reads rendered a flat glob (`{{.SchemaID}}/*.parquet`) and never consulted the manifest, so deleting a manifest-listed parquet silently dropped its rows from results (committed red in 5bd02c3, reproduced live). The fix:

- New `ParquetSource` engine seam (`WithParquetSource`): with no caller-supplied render hint, reads scan the **manifest-listed objects as an explicit path list** — a missing listed object now fails the scan instead of vanishing from a glob expansion.
- On a failed read over source-authored paths the engine lazily probes `MissingKeys` and classifies storage loss as `ErrParquetSetInconsistent` (typed carrier: schema + missing keys). Classification is **by storage state, never DuckDB driver text**.
- `ErrParquetSetInconsistent` is **non-degradable** (joins `ErrSchemaMetadataCacheRequired` at both engine fallback sites): a Postgres-only fallback would return exactly the silent-loss answer the classification exists to prevent. Counter-argument noted: PG currently retains all flushed rows, so a degraded answer would be complete *today* — but that couples read correctness to a retention policy tiering is designed to eventually break. One-line policy toggle if adjudicated otherwise.
- Direction contract: **manifest ⊆ live objects is required; live ⊄ manifest is tolerated** (extra unlisted objects stay invisible; reconciliation is #203). Safe against both CDC windows — flusher and init copy final objects *before* listing them.
- Precedence: an explicit `S3ParquetPathTemplate` still directs the scan (#184's pinned contract); manifest resolution is the no-hint default, with a legacy-glob fallback for never-flushed schemas. Resolved paths feed the plan-cache scope hash so a changed file set recompiles.
- **Scope note**: server/factory wiring is deliberately absent. `forma.Config` has no manifest location/bucket/S3-client surface (the path template arrives per-request), so wiring the source in `factory` means designing new public config — spun out as a follow-up rather than invented here. The fix is live at the engine seam and in the production harness.

### Production change B — branch-classified read-path sentinels (`internal/federated/errors.go`)

- `ErrDuckDBUnavailable` (client nil, breaker open — rejected before DuckDB; degradable)
- `ErrFederatedReadFailed` (execute/scan/iterate failures — corrupt bytes, truncation, schema mismatch, credential rejection; degradable)
- `ErrParquetSetInconsistent` (+ typed `ParquetSetInconsistentError`; non-degradable)

All wraps preserve the existing stable message prefixes; raw driver errors stay in the chain for logs. Package-level placement follows the `ErrSchemaMetadataCacheRequired` precedent (#185); the public `forma/errors.go` entity surface is untouched.

## Scenario → test map (all in `internal/e2e_harness/production/`)

| # | Scenario | Test | Nature |
|---|---|---|---|
| 1 | Corrupt bytes (mid-file flip, footer intact) | `TestParquetCorruption_CorruptBytes` | contract |
| 2 | Missing manifest-listed file | `TestManifestConsistency_MissingListedParquet` | contract (red-first fix) |
| 3 | Permission error (bad httpfs creds → real RustFS SigV4 rejection) | `TestParquetPermission_BadCredentials` | contract |
| 4 | Truncated to half (footer gone) | `TestParquetCorruption_Truncated` | contract |
| 5 | Wrong-schema parquet (manifest-listed, binder error) | `TestParquetCorruption_WrongSchemaFile` | contract |
| 6 | Empty 0-row valid parquet (fabricated — exporter refuses 0-row) | `TestParquetCorruption_EmptyParquetFile` | contract |
| 7 | One good + one bad in the scan set | `TestManifestConsistency_OneGoodOneBadFile` | characterization (off) + contract (on) |
| 8 | Real S3 stop → restore → clean recovery | `TestS3Restart_CleanRecovery` | contract (restore half is new; halted half overlaps #185) |
| 9 | Real Postgres stop + parquet-only tiers | `TestPostgresHalt_ParquetTiersError` | contract |

Shared shape: seed → precondition (`UseDuckDB` + oracle match) → inject real fault → degraded OFF: bounded (30s) error with `errors.Is` classification → degraded ON: oracle-complete Postgres-only fallback with the fallback recorded on the plan. All new tests pin `WithDuckMaxConnections(1)` (#245).

### Success criteria disposition

- *Classified, actionable error*: `errors.Is`-assertable sentinels; the inconsistency error names schema + missing keys per docs/error-handling.md read-path style.
- *Corrupt parquet does not crash*: every corruption scenario returns an ordinary error (no `recover()` exists in the path; the duckdb driver returns errors, verified live).
- *Degraded mode recovers partial results where possible*: pinned oracle-complete for S1/S3/S4/S5/S7/S8; pinned NOT possible for S2 (non-degradable by design) and S9 (fallback is Postgres-only).
- *Clean recovery after restoration*: S3 restart (new `RestartS3` verb), Postgres resume (new `HaltPostgres`/`ResumePostgres` split), credential restore (S3) — all assert the DuckDB route returns with oracle-complete results.

## Harness additions (test infrastructure only)

`RestartS3` (base/cluster/env — rebinds S3 client, DuckDB httpfs endpoint, CDC config), `HaltPostgres`/`ResumePostgres` (split from `RestartPostgres`; the pool deliberately stays open so failures are genuine network errors), `ReopenDuckDBWithS3Creds` (httpfs-session credential override), `RegisterParquetInManifest`, byte-mutation fault helpers, `seedMultiParquet`.

## Characterization & adjudications (details in issue comment)

- **S7 degraded-off**: the scan is all-or-nothing (no `ignore_errors`/`union_by_name`); good parquet data is recovered via the PG fallback only, never via a partial parquet read. Follow-up filed for partial-read resilience.
- **S3**: bad-credential session is the only permission fault that can reach DuckDB httpfs on the single-credential RustFS store; RustFS SigV4 enforcement verified live.
- **Flush ordering residual**: mark-flushed precedes manifest-append, so manifest-driven reads have a sub-second window where a flushing batch is neither hot nor listed. Not a #187 fault vector (flushes here are synchronous); follow-up filed.

## Verification

- `make lint` (golangci-lint v1.64.8) and `make test` green at every commit.
- Full production suite (`-tags=e2e`) green twice consecutively; federated `-short` suite and e2e smoke green (CI-equivalent set).
- New scenario tests green individually; S2 committed red first with the live reproduction, green after the fix.


iceboundrock (commented):
## Findings

1. **P1 - Invalid explicit path templates can query a different dataset.** `internal/federated/parquet_source.go:44-51` treats a non-empty hint as absent whenever `duckDBParquetPathsForQuery` returns no paths. `duckDBParquetPathsForQuery` silently discards `RenderS3ParquetPath` errors at `internal/federated/duckdb_query_build.go:247-250`. With a `ParquetSource` installed, a malformed explicit `S3ParquetPathTemplate` therefore falls through to manifest paths instead of failing the requested query. Preserve the explicit-hint precedence contract by returning the render error; add a regression test for an invalid template.

2. **P1 - Scenario 9 has no classified, actionable failure.** `internal/e2e_harness/production/postgres_halt_e2e_test.go:42-62` only asserts a non-nil error. The dirty-ID fetch still returns an unclassified `fetch dirty ids: ...` error in `internal/federated/duckdb_query.go:160-162`. Issue #187 requires every failure mode to produce a classified, actionable error. Define the appropriate PostgreSQL/read-source classification and assert it in all three S9 assertions.

3. **P2 - URI-form manifest entries are broken by the new query source.** `internal/manifest/query_source.go:65` always prefixes `Bucket`, and `:82` sends the original value to `Exists`. Existing manifest parsing explicitly accepts `s3://bucket/base.parquet` entries (`internal/manifest/manifest_test.go:23`), producing invalid `s3://<bucket>/s3://...` scan paths and unusable existence probes. Normalize or reject absolute URI entries consistently, with coverage for the accepted format.

4. **P2 - Missing-key classification is not tied to the scanned manifest snapshot.** `QuerySource.Paths` loads one manifest, but `MissingKeys` reloads it at `internal/manifest/query_source.go:76-89`; `classifyDuckDBReadError` then uses that later result (`internal/federated/parquet_source.go:63-67`). A concurrent compaction/flush can remove the failed key from the newer manifest or add an unrelated missing key, causing a generic error or false inconsistency classification. Carry the resolved key set/snapshot from path resolution into the error-path probe.

5. **P2 - The wrong-schema scenario covers only column names, not incompatible types.** `internal/e2e_harness/production/parquet_corruption_e2e_test.go:95-98` writes only `wrong_col` and `other_col`. #187 specifies a parquet with different column names/types. Include a same-name, incompatible-type fixture to exercise the type-mismatch half.

## Spec

- #187 scenarios 1, 2, 3, 4, 6, 7, and 8 are covered by real storage/container faults.
- The issue/adjudication departures are correctly applied: S2 is deliberately non-degradable, S3 uses bad HTTPFS credentials, and S7 recovers through PostgreSQL rather than a partial parquet scan.
- Finding 2 misses the explicit success criterion: every mode must yield a classified, actionable error. Finding 5 leaves the stated names/types S5 input only partially exercised.
- No finding for absent factory/server wiring: the issue adjudication explicitly defers that public-config design to #250.

## Standards

- `internal/e2e_harness/production/env.go` is 527 lines, exceeding the 500-line limit in `AGENTS.md:68` and `coding-standard.md:7-11`.
- `DBFederatedQueryEngine.Query` in `internal/federated/engine.go:105-216` is 112 lines, exceeding the 100-line function limit in the same rules.
- The new `return err` propagation at `internal/federated/duckdb_query.go:96` and in `internal/manifest/query_source.go:53,78` does not add the required local error context (`AGENTS.md:69`, `coding-standard.md:22-33`).

## Assessment

Not ready to merge. Resolve the two P1 correctness/spec issues before approval; P2 issues should be addressed in this PR because they are introduced by the manifest-driven read path.

## Observations

- The path-set contribution to the plan-cache scope correctly prevents stale SQL literals after the manifest changes.
- The storage-state classification avoids brittle DuckDB driver-text matching.
- S2 remains non-degradable at both page and count fallback sites, matching the adjudicated consistency policy.

## Verification

- Reviewed all 9 commits (`c87bd7a` through `17bd51b`) against `main`, issue #187, and its adjudication comment.
- Reviewed `AGENTS.md`, `coding-standard.md`, and `docs/error-handling.md`.
- GitHub checks currently pass: Test, Lint, E2E Smoke Tests, and Benchmark Smoke.
- Did not rerun Docker-backed E2E locally.

## Summary

5 findings: 2 P1 and 3 P2. Primary blocker: an invalid explicit Parquet template can silently execute against manifest paths rather than the requested source.

iceboundrock:
Review round 1: all 5 findings verified against the code and accepted; all fixed in 56f9035, together with the three Standards items. Point-by-point:

**F1 (P1, invalid explicit template falls through to manifest paths) — fixed.** `duckDBParquetPathsForQuery` now returns the render error instead of swallowing it, wrapped with `forma.ErrInvalidInput` (a caller-supplied template is caller input; per docs/error-handling.md this is the write-path-style user-facing class, and it is excluded from degraded absorption — masking a bad template behind a PG-only fallback would be the same silent-wrong-dataset failure in different clothes). `resolveParquetPaths` propagates it before any source consultation. Regression: `TestParquetSource_InvalidHintTemplateErrs` pins error + source-not-consulted + duck-not-executed + non-degradable. Note this also hardens the pre-source behavior: a broken template used to silently render without S3 paths; it now errors for hint-only engines too.

**F2 (P1, S9 unclassified) — fixed.** New `ErrPostgresReadFailed` sentinel, branch-classified at the two Postgres seams of the federated read: the dirty-ID consistency fetch and `queryPostgresOnly` (which also classifies the degraded fallback's own failure). `TestPostgresHalt_ParquetTiersError` now asserts `errors.Is(err, ErrPostgresReadFailed)` on all three probes (off / degraded-on / warm+cold). Unit: `TestSentinel_DirtyFetchFailureIsPostgresReadFailed`.

**F3 (P2, URI-form manifest entries) — fixed.** `QuerySource.Paths` passes absolute `s3://` entries through unchanged (no double prefix); probing treats foreign-bucket URIs as unprovable and skips them — an unprovable key must not fabricate inconsistency. Tests: `TestQuerySource_PathsPassesURIEntriesUnchanged`, `TestQuerySource_MissingInSkipsGlobsAndForeignBuckets`.

**F4 (P2, classification not tied to the scanned snapshot) — fixed, by interface rework.** `MissingKeys(schemaID)` (which re-loaded the manifest) is replaced by `MissingIn(ctx, scanned []string)`: the engine passes the exact path set the failed scan used, and the probe never touches the manifest again — a concurrent flush/compaction can no longer hide the lost key or surface an unrelated one. Glob entries (the never-flushed fallback) are skipped as unprovable. Pinned by the unit assertion that the probe received exactly the scanned set.

**F5 (P2, wrong-schema covers names only) — fixed.** `TestParquetCorruption_WrongTypeFile` adds the type half: all column names match the real export, `row_id`/`changed_at` carry incompatible VARCHAR values — schema unification (or the UUID cast) fails and classifies `ErrFederatedReadFailed`, degraded-on recovers oracle-complete.

**Standards — all three fixed.** `env.go` split to 409 lines (`env_infra.go` carries the restart/halt/resume verbs, pure move); `DBFederatedQueryEngine.Query` is 90 lines after extracting `degradeToPostgresOnly` (the two duplicated fallback blocks) and `degradableFederatedError` (the exclusion predicate — which is also where F1's `ErrInvalidInput` exclusion lives); the flagged `return err` sites now carry local context (`resolve parquet paths:`, `resolve manifest paths:`; `MissingIn` no longer has the flagged load path).

One behavior note reviewers should see: `degradeToPostgresOnly` wraps a failed fallback with `degraded postgres-only fallback:` context — previously the raw `perr` surfaced. `errors.Is` chains are unchanged.

Verification: `make lint` + `make test` green; full production e2e suite (`-tags=e2e`) green post-fix; the reworked scenarios (S2/S5/S7/S9, wrong-type, permission) green individually.

Issue #175, opened 2026-07-10
[E2E] Entity lifecycle: create → update → delete → flush → query


iceboundrock:
Part of epic #172.

## Task

Prove the full entity lifecycle is correct across CDC boundaries. The highest-priority correctness test because the current code risks **delete resurrection** after flush.

### 1. Create → flush → query
- Write entity → CDC flush → federated query
- Assert exactly one visible row with correct values

### 2. Create → update → flush → query
- Write entity → update it → CDC flush → federated query
- Assert latest version wins, no duplicate

### 3. Create → delete → query (pre-flush)
- Write entity → soft-delete → federated query
- Assert dirty tombstone hides the row from all tiers

### 4. Create → delete → flush → query (delete resurrection risk)
- Write entity → flush (row now in cold) → soft-delete → flush → federated query
- **Critical**: tombstone must be exported to Parquet
- **Critical**: cold version must NOT reappear after tombstone is flushed
- Current analysis: delete removes `entity_main` but CDC export inner-joins to main, so tombstone may never reach Parquet

### 5. Create → delete → restore → flush → query
- Write → delete → restore → flush → query
- Assert restored row visible, no zombie tombstone

### 6. Cross-restart stability
- After any lifecycle stage, restart containers and assert idempotent results

## Success criteria

- Dirty shadow hides cold versions before flush
- After flush, exactly one latest visible version remains
- A deletion stays invisible before and after flush, restart, retry, and compaction
- Row reuse after delete works correctly

PR #253, opened 2026-07-14
feat(compaction): dirty-ratio physical rewrite + before/after equivalence e2e (#188)

### LINKED ISSUE #188: [E2E] Compaction before/after equivalence
Part of epic #172.

## Task

Prove that compaction produces query-equivalent results and correct manifest/object state. Production compaction currently only relabels manifest entries (no physical merge), and size-based promotion cannot trigger because CDC-created entries leave `SizeBytes=0`.

### 1. Promotion path
- Generate delta entries with realistic `SizeBytes` metadata → run compactor → verify promotion
- Assert: former delta entries are now listed as base entries in manifest
- Assert: federated query before and after compaction returns identical results

### 2. Rewrite path (when implemented)
- Assert dirty-ratio trigger fires correctly
- Assert new base file is created with correct merged content
- Assert old delta files are cleaned up

### 3. No-op path
- Run compactor on schema with no deltas
- Assert: no manifest changes, no new objects

### 4. Low dirty ratio skip
- Run compactor on schema with small deltas
- Assert: compaction skipped (or correctly promoted)

### 5. Manifest atomicity
- Force ETag conflict during compaction save
- Assert: retry with backoff, no lost entries
- Assert: manifest version advances monotonically

### 6. After compaction query consistency
- Full lifecycle: seed → flush → compact → query
- Compare query results (exact row IDs, values) before and after compaction
- Assert: identical results

## Success criteria

- Federated query results are bit-for-bit identical before and after compaction
- Manifest correctly reflects object inventory
- No orphan objects after compaction
- No duplicate manifest entries

iceboundrock:
Closes #188 (epic #172, Phase 4).

## What this delivers

The dirty-ratio compaction rewrite goes from stub (`RewritePending`, no-op) to a real full-schema physical merge, plus the two production defects the issue named, plus the six-scenario e2e suite proving before/after query equivalence.

### Production changes

1. **SizeBytes fix** (`a0b5853`): CDC flush and cdc-init now stat the final object and record `FileEntry.SizeBytes` (best-effort — the size only feeds the promotion heuristic). Before this, every CDC-written entry carried `SizeBytes=0` and size-based promotion could never trigger in production.
2. **Byte-precise promotion threshold** (`cf55df8`): `deltaTotalMB = bytes/(1024*1024)` integer truncation meant sub-MB delta tiers rounded to 0 MB forever. `CompactionConfig.TargetBaseSizeBytes` (derived from the MB flag when unset) is compared against raw bytes.
3. **The rewrite** (`c784034`): DuckDB merges the schema's **complete** base+delta parquet set per `row_id` with the read path's exact LWW fold — `changed_at DESC, deleted_at DESC NULLS LAST, row_id ASC` (`source_tier_priority` is constant within cold and drops out) — drops tombstones (safe only under the full-inventory invariant), normalizes survivor `deleted_at` to 0 so equal-ver_ts ties stay base-wins (#183), and carries `changed_at` verbatim (#210).
4. **Tool wiring** (`f8abf52`): `tools compactor` builds the merge engine from the CDC DuckDB export flags; `RewriteApplied` results log merge counters and the new base key.

### Commit ordering & atomicity (manifest-driven reads, #187)

merge to `_tmp` → copy to a **UUID-named** base key (`base-<uuidv7>.parquet`, never reuses a listed key; range naming would let repeated rewrites overwrite an in-read object) → manifest splice under the loaded ETag → best-effort deletion of the now-unlisted sources. The manifest save is the only atomic step: a real conditional-put conflict (`If-Match`; rustfs honors it with 412 — spike-verified) surfaces as retryable `ErrConcurrentModification`, and the retry re-merges from a fresh manifest load, dropping the failed attempt's staged base. Crash windows on either side of the commit leave only unlisted orphans (#203 scope).

Concurrency vs flush is **ETag-only** — no advisory lock, no new PG dependency in the tool: a delta appended between load and save forces the 412→retry path, which folds it in (pinned end to end by the atomicity scenario).

### E2E coverage (issue scenarios → tests)

| # | Scenario | Test |
|---|----------|------|
| 1 | Promotion path (real SizeBytes) | `TestCompactionPromotionEquivalence` — relabel only, zero parquet-object changes, promotion precedence over the rewrite trigger |
| 2 | Rewrite path | `TestCompactionRewriteEquivalence` — trigger, merged content record-level, sources deleted, hot tier untouched, second-pass Noop |
| 3 | No-op | `TestCompactionNoopPaths/NoDeltas` — `assertStateUnchanged` all surfaces |
| 4 | Low dirty ratio | `TestCompactionNoopPaths/LowDirtyRatio` — computed-and-rejected `DirtyRatio` positive control, delta stays delta |
| 5 | Manifest atomicity | `TestCompactionManifestAtomicity` — real 412 via `PausingS3` on the manifest save + concurrent real flush; no lost entries, monotonic versions, no first-attempt orphan |
| 6 | Full lifecycle | `TestCompactionFullLifecycleEquivalence` — deleted entity physically absent from the new base, invisible under every tier hint (#175 criterion transfer) |

Edge coverage: multi-version LWW fold across an equal-ver_ts base/delta pair plus newer delta (`TestCompactionRewriteMultiVersionLWW`), all-tombstone schema-empties-out with the zero-row base entry kept listed as the `QuerySource` glob-fallback guard (`TestCompactionRewriteAllTombstones`). Every scenario asserts the bit-for-bit criterion twice over: literal engine-vs-itself snapshots across the boundary (`assertResultsIdentical`) AND oracle equality on both sides, plus `manifest == S3 inventory` set equality (no orphans, no duplicates, tmp excluded per #226).

Unit level: the merge SQL shape is pinned verbatim (`merge_sql_test.go`), the fold semantics run on real in-memory DuckDB over local parquet fixtures shaped like production exports (`merge_test.go`), and the splice/conflict-retry/staged-cleanup orchestration is pinned with fakes (`rewrite_test.go`). The nil-Merger `RewritePending` degradation stays pinned for manifest-only invocations.

### Tripwire replacement

`assertCompactionPreservesDeletion` (lifecycle) is deleted in the same commit that lands the rewrite, per its own failure message; every other #175 assertion is intact and the lifecycle suite is regression-clean.

### Harness additions

`RunCompactionWith(CompactionOverrides{S3, TargetBaseSizeBytes, DirtyRatioPct})` wires the full compactor (merge engine via `cdc.NewDuckExporter`, decoratable S3); `PausingS3` gains a `KeyContains` predicate; `S3ObjectClient` gains `HeadObject` (new `S3OpHead` for the decorators); `NewS3ManifestProvider` widened to `manifest.S3Client`.

## Adjudications (also commented on the issue)

- **Old-file deletion**: immediate best-effort deletion after the manifest commit (user decision during planning). Residual: an in-flight query that resolved the pre-swap object list can race the deletion and fail loudly (`ErrParquetSetInconsistent` class) within a window of one query duration. Grace-period GC in #203 is the long-term fix.
- **"No orphan objects"**: asserted as `manifest == live parquet set` (tmp excluded, #226 tolerance); crash-window orphans remain #203 scope.

## Verification

- `make lint` clean (golangci-lint v1.64.8)
- `make test` full unit suite green (includes new merge/rewrite tests on real DuckDB)
- `make test-e2e-production` full suite green (110s), including the seven compaction tests and the lifecycle regression
- Red-first discipline: the five rewrite-dependent e2e tests were committed failing (`99a3524`, all `outcome=rewrite_pending`) before the implementation flipped them green


iceboundrock:
## Spec

**Fail**: issue #188 adjudication section 6 specifies an ETag-only retry path for a real `412` conflict. `saveManifestChecked` maps every failed conditional save that returns an empty ETag to `ErrConcurrentModification` (`internal/compaction/compactor.go:265-270`), including authorization, transport, and S3 service failures. Retry only a confirmed precondition failure; surface other errors with context, and handle ambiguous outcomes without deleting an object that may have been committed.

iceboundrock:
## Standards

**Fail** against `coding-standard.md` section 1 and `AGENTS.md:68`:

- `cmd/tools/compactor.go:16-166`: `runCompactor` is 151 lines, exceeding the 100-line function limit. Extract configuration/building responsibilities.
- `internal/cdc/flush_batch_test.go:511`: the file is 511 lines, exceeding the 500-line source-file limit. Split the new test coverage or move helpers.

Also, `coding-standard.md` section 3 requires contextual error wrapping. Raw propagated errors remain in modified paths including `internal/compaction/compactor.go:245`, `internal/compaction/rewrite.go:90`, and merge-SQL validation returns in `internal/compaction/merge_sql.go:54,84,98`.

iceboundrock:
## Verification

- `GOCACHE=/var/folders/37/2bpjmt0129z96c4plfy52gkw0000gn/T/opencode/forma-pr253-compaction-gocache GOFLAGS=-buildvcs=false go test ./internal/compaction` passed.
- `GOCACHE=/var/folders/37/2bpjmt0129z96c4plfy52gkw0000gn/T/opencode/forma-pr253-tools-gocache GOFLAGS=-buildvcs=false go test ./cmd/tools` passed.
- `git diff --check main...origin/pr-253` passed.
- GitHub checks are green: Lint, Test, E2E Smoke Tests, and Benchmark Smoke.

The production e2e suite was not rerun locally because it requires Docker/testcontainers; its PR CI evidence is captured above.

iceboundrock:
## Observations

- The complete base+delta source inventory and LWW ordering are aligned with the documented federated reader fold for the covered tombstone and equal-version cases.
- The immediate best-effort deletion of pre-swap sources retains the documented, accepted in-flight reader 404 window. This is an operational residual for #203, not an additional finding against this PR.
- The new tests exercise confirmed 412 retry, but do not model a post-commit response failure. Add a regression test where `SaveManifest` commits the replacement then returns a transport error; it must never delete the listed `finalKey`.

iceboundrock:
## Findings

1. **Critical** `internal/compaction/compactor.go:265-270`, `internal/compaction/rewrite.go:86-90`: any `SaveManifest` error with a prior ETag is classified as `ErrConcurrentModification`. This includes an ambiguous network/timeout result after S3 may already have committed the conditional PUT. `runRewrite` then deletes `finalKey`, which can now be manifest-listed; the retry observes no deltas and leaves a manifest entry pointing to a missing base object. Classify only a confirmed HTTP 412 as a CAS conflict. For an ambiguous write, reload the manifest and retain the staged key if that commit is present before retrying or cleaning up.

2. **High** `cmd/tools/compactor.go:111-112`: the AWS SDK client can use its full default credential chain, but DuckDB receives only `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Role/profile/web-identity credentials, and STS credentials requiring `AWS_SESSION_TOKEN`, can make manifest/copy calls succeed while DuckDB cannot read or write the parquet set. Resolve and pass the complete AWS credential set, including session token, or configure DuckDB to use the credential chain.

iceboundrock:
## Summary

**Request changes.** The PR has one critical manifest-atomicity/data-loss risk, one high production-credential wiring risk, one spec mismatch, and documented size-limit/error-wrapping violations. CI and focused local tests are green, but they do not cover the ambiguous successful-write failure mode.

iceboundrock:
## Assessment

The physical merge approach, full-inventory fold, UUID final keys, and pre/post equivalence coverage are directionally sound. However, the manifest save path cannot distinguish a confirmed ETag conflict from an unknown conditional-PUT outcome. Its current cleanup can remove the newly published base and make a committed manifest unreadable. This is a blocking durability failure and requires changes before merge.

A formal GitHub request-changes review cannot be submitted because this is the reviewer's own PR; this assessment comment is intentionally used instead.

iceboundrock:
## Review round 1 — resolution (345978f)

### Findings 1 (Critical) — fixed
Confirmed against the code: `manifest.Save` advances `Version`/`UpdatedAtMs` **before** the conditional put, so an ambiguous transport failure could indeed follow a committed write, and the old `newEtag=="" && etag!=""` heuristic plus unconditional staged-base deletion produced exactly the broken state described (committed manifest pointing at a deleted object; retry Noops on the spliced manifest and never repairs it).

Resolution:
- `isPreconditionFailed` classifies only a **confirmed** rejection — smithy `PreconditionFailed` API error or an HTTP 412 response — as `ErrConcurrentModification` (providers may also pre-classify via the sentinel). The heuristic is gone; all other save failures surface wrapped and non-retryable.
- `runRewrite` deletes the staged base only on the confirmed-412 branch, where it is provably unlisted. On any other failure the staged key is **retained** and embedded in the error as the operator pointer (#197 pattern).
- On the "reload and check whether the commit is present" option: the fix retains the staged key **unconditionally** on ambiguity instead of probing. It is strictly safer (no dependency on the probe itself succeeding mid-outage) and the next pass converges either way — a committed swap loads as a healthy no-delta manifest (Noop), an uncommitted one re-merges and the stale staged object stays an unlisted orphan for #203.

Pinned by `TestCompactor_Rewrite_AmbiguousSaveRetainsStagedBase` (no retry, no deletion, key in the error) and the requested `TestCompactor_Rewrite_CommittedSaveWithLostResponse` (save commits, response lost: listed `finalKey` never deleted, second pass Noop over the healthy committed manifest). The e2e `TestCompactionManifestAtomicity` now also validates the classifier against a real rustfs 412 end to end.

### Findings 2 (High) — fixed
`resolveMergeCredentials` resolves the full default chain (`LoadDefaultConfig` → `Credentials.Retrieve`, env-var fallback when the chain yields nothing, e.g. anonymous local stores) and hands DuckDB the complete triple; `NewDuckExporter` gains `s3_session_token` support (`CDCConfig.S3SessionToken`, `AWS_SESSION_TOKEN` fallback, same embed validation as key/secret). Note the session-token gap was systemic — no CDC tool ever set `s3_session_token`, so env-provided STS credentials were broken for flush/init too; the exporter-side fix covers all callers, while full-chain resolution is wired for the compactor subcommand (extending it to `cdc-flush`/`cdc-init` would be a separate change if wanted).

### Spec — resolved by the Findings 1 fix
Retry now happens only on a confirmed precondition failure; other errors surface with context; ambiguous outcomes never delete a possibly-committed object.

### Standards — fixed
- `runCompactor` 151 → 47 lines (`parseCompactorFlags` / `resolveMergeCredentials` / `openMergeEngine` / `logCompactionResult`).
- `flush_batch_test.go` 511 → 406 lines (`TestUpdateManifest_*` moved to `flush_manifest_test.go`).
- Raw propagation wrapped at the cited sites (`applyPromotion` save, `runRewrite` swap error, all three merge-SQL validator returns).

### Observations bullet 3 — test added
`TestCompactor_Rewrite_CommittedSaveWithLostResponse`, as above.

Verification: `make lint` clean, full unit suite green (26 packages), `go vet -tags=e2e ./...` clean, and the seven-test compaction e2e suite green including the real-412 atomicity scenario.

PR #254, opened 2026-07-14
test(e2e): schema evolution across Parquet generations — union_by_name reads + pre-read invariant validator (#189)

### LINKED ISSUE #189: [E2E] Schema evolution across Parquet generations

iceboundrock:
Closes #189. Part of epic #172 (Phase 4).

## What this delivers

**E2E: schema evolution across Parquet generations.** Every scenario drives the real pipeline twice: base tier exported under a v1 schema (`RunInit`), the Env then evolves in place, and delta/hot generations land under v2 — so the manifest lists physically differently-shaped parquet files for one schema and the federated read must resolve the union. All five issue scenarios are oracle-checked with DESCRIBE generation-shape preconditions and `Plan.Routing.UseDuckDB` cold-read proofs:

| Scenario | Test | Pre-fix (red-first snapshot) |
|---|---|---|
| 1 added column | `TestSchemaEvolutionAddedColumn` | RED — binder error (`Referenced column "score" not found`) |
| 2 removed column | `TestSchemaEvolutionRemovedColumn` | GREEN — characterization (see below) |
| 3 changed type int→float | `TestSchemaEvolutionChangedType` | RED — **silently wrong data** (see below) |
| 4 EAV→bound promotion | `TestSchemaEvolutionEAVToBoundPromotion` | GREEN — anchor, committed first |
| 5 mixed generations | `TestSchemaEvolutionMixedGenerations` | RED — binder error |

**Production fix** (commit 3): `read_parquet(..., union_by_name=true)` at all three scan sites — `s3_source`, the semijoin pushdown (both `advanced_query_template_duckdb.go`), and the compaction merge (`merge_sql.go`, so mixed-generation compaction normalizes instead of stalling delta GC) — plus a **pre-read system-column invariant validator** (`internal/federated/parquet_schema_validation.go`) that keeps #187's corruption detection loud (details below).

**Harness**: new verb `Env.EvolveSchema(ctx, newSchemaDir)` — swaps the live registry (write/flush path) and metadata cache (read/oracle path) and drops the memoized engine/EntityManager, modelling a production restart under updated metadata. The oracle's `applyEvent` now folds events referencing attributes absent from the *current* generation to absent instead of erroring (removal support; typos still can't reach it — the write path rejects unknown attributes under the generation that applied them).

## Two notable red-first findings

1. **Type change corrupted data silently, not loudly.** Without `union_by_name`, DuckDB coerces every file to the *first* file's schema: the v2 delta's DOUBLE `score` was cast to the v1 base's INTEGER, destroying fractional values with totals intact — only the oracle caught it (4 attribute mismatches). This is the strongest motivation for widening over first-file-wins.
2. **Column removal was already green** — both the lone-base read (explicit projection ignores extra columns) and the mixed read (DuckDB projection pushdown never requests the dropped column from any file). Pinned as characterization.

## Scenario 3 contract adjudication

The issue allowed "type coercion is predictable **or** explicitly rejected". Landed contract: **predictable widening** — `union_by_name` resolves INTEGER∪DOUBLE to DOUBLE (lossless), matching the oracle's numeric-family float64 normalization. Genuinely incompatible unions (e.g. VARCHAR∪INTEGER) still fail as classified read errors.

## #187 re-adjudication (assertions untouched, docstrings updated)

`union_by_name` NULL-fills absent columns, which would flip #187's wrong-schema contract from classified-error into **silent row loss**: the rogue file's rows get NULL `row_id`, drop out of the dirty anti-join, and the query succeeds while ignoring the file. There is no error to hook, so the validator must run **pre-read**: it probes each concrete path's footer once (cached forever — parquet objects are write-once) and enforces the export invariant `row_id UUID / changed_at BIGINT / deleted_at BIGINT`, failing classified (`ErrFederatedReadFailed`, degradable) before the scan.

The corrupt/evolution split is: **schema violation → pre-read validator; byte corruption → unchanged execution-path classifier** (an unreadable footer is inconclusive by design). Effects on the merged #187 suite:

- `TestParquetCorruption_WrongSchemaFile` / `_WrongTypeFile`: same assertions pass; the failure source moved from the DuckDB binder to the validator (docstrings updated).
- `_CorruptBytes` / `_Truncated` / `_EmptyParquetFile` / both `TestManifestConsistency_*`: untouched and green.
- #188 compaction suites: untouched and green (`union_by_name` is a no-op for same-generation merges).

## Follow-ups (opened after merge)

1. Never-flushed attribute on cold-only reads: an attribute added to the schema before any v2 flush is absent from the entire cold union → still a loud classified, degradable failure. A projection NULL-guard has a plan-cache poisoning hazard (the projection would depend on the cold column set, which the shape hash does not include).
2. Manifest schema stamping (`FileEntry` column-set fingerprint) as a cross-process validator cache.
3. Mixed-generation compaction e2e (extend the #188 equivalence harness; the production merge already handles it via this PR's `union_by_name`).

## Verification

- `make test` (unit, all packages) — green
- `make lint` — clean
- `go test -tags=e2e -run 'TestSchemaEvolution'` — 5/5 green
- `go test -tags=e2e -run 'TestParquetCorruption|TestManifestConsistency|TestCompaction'` (#187/#188 regression) — 14/14 green
- `make test-e2e-production` (full suite) — green (100s, all suites)


iceboundrock:
## Findings

1. **P1 - compaction can silently discard malformed-source rows.** `internal/compaction/merge_sql.go:62-72` enables `union_by_name` for merges without applying the new system-column invariant validation. If one source lacks `row_id`, DuckDB NULL-pads it; `ROW_NUMBER() OVER (PARTITION BY row_id)` then folds every such row into one partition. `internal/compaction/rewrite.go:76-109` commits the replacement manifest and deletes those source objects, making this permanent data loss. Validate every compaction source before merge and add a missing-`row_id` compaction regression test.

2. **P1 - an explicit glob bypasses the wrong-schema protection.** `internal/federated/parquet_schema_validation.go:54-59` intentionally skips globs, while `internal/federated/parquet_source.go:39-54` gives an explicit `S3ParquetPathTemplate` precedence over manifest paths. A malformed file matched by a hinted glob is therefore read with `union_by_name`; its NULL `row_id` is filtered by the anti-join and its rows vanish silently. Enumerate and validate glob matches, or reject glob scans while the invariant is required. Add an E2E regression case.

3. **P2 - an open circuit breaker still reaches DuckDB/S3.** `internal/federated/duckdb_query.go:94-109` executes the per-object `DESCRIBE` probes before the breaker guard at `:126-130`. This violates the existing rejection contract and can block on storage while the breaker is open. `fakeDuckDBExecutor` excludes `DESCRIBE` from `calls` (`internal/federated/engine_test.go:87-98`), masking the issue in `TestBreakerRecordsExecutionFailures`. Move the breaker gate ahead of validation and count all executor calls in the regression test.

## Spec

- **Missing renamed-column behavior.** #189 requires handling “added/removed/**renamed** columns, changed types.” `union_by_name` only unifies equal physical names (`internal/sqlgen/advanced_query_template_duckdb.go:34-41`). A logical rename leaves old-generation values under the old name and the current projection selects the new name, returning NULL for old rows. `TestSchemaEvolutionMixedGenerations` models remove-plus-add instead (`schema_evolution_e2e_test.go:358-405`), so it neither preserves nor explicitly rejects a rename. Define and implement the rename contract, then test it.

- **Scenario fixtures do not match the requested cardinalities.** #189 specifies added-column coverage as v1=3 attributes to v2=4, and removed-column coverage as v1=4 to v2=3. The PR exercises 2 to 3 (`schema_evolution_e2e_test.go:18-52,193-205`) and 3 to 2 (`:245-259`). The mechanism is exercised, but the stated coverage is not.

## Standards

- `internal/federated/engine_test.go` is now 535 lines, exceeding the 500-line source-file limit in `coding-standard.md:7-11` and `AGENTS.md:68`.
- `internal/federated/duckdb_query.go:107-109` returns the validator error bare. `coding-standard.md:22-33` and `AGENTS.md:69` require contextual wrapping.
- `evolutionProfile`, `soleParquetOf`, and `validSystemCols` (`schema_evolution_helpers_e2e_test.go:69,128`; `parquet_schema_validation_test.go:37`) are private helpers not named verb-first, contrary to `coding-standard.md:52-56`.

## Assessment

**Request changes.** The first two findings introduce silent data loss on supported read/compaction paths, which conflicts with the PR stated goal of preserving loud corruption detection.

## Observations

- The new #187 validator and regression coverage are justified protective scope: `union_by_name` otherwise changes a previously loud schema-corruption failure into silent row loss.
- `git diff --check main 84bdf1d2f1c9cefa41731818eb4b8c11aa396a3a` reports seven trailing-whitespace additions in `docs/federated-query/design.md:174-179,189`.

## Verification

Passed locally against `84bdf1d`:

- `go test ./internal/federated`
- `go test ./internal/compaction`
- `go test ./internal/sqlgen`
- `go test -tags=e2e -run "^TestSchemaEvolution" ./internal/e2e_harness/production -timeout=30m`

Not run: `make test`, `make lint`, corruption/manifest/compaction E2E regression suite, or the full production E2E suite.

## Summary

Three actionable production findings: two P1 silent-data-loss paths and one P2 breaker regression. The PR also misses rename coverage/behavior from #189 and has three documented standards violations. The focused unit and schema-evolution E2E tests pass, but the P1 issues need resolution before merge.

iceboundrock:
Review round 1 — all three findings fixed, both spec gaps closed, standards addressed; one observation answered with reasoning (f4944b4).

**P1-1 (compaction silently discards malformed-source rows) — fixed.** The invariant moved to a shared leaf package `internal/parquetcheck` (single source of truth for both consumers, no drift), and `DuckMerger.MergeToTmp` now validates every source against it **before any write**: a violation aborts the merge naming the object and column, so `runRewrite` never reaches the manifest swap or the source deletions. An unreadable footer stays inconclusive there too — the merge read must consume every file and fails loudly on it, unchanged. Regression: `TestDuckMerger_RejectsWrongSchemaSource` (real DuckDB over local fixtures: missing-`row_id` source rejected, no tmp object staged, positive control merges clean).

**P1-2 (explicit glob bypasses wrong-schema protection) — fixed** via the enumerate-and-validate option: the validator expands glob paths through DuckDB's `glob()` and validates each concrete match with the same write-once cache; an unlistable glob is inconclusive (defers to the main read's classifier, mirroring unreadable footers). Rejecting glob scans was not viable — it would break #184's explicit-hint contract and the manifest-empty fallback. Unit: `TestParquetSchemaValidator_GlobExpandsAndValidatesMatches`, `_UnlistableGlobIsInconclusive`. E2E regression as requested: `TestParquetCorruption_WrongSchemaFile_GlobHint` — the rogue file is deliberately NOT manifest-registered, so only the hinted glob can reach it; classified `ErrFederatedReadFailed`, degradable, with a healthy glob-hint positive control.

**P2 (open breaker still reaches DuckDB/S3) — fixed.** The gate moved to the top of `StreamDuckDBFederatedQuery`: an open breaker now rejects before path resolution, before any `DESCRIBE`/`glob()` probe, and before the main scan. On the test-fake accounting: probes are counted in a separate `describeCalls` field rather than folded into `calls` — folding them in would silently shift the existing pins on main-scan call counts (`TestParquetSource_PathsRenderIntoScan`, `TestBreakerRecordsExecutionFailures`); the new `TestBreakerOpenRejectsBeforeSchemaProbes` asserts all three counters (probes, scans, path resolutions) are zero while open.

**Spec — renamed-column behavior — defined and pinned.** Contract: a logical rename (same attributeID, new name) behaves as **remove+add over parquet generations** — parquet columns are physical-name-keyed and `union_by_name` unifies by name only, so old-generation values stay under the old column, which the v2 projection never selects; values do not migrate. `TestSchemaEvolutionRenamedColumn` (score→points, attributeID 3 retained) pins the shape divergence per generation and that a `points`-filter excludes every v1 row. The tier asymmetry — hot-tier EAV storage is attr_id-keyed, so a v1-written row still hot at evolution time would surface its value under the new name from pg_source — is documented in the test comment; this recipe keeps v1 rows parquet-resident, the generation boundary #189 is about.

**Spec — scenario cardinalities — fixed.** The evolving fixtures gained a third `label` attribute: added-column now runs v1=3 → v2=4 and removed-column v1=4 → v2=3, as the issue states.

**Standards — all three fixed.** `engine_test.go` split by pure move of the shared fakes into `engine_fakes_test.go` (413/112 lines); the validator error is wrapped with context at the call site (`pre-read parquet schema validation: …`); private helpers renamed verb-first (`buildEvolutionProfile`, `requireSoleParquet`, `buildValidSystemCols`, plus new `buildLabeledExtras`).

**Observation — trailing whitespace in design.md — no change, with reasoning.** The §5 block (and most of the file — 136 lines carry trailing whitespace) uses trailing double-space markdown hard breaks; every untouched neighboring line in the block has them, and the added lines match that style. Stripping only the new lines would leave the block mixed; a whole-file normalization is a separate cleanup if wanted.

Verification: `make test` green, `make lint` clean, unit suites for federated/compaction/sqlgen/cdc green, and the e2e set (`TestSchemaEvolution*` now 6 scenarios, `TestParquetCorruption*` incl. the new glob-hint case, `TestManifestConsistency*`, `TestCompaction*`, breaker suites) green — 23/23 (37s).

iceboundrock:
CI red after f4944b4 — root-caused and fixed in d1fcd6d.

**Mechanism** (`TestTruthPassOracleFiltersPureEAVAttribute` and the benchmark `-short` suite): the round-1 P1-2 fix taught the validator to expand glob paths, which extended pre-read validation to the benchmark harness's glob-hinted reads — and benchmark parquet is the legacy CSV-sniffed shape: `read_csv_auto` types `row_id` as **VARCHAR**, not UUID (verified by probing the sniffer directly), so every benchmark file failed the invariant and every benchmark engine read went loud. This is exactly why the hardcoded benchmark projections carry `CAST(row_id AS UUID)`.

**Fix**: benchmark schema IDs (100-102) are exempt from pre-read validation, mirroring the existing special-casing (`injectSchemaProjections`, the EAV-join gate). The invariant codifies the *production* exporters, which can never write into that range (`ValidateFixtureSchemaID` reserves it); the benchmark shape is scheduled for retirement (#196/#198), and re-shaping the CSV writer would not fix the pre-generated heavy-live datasets in S3 anyway. Pinned by `TestBenchmarkSchemaIDsSkipSchemaValidation` (schema 100 + would-be-violating `row_id VARCHAR` probe answer → main scan runs, no error).

Verification: the failing test reproduced locally pre-fix and passes post-fix; full CI-parity `federated/... -tags=e2e -short` suite green; production evolution/corruption/compaction e2e re-run green (the gate only adds a benchmark-range condition — production IDs 20-22 unaffected).

PR #258, opened 2026-07-15
cdc: retire the no-attribute-cache generic export path (#193)

### LINKED ISSUE #193: cdc: align or retire the no-attribute-cache generic export path (value_text-only)
Follow-up from PR #191 (#173, epic #172).

## Problem

PR #191 fixed the schema-aware CDC export to read the numeric EAV family from `value_numeric` (bool as 1/0, dates as epoch ms) matching the mainline write path. The **generic** export path — taken when no schema attribute cache is available (`resolveSchemaAttrCache` returns nil and the exporter falls back to a generic projection) — still writes `value_text`-only `attributes` structs.

Today this is latent: the production federated reader requires a metadata cache (`ErrSchemaMetadataCacheRequired`), so no-cache schemas can't be queried anyway. But the path silently produces parquet that the production reader could never consume correctly, and it will bite whoever wires a schema without registry JSON.

## Suggested fix

Either align the generic path with the schema-aware storage conventions, or retire it (fail fast when no attribute cache is available) — decide which contract we actually want.

## Acceptance

- The generic path either produces reader-compatible output or errors loudly.
- A regression test pins the chosen behavior.

iceboundrock:
Closes #193.

## Problem

The CDC export had a **generic** projection path taken whenever a schema's attribute metadata cache couldn't be resolved (`resolveSchemaAttrCache` returned nil). It emitted `value_text`-only `attributes` structs — parquet the production federated reader could never consume correctly, since the reader already fails fast with `ErrSchemaMetadataCacheRequired` when no cache is loaded. Latent today, but it would bite whoever wired a schema without registry JSON.

## Decision (see design spec)

**Retire the generic path — fail fast, aborting the whole run via pre-flight validation.** Aligning the output was infeasible (numeric-family typing — bool→1/0, dates→epoch-ms — can't be derived without the cache) and pointless (no reader consumes a no-cache schema). This makes the write path symmetric with the read path.

## What changed (`internal/cdc` only)

- **Shared resolver + sentinel** (`errors.go`): `resolveRequiredAttrCache` treats a nil registry, a lookup error, or an empty cache (`len==0`, mirroring the reader) as unresolvable, carrying the plain operator sentinel `ErrSchemaAttrCacheUnavailable` (not `forma.ErrInvalidInput`).
- **Flush pre-flight** (`flusher.go`): `processSchemas` resolves every schema's cache up front and aborts before any side effect; `executeFlush` reads the pre-resolved cache instead of consulting the registry.
- **Init pre-flight** (`init.go`): symmetric in `processInitSchemas`; the old `resolveSchemaAttrCache` is deleted.
- **Defense in depth** (`duckdb_exporter.go`): `buildExportSQLPlan` hard-errors on an empty cache; the dead generic-projection code (`buildGenericExportSQL`, `defaultMainColumns` field, `defaultDeltaMainColumns`/`defaultBaseMainColumns`, orphaned `prefixColumns`) is removed. The schema-aware/projected path is unchanged.
- **Docs**: design spec + implementation plan under `docs/superpowers/`.

Pre-flight is all-or-nothing: an unresolvable schema aborts the entire flush/init run with zero parquet writes and zero `change_log` marks. Only schemas actually in the work set are validated, so a no-work schema never trips the abort.

## Tests

- `TestResolveRequiredAttrCache` — the four resolver cases.
- `TestBuildExportSQL_ErrorsWithoutAttrCache` / `TestBuildBaseExportSQL_ErrorsWithoutAttrCache` — exporter hard-error.
- `TestProcessSchemas_AbortsWhenSchemaCacheUnavailable` + `TestProcessInitSchemas_AbortsWhenSchemaCacheUnavailable` — pin the pre-flight abort (sentinel via `errors.Is`, schema named, zero side effects). These are cited in the `multi_schema_isolation` e2e comment.
- `make test` PASS across the module; `make lint` clean (golangci-lint v1.64.8).

## Review notes

Built via a per-task spec+quality review loop plus a broad whole-branch review (verdict: ready to merge). Two cosmetic Minors remain (non-blocking): a double-`%w` chain renders the sentinel text mid-message; the two `*ErrorsWithoutAttrCache` tests use `schemaID=1`, a weak witness that the message names the id.

### DIFF
--- a/docs/superpowers/plans/2026-07-14-cdc-193-retire-generic-export.md
+++ b/docs/superpowers/plans/2026-07-14-cdc-193-retire-generic-export.md
@@ -1,0 +1,660 @@
+# Retire the No-Attribute-Cache Generic CDC Export Path (#193) — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make CDC export (delta flush + base init) fail fast when a schema's attribute metadata cache is unresolvable, aborting the whole run before any side effect, and delete the dead `value_text`-only generic projection.
+
+**Architecture:** A shared `resolveRequiredAttrCache` helper (new `internal/cdc/errors.go`) treats a registry lookup error or an empty cache as a hard failure carrying the sentinel `ErrSchemaAttrCacheUnavailable`. Both schema loops (`processSchemas` for flush, `processInitSchemas` for init) run a **pre-flight pass** that resolves every schema's cache up front and aborts on the first failure — nothing flushes. The per-schema warn+fallback is removed and the export SQL builder hard-errors on an empty cache as defense in depth, letting the generic SQL code path be deleted.
+
+**Tech Stack:** Go, DuckDB (`postgres_query` export), testify (`require`), zap logging. Package `internal/cdc`.
+
+## Global Constraints
+
+- Source files ≤500 lines, functions ≤100 lines (`coding-standard.md`).
+- Always wrap errors with context: `fmt.Errorf("...: %w", err)` — never bare `return err`.
+- This is an operator/CDC path error, NOT HTTP write-path validation: the sentinel is a plain error, **NOT** wrapped in `forma.ErrInvalidInput` (`docs/error-handling.md`).
+- Error messages must name the logical value (attribute metadata cache), the `schema_id`, and the expected state / remedy.
+- Match errors with `errors.Is`/`errors.As`, never string comparison.
+- Test env for single-test runs (mirrors the Makefile): `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false`.
+- Module path: `github.com/lychee-technology/forma`. Type: `forma.SchemaAttributeCache map[string]forma.AttributeMetadata`.
+
+---
+
+### Task 1: Sentinel error + shared required-cache resolver
+
+**Files:**
+- Create: `internal/cdc/errors.go`
+- Create: `internal/cdc/errors_test.go`
+- Modify: `internal/cdc/mocks_test.go` (add `stubSchemaRegistry` test helper)
+
+**Interfaces:**
+- Produces: `var ErrSchemaAttrCacheUnavailable error` (sentinel) and
+  `func resolveRequiredAttrCache(reg forma.SchemaRegistry, schemaID int16) (forma.SchemaAttributeCache, error)`.
+  Returns `(nil, err)` wrapping the sentinel when `reg == nil`, when
+  `reg.GetSchemaAttributeCacheByID` errors, or when the returned cache is
+  empty (`len == 0`). Returns `(cache, nil)` when the cache is non-empty.
+- Produces (test helper): `type stubSchemaRegistry struct{ cache forma.SchemaAttributeCache }` implementing `forma.SchemaRegistry`, returning `cache` with no error.
+
+- [ ] **Step 1: Add the `stubSchemaRegistry` test helper to `mocks_test.go`**
+
+Append to `internal/cdc/mocks_test.go` (after the `errorSchemaRegistry` methods):
+
+```go
+// stubSchemaRegistry returns a fixed attribute cache with no error. A nil/empty
+// cache field models a schema whose lookup succeeds but carries no attributes.
+type stubSchemaRegistry struct{ cache forma.SchemaAttributeCache }
+
+func (r stubSchemaRegistry) GetSchemaAttributeCacheByName(string) (int16, forma.SchemaAttributeCache, error) {
+	return 0, r.cache, nil
+}
+
+func (r stubSchemaRegistry) GetSchemaAttributeCacheByID(int16) (string, forma.SchemaAttributeCache, error) {
+	return "", r.cache, nil
+}
+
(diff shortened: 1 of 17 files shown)


iceboundrock:
## Findings

### P1 - `Runner.RunOnce` bypasses the required pre-flight

**Location:** `internal/cdc/runner.go:177-188`

`Runner.RunOnce` loops over `flushCtx.processSchema` directly. The new cache resolution and `attrCaches` population exist only in `schemaFlushContext.processSchemas` (`internal/cdc/flusher.go:264-276`). `executeFlush` then reads `c.attrCaches[schemaID]` (`flusher.go:376`), so this public path passes a nil cache to the exporter even when the registry is valid.

**Impact:** Every runner-backed flush fails at the new defense-in-depth check. The production E2E smoke job demonstrates this for schemas 20 and 21 with `attribute metadata cache is required but empty`.

**Required change:** Route the runner through `flushCtx.processSchemas(ctx, schemaIDs)`, or perform the identical all-schema pre-flight before its loop. Add a `Runner.RunOnce` regression test covering a populated cache and an unresolved schema with zero processing side effects.

### P3 - Cache-resolution errors do not state the expected/remedial state

**Location:** `internal/cdc/errors.go:26-34`; `internal/cdc/duckdb_exporter.go:231`

The errors identify the schema and missing cache but do not tell the operator that a registered, non-empty cache is required or how to remediate it. Include that expectation and the remedy while preserving `errors.Is(err, ErrSchemaAttrCacheUnavailable)`.

iceboundrock:
## Spec

The implementation correctly retires the generic projection, resolves caches before the package-level flush/init loops, preserves the schema-aware projection, and removes the dead generic SQL path. No material scope creep found.

The design explicitly covers both delta entry points: `RunOnce` and `Runner.RunOnce`. The runner bypass described in Findings therefore leaves the all-or-nothing pre-flight contract incomplete.

The design also requires the error to name the logical cache, `schema_id`, and a remedy: register the schema JSON or remove the schema from the flush set. The new errors name the cache/schema but omit the expected non-empty registered state and remedy.

The acceptance tests exercise helper-level pre-flight (`processSchemas` and `processInitSchemas`), but they do not protect the public runner path that regressed.

iceboundrock:
## Standards

- `docs/error-handling.md:33-39` requires messages to name the logical value, offending identifier, and expected state. `internal/cdc/errors.go:26-34` lacks the expected registered, non-empty metadata-cache state; the exporter defense error at `duckdb_exporter.go:231` has the same omission.
- `coding-standard.md:77-80` requires verification after code changes. The failing production E2E check means the current verification outcome is not acceptable for merge.

No other concrete documented-standard violations found. Formatter and lint-only observations were excluded.

iceboundrock:
## Assessment

**Request changes.** The P1 breaks the reusable CDC runner, which is exercised by the production E2E harness. This prevents valid metadata caches from producing Parquet and blocks the intended contract.

The P3 diagnostic improvement is non-blocking, but should be addressed with the P1 because this PR introduces the new operator-facing failure contract.

iceboundrock:
## Observations

- The package-level `cdc.RunOnce` correctly delegates to `processSchemas`, so its pre-flight behavior is intact.
- The sentinel and `errors.Is` coverage for lookup-error and empty-cache cases are appropriate.
- Generic SQL removal and the exporter empty-cache guard are aligned with the chosen fail-fast design.
- The new helper-level abort tests validate early return, but the runner-specific integration seam was missed.

iceboundrock:
## Verification

Reviewed the PR diff against merge-base `f81a9cb4b161e82bb69068b5526ecde3c533effe`, issue #193, and the PR design specification.

GitHub checks at review time:
- `Test`: pass
- `Lint`: pass
- `Benchmark Smoke`: pass
- `E2E Smoke Tests`: fail

The failed E2E job reports the exact regression for schemas 20 and 21: `build export SQL ... attribute metadata cache is required but empty`. Re-run the production E2E smoke suite after repairing `Runner.RunOnce`.

iceboundrock:
## Summary

One blocking P1 and one non-blocking P3 found. Fix `Runner.RunOnce` to execute the common pre-flight, add runner-level regression coverage, improve operator remediation text, then re-run the failing E2E smoke suite.

iceboundrock:
## CI fix — `75d5545`

The production e2e suite failed with:

```
schema 20: execute flush: ... build export SQL for schema 20: attribute metadata cache is required but empty
```

**Root cause.** There are two CDC flush entrypoints that each build a `schemaFlushContext` and iterate schemas:
- `cdc.RunOnce` (flusher.go) → delegates to `processSchemas`, where the new pre-flight populates `flushCtx.attrCaches`.
- `cdc.Runner.RunOnce` (runner.go) → had its **own inline loop** calling `processSchema` directly, skipping the pre-flight.

Since this PR removed `executeFlush`'s own registry lookup in favor of the pre-resolved `c.attrCaches[schemaID]`, the Runner path passed a **nil** cache to the exporter, which then hard-errored on *every* schema (even ones with valid caches like `e2e_simple`/20). The production harness flushes via `cdc.NewRunner(...).RunOnce(...)` (`internal/e2e_harness/production/cdc.go:83-85`), so the whole suite tripped; the "fault never fired" failures were downstream of the same early abort.

**Fix.** `Runner.RunOnce` now delegates to `processSchemas` — it runs the same pre-flight and de-duplicates the loop. My plan/spec missed that the Runner path duplicated the flush loop rather than sharing it.

**Verification (local, red→green).** The exact failing suite reproduced the error on the prior commit, then passed after the fix:
```
ok  github.com/lychee-technology/forma/internal/e2e_harness/production  102.267s
```
`make test` and `make lint` (golangci-lint v1.64.8) both clean.

Issue #195, opened 2026-07-10
federated: restore single-scan s3_source via version-aware predicate pushdown


iceboundrock:
Follow-up from PR #191 (#173, epic #172).

## Problem

PR #191 (commit bd1c829) fixed filter-before-LWW-dedup resurrection: the `s3_source` predicate is now applied as a row_id **semijoin** (any version of a row matching lets all versions enter dedup; the visible-filter applies the predicate to the LWW winner). Correct, but it scans the parquet glob **twice** (once for the semijoin row_id set, once for the main read).

## Suggested fix

Only if profiling shows it matters at realistic tier sizes: a version-aware predicate pushdown (e.g. pushing the predicate only where it cannot change the LWW winner, or a window-function formulation) could restore single-scan performance without reintroducing the resurrection bug.

## Acceptance

- Benchmark evidence first (heavy preset warm/cold latency, before/after).
- Any rewrite keeps the filter-before-LWW regression coverage green (the production harness smoke + the #178 suite once it lands).
- Placeholder count/order stability preserved (plan cache + arg binding).

iceboundrock:
Gate update from #212 (PR #216, merged): the keyset cursor is now applied post-dedup in the `visible` CTE and is deliberately NOT pushed into the `s3_source` row_id semijoin — `ranked` windows all versions before the cursor filters (correctness first; recovering that pushdown belongs to this rewrite). The single-scan rewrite must preserve both resurrection invariants; `TestFilterLWW` + `TestKeysetLWW` (production e2e) are the mandatory regression gate.

iceboundrock:
## Benchmark evidence (gate per this issue's "profiling first" acceptance)

Harness: `internal/federated/singlescan_pushdown_test.go` (PR #267) — three `s3_source` formulations wrapped in the production `ranked`/`visible` pipeline, run on identical parquet shards (8 files, 3 versions/row, #173 resurrection + tombstone traps encoded). All three forms first pass an equivalence test (identical visible sets, with the no-pushdown post-LWW-only control as ground truth), so the comparison is apples-to-apples.

Machine: Apple M1 Pro, local NVMe, in-memory DuckDB (`duckdb-go/v2 v2.5.6`), `-benchtime 5x`.

| scale (logical rows ×3 versions) | predicate | semijoin (current, 2 scans) | QUALIFY single-scan | post-LWW-only (no pushdown) |
|---|---|---|---|---|
[...]

Issue #263, opened 2026-07-16
ci: k6 Smoke latency-ceiling produces flaky false failures on shared runner


iceboundrock:
## Problem

The **k6 Smoke** CI job intermittently fails on its latency ceiling while all functional gates pass, producing false red builds. Most recent occurrence: PR #261, run 29464678359 (passed on a plain re-run with no code change).

Failing run's thresholds:
```
forma_query_success  ✓ rate=100.00%  (>0.97)
http_req_failed      ✓ rate=0.00%    (<0.03)
forma_query_duration ✗ p(95)=11.66s  (<8000ms)
http_req_duration    ✗ p(95)=11.51s  (<8000ms)   → k6 exit 99
checks: 117/117 succeeded
```

Distribution: `med=31ms, p(90)=5.3s, p(95)=11.5s, max=14.45s` over ~62 samples. The tail blows up under CPU contention on the shared 2-core GitHub runner (PG + RustFS + server + k6 all on 2 cores); with so few samples, a couple of slow outliers move p95 past the ceiling.

## Why it's noise, not a regression

This gate was deliberately redefined (during #190 / #243, commit 29407f8) as a **functional** gate with a *loose* latency ceiling (`p95<8s`) meant only to catch hangs/catastrophic slowness — precise latency SLA is owned by `benchmark-smoke` on dedicated hardware. The functional thresholds are the meaningful signal and consistently pass. The `p95<8s` ceiling has now produced false failures across multiple PRs.

## Options

1. **Drop the latency thresholds from the CI k6 gate** entirely; keep only the functional gates (`forma_query_success`, `http_req_failed`, checks). Latency stays observable in the output but non-blocking. (Aligns with the gate's stated intent.)
2. **Widen the ceiling** (e.g. `p95<20s`) so it only trips on true hangs.
3. **Increase sample size / VU-duration** so p95 is less outlier-sensitive (costs CI minutes).

Recommendation: option 1 (or 2), since the ceiling is explicitly not a real SLA.

## Scope

- `tests/e2e/` k6 smoke script/thresholds + `.github/workflows/ci.yml` k6 job.
- Lineage: #190 / #243 (where this gate was introduced and last tuned).

/label bug, priority:low

iceboundrock:
## Investigation: the "noise" framing does not hold — this is a deterministic server-side perf bug

[...]

PR #267, opened 2026-07-16
test(federated): #195 scan-form equivalence + benchmark harness (rewrite rejected by evidence)

iceboundrock:
## What

Adds the #195 evidence harness as a permanent regression guard: `internal/federated/singlescan_pushdown_test.go` — an equivalence test proving three `s3_source` predicate formulations (production row_id semijoin, QUALIFY single-scan candidate, no-pushdown post-LWW control) return identical visible sets on data encoding the #173 resurrection and tombstone traps, plus `BenchmarkScanForms` for measuring them.

**No production code changes.** The #195 rewrite itself was rejected by the evidence this harness produced (see the closing comment on #195): the current semijoin is 1.5×–4.4× faster than both single-scan candidates at 1M/10M-row scale, with the gap widening as data grows.

## Why keep the harness

- The equivalence test runs in the normal unit suite (30k logical rows, ~2s) and pins the filter-after-LWW contract at the SQL-formulation level: any future s3_source change that resurrects stale versions or surfaces tombstones fails it loudly, with the trivially-correct no-pushdown form as ground truth.
- The prepared-args test guards the `?`-placeholder binding position a future QUALIFY revival would rely on.
- `BenchmarkScanForms` (env `FORMA_SCANFORM_ROWS`) is the ready-made rig for re-litigating #195 if tier shapes ever change.

## Verification

- `go test ./internal/federated -run TestScanForms -v` — all PASS
- `make test` — full unit suite green
- `make lint` — clean (golangci-lint v1.64.8)
- Benchmarks executed clean at 200k (smoke), 1M, and 10M logical rows

Refs #195 (closed by evidence; this PR ships the measurement harness).

### DIFF
--- a/internal/federated/singlescan_pushdown_test.go
+++ b/internal/federated/singlescan_pushdown_test.go
@@ -1,0 +1,280 @@
+package federated
+
+import (
+	"database/sql"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+
+	_ "github.com/duckdb/duckdb-go/v2"
+	"github.com/stretchr/testify/require"
+)
+
+// #195 evidence harness: compares s3_source predicate-pushdown formulations
+// on identical parquet data. Each form is wrapped in the same
+// ranked/visible pipeline as AdvancedQueryTemplateDuckDB, with the hot tier
+// and dirty set elided — the #195 rewrite only touches how s3_source
+// narrows the dedup input, and dirty exclusion removes whole row_id
+// partitions identically under every form. The post-lww-only form applies
+// no pushdown at all, so it is trivially correct (pure filter-after-LWW)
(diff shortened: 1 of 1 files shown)


iceboundrock:
CI status: Lint / Test / Benchmark Smoke / E2E Smoke all green. **k6 Smoke red twice with the #263 signature** — functional gates pass both runs (136/136 then 72/72 checks, query_success 100%, http_req_failed 0%), only the `forma_query_duration p(95)<8000ms` latency ceiling crossed (8.35s, then 10.68s on re-run; med 30–79ms). This PR adds one Go test file (`internal/federated/singlescan_pushdown_test.go`) that is not in the k6 path, so the failure is the known 2-core-runner contention issue tracked in #263, not a regression from this diff. Leaving the merge decision to the reviewer rather than re-run-until-green.

Issue #269, opened 2026-07-16
bug(sqlgen): top-level OR composite renders unparenthesized on PG optimized-query path — schema_id constraint lost, total_records polluted


iceboundrock:
## Problem

The PostgreSQL optimized-query path renders a top-level multi-child `OR` composite **without outer parentheses**, then the template joins it with `AND`:

- `hybridStyle.outerParensOnMulti: false` — `internal/sqlgen/condition_walker.go:87-93`
- Template injection sites: `internal/advanced_query_template.go:12,16,32,40` — all four are bare `... schema_id = {{.SchemaID}} AND {{.Anchor.Condition}}`

Captured live from a running server (Postgres `log_min_duration_statement`), an OR-of-50 condition renders as:

```sql
WHERE t.schema_id = $1 AND (EXISTS (...)) OR (EXISTS (...)) OR ... OR (EXISTS (...))
```

By operator precedence this parses as `(t.schema_id = $1 AND branch1) OR branch2 OR ... OR branchN` — **every OR branch after the first scans the EAV table unconstrained by `schema_id`**.

## Impact

1. **Correctness (page totals / pagination drift):** the anchor CTE can admit `row_id`s from *other schemas* whose EAV rows happen to match an OR branch (`x.schema_id = t.schema_id` correlates to the *row's* schema, not the requested one; attr_ids are reused across schemas). Downstream `main_data`/`eav_aggregated` joins re-filter by `ltbase_schema_id` (`advanced_query_template.go:106,124`), so final records stay schema-correct — but `COUNT(*) OVER() AS total` (line 72) counts the **polluted anchor**, so `total_records`/`total_pages` can be inflated, and `LIMIT/OFFSET` paginates over the polluted row_id set, which can push real rows off a page.
2. **Performance:** the unconstrained branches turn the anchor scan into a full-EAV-table scan (all schemas) with up to N correlated EXISTS subplans per row. This is a large contributor to the multi-second `visit` list latency measured in #268 (91s cold / 10-12s warm for a 2000-row schema, local M-series; 3.4-13.5s in CI).

[...]

PR #270, opened 2026-07-16
perf(relations): batch parent enrichment lookup via set-based ANY anchor (#268, closes #263)

### LINKED ISSUE #263: ci: k6 Smoke latency-ceiling produces flaky false failures on shared runner

### LINKED ISSUE #268: perf: relation enrichment issues OR-of-N correlated EXISTS parent lookup on every list page
## Problem

Relation enrichment (`x-relation`) issues a pathological parent-lookup query on **every list/query response page** of any child schema. For the e2e `visit` schema (the only one with an `x-relation`: `contactSnapshot` → `lead` via `leadId`), a plain `GET /api/v1/visit?items_per_page=50` takes **3–13 seconds** against ~2000 rows on the CI runner, while the identical query shape on `lead`/`log` (no relations) takes 10–60ms.

Discovered while investigating #263: the k6 Smoke latency-ceiling "flake" is actually this bug — the slow tail is 100% visit requests in both failing *and* passing runs (failing run 29464678359: visit med 3408ms / max 9539ms, 20/21 samples >1s; green run 29475649847: med 3381ms / max 13514ms, 21/24 >1s; lead/log max ≤63ms in both). Latency scales with `items_per_page` (ipp=20 ≈ 2–3s, ipp=50 ≈ 5–7s, ipp=100 ≈ 9.5s). Full attribution evidence: https://github.com/Lychee-Technology/forma/issues/263#issuecomment-4989862190

## Root cause

1. List path enriches all relations when no attrs requested — `internal/entity_query_service.go:108`.
2. `fetchParents` (`internal/entity_relation_service.go:123-143`) builds an **OR-composite of N `KvCondition{ParentIDAttr = fkValue}`**, one per distinct FK on the page (N ≈ items_per_page, ≤100), and runs it through `QueryPersistentRecords` on the parent schema.
3. The parent id (`lead.id`) is an EAV attribute (attributeID 26, text), so every leaf renders as a **correlated** `EXISTS (SELECT 1 FROM <eav> x WHERE x.schema_id=$ AND x.row_id=t.row_id AND x.attr_id=$ AND x.value_text=$)` (`emitEAVLeaf`, `internal/postgres_persistent_repository_query.go:327-347`).
4. Net effect: the anchor scan evaluates up to N correlated subplans per candidate row (~2000 rows × ~50 branches). Additionally, the OR-of-N clause's bind-arg count varies with page size, so the OLTP plan cache (`optimizedQueryShapeKey` includes `len(args)`) misses per distinct page size.

Impact: every schema using `x-relation`, on every list/advanced-query/single-get response (single-get pays 1 branch and is fast; lists pay the full page width).

## Fix (designed)

Replace the OR-of-N condition with **one set-based lookup** — no public Condition-DSL change, no DuckDB/sqlgen dual-path change:

- Add internal repository method `QueryPersistentRecordsByAttrValues(ctx, tables, schemaID, attr, values []string, limit)` (interface: `internal/model/interfaces.go`; impl in new file `internal/postgres_persistent_repository_by_attr_values.go`).
- It injects a set-based anchor clause into the existing optimized-query template (`internal/advanced_query_template.go`), reusing `runOptimizedQuery` hydration unchanged:
  - EAV text/uuid parent id → `t.attr_id = $2 AND t.value_text = ANY($3)` — a single range scan on the existing index `(schema_id, attr_id, value_text, row_id)` (`cmd/tools/init_db.go:211`);
  - hot-column parent id → `m.<col> = ANY($2)` with the main-table anchor;
  - numeric EAV → `value_numeric = ANY($3)`; other value types → explicit error.
- `fetchParents` calls it once per relation with the page's distinct FK set. Enrichment semantics (enrich-all-when-unspecified) unchanged.

Expected outcome: visit lists return to the same latency class as lead/log; the #263 p95 ceiling gains ~100× headroom and stops flaking. Constant clause shape also ends the per-page-size plan-cache misses.

## Lineage

- Surfaced by the k6 functional gate introduced in #190/#243; false-flake framing in #263.

iceboundrock:
## Summary

Fixes the root cause behind the k6 Smoke latency-ceiling "flake" (#263): it was never shared-runner noise but a deterministic relation-enrichment performance bug (#268). None of #263's three threshold options are taken — **thresholds are unchanged**; the `p95<8s` ceiling is precisely what caught this bug and now returns to its intended hung-server-guard role with ~100× headroom.

## Root cause (#268)

`visit` is the only e2e schema with an `x-relation` (`contactSnapshot` → `lead` via `leadId`). Every list/query page enriches relations; `fetchParents` expanded one `KvCondition{id=X}` per distinct FK into an OR-of-N composite. Each leaf renders as a correlated `EXISTS` against the EAV table, and the composite renders unparenthesized (#269), so the `schema_id` constraint only guards the first branch — the anchor became a full EAV scan (110k rows, all schemas) with up to N=items_per_page hashed subplans per row, once per page.

Measured attribution (both a failing and a "green" CI run): `visit` lists med 3.4s / max 13.5s while `lead`/`log` served in 10-60ms at the same timestamps. The "flake" was sampling luck over which schemas landed in the 62-sample p95 window. Evidence: https://github.com/Lychee-Technology/forma/issues/263#issuecomment-4989862190

## Fix

`fetchParents` now makes one set-based lookup through a new internal repository method:

- `QueryPersistentRecordsByAttrValues(ctx, tables, schemaID, attr, values, limit)` (`internal/model/interfaces.go`, impl `internal/postgres_persistent_repository_by_attr_values.go`) injects a constant-shape anchor clause — EAV text/uuid: `t.attr_id = $2 AND t.value_text = ANY($3)`; plain-text hot column: `m.<col> = ANY($2)`; numeric EAV: `value_numeric = ANY($3)`; anything else errors — and reuses `runOptimizedQuery` hydration unchanged.
- Public Condition DSL and the DuckDB/federated dual path are untouched. Enrichment semantics (enrich-all-when-unspecified) unchanged.
- Bonus: constant clause shape ends the per-page-size OLTP plan-cache misses the variable OR arity caused (`optimizedQueryShapeKey` includes arg count).

## Verification

**Anchor EXPLAIN ANALYZE** (seeded CI dataset, 2000 records/schema, 110k EAV rows, 50 ids):

| | plan | time |
|---|---|---|
| before | Seq Scan 110,400 rows × 50 hashed SubPlans (schema_id guard lost to #269) | **2189ms** |
| after | Index Only Scan on `(schema_id, attr_id, value_text, row_id)`, 36 buffers | **0.42ms** |

Identical result sets (50 row_ids both sides).

**End-to-end** (local M-series, same DB/data, before → after):

| query | before | after |
|---|---|---|
| `GET /visit?items_per_page=20` | 91.0s (cold) | 9.8ms |
| `GET /visit?items_per_page=50` | 11.95s | 14.8ms |
| `GET /visit?items_per_page=100` | 10.45s | 23.4ms |
| `GET /visit?...sort_by=createdAt` | 5-7s | 20ms |
| `GET /lead?items_per_page=50` (unchanged path) | 98.7ms | 11.8ms* |

\* lead also improves because the polluted OR branches previously scanned other schemas' EAV rows too (see #269) — cross-traffic contention is gone.

**Correctness**: `contactSnapshot` cross-checked against parent `lead.contact` for sampled pages — 5/5 match; existing `TestEntityManager_Get` enrichment + projection-exclusion tests stay green.

**Local k6 smoke** (fixed server): p95 **60ms** / max 82ms (was p95 11.5s / max 14.45s), 976/976 checks, 520 iterations vs ~62 (throughput recovered 8×). Per-tag breakdown confirms visit ≈ lead ≈ log across every qtype.

**Tests**: TDD red-first — repository tests pin the `ANY($n)` clause/args (incl. empty-values skip, hot-column anchor, numeric array, unknown-attr error); an entity-manager test pins exactly one batched parent lookup carrying all distinct FKs and zero condition-based lead queries. `make test` and `make lint` (golangci-lint v1.64.8) green.

## Observability (#263 follow-through)

k6 requests now carry `qtype` + `schema` tags, so `http_req_duration{qtype:…}`/`{schema:…}` submetrics attribute any future tail without downloading the raw report. Thresholds untouched. `tests/e2e/README.md` threshold table updated from the pre-#190 route-split model to the current functional-gate model.

## Related

- #269 (filed during this work): top-level OR composites render unparenthesized on the PG optimized-query path — correctness (total_records pollution, pagination drift) + perf for user-supplied OR conditions. Kept out of scope here; enrichment no longer emits OR composites at all.

Closes #268
Closes #263

### DIFF
--- a/internal/entity_manager_test.go
+++ b/internal/entity_manager_test.go
@@ -6,7 +6,6 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
-	"reflect"
 	"sort"
 	"strings"
 	"testing"
@@ -515,13 +514,6 @@ type mockPersistentRecordRepository struct {
 	atomicDeleteFailAt int
 }
 
-type mockByAttrValuesCall struct {
-	schemaID int16
-	attr     string
-	values   []string
-	limit    int
-}
-
 type mockFederatedQueryEngine struct {
 	lastTables model.StorageTables
 	lastQuery  *model.FederatedAttributeQuery
@@ -643,62 +635,6 @@ func cloneRecordStore(src map[int16]map[uuid.UUID]*model.PersistentRecord) map[i
 		cloned[schemaID] = inner
 	}
 	return cloned
-}
-
-func (m *mockPersistentRecordRepository) QueryPersistentRecordsByAttrValues(ctx context.Context, tables model.StorageTables, schemaID int16, attr string, values []string, limit int) (*model.PersistentRecordPage, error) {
-	m.byAttrValuesCalls = append(m.byAttrValuesCalls, mockByAttrValuesCall{
-		schemaID: schemaID,
-		attr:     attr,
-		values:   append([]string(nil), values...),
-		limit:    limit,
-	})
-
-	valueSet := make(map[string]struct{}, len(values))
-	for _, v := range values {
-		valueSet[v] = struct{}{}
-	}
-
-	schemaRecords := m.records[schemaID]
-	rowIDs := make([]uuid.UUID, 0, len(schemaRecords))
-	for id := range schemaRecords {
-		rowIDs = append(rowIDs, id)
-	}
-	sort.Slice(rowIDs, func(i, j int) bool {
-		return rowIDs[i].String() < rowIDs[j].String()
-	})
-
-	selected := make([]*model.PersistentRecord, 0, len(values))
-	for _, id := range rowIDs {
-		record := schemaRecords[id]
-		matched := false
-		for _, text := range record.TextItems {
-			if _, ok := valueSet[text]; ok {
-				matched = true
-				break
-			}
-		}
-		if !matched {
-			for _, eav := range record.OtherAttributes {
-				if eav.ValueText == nil {
-					continue
-				}
-				if _, ok := valueSet[*eav.ValueText]; ok {
-					matched = true
-					break
-				}
-			}
-		}
-		if matched {
-			selected = append(selected, record)
-		}
-	}
-
-	return &model.PersistentRecordPage{
-		Records:      selected,
-		TotalRecords: int64(len(selected)),
-		TotalPages:   1,
-		CurrentPage:  1,
-	}, nil
 }
 
 func (m *mockPersistentRecordRepository) BatchInsertPersistentRecords(ctx context.Context, tables model.StorageTables, records []*model.PersistentRecord) error {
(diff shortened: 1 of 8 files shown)


iceboundrock:
## Verification — CI run 29485911618, all 5 checks green

**k6 Smoke on the actual shared 2-core runner** (the environment where #263 kept tripping), same thresholds:

| metric | #263 failing run | this PR |
|---|---|---|
| `http_req_duration` p95 | 11.51s ✗ | **231.6ms** ✓ |
| `forma_query_duration` p95 | 11.66s ✗ | **378.0ms** ✓ |
| max | 14.45s | 728.8ms |
| iterations completed | ~62 | **449** (throughput recovered ~7×) |

The `p95<8s` ceiling now carries ~20-30× headroom on the shared runner without any threshold change. Benchmark Smoke, E2E Smoke, Lint, and unit tests (`-race`) all pass on the same run.

iceboundrock (commented):
## Standards

### Hard violations

- `internal/entity_manager_test.go:1209-1311`: `TestEntityManager_QueryEnrichesRelationsWithSingleBatchedParentLookup` is 103 lines. `coding-standard.md:7-11` rejects functions/methods over 100 lines.
- `internal/entity_manager_test.go` (1,368 lines; +173 in this PR): the changed Go source remains far above the 500-line limit. `coding-standard.md:7-11` rejects resulting source files over 500 lines; `:93-98` requires immediate-area improvement when touching nonconforming code.
- `internal/postgres_persistent_repository_by_attr_values.go:43-46`: bare `return nil, err` drops caller context. `coding-standard.md:20-33` and `AGENTS.md:68-71` require contextual wrapping.
- `internal/postgres_persistent_repository_by_attr_values.go:32,36,40,75-76,92-93,99-100`: new read-path errors omit required message fields: 32/36/40 lack logical value type and explicit expected state; 75-76, 92-93, and 99-100 lack explicit expected state. `docs/error-handling.md:20-31,33-39` requires plain operator-visible read errors whose messages name logical type, offending attribute/column, and expected state.

### Judgment

None.

iceboundrock:
## Findings

- P3 `tests/e2e/k6/scenarios.ts:409-411`: the `setup()` health-check request does not pass `qtype` and `schema` tags, while the updated README states that every request is tagged. Add a stable tag pair such as `qtype: setup_health_check` / `schema: lead` so this request is attributable if it is included in k6 HTTP metrics, and so the documentation remains exact.

No release-blocking behavioural defect was identified in the set-based parent lookup.

iceboundrock:
## Spec

Reviewed against #268 and #263. The implementation meets the requested scope:

- `QueryPersistentRecordsByAttrValues` is internal and reuses the optimized-query hydration path.
- Text/UUID EAV, plain text hot-column, and numeric EAV anchors use the specified `ANY(...)` forms; unsupported value types return errors.
- Relation enrichment makes one lookup per relation using the page's distinct foreign-key values.
- The public condition DSL and DuckDB/federated path are unchanged; #269 remains out of scope.
- k6 thresholds are unchanged, request tags are added for all VU scenario requests, and the README describes the functional gate.

No requirement omission, incorrect implementation, or scope creep found beyond the tagging consistency item in Findings.

iceboundrock:
## Assessment

The core performance fix is narrowly scoped and preserves the existing optimized-query execution and record hydration path. The set-based anchor removes the variable-arity OR composite that triggered the pathological relation-enrichment query.

There is no functional blocker from this review. However, the documented standards violations in the separate **Standards** comment need resolution before approval: the modified test file/function exceed the repository size limits, and new error handling/messages do not meet the repository's contextual-error and read-consistency message rules.

iceboundrock:
## Observations

- The anchor clauses have a constant SQL shape and a fixed argument count per storage path, which restores optimized-query render-plan reuse across page sizes.
- The only current `x-relation` schema is `visit`; its parent `lead.id` is a text EAV attribute, directly exercising the new indexed text-anchor path.
- The new tests cover text EAV, hot-column, numeric EAV, empty values, unknown attributes, and the one-batched-lookup relation behaviour.
- This PR intentionally does not correct the independent top-level OR parenthesization defect tracked in #269.

iceboundrock:
## Verification

Fresh local verification in detached `bab7a52` worktree:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal -run '^TestQueryPersistentRecordsByAttrValues' -count=1 -v` passed: 5 tests.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal -run '^(TestQueryPersistentRecordsByAttrValues|TestEntityManager_QueryEnrichesRelationsWithSingleBatchedParentLookup)$' -count=1 -v` passed: entity-manager batching test.
- `bun run build-k6` passed.
- `git diff --check main...origin/pr/270` produced no whitespace errors.

GitHub Actions run `29485911618` is green: Test, Lint, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

iceboundrock:
## Summary

Review result: 1 low-priority observability/documentation finding and 4 documented standards violations; no spec-compliance or core functional defect identified.

The performance approach is sound and the focused tests plus CI checks are green. Resolve the Standards findings and the untagged setup health request before approval.

iceboundrock:
## Review response — all actionable items resolved in 0ba01c4

### Findings (P3, setup health check untagged)
Fixed. `setup()`'s health check now carries `qtype: 'setup_health_check'` / `schema: 'lead'` (`scenarios.ts`), so every request the script issues is attributable and the README claim is exact.

### Standards

1. **Test function 103 lines > 100** — Fixed. The test moved to `internal/entity_relation_service_test.go` with the seeding extracted into `seedVisitEnrichmentFixture` (49 lines, `t.Helper()`); the test function is now 60 lines. Assertions are unchanged.
2. **`entity_manager_test.go` (1,368 lines) above the 500-line limit** — Immediate area resolved per Incremental Improvement Mode: the relation-enrichment test, the mock's `QueryPersistentRecordsByAttrValues` method, and the call recorder now live in the new `internal/entity_relation_service_test.go` (197 lines). This PR's net change to the oversized file is **+1 line** (the `byAttrValuesCalls` struct field, which must sit with the mock's type declaration). The file's pre-existing 1,195 lines are out of scope here; filed **#271** with a concrete per-service split plan.
3. **Bare `return nil, err` at `by_attr_values.go:43-46`** — Fixed: wrapped as `failed to query records by attr %q for schema %d: %w`.
4. **Read-path error messages missing required fields** — Fixed. All six sites now name the logical value type, the offending attribute/column, and the expected state, e.g.:
   - `attribute not in metadata cache for schema %d (expected a registered attribute with a value type)`
   - `unsupported main-column binding for %s attribute %q: column %s with encoding %q (expected an unencoded text/uuid column for batch value lookup)`
   - `invalid value %q for %s attribute %q: not parseable as float64 (expected a decimal value_numeric operand): %w`
   - `unsupported value_type %s for attribute %q: no batch value column (expected text, uuid, or a numeric family type)`

`make test` and `make lint` green after the changes; the k6 bundle rebuilds cleanly.

### Spec / Observations / Verification / Summary
No action needed; two confirmations for the record:

- #269 (top-level OR parenthesization) intentionally stays out of scope — after this PR, enrichment emits no OR composites at all, so the remaining exposure is user-supplied top-level OR conditions on the PG optimized-query path, tracked there.
- The Observations point about constant clause shape restoring render-plan reuse is the intended secondary effect: `optimizedQueryShapeKey` includes the arg count, so the old variable-arity OR forced one render miss per distinct page size.

Issue #222, opened 2026-07-12
sqlgen: retire the toy-schema projection defaults in injectDuckDBTemplateParams (phantom ltbase_* S3 aliases)


iceboundrock:
Follow-up from PR #221 / issue #214 (noted there as out of scope).

## Problem

`injectDuckDBTemplateParams` (`internal/sqlgen/duckdb_template_renderer.go:167-200`) injects fixed toy-schema defaults for every projection param a caller did not set: `S3SourceSelect`, `PGSourceSelect`, `PGGroupBy`, `EAVPivotSelect`, `EAVPivotAttrs`, `HasEAVPivot`, and `OuterSelect`.

The `S3SourceSelect` default (`duckdb_template_renderer.go:170`) projects `ltbase_created_at AS created_at`, `ltbase_updated_at AS ver_ts`, `ltbase_deleted_at AS deleted_ts` — columns that do not exist in any parquet produced by the CDC exporter. The runtime projection is `changed_at AS created_at/ver_ts`, `deleted_at AS deleted_ts` (`buildS3Projection`, `internal/sqlgen/duckdb_schema_projection.go:141-152`), which #221 pinned with anti-drift guards on both the runtime template and the design doc. The other defaults hardcode the same retired toy schema (name/age/tag, EAV attr 205).

## Why it matters

- **It is a second, still-live copy of a fallback the project already retired.** `injectSchemaProjections` (`internal/federated/duckdb_query.go:425-433`) fails fast with `ErrSchemaMetadataCacheRequired` when no schema cache is loaded, and its comment explicitly documents why the toy-schema projection was removed there: its column set cannot be scanned by `duckDBScanBuffers` (the #147 positional-scan contract) — "a hard error at best, misaligned rows at worst". The renderer-level defaults reproduce exactly that projection.
- **A latent error-swallow lands precisely on these defaults.** `injectSchemaProjections` discards the projection error: `if projErr != nil || sp == nil { return hit, nil }` (`duckdb_query.go:437-440`), leaving the params unset so the renderer silently substitutes the phantom projection. `BuildSchemaProjection` currently has a single `return sp, nil` path, so this is not triggerable today — but `ProjectionCache.GetOrBuild` returns an error by contract, and any future error path would silently degrade into a misaligned toy projection instead of failing.
- **Post-#221 the defaults are actively misleading.** The repo now carries executable guards asserting the real aliases; grepping for `ltbase_updated_at` still hits this production file.

## Reachability

[...]

Issue #220, opened 2026-07-12
refactor: split oversized files — e2e_harness/federated/query.go (1067), httpapi/server.go (747)


iceboundrock:
Recorded from PR #218's standards review (coding-standard.md: files <=500 lines, functions approaching 80 lines should be refactored).

Both files exceeded the limits before #218; that PR only made one-line/expression-level edits to them, so the split was deferred per the incremental-improvement rule.

## internal/sqlgen/duckdb_schema_projection.go (712 lines)

Natural seam: the benchmark-specific projections are a separate concern from the production schema projection.

- Move `BuildBenchmarkS3Projection`, `BuildBenchmarkOuterSelect`, `BuildBenchmarkProjections` (93 lines, above the 80-line refactoring trigger — the per-schema descriptor tables can become package-level vars or a lookup), `benchmarkAttr`, `eavJSONAttr`, `benchmarkEAVJSONArray` into a new `duckdb_benchmark_projection.go`.
- The production side (`BuildSchemaProjection`, pivot/outer-select builders, `duckDBAttrCast`) stays.

## internal/e2e_harness/federated/query.go (1188 lines)

Natural seam: tier-query construction vs execution/scan.

- Extract the parquet/PG tier-query builders (`buildParquetTierQuery`, `buildFederatedCombinedQuery`, `buildFederatedQueryCountSQLDynamic`, tradeTime filter-clause helpers) into e.g. `query_build.go`.
- Keep row scanning/normalization (`normalizeBenchmarkTradeTimeValue` etc.) and public harness entry points in `query.go`, or split scan into `query_scan.go` if either half still exceeds 500.

Pure moves, no behavior change; existing unit pins in `query_unit_test.go` and the sqlgen projection tests are the guardrails.

Refs #218.

iceboundrock:
[...]

PR #273, opened 2026-07-16
fix(federated): read dotted (nested) attributes through their parquet column aliases (#260)

### LINKED ISSUE #260: [federated] DuckDB read of schemas with nested (dotted) attributes fails with Binder Error
## Summary

A federated DuckDB read of a schema whose attributes contain dots (nested objects, e.g. `contact.annualIncome`) fails with a DuckDB `Binder Error`, because the projection emits the dotted attribute name **unquoted** — DuckDB then parses `contact.annualIncome` as `table.column` instead of a single column named `contact.annualIncome`.

## Repro

Server started with the federated DuckDB engine enabled (`DUCKDB_ENABLED=true` + S3 config), data flushed to delta parquet, then:

```
POST /api/v1/advanced_query
{"schema_name":"lead","condition":{"l":"and","c":[]},"page":1,"items_per_page":3,
 "federated":{"enabled":true,"include_execution_plan":true,
   "preferred_tiers":["warm","cold"],
   "s3_parquet_path_template":"s3://forma-cdc/delta/{{.SchemaID}}/*.parquet"}}
```

Response:

```
duckdb federated query: execute duckdb query: federated read failed:
Binder Error: Referenced table "contact" not found!
Candidate tables: "read_parquet"
LINE 12: ... contact.annualIncome, contact.annualIncomeCurrency, ...
```

`log` (a flat schema) reads cleanly and returns correct rows with `used_duckdb=true`; `lead` and `visit` (nested attributes) both fail.

## Where

The DuckDB projection builder emits attribute names directly into the `SELECT` list (`internal/sqlgen/duckdb_schema_projection.go` / `advanced_query_template_duckdb.go`). Dotted names must be quoted (`"contact.annualIncome"`) so DuckDB treats them as column identifiers.

## Impact

The federated/DuckDB read path is unusable for any schema with nested attributes. It was surfaced by the #190 e2e work, which enabled the DuckDB engine in the standalone server for the first time; the e2e DuckDB assertion targets the flat `log` schema to work around it.

## Context

Discovered while implementing #190 (PR #259). Not caused by that PR — pre-existing behavior first exercised there.

iceboundrock:
Closes #260.

## Problem
A federated DuckDB read of any schema with dotted attribute names failed with
`Binder Error: Referenced table "contact" not found`: the reader injected the
raw dotted name into every SQL fragment, so DuckDB parsed `contact.annualIncome`
as `table.column`.

## Root cause is two layers deep
Quoting alone (the issue's suggestion) cannot fix it: the CDC exporter has
always written parquet columns through a dot→underscore fold
(`contact_annualIncome`), so a quoted `"contact.annualIncome"` would miss the
physical column. And the logical WHERE clause renders both against raw
`read_parquet` (physical columns) and against the `visible` CTE (unified
columns), forcing unified names == physical names.

## Fix
- Extract the writer's fold into a shared contract `sqlgen.ParquetAttrColumn`;
  the CDC exporter delegates to it (byte-identical — existing S3 parquet needs
  no migration).
- Every reader emission site resolves through it: S3/PG projections, EAV
  pivot, outer select, DuckDB WHERE (`resolveDuckDBColumn`), non-keyset
  ORDER BY EAV branch. Keyset is out of scope (system columns only, validated
  upstream). Metadata maps stay keyed by logical names; fragment ordering stays
  by logical name (positional UNION ALL unchanged).
- Fail fast on lossy-fold collisions (`contact.name` vs `contact_name`) on
  both writer and reader, and propagate the reader guard through the engine
  path (`injectSchemaProjections` previously swallowed projection errors into
  the retired toy defaults, #222).
- Flat schemas render byte-identically (fold is identity without dots).

## Tests
- Red-first production E2E: new `e2e_nested` fixture (column-bound + EAV-only
  dotted attrs), DuckDB-forced warm+cold reads with dotted WHERE/ORDER BY and
  `UseDuckDB` gates, plus all-tier merge parity including unflushed hot rows.
- Test-harness fix surfaced by greening: the oracle now flattens nested event
  attributes into dotted names (mirroring the production write path); the
  flatten is behavior-neutral for flat fixtures.
- In-process DuckDB regression over real parquet with physical column names.
- Rendering unit tests for every fragment, collision-guard tests (unit +
  engine-level propagation).
- Full sweep green: `make test`, `make lint`, production E2E suite, federated
  `-short` + benchmark suites, infra smoke.

## Docs
- `docs/federated-query/design.md` §4.4: attribute → column naming contract.
- `tests/e2e/README.md`: flat-schema workaround retired; `--schema lead` now
  works with `--require-duckdb`.

## Follow-ups (non-blocking, triaged in final review)
- `SchemaProjection.EAVAttrsJSONSelect` is a pre-existing dead field — fold
  into #220's cleanup.
- Oracle flatten keeps `[]any` values whole (no fixture emits arrays; fails
  loudly, not silently, if one ever does).


iceboundrock:
## Findings

### P1: Folded attribute names can collide with Parquet system columns

`internal/cdc/export_sql_builder.go:119-143` and `internal/sqlgen/duckdb_schema_projection.go:126-138` reject only collisions between two attributes. They do not reserve the columns emitted by the CDC base projection, including `row_id`, `schema_id`, `changed_at`, and `deleted_at`.

A valid nested schema property such as `row.id` flattens to the logical attribute `row.id`, then folds to `row_id`. The cache validation currently permits it because it checks only duplicate IDs and main-column bindings. CDC then emits both the system `row_id` and the attribute alias `row_id`; downstream DuckDB projections can fail as ambiguous or read the system value as the attribute.

Seed the alias collision set with all CDC/CTE system-column names (and ideally validate this while registering the schema). Add regression coverage for a nested attribute such as `row.id`.

The existing collision tests cover `contact.name` versus `contact_name`, but not a folded attribute versus a reserved system column.

iceboundrock:
## Spec

Issue #260 requires federated DuckDB reads of nested/dotted attributes to stop treating `contact.annualIncome` as `contact`.`annualIncome`, while retaining correct behavior for the affected nested schemas and flat schemas.

Assessment: the shared physical-column mapping is an appropriate implementation. It matches the existing CDC Parquet representation, so aliasing to `contact_annualIncome` is safer than only quoting the logical name. The unit and production E2E coverage exercises projection, dotted WHERE, EAV WHERE, ordering, warm/cold reads, and all-tier merging. No unrelated scope creep found.

Residual coverage gap: the new production E2E uses a synthetic nested schema rather than the issue-named `lead` and `visit` schemas. The generic test provides meaningful mechanism coverage, but it would not detect metadata-specific regressions in those shipped schemas.

iceboundrock:
## Standards

`internal/sqlgen/duckdb_schema_projection.go` is now 730 lines. This violates the 500-line source-file limit in `coding-standard.md` §1 and `AGENTS.md` lines 66-71. The PR modifies this already oversized file but does not split the immediate projection/aliasing area.

Please extract cohesive projection construction helpers before merge; the new alias-validation and source-projection logic is a natural boundary.

iceboundrock:
## Assessment

**Request changes.** The dotted-attribute implementation is well targeted and verified, but the P1 reserved-column collision permits a valid schema to generate ambiguous/corrupt Parquet projections. The projection file-size violation also needs resolution under the repository coding standard.

After reserving system column aliases, adding regression coverage, and splitting the modified projection area, this should be ready to merge.

iceboundrock:
## Observations

- Extracting `sqlgen.ParquetAttrColumn` makes the existing CDC writer representation an explicit reader/writer contract and preserves compatibility with already-flushed Parquet.
- Projection errors now propagate through `injectSchemaProjections` rather than silently falling through, which is the correct behavior for a read-path storage-contract inconsistency.
- The new attribute-to-attribute collision guard executes when building CDC or reader projections. Moving it to schema registration would reject an unusable schema before it accepts hot-tier writes, rather than at its first flush/query.

iceboundrock:
## Verification

Executed against PR head `5ec9dd460944118db7b231caf3d3ddab99b3482e` in an isolated worktree:

- `git diff --check origin/main...origin/pr/273`
- `GOCACHE=... GOFLAGS=-buildvcs=false go test ./internal/sqlgen ./internal/cdc ./internal/federated`
- `make test`
- `make lint`
- `GOCACHE=... GOFLAGS=-buildvcs=false go test -tags=e2e -v ./internal/e2e_harness/production -run TestFederatedReadNestedDottedAttributes -timeout=5m`

All commands passed. The focused production E2E completed successfully with Postgres and RustFS testcontainers.

iceboundrock:
## Summary

Review of #273 against issue #260 and repository standards found:

- 1 P1 correctness finding: folded attributes can collide with reserved Parquet system columns.
- 1 standards violation: `internal/sqlgen/duckdb_schema_projection.go` exceeds the 500-line limit.
- Spec alignment is otherwise strong; no unrelated scope creep was identified.
- Unit, lint, and the focused production dotted-attribute E2E pass.

Do not merge until the reserved-column collision and file-size violation are addressed.

iceboundrock:
## Review response

### P1: folded attribute names colliding with system columns — fixed (e961f7b)

Verified the failure mode: an attribute `row.id` folds to `row_id`, the writer emits it alongside the system `row_id` in the COPY select, and the reader's s3_source/pg_source projections duplicate it again — the prior guards only checked attr-vs-attr.

Fix consolidates all folding validation into one shared contract, `sqlgen.ValidateParquetAttrColumns`:

- **Reserved set** covers the parquet export columns (`schema_id`, `row_id`, `changed_at`, `deleted_at`, `ltbase_row_id/schema_id/created_at/updated_at/deleted_at`), the unified-CTE system columns (`created_at`, `ver_ts`, `deleted_ts`), the dedup machinery (`source_tier_priority`, `rn`), and the outer-select tail (`attributes_json`, `total_records`, `total_pages`, `current_page`).
- **Wired at three layers**, per the Observations comment's registration suggestion — implemented, not deferred: `schemameta.validateSchemaAttributeCache` (covers both the file registry and the DB loader, so an unusable schema is rejected before it accepts hot-tier writes), plus the CDC writer and the federated reader as defense in depth. The engine-level propagation test from the previous round guarantees the reader-side rejection reaches the caller.
- **Regression coverage**: `row.id` at all three layers (`TestFileSchemaRegistryRejectsReservedFoldedColumnAtLoad`, `TestBuildSchemaDrivenProjectionReservedColumn`, `TestBuildSchemaProjection_ReservedColumnFails`), a reserved-set table test (8 folded/literal variants incl. `rn`, `source_tier_priority`), a fold-collision-at-load test, and a negative case pinning that camelCase (`createdAt`) does not fold onto snake_case system columns.

`make test` confirms no existing fixture/schema trips the new registration check.

### Standards: 500-line violation — fixed (6cd583d)

Split along the seam already recorded in #220: the benchmark-schema (100-102) projections moved to `duckdb_benchmark_projection.go` (290 lines), leaving `duckdb_schema_projection.go` at 447. Pure mechanical move, no behavior change; existing projection tests are the guardrail. The #220 remainder (`internal/e2e_harness/federated/query.go`) is untouched here — different subsystem, still tracked there.

### Spec residual gap: synthetic fixture vs shipped `lead`/`visit` — responding, no change

`lead`/`visit` live in the bun/TS black-box stack (`tests/e2e/schemas/`), not the Go production harness, so the harness can't exercise them without duplicating their metadata into a second fixture set. The shipped-schema coverage for exactly those schemas is `bun run federated-check -- --schema lead --require-duckdb` (README updated in this PR to drop the flat-schema workaround) — operator-run against the real server + real schema files, which is where a metadata-specific regression in `lead`/`visit` would surface. The synthetic `e2e_nested` fixture keeps the mechanism (column-bound + EAV dotted attrs) in CI. If you want `lead`/`visit` mirrored into the production harness fixtures as a CI gate, I'd file it as a follow-up rather than grow this PR.

### Verification of this round

`make test` (0 failures), `make lint` clean, `go test ./internal/sqlgen ./internal/federated ./internal/cdc ./internal/schemameta` green, and the focused production e2e `TestFederatedReadNestedDottedAttributes` re-run green at `6cd583d` — the registration-time check now also runs against every harness fixture at env setup, so the e2e doubles as its integration proof.

Issue #274, opened 2026-07-17
deleted_ts encoding asymmetry: base exports 0, delta exports NULL — equal-ver_ts ties resolved by a NULLS LAST accident, not an explicit rule


iceboundrock:
Follow-up to #210 (fix direction adjudicated 2026-07-16: option 1, init exports `ltbase_updated_at AS changed_at`; this asymmetry was explicitly scoped out).

## Current state

- Base export COALESCEs the tombstone column: `COALESCE(m.ltbase_deleted_at, 0) AS deleted_at` (`internal/cdc/init_exporter.go`) → live base rows carry `deleted_ts = 0`.
- Delta export passes it through raw: `cl.deleted_at` (`internal/cdc/duckdb_exporter.go:170`) → live delta rows carry `deleted_ts = NULL`.
- The federated dedup ranks `ver_ts DESC, source_tier_priority DESC, deleted_ts DESC, row_id ASC` (`internal/sqlgen/advanced_query_template_duckdb.go`); all parquet rows share tier priority 1. On an equal-ver_ts tie the winner falls to `deleted_ts DESC`, where DuckDB's default NULLS LAST sorts `0` before `NULL` — the base copy wins by an encoding accident, not a designed precedence (adjudication A3 arm A on #183, characterized in `tiebreak_lww_e2e_test.go`).

## Why this is benign today (and why it still deserves a deliberate rule)

Since #210, cdc-init stamps base ver_ts from `ltbase_updated_at`, which is written from the same clock read as `change_log.changed_at` in the same transaction. An equal-ver_ts base/delta tie therefore only arises between copies encoding the **same version with identical attributes** (create→flush→init with no intervening update, or a base copy tying its own newest flushed delta) — whichever copy wins, the served row is identical. There is no longer a reachable divergent-value tie through production export paths.

But the resolution rule remains an accident of encoding:

1. Any future change to either export's deleted_at encoding (or to DuckDB's NULL ordering defaults) silently flips tie winners.
2. Equal-timestamp **divergent** versions of a row (two same-ms writes) are still unranked below `row_id ASC` — undefined winner.
3. `testEqualVertsTombstoneWins` (#183) pins the tombstone side of the same key: a tombstone's `deleted_ts = T1` beating live `0`/`NULL` is load-bearing and must survive any normalization.

## Options

[...]

Issue #200, opened 2026-07-10
sqlgen: federated date/datetime predicates bind time.Time as CAST(? AS TIMESTAMP) against epoch-ms BIGINT attribute columns


iceboundrock:
Follow-up from #194 (PR #191, epic #172).

## Problem

The federated DuckClause handles date/datetime predicate parameters as DuckDB TIMESTAMP: `ToDuckDBParam` / `MapValueTypeToDuckDBType` (`internal/sqlgen/duckdb_type_mapper.go:32-34,90-103`) bind a `time.Time` wrapped in `CAST(? AS TIMESTAMP)`, and `parseDuckDBRawParam` (`internal/sqlgen/dualpath_sql_helpers.go:128-134`) converts epoch-ms literals to `time.Time` for the same purpose.

But every date/datetime attribute column in the federated CTEs is epoch-ms BIGINT — three sides agree on that convention:

- EAV pivot: `castEAVValue` emits `TRY_CAST(value_numeric AS BIGINT)` (aligned in cb0b1bf / #173);
- main-column unix_ms export: `castMainValue` emits `TRY_CAST(<col> AS BIGINT)` (aligned in #194);
- reader projection: `duckDBAttrCast` casts date attrs with `CAST(attr AS BIGINT)` (`internal/sqlgen/duckdb_schema_projection.go`).

So any federated query with a date/datetime filter compares `<BIGINT> <op> CAST(? AS TIMESTAMP)` — a DuckDB binder error. This affects both EAV-only date attributes (e.g. `e2e_wide.born`/`seen`) and main-column-bound ones (`joined`/`touched`). It is currently unexercised: the production smoke battery deliberately has no date filter (only a sort on `touched`, which references the column without binding a parameter).

## Suggested fix

Normalize date/datetime predicate parameters on the DuckClause side to epoch-ms int64 and emit a plain BIGINT comparison (`attr >= ?`), matching the storage convention. Verify against rendered SQL, not just unit expectations (cf. the #139 lesson).

Sync points to update alongside:
- pinned characterization tests that pin `CAST(? AS TIMESTAMP)`: `internal/sqlgen/predicate_characterization_test.go`, `duck_clause_test.go`, `duckdb_type_mapper_test.go`, `duckdb_sql_generator_test.go`;
[...]

PR #275, opened 2026-07-17
fix(cdc): #210 stamp base export ver_ts from ltbase_updated_at, not created_at

### LINKED ISSUE #210: cdc-init exports created_at as LWW version timestamp: backfilled base loses to previously flushed updates (deterministic stale reads) and ties with same-timestamp versions

iceboundrock:
Closes #210.

## Problem

`cdc-init` exported `m.ltbase_created_at AS changed_at`, so a base row carrying post-create attribute state ranked at its **creation** time in the federated LWW dedup (`ver_ts DESC, source_tier_priority DESC, deleted_ts DESC, row_id ASC`; all parquet rows share tier priority 1). Two failure modes, both probe-confirmed in the issue:

- **FM1 — deterministic stale read**: create(T1) → update v2(T2) → flush (delta v2@T2) → update v3(T3) → init (base v3@**T1**) → onboarding change_log cleanup (#176). The stale delta v2 outranks the fresh base v3 on every read until the row's next mutation.
- **FM2 — divergent equal-ver_ts tie**: create(T1) → flush (delta v1@T1) → update v2(T2) → init (base v2@**T1**). Both copies tie on ver_ts with different attributes; base won only through the `deleted_ts` 0-vs-NULL NULLS LAST accident (#183's A3 arm-A characterization, comment-pinned "revisit when #210 lands").

## Fix (adjudicated direction 1)

One production line: `init_exporter.go` now stamps `m.ltbase_updated_at AS changed_at`.

Why this is semantically exact, not just newer-is-better:
- Every write path (Insert/Update/BatchInsert/BatchUpdate, soft delete via the update statement) stamps `ltbase_updated_at` **and** `change_log.changed_at` from the **same clock read in the same transaction** (`postgres_persistent_repository.go:156/173`, `_batch.go:76/85`). The base copy therefore ranks exactly at its newest flushed delta — never below a previously flushed older version, never above a later hot write.
- `GREATEST(created, updated)` is unnecessary: `updated_at >= created_at` always holds (equal at insert).
- Hard deletes remove the row entirely and init exports `activeOnly`, so no path leaves `updated_at` stale.

Both failure modes vanish: FM1 because base now carries T3 > T2; FM2 because the intervening update lifts base to T2 — equal-ver_ts base/delta ties now only occur between copies encoding the **same version with identical attributes** (winner-indifferent).

## Commits (red-first)

1. `test(cdc)`: FM1 e2e `TestInitVerTsLWW/fresh_base_beats_stale_delta` through production verbs only — failed on main with the issue's exact diff (`expected fresh-v3/500000, actual stale-v2/400000`).
2. `fix(cdc)`: the one-line stamp change + flipped pinned expectation in `export_sql_shared_test.go` (verified red before the fix).
3. `test(e2e)`: ripple — see below.

## Test ripple

- `tiebreak_lww_e2e_test.go` (#183): `testEqualVertsDivergentProbe` **rewritten** as `testInitRestampBeatsStaleDelta` — same construction, but it now pins the fix (base wins by strict recency; A3 arm A retired). It routes through `AssertQueryMatches` since the oracle no longer has a tie to beg the question on; the rn=1 exactly-once invariant stays. `equal_verts_identical_copies` and `equal_verts_tombstone_wins` are untouched (still real ties, by same-version identity resp. restamp).
- `full_type_parquet_e2e_test.go`: base `changed_at` now pinned to `ltbase_updated_at`.
- `three_tier_merge_e2e_test.go` (#177): assertions unchanged (still green); the "base copy written LAST carries the OLDEST ver_ts" premise text updated.
- Comment sweep over remaining `#210` references: strictly-newer guards (`assertStrictlyNewer` etc.) kept — equal-timestamp **divergent** versions of a row remain unranked below `row_id ASC`; only comments stating the old init stamp as current fact were updated.
- Compaction needs **no change**: `merge_sql.go` passes `changed_at` through verbatim and its LWW fold now simply picks the correct version.

## Verification

- `make test` green, `make lint` clean (pinned v1.64.8).
- Targeted e2e: `TestInitVerTsLWW | TestWarmColdTiebreak | TestThreeTierMerge | TestNullAndBoundaryRoundTripAcrossTiers | TestCompactionRewrite* | TestFilterLWW*` — green.
- Full `make test-e2e-production` green (106s); `federated -tags=e2e -short` + benchmark green.

## Follow-up

- #274: the `deleted_ts` 0(base)-vs-NULL(delta) encoding asymmetry — post-fix ties are winner-indifferent, but the resolution still rests on a NULLS LAST accident rather than an explicit rule.

## Operational note

Existing pre-generated heavy-live datasets carry base files stamped under the old create-time contract. Rows updated-after-create before their init keep the old (losable) ver_ts until re-generated (same precedent as #200's dataset regeneration note). New inits are correct immediately; no migration of already-exported base files is performed here.

### DIFF
--- a/internal/cdc/export_sql_shared_test.go
+++ b/internal/cdc/export_sql_shared_test.go
@@ -87,7 +87,7 @@ func TestBuildExportSQL_CommonSemanticsAcrossModes(t *testing.T) {
 			},
 			expectedMemoryLimit:  "6GB",
 			expectedModeTable:    "",
-			expectedTimeSlotExpr: "m.ltbase_created_at AS changed_at",
+			expectedTimeSlotExpr: "m.ltbase_updated_at AS changed_at",
 			expectActiveOnly:     true,
 		},
 	}
--- a/internal/cdc/init_exporter.go
+++ b/internal/cdc/init_exporter.go
@@ -22,7 +22,11 @@ func (e *DuckExporter) ExportBaseFileToTmp(ctx context.Context, cfg CDCConfig, p
 // buildBaseExportSQL constructs the DuckDB SQL for exporting base files.
 // Key differences from delta export:
 // 1. No change_log table involved
-// 2. Uses ltbase_created_at as changed_at (the reader's version timestamp)
+// 2. Uses ltbase_updated_at as changed_at (the reader's version timestamp):
+//    the base row IS the state as of its last update, and every write path
+//    stamps ltbase_updated_at and change_log.changed_at from the same clock
+//    read, so the base copy ranks exactly at its newest flushed delta — never
+//    below a previously flushed older version (#210)
 // 3. Queries entity_main directly for row metadata
 func buildBaseExportSQL(pgConnStr string, s3TmpPath string, cfg CDCConfig, schemaID int16, rowIDs []uuid.UUID, attrCache forma.SchemaAttributeCache) (string, string, string, error) {
 	plan, err := buildExportSQLPlan(exportModeSpec{
@@ -31,7 +35,7 @@ func buildBaseExportSQL(pgConnStr string, s3TmpPath string, cfg CDCConfig, schem
 		useChangeLog:       false,
 		schemaIDSelect:     "m.ltbase_schema_id AS schema_id",
 		rowIDSelect:        "m.ltbase_row_id AS row_id",
-		timeSlotSelect:     "m.ltbase_created_at AS changed_at",
+		timeSlotSelect:     "m.ltbase_updated_at AS changed_at",
 		deletedAtSelect:    "COALESCE(m.ltbase_deleted_at, 0) AS deleted_at",
 	}, pgConnStr, s3TmpPath, cfg, schemaID, 0, rowIDs, attrCache)
 	if err != nil {
--- a/internal/cdc/init_exporter.go
+++ b/internal/cdc/init_exporter.go
@@ -2,6 +2,7 @@ package cdc
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/google/uuid"
 	"github.com/lychee-technology/forma"
@@ -21,13 +22,13 @@ func (e *DuckExporter) ExportBaseFileToTmp(ctx context.Context, cfg CDCConfig, p
 
 // buildBaseExportSQL constructs the DuckDB SQL for exporting base files.
 // Key differences from delta export:
-// 1. No change_log table involved
-// 2. Uses ltbase_updated_at as changed_at (the reader's version timestamp):
-//    the base row IS the state as of its last update, and every write path
-//    stamps ltbase_updated_at and change_log.changed_at from the same clock
-//    read, so the base copy ranks exactly at its newest flushed delta — never
-//    below a previously flushed older version (#210)
-// 3. Queries entity_main directly for row metadata
+//  1. No change_log table involved
+//  2. Uses ltbase_updated_at as changed_at (the reader's version timestamp):
+//     the base row IS the state as of its last update, and every write path
+//     stamps ltbase_updated_at and change_log.changed_at from the same clock
+//     read, so the base copy ranks exactly at its newest flushed delta — never
+//     below a previously flushed older version (#210)
+//  3. Queries entity_main directly for row metadata
 func buildBaseExportSQL(pgConnStr string, s3TmpPath string, cfg CDCConfig, schemaID int16, rowIDs []uuid.UUID, attrCache forma.SchemaAttributeCache) (string, string, string, error) {
 	plan, err := buildExportSQLPlan(exportModeSpec{
 		defaultMemoryLimit: "8GB",
@@ -39,7 +40,7 @@ func buildBaseExportSQL(pgConnStr string, s3TmpPath string, cfg CDCConfig, schem
 		deletedAtSelect:    "COALESCE(m.ltbase_deleted_at, 0) AS deleted_at",
 	}, pgConnStr, s3TmpPath, cfg, schemaID, 0, rowIDs, attrCache)
 	if err != nil {
-		return "", "", "", err
+		return "", "", "", fmt.Errorf("failed to build base export SQL plan for schema %d: %w", schemaID, err)
 	}
 
 	return plan.sql, plan.mainQuery, plan.eavQuery, nil
--- a/internal/e2e_harness/production/filter_lww_e2e_test.go
+++ b/internal/e2e_harness/production/filter_lww_e2e_test.go
@@ -80,12 +80,13 @@ func testStaleFilterDeltaGenerations(ctx context.Context, t *testing.T, env *Env
 
 // testStaleFilterBaseVsDelta is issue #178 scenario 1 in the base-vs-delta
 // layout: v1 (matching) in base via init + onboarding cleanup, v2
-// (non-matching) in a delta flushed afterwards. Deterministic under #210:
-// base ver_ts = create time < delta ver_ts = update time.
+// (non-matching) in a delta flushed afterwards. Deterministic ordering: base
+// ver_ts = v1's ltbase_updated_at (#210), equal to its create time since the
+// row was never updated before init, < delta ver_ts = v2's update time.
 func testStaleFilterBaseVsDelta(ctx context.Context, t *testing.T, env *Env, wide SchemaRef) {
 	creates := buildOpenCreates(wide, 5)
 	mustApplyEvents(ctx, t, env, "apply creates", creates...)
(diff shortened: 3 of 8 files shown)


iceboundrock (commented):
## Findings

- **P2 - nondeterministic regression test:** `internal/e2e_harness/production/init_ver_ts_lww_e2e_test.go:60,65` requires `v2.ChangedAt < v3.ChangedAt`, but write timestamps come from independent `time.Now().UnixMilli()` calls and the repository does not enforce a per-row strictly increasing clock. When the two updates land in one millisecond, `assertStrictlyNewer` fails before exercising the init/LWW assertion. Make the fixture establish a strict timestamp ordering deterministically (for example, wait/retry before the next write), rather than treating an equal timestamp as an unrelated test failure.

## Spec

Pass. The change implements #210 direction 1 exactly: base export now uses `m.ltbase_updated_at AS changed_at`; the production-verb test covers the stale-delta sequence, and the prior divergent-tie probe is converted to a strict-recency assertion. No scope creep found.

## Standards

- `internal/cdc/init_exporter.go:42` returns the `buildExportSQLPlan` error without context. This is in the modified function and violates `AGENTS.md:69` and `coding-standard.md:22-32`; wrap it with operation context.
- `gofmt -d` changes the newly edited numbered comment in `internal/cdc/init_exporter.go`. Run `gofmt` on the touched Go files, as required by `coding-standard.md:77-80`.

## Assessment

The production fix is minimal and correct for #210. Requesting the deterministic E2E fixture adjustment before merge; the error-wrapping and formatting cleanup should land with it.

## Observations

The new comment's absolute “never below” wording assumes wall-clock monotonicity. `nowMillis()` is raw `time.Now().UnixMilli()` with no transactional monotonicity guarantee, so a system-clock rollback remains an existing LWW limitation. This is not a regression introduced by the PR and is outside #210's requested fix.

## Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/cdc` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/cdc -run '^TestBuildExportSQL_CommonSemanticsAcrossModes$' -v` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -tags=e2e -run '^TestInitVerTsLWW$' ./internal/e2e_harness/production -timeout=5m -v` passed.
- `golangci-lint run --timeout=5m` passed.
- `git diff --check origin/main...origin/pr-275` passed.

## Summary

One P2 test-determinism finding and two standards cleanup items; #210 specification review passes.

iceboundrock:
## Review round-1 response (a16327a)

**P2 — nondeterministic regression test (`init_ver_ts_lww_e2e_test.go:60,65`)** — Fixed. Added `waitClockPast` (in `filter_lww_helpers_test.go`, next to `assertStrictlyNewer`): it blocks until the test-process wall clock is strictly past the prior event's `ChangedAt`. The repository stamps `changed_at` from this same process clock (`nowMillis`, `postgres_persistent_repository.go:63-67`, and the harness runs the `EntityManager` in-process), so the next write is guaranteed a strictly later millisecond — the ordering is now established deterministically rather than asserted after the fact. `assertStrictlyNewer` stays as the documented invariant check; it can no longer fire spuriously.

Also applied to `testInitRestampBeatsStaleDelta` (`tiebreak_lww_e2e_test.go:111`), which is in this PR's diff and had the identical create→update same-millisecond race. The pre-existing `assertStrictlyNewer` call sites (`filter_lww_*`, `keyset_lww`) share the same latent race but are outside this PR's diff — can file a follow-up to sweep them onto `waitClockPast` if wanted.

**Standards — bare error return (`init_exporter.go:42`)** — Fixed: wrapped as `failed to build base export SQL plan for schema %d: %w`. Note the delta-side sibling `buildExportSQL` (`duckdb_exporter.go:172-174`) has the same bare return; that's pre-existing outside this diff and left for the same follow-up sweep.

**Standards — gofmt** — Fixed: the renumbered comment in `init_exporter.go` is now gofmt-clean. For the record, `gofmt -l` also flags `boundary_roundtrip_e2e_test.go`, `full_type_parquet_e2e_test.go`, and `list_roundtrip_e2e_test.go` — I diffed against `origin/main` and that dirt is byte-identical there (pre-existing, not introduced by this PR's edits), so it was left untouched to keep the diff scoped.

**Observation — wall-clock monotonicity** — No change, agreeing with your own scoping. One clarification on the comment's wording: the "ranks exactly at its newest flushed delta" claim rests on same-transaction *equality* of `ltbase_updated_at` and `change_log.changed_at` (one clock read), which holds regardless of clock direction. What a clock rollback can break is *cross-version* ranking — and that is the LWW system's global `changed_at`-ordering assumption, shared by every tier and untouched by this PR.

**Verification** — `go test ./internal/cdc` green; `make lint` clean (pinned v1.64.8); `gofmt -l internal/cdc` clean; `go test -tags=e2e -run 'TestInitVerTsLWW|TestWarmColdTiebreak' ./internal/e2e_harness/production` green.

Issue #212, opened 2026-07-11
federated: keyset cursor filters versions before LWW dedup — stale version can resurrect on paginated reads


iceboundrock:
Found while implementing #178 (epic #172), via the plan's keyset decision-gate probe. Adjacent to #183 (sort/pagination stability), which does not cover version resurrection.

## Problem

The advanced DuckDB template applies the keyset cursor predicate in the `ranked` CTE **before** LWW dedup (`internal/sqlgen/advanced_query_template_duckdb.go:83-85` — cursor injected at line 84, `rn = 1` selected only afterwards in `visible` at line 91). SQL `WHERE` evaluates before window functions, so the cursor filters row **versions**, not LWW winners. For a cursor over a mutable sort attribute, a superseded version that satisfies the cursor can win `rn = 1` and resurrect — the keyset-cursor twin of the filter-before-LWW bug fixed for logical predicates in #173 (bd1c829) and pinned by #178.

## Reproduction (production e2e harness, real DuckDB engine)

Row A: v1 `count=900` (delta #1) → v2 `count=100` (delta #2, both flushed, strictly increasing changed_at). Rows B (`count=400`) and C (`count=600`) stable. Query sorted by `count` ascending with cursor after `count=500`:

```go
page, err := env.Query(ctx, Query{
    Schema: wide,
    Sorts:  []Sort{{Attr: "count"}},
    Keyset: &model.KeysetCursor{
        Columns: []model.KeysetColumn{{Attribute: "count", Direction: forma.SortOrderAsc}},
        Values:  []any{float64(500)},
        Mode:    model.KeysetCursorModeAfter,
    },
    Limit: 10,
})
```

Correct result: exactly `[C(600)]` — A's live version (100) sorts before the cursor. Observed:

```
keyset page = [
  {row_id=...7cd9}   <- row-c (count=600), correct
  {row_id=...716b}   <- row-a v1 (count=900), STALE — must not appear
]
```

## Rendered-SQL evidence (captured from page.Plan.Sources)

```sql
ranked AS (
  SELECT *,
    ROW_NUMBER() OVER (
      PARTITION BY row_id
      ORDER BY ver_ts DESC, source_tier_priority DESC, deleted_ts DESC, row_id ASC
    ) AS rn
  FROM unified
  WHERE
    ((count > $1)) AND        -- $1 = 500   <- KEYSET CURSOR, pre-dedup
    1=1
),
visible AS (
  SELECT *
  FROM ranked
[...]

PR #278, opened 2026-07-17
chore(federated): #217 delete the dead $n-placeholder keyset codegen

### LINKED ISSUE #217: federated: delete the dead $n-placeholder keyset codegen in internal/federated/keyset.go — latent #161 mixed-placeholder trap
Follow-up from #212 (fixed in PR #216), flagged by that PR's final review.

## Problem

PR #216 unified the live keyset SQL generation (`internal/sqlgen/keyset_where.go`) on positional `?` placeholders, because DuckDB treats `$n` as an absolute parameter index and shifts every `?` after it — the #161 mixed-placeholder mis-binding. But `internal/federated/keyset.go` still carries a second, older copy of the same codegen that emits `$n` with a `paramOffset`:

- `generateKeysetWhereClause` (`keyset.go:27-77`) — `$n` + `paramOffset`, one arg per column (no per-occurrence repetition)
- `keysetComparisonOp` (`keyset.go:81-96`) — only called by the above
- `buildKeysetOrderBy` (`keyset.go:176-189`)
- `extractCursorFromRecord` → `recordColumnValue` → `eavColumnValue` (`keyset.go:98-147`) — cursor-extraction chain whose EAV leg is a stub that always returns `nil`

None of these has a non-test caller — they are kept alive solely by `internal/federated/keyset_test.go`. Verified by grep across the repo: zero production references.

## Why it matters

The `$n` copy is a latent trap, not just dead weight: anyone who later wires the paginated-engine branch (or any new call site) to build SQL from this package's `generateKeysetWhereClause` silently reintroduces the exact mixed-placeholder mis-binding #212 just eliminated. Its arg contract also diverges from the live one (`internal/sqlgen` now appends prefix values once per placeholder occurrence — n(n+1)/2 args for n columns; this copy appends one per column), so the two functions with the same name have incompatible semantics.

## What stays

`validateKeysetColumns` / `isSupportedKeysetColumn` (`keyset.go:149-171`) are NOT dead: `executeFederatedKeysetQuery` (`pagination.go:153`) uses them, and that branch is reachable via the benchmark harness (`internal/e2e_harness/federated/benchmark/execute.go:239-243` sets `KeysetEnabled: true` and calls `ExecuteFederatedPaginatedQuery`). Keep them (or relocate them if `keyset.go` empties out around them).

## Suggested change

- Delete the six dead functions above and their cases in `internal/federated/keyset_test.go`.
- Keep `validateKeysetColumns` / `isSupportedKeysetColumn` and their tests.
- `go build ./... && make test` as the gate; no behavior change expected on any path.

## Relations

- #212 / PR #216 — the fix that made the `$n` copy divergent
- #161 — the original mixed-placeholder mis-binding class
- #195 — the single-scan rewrite should find exactly one keyset codegen when it lands

iceboundrock:
Closes #217.

## What

Deletes the six dead functions in `internal/federated/keyset.go` — the older `$n`-placeholder copy of the keyset codegen — and their tests. Re-verified against current main before deleting: zero production callers for all six (`generateKeysetWhereClause`, `keysetComparisonOp`, `extractCursorFromRecord`, `recordColumnValue`, `eavColumnValue`, `buildKeysetOrderBy`); they were kept alive solely by `keyset_test.go`.

## Why (from the issue)

The `$n` copy is a latent #161 trap: DuckDB treats `$n` as an absolute parameter index and shifts every `?` after it — the exact mixed-placeholder mis-binding #212/PR #216 eliminated from the live generator. Its arg contract had also diverged from `internal/sqlgen/keyset_where.go` (one arg per column vs one per placeholder occurrence), so two same-named functions had incompatible semantics. Any future call site wired to this copy would silently reintroduce the bug.

## What stays (verified live)

- `validateKeysetColumns` / `isSupportedKeysetColumn` — production caller `pagination.go:155` (`executeFederatedKeysetQuery`, reachable via benchmark `KeysetEnabled`).
- `validateKeysetTiebreak` (added by #183, post-issue) — production callers `pagination.go:158` and `engine.go:123`.
- All 8 tests for the validators.

`keyset.go` gains a header comment pointing at the single remaining codegen (`internal/sqlgen/keyset_where.go`) and why this copy was removed.

## Ripple

`internal/e2e_harness/production/pagination_mutation_e2e_test.go:71` comment no longer names the deleted `extractCursorFromRecord`.

## Verification

- `go build ./...`, `make test`, `make lint` (pinned v1.64.8) — all green.
- Repo-wide grep for the six names: remaining hits are only the distinct same-named live implementations in `internal/sqlgen` (untouched).
- No behavior change expected on any path; diff is -325/+9.


iceboundrock:
## Findings

No actionable findings.

## Spec

Meets #217: the six unreferenced `$n`-placeholder generator/cursor helpers and their tests are removed; live keyset validators and their tests remain. The only non-deletion change correctly removes the stale reference to the deleted helper in `internal/e2e_harness/production/pagination_mutation_e2e_test.go:71`.

## Standards

No documented standards violations. Changed files remain within size limits and introduce no error-handling, naming, nesting, or layering problems.

## Assessment

Acceptable dead-code removal that eliminates the latent mixed-placeholder regression path without changing the live generator in `internal/sqlgen/keyset_where.go`.

## Observations

`internal/federated/keyset.go:8-14` now gives maintainers a clear single-source-of-truth pointer and explains why positional `?` placeholders are required.

## Verification

Reviewed the full diff, issue #217, and production call sites for retained validators and live SQL generation. `gh pr checks --required` reports no required checks configured for this branch; author-reported build/test/lint results were not independently rerun.

## Summary

No findings; the PR satisfies #217 and retains the live validation seams.

Issue #194, opened 2026-07-10
cdc: main-column-bound date/datetime (unix_ms) encoding mismatch between exporter and reader


iceboundrock:
Follow-up from PR #191 (#173, epic #172).

## Problem

For **main-column-bound** date/datetime attributes with `unix_ms` encoding, the CDC DuckDB exporter converts values to DuckDB DATE/TIMESTAMP types in the parquet output, while the production federated reader expects raw epoch-ms BIGINT. Warm/cold reads of such attributes would decode wrong values or fail to bind.

Not currently exercised: the production harness fixtures deliberately avoid main-column-bound dates (`e2e_wide` keeps date/datetime as EAV-only), which is a known coverage gap noted in the PR.

## Suggested fix

Same alignment treatment as PR #191's EAV numeric fix (commit cb0b1bf): make the exporter emit exactly what the reader consumes, verified against both mainline write and read sides — do not assume, verify the storage convention first (cf. the #139 lesson).

## Acceptance

- `e2e_wide` gains a main-column-bound date/datetime attribute and the production smoke passes across all three tiers.
- Byte/value parity test between hot and warm/cold reads of the bound date attribute.

PR #279, opened 2026-07-17
fix(cdc): #219 normalize non-unix_ms date main columns to epoch-ms on export

### LINKED ISSUE #219: federated: non-unix_ms date main-column encodings misalign with the reader's CAST(attr AS BIGINT) contract
Recorded from #200 (out of scope there).

ISO8601 text-encoded date main columns, and the default TRY_CAST(<col> AS DATE/TIMESTAMP) branch of castMainValue (internal/cdc/duckdb_exporter.go:413), export values the federated reader projection cannot consume: duckDBAttrCast casts every date/datetime attribute with CAST(attr AS BIGINT), and (since #200 / PR #218) the DuckClause binds epoch-ms BIGINT params. Any schema using a non-unix_ms date encoding would fail federated reads/filters. No fixture uses those encodings today.

Options: normalize castMainValue's date export to epoch-ms for all encodings, or reject non-unix_ms date column bindings at schema registration.

Refs #200, #194, #173.

iceboundrock:
Closes #219. Refs #200, #194, #173.

## Direction changed after review: normalize, don't reject

The first cut rejected non-unix_ms date main-column bindings at registration. Review surfaced that this would also forbid **iso8601 date columns on the Postgres/OLTP path**, which are a legitimately-supported, tested capability (`parseDateValue` → RFC3339 text; `TestHybrid_DateTimeISO8601Encoding`). After tracing the corruption mechanism, the adjudicated direction is to **normalize the export** so the OLTP capability is preserved *and* the federated path serves it correctly.

## The corruption mechanism

An iso8601 datetime is stored as an RFC3339 **text** string (`transform.storeInMainColumn`). On the OLTP read path this is compared text-to-text — correct. But:

1. **CDC flush** (`castMainValue`): only `unix_ms` emitted epoch-ms `BIGINT`; iso8601/default fell to `TRY_CAST(col AS DATE/TIMESTAMP)` → a native DATE/TIMESTAMP parquet column.
2. **Federated read** (`duckDBAttrCast`): projects every date attr as `CAST(attr AS BIGINT)` → `CAST(DATE AS BIGINT)` = days-since-epoch, `CAST(TIMESTAMP AS BIGINT)` = micros-since-epoch.
3. **Predicate bind** (`ToDuckDBParam`, #200): epoch-**ms** int64.

So on the warm/cold tiers an iso8601 date came back off by 10³–10⁸× from the epoch-ms the filter compared against — silent wrong rows, no error. And since **CDC flushes every schema**, this hit any iso8601 (or default) date column once flushed, regardless of query routing.

## Fix

`castDateMainValue` normalizes every date/datetime encoding to epoch-ms `BIGINT`:
- **iso8601** (RFC3339 text): `epoch_ms(TRY_CAST(col AS TIMESTAMP))`
- **unix_ms / default** (epoch millis in a bigint column): `TRY_CAST(col AS BIGINT)`

The reader is unchanged — it now receives a consistent epoch-ms BIGINT from all encodings on every tier. No registration rejection; the OLTP iso8601 path and its tests are untouched.

## Tests

- `TestCastMainValue_DateAndDatetimeEncodings` extended: iso8601 and default now assert epoch-ms output (old expectations pinned the buggy native casts).
- **New `TestCastDateMainValue_ExecutesToEpochMs`** runs the emitted expressions against a real in-process DuckDB: iso8601 text `'2024-01-02T03:04:05Z'`, unix_ms bigint, and default bigint all evaluate to `1704164645000` and read back unchanged through `CAST(attr AS BIGINT)`.

## Verification

- `make test` — 26 packages ok, zero FAIL. `make lint` clean (pinned v1.64.8), `gofmt` clean.
- `go test -tags=e2e -short ./internal/e2e_harness/federated/...` green (116s + 50s).

### DIFF
--- a/internal/cdc/cast_date_duckdb_test.go
+++ b/internal/cdc/cast_date_duckdb_test.go
@@ -1,0 +1,68 @@
+package cdc
+
+import (
+	"database/sql"
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCastDateMainValue_ExecutesToEpochMs proves the #219 normalization end to
+// end against a real DuckDB engine: the expression castMainValue emits for each
+// date/datetime encoding evaluates to the SAME epoch-ms BIGINT, and that value
+// reads back unchanged through the federated reader's CAST(attr AS BIGINT)
+// projection. Pre-#219 the iso8601/default branches produced native
+// DATE/TIMESTAMP columns that CAST(... AS BIGINT) reinterpreted as
+// days/microseconds — a silent 10^3-10^8x scale error on the warm/cold tiers.
+func TestCastDateMainValue_ExecutesToEpochMs(t *testing.T) {
+	db, err := sql.Open("duckdb", ":memory:")
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, db.Close()) })
+
+	// 2024-01-02T03:04:05Z == 1704164645000 ms since the epoch.
+	const wantEpochMs int64 = 1704164645000
+
+	cases := []struct {
+		name string
+		meta forma.AttributeMetadata
+		// source SQL column value expression matching how the write path stores
+		// this encoding: iso8601 -> RFC3339 text; unix_ms/default -> epoch-ms bigint.
+		sourceExpr string
+	}{
+		{
+			name:       "iso8601_text",
+			meta:       forma.AttributeMetadata{ValueType: forma.ValueTypeDateTime, ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumnText02, Encoding: forma.MainColumnEncodingISO8601}},
+			sourceExpr: "'2024-01-02T03:04:05Z'",
+		},
+		{
+			name:       "unix_ms_bigint",
+			meta:       forma.AttributeMetadata{ValueType: forma.ValueTypeDateTime, ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumnBigint02, Encoding: forma.MainColumnEncodingUnixMs}},
+			sourceExpr: "1704164645000::BIGINT",
+		},
+		{
+			name:       "default_bigint",
+			meta:       forma.AttributeMetadata{ValueType: forma.ValueTypeDate, ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumnBigint02}},
+			sourceExpr: "1704164645000::BIGINT",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Export side: evaluate the exporter's cast over the stored value.
+			exportExpr := castMainValue("m", tc.meta)
+			var exported int64
+			require.NoError(t,
+				db.QueryRow("SELECT "+exportExpr+" FROM (SELECT "+tc.sourceExpr+" AS m) t").Scan(&exported),
+				"export expr %q must evaluate", exportExpr)
+			require.Equal(t, wantEpochMs, exported, "export must yield epoch-ms")
+
+			// Read side: the federated reader casts the parquet column with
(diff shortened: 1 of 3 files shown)


iceboundrock:
## Findings

- P1: `internal/schemameta/loader.go:52-70` exposes `MetadataCache.RegisterSchema`, which stores a supplied cache without calling `validateSchemaAttributeCache`. A caller can therefore register an `iso8601` or default-encoded Date/DateTime main-column binding and retain the broken CDC export/federated-read combination this PR is intended to reject. The PR description states that all three registration entry points funnel through the validator, but this direct registration path does not. Validate before mutating cache state and add a regression test through `RegisterSchema`.

## Spec

Partial. The file/DB metadata loading route now rejects the invalid binding as #219 requires, but direct metadata-cache registration remains a bypass, so registration-time enforcement is incomplete.

## Standards

`internal/schemameta/validation.go:51` adds a bare `return err`. This violates the contextual error-wrapping rule in `AGENTS.md:69` and `coding-standard.md:22-33`; wrap it at the `validateSchemaAttributeCache` boundary. The selected plain operator-error class and detailed message otherwise match `docs/error-handling.md:20-45`.

## Assessment

Changes requested. The validator logic and tests are sound for the file-loader route, but the public registration bypass leaves the reported federated inconsistency reachable.

## Observations

Sorted attribute iteration makes the reported first violation deterministic, and the test matrix correctly covers `iso8601`, default encoding, `unix_ms`, EAV storage, and unrelated encodings.

## Verification

Reviewed the full PR diff, #219, `validation.go`, and the PR-head `loader.go` registration path. `gh pr checks --required` reports no required checks configured for this branch; author-reported test/lint results were not independently rerun.

## Summary

One P1 finding and one standards violation: route the new validation through `RegisterSchema`, wrap its propagated error, and add the direct-registration regression test.

iceboundrock:
## Review response — direction changed to normalize (d1da297)

**P1 (RegisterSchema bypass) + the rejection approach — resolved by pivoting.** Verifying the P1 finding surfaced that closing the `RegisterSchema` bypass would also forbid iso8601 date columns on the **Postgres/OLTP path**, which are legitimately supported and tested (`parseDateValue` → RFC3339 text; `TestHybrid_DateTimeISO8601Encoding`, `TestHybrid_MainColumn_AnchorTrue`). So the whole registration-rejection direction was too broad: iso8601 date columns aren't broken on OLTP, only on the federated read path.

Traced the actual corruption: `castMainValue` exported non-unix_ms dates as native DATE/TIMESTAMP parquet columns, which the reader's `CAST(attr AS BIGINT)` reinterpreted as days/micros — off from the epoch-ms convention by 10³–10⁸×. Because CDC flushes every schema, that's corrupt lakehouse data the moment an iso8601 date column is flushed, regardless of routing.

Adjudicated fix: **normalize the export** instead of rejecting the encoding. `castDateMainValue` now maps every date/datetime encoding to epoch-ms BIGINT (iso8601 → `epoch_ms(TRY_CAST(col AS TIMESTAMP))`; unix_ms/default → `TRY_CAST(col AS BIGINT)`). This preserves the OLTP iso8601 capability and fixes the federated path in one place. The registration validator and its RegisterSchema-bypass concern are moot — there's no date-encoding rejection anymore.

**Standards (bare `return err` in the old `validateSchemaAttributeCache`) — moot.** That code is reverted; `validation.go` now matches main.

**Note for follow-up:** the P1 comment also exposed a genuine, independent gap — `MetadataCache.RegisterSchema` never runs `validateSchemaAttributeCache` at all, so it bypasses the #260 folding-collision guard too. That's orthogonal to #219; happy to file it separately if you want the public registration path to enforce that invariant.

Proof the normalized expression is correct against a real engine: new `TestCastDateMainValue_ExecutesToEpochMs` runs iso8601/unix_ms/default through in-process DuckDB — all yield `1704164645000` and round-trip through `CAST(attr AS BIGINT)`. `make test`/`make lint`/federated `-short` e2e green.

PR #280, opened 2026-07-17
fix(sqlgen): #222 retire the toy-schema projection defaults in injectDuckDBTemplateParams

### LINKED ISSUE #222: sqlgen: retire the toy-schema projection defaults in injectDuckDBTemplateParams (phantom ltbase_* S3 aliases)

iceboundrock:
Closes #222. Refs #214, #221, #260.

## Problem

`injectDuckDBTemplateParams` (`internal/sqlgen/duckdb_template_renderer.go`) substituted a retired toy-schema projection — `name/age/tag`, EAV attr 205, phantom `ltbase_created_at/updated_at/deleted_at` parquet aliases that don't exist in any CDC-produced parquet — for each of the seven schema-projection params a caller left unset. This was a second live copy of the exact fallback `injectSchemaProjections` (`internal/federated/duckdb_query.go`) already fails fast on: its column set can't be scanned by `duckDBScanBuffers` under the #147 positional-scan contract, so a missing param was "a hard error at best, misaligned rows at worst".

## Fix

Delete the seven defaults; add `requireProjectionParams` — an advanced-template render (`BuildDuckDBQuery` dual + legacy nil paths, and `CompileDuckDBQuery`) now fails fast naming every missing projection key. Production always sets all seven via the engine's `injectSchemaProjections` (`BuildSchemaProjection` for schema-cache queries, `BuildBenchmarkProjections` for benchmark schemas), so **no production path changes behavior** — the guard only bites callers that never had a valid projection.

### Issue item 3 already delivered by #260

The issue also asked that `injectSchemaProjections` propagate `projErr` instead of swallowing it. #260 (PR #273, `ec9add2`) already did this — `duckdb_query.go:253-272` propagates and hard-fails on an empty cache, pinned by `TestInjectSchemaProjectionsPropagatesCollision`. No further change there.

## Tests

The renderers now refuse an empty param map, so every test that drove the advanced template on the toy defaults derives real projection params from `BuildSchemaProjection`:

- New shared sqlgen helper `projection_params_test.go` (`testProjectionCache` / `applyProjectionParams` / `withTestProjection`); `compiledParityParams` now takes `t` and injects the fixture projection, which fixes the bulk of the sqlgen call sites in one place.
- `internal/health_and_renderer_test.go` and `internal/federated/duckdb_federated_integration_test.go` inline the same fixture (cross-package `_test` helpers aren't importable).
- New guard tests pin the missing-param rejection on both `BuildDuckDBQuery` and `CompileDuckDBQuery` (all keys named; partial-set names only the absent ones; legacy nil path guarded too).

## Acceptance (grep condition corrected)

The issue's original grep (`ltbase_updated_at\|ltbase_deleted_at` → no hits outside absence-asserting tests) was written before #221/#260 landed and is now stale. The real acceptance line: **`duckdb_template_renderer.go` has zero hits.** Remaining `internal/sqlgen` hits are all legitimate — the reserved-column set (`parquet_attr_column.go:60-61`) and the real runtime projections (`duckdb_schema_projection.go:270-271`, `duckdb_benchmark_projection.go:91-92`), which must stay.

## Verification

- `make test` — all 26 packages ok, zero FAIL. `make lint` clean (pinned v1.64.8).
- `go test -tags=e2e -short ./internal/e2e_harness/federated/...` green (render-path regression gate; 111s + 47s).
- `grep -n 'ltbase_updated_at\|ltbase_deleted_at' internal/sqlgen/duckdb_template_renderer.go` — no hits.

### DIFF
--- a/internal/federated/dotted_attrs_read_test.go
+++ b/internal/federated/dotted_attrs_read_test.go
@@ -147,3 +147,23 @@ func TestInjectSchemaProjectionsPropagatesCollision(t *testing.T) {
 	// propagated error must leave them untouched.
 	require.NotContains(t, sqlParams, "S3SourceSelect")
 }
+
+// TestApplySchemaProjection_NilProjectionFails pins the nil-projection guard
+// (#222/#280 review): a nil projection with no error must be refused, never
+// treated as success. The real schemaProjection seam never yields sp==nil
+// without an error, so the guard is exercised directly.
+func TestApplySchemaProjection_NilProjectionFails(t *testing.T) {
+	sqlParams := map[string]any{}
+	err := applySchemaProjection(sqlParams, 42, nil, nil)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "42")
+	require.NotContains(t, sqlParams, "S3SourceSelect", "a refused projection must leave params unset")
+
+	// A non-nil projErr is still wrapped with schema context.
+	perr := applySchemaProjection(sqlParams, 42, nil, errTestProjection)
+	require.ErrorIs(t, perr, errTestProjection)
+	require.ErrorContains(t, perr, "42")
+	require.NotContains(t, sqlParams, "S3SourceSelect")
+}
+
+var errTestProjection = fmt.Errorf("boom")
--- a/internal/federated/duckdb_federated_integration_test.go
+++ b/internal/federated/duckdb_federated_integration_test.go
@@ -128,6 +128,24 @@ func TestBuildDuckDBQuery_AdvancedTemplate(t *testing.T) {
 		"Offset":               0,
 		"PageSize":             20,
 	}
+	// The advanced template refuses to render without schema projection params
+	// (#222 retired the toy defaults); derive them like injectSchemaProjections
+	// does, from the same fixture shape design_doc_sql_test.go pins.
+	sp, err := sqlgen.BuildSchemaProjection(1, forma.SchemaAttributeCache{
+		"name": {AttributeID: 1, ValueType: forma.ValueTypeText,
+			ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumn("text_01")}},
+		"age": {AttributeID: 2, ValueType: forma.ValueTypeInteger,
+			ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumn("integer_01")}},
+		"tag": {AttributeID: 205, ValueType: forma.ValueTypeText},
+	})
+	require.NoError(t, err)
+	params["S3SourceSelect"] = sp.S3SourceSelect
+	params["PGSourceSelect"] = sp.PGSourceSelect
+	params["PGGroupBy"] = sp.PGGroupBy
+	params["EAVPivotSelect"] = sp.EAVPivotSelect
+	params["EAVPivotAttrs"] = sp.EAVPivotAttrs
+	params["HasEAVPivot"] = sp.EAVPivotAttrs != ""
+	params["OuterSelect"] = sp.OuterSelect
 	sql, args, err := sqlgen.BuildDuckDBQuery(tmpl, params, q, []uuid.UUID{}, nil)
 	require.NoError(t, err)
 	require.NotEmpty(t, sql)
--- a/internal/federated/duckdb_query.go
+++ b/internal/federated/duckdb_query.go
@@ -252,15 +252,24 @@ func (e *DBFederatedQueryEngine) injectSchemaProjections(sqlParams map[string]an
 
 	// Production schema: compute projections from the attribute cache
 	sp, hit, projErr := e.schemaProjection(schemaID, cache)
+	return hit, applySchemaProjection(sqlParams, schemaID, sp, projErr)
+}
+
+// applySchemaProjection writes the seven schema-projection params into sqlParams
+// from a computed projection, or refuses. Both a non-nil projErr and a nil
+// projection are hard failures: with the toy-schema defaults retired (#222)
+// there is nothing to fall through to, so an advanced-template render must fail
+// fast rather than emit unset projection params. Extracted so the nil-guard is
+// unit-testable (the real schemaProjection seam never yields sp==nil without an
+// error).
+func applySchemaProjection(sqlParams map[string]any, schemaID int16, sp *sqlgen.SchemaProjection, projErr error) error {
 	if projErr != nil {
 		// A non-nil projection error is a hard data-contract failure (e.g. the
-		// alias-collision guard in BuildSchemaProjection). Propagate it so the
-		// query fails fast instead of silently falling through to the retired
-		// toy-schema projection defaults (#222/#260).
-		return hit, fmt.Errorf("build schema projection for schema %d: %w", schemaID, projErr)
+		// alias-collision guard in BuildSchemaProjection). Propagate it.
+		return fmt.Errorf("build schema projection for schema %d: %w", schemaID, projErr)
 	}
 	if sp == nil {
-		return hit, nil
+		return fmt.Errorf("schema projection for schema %d is nil with no error; refusing to render with an undefined projection", schemaID)
 	}
 	sqlParams["S3SourceSelect"] = sp.S3SourceSelect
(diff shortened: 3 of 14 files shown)


iceboundrock:
## Findings

- P2: `internal/federated/duckdb_query.go:262-264` still treats `sp == nil` as success (`return hit, nil`). #222 explicitly requires `injectSchemaProjections` to propagate both `projErr` and a nil projection rather than silently continue. This PR relies on the advanced-template guard to fail later, but the engine seam itself still returns success and non-advanced/custom templates are not protected. Return a contextual error for nil `sp` and add a regression test.

## Spec

Partial. The retired toy projection defaults are removed, tests now provide real projections, and missing advanced-template projection params fail loudly. The required nil-projection branch in `injectSchemaProjections` remains unimplemented despite the PR statement that #260 completed it.

## Standards

- `internal/sqlgen/duckdb_template_renderer.go:33-137`: the modified `BuildDuckDBQuery` is now 105 lines, exceeding the 100-line limit in `coding-standard.md:7-11`.
- `internal/sqlgen/duckdb_template_renderer.go:92,130` and `internal/sqlgen/duckdb_compiled_query.go:72` add bare error returns, contrary to `AGENTS.md:69` and `coding-standard.md:22-33`. Add operation context.
- `internal/sqlgen/projection_params_test.go:21,48`: private helper names `testProjectionCache` and `withTestProjection` are not verb-first as required by `coding-standard.md:52-56`.

## Assessment

Changes requested. The primary cleanup is directionally correct and removes the unsafe fallback, but the incomplete nil-projection contract and the documented standards regressions should be resolved before merge.

## Observations

The sorted `projectionParamKeys` list produces deterministic errors and centralizes the seven-key contract. Consider a table-driven test that begins with a valid map and removes each key individually; current tests cover all-missing and one partial case but do not isolate every required key. The two cross-package test fixtures duplicate more than 15 lines; this is a maintainability observation, not a blocker.

## Verification

Reviewed the full PR diff, #222, the PR-head renderer, and the PR-head `injectSchemaProjections` implementation. `gh pr checks --required` reports no required checks configured for this branch; author-reported unit/lint/e2e results were not independently rerun.

## Summary

One P2 spec finding and three standards findings. Fail on nil schema projections, add its regression test, then address the function-size, error-context, and helper-naming violations.

iceboundrock:
## Review response (90bc1da)

**P2 — nil projection treated as success** — Fixed. Extracted the projection-application logic into `applySchemaProjection(sqlParams, schemaID, sp, projErr)`, which now refuses a nil projection with a contextual error, symmetric with the `projErr` and empty-cache guards. The engine seam is unchanged behaviorally on every reachable state: `schemaProjection` → `GetOrBuild` never yields `sp==nil` without an error (`BuildSchemaProjection` returns non-nil on success; `GetOrBuild` only caches non-nil), so this is defense-in-depth. New `TestApplySchemaProjection_NilProjectionFails` exercises the guard directly (nil→error, params left unset) plus the projErr wrap path.

Note on your spec observation: #260 delivered the `projErr` propagation (that part was done); the `sp == nil` branch was the remaining piece and is now closed. My PR statement conflated the two — the nil branch was not covered by #260.

**Standards — `BuildDuckDBQuery` > 100 lines** — Fixed: the `dual == nil` fallback is extracted into `buildDuckDBQueryLegacy`; `BuildDuckDBQuery` is now 72 lines.

**Standards — bare error returns (`renderer:92,130`, `compiled_query:72`)** — Fixed: the three `requireProjectionParams` returns now wrap with the entry point (`build DuckDB query:` / `compile DuckDB query:`), which also disambiguates the direct vs compiled render path. `%w` preserves the inner message, so the missing-key/`#222` assertions still hold.

**Standards — helper names not verb-first** — Fixed: `testProjectionCache` → `buildTestProjectionCache`, `withTestProjection` → `injectTestProjection`.

**Observation — table-driven per-key isolation test** — not added; the existing all-missing + one-partial cases plus the deterministic `projectionParamKeys` list cover the contract, and a per-key sweep would be marginal. Happy to add if you'd prefer the exhaustive form.

**Verification** — `make test` (26 pkgs, 0 FAIL), `make lint` clean, `gofmt` clean. Federated `-short` e2e green (116s + 48s). One full-suite run flaked at `TestSuite_Integration` under load (181s vs the normal 116s); it passes in isolation (3.09s) and on a clean full re-run — the known e2e-under-contention flake, not a regression (`applySchemaProjection` is behavior-equivalent on all reachable states).

Issue #283, opened 2026-07-17
Split internal/transform/attribute_converter.go (583 lines, over the 500-line cap)


iceboundrock:
`internal/transform/attribute_converter.go` was already over the 500-line cap before #205 (558 lines) and the #205 write-path fix added +25 (now 583). The branch honored the cap where it crossed a threshold itself (`transformer.go` was split into `typed_value.go` at 503), but retro-splitting a pre-existing violator inside a bugfix branch would have been unrelated-rewrite scope creep — so the split lands here instead (per coding-standard: fix the immediate area, file the follow-up).

Natural seam from the #205 final review: the converter half (`ToEAVRecord`/`FromEAVRecord`/`ToEAVRecords`/`FromEAVRecords` + their small helpers) vs the extract-value half (`extractValueFromEAVRecord` + the `to*ForEAV` conversion helpers). Pure mechanical move, no signature changes; `git diff --stat` should show deletions ≈ additions.

Refs #205 (review finding m3).

iceboundrock:
## Already resolved — closing as stale

Verified against `main` @ `3ec288d`:

```
$ wc -l internal/transform/attribute_converter.go
479 internal/transform/attribute_converter.go
```

The file is **479 lines**, under the 500-line cap — down from the 583 recorded here. The exact seam this issue proposed (converter half vs extract-value half) was already carved out as a side effect of #204 / PR #288:

```
internal/transform/attribute_converter.go            # ToEAVRecord/FromEAVRecord/ToEAVRecords/FromEAVRecords
internal/transform/attribute_value_extraction.go     # extractValueFromEAVRecord + to*ForEAV helpers
internal/transform/attribute_value_extraction_test.go
```

`git log` on the file confirms the shrink landed in `579e78a` (#204, PR #288).

No cap violation remains and the proposed split exists, so there is nothing left to do here. Note for readers arriving from #294: that issue changes `FromEAVRecords` in `attribute_converter.go`, which is now comfortably under cap.

Remaining >500-line non-test files on `main`, for the record:

| lines | file | tracked in |
|---|---|---|
| 1067 | `internal/e2e_harness/federated/query.go` | #220 |
| 747 | `internal/httpapi/server.go` | #220 (added) |
| 501 | `internal/e2e_harness/federated/harness.go` | — (borderline) |

Issue #282, opened 2026-07-17
HTTP JSON decoding rides float64: integer payloads above 2^53 are rounded at the door (UseNumber decision)


iceboundrock:
Surfaced during #205. The write-path fix there makes bound bigint / unix_ms dates carry exact int64 end to end **for Go-API callers** (and string payloads): `numutil.Int64Exact` handles `int64`, `json.Number`, and integer strings before any float64 conversion.

But the HTTP entrypoint decodes request bodies with a plain `json.NewDecoder(r.Body).Decode(v)` (`internal/httpapi/server.go:619`) and never calls `UseNumber()`, so every JSON number arrives as `float64` before the transform chain ever sees it. For HTTP clients the practical integer ceiling is still 2^53 — the precision is lost at the door, outside the reach of #205's fix.

Decision needed (per #205 acceptance criterion 1's alternative branch):

1. **Adopt `dec.UseNumber()`** at the decode site so integers arrive as `json.Number` (which the transform chain already handles exactly). Requires auditing every consumer of the decoded `map[string]any` — JSON-schema validation, relation handling, query condition parsing, anything that type-switches on `float64` — since the dynamic type of numeric values changes globally. Non-trivial blast radius; needs its own test sweep.
2. **Document the 2^53 ceiling as the public HTTP API contract** for numeric payloads, and point Go-API/int64-exact users at the library path.

Adjacent note from the #205 final review (same exactness-coverage theme, Go API side): `int64ExactForEAV` tolerates `*int64` but not `*int`/`*int32` pointer shapes (which `toFloat64ForEAV` accepts), so those exotic inputs still ride the float64 fallback — identical to pre-#205 behavior, just not exact. Worth folding into whichever option is chosen.

Refs #205.

iceboundrock:
Scope update: PR #284's review round 1 (owner P1 finding) resolved the decision for the **entity write path** — `readEntityJSONBody` now decodes create/update payloads with `UseNumber` (commit `cd67796`), with a `toBool` json.Number case preserving numeric-truthiness payloads and handler-level regressions at ±MaxInt64. Remaining scope here: whether non-entity payloads (query conditions — overlaps #281's comparison-side audit) should follow, and the `*int`/`*int32` pointer-shape note for Go-API callers. The 2^53-ceiling documentation branch is no longer needed for writes.

iceboundrock:
## Decision: adopt `dec.UseNumber()` (option 1)

[...]

Issue #281, opened 2026-07-17
bigint predicate and keyset-cursor bindings above 2^53: comparison-side float64 exactness audit


iceboundrock:
Follow-up to #205, which fixed the **projection** path: bound BIGINT columns now survive storage and federated reads exactly across the full int64 range. The **comparison** paths were deliberately out of scope and deserve their own audit now that columns can legally hold exact values above 2^53:

- **WHERE predicates**: query condition values that arrive as float64 (e.g. via HTTP JSON, see the UseNumber issue) bind against now-BIGINT columns; `BIGINT col = DOUBLE param` unifies to DOUBLE in DuckDB, so an equality/range filter at e.g. `2^53+1` can mismatch rows that the projection returns exactly.
- **Keyset cursors**: `generateKeysetWhereClause` (`internal/sqlgen/keyset_where.go:47-51`) binds cursor values raw. Values sourced from `Int64Items` bind as int64 → exact BIGINT comparison (already correct). But if any continuation-token deserialization path yields float64, a `BIGINT col > DOUBLE param` comparison at a >2^53 page boundary could skip or duplicate one row (#205 final review, finding M-2 — hypothetical today, magnitude no worse than pre-#205 when the column itself was rounded).

Scope: trace how condition/cursor values reach the DuckDB and PG binders for bigint attrs, add >2^53 boundary probes for both dialects (filter + keyset continuation), and fix or document what they reveal.

Refs #205.

iceboundrock:
## Audit result: where bigint comparison values lost exactness (and what was done)

The issue asked to *trace* the comparison-value path and fix or document what it revealed. The trace is below; one live bug was found and fixed, one ceiling was documented, and the hypothesised keyset risk (M-2) was resolved as not-currently-reachable and fenced.

### The path

Condition values arrive as the `"op:value"` **string** DSL and stay strings end to end. So **#282 / `UseNumber` is a no-op for comparisons** — it fixes the write/validation path, not this one. Every float64 hop that mattered was inside `internal/sqlgen`.

### Fixed — predicates disagreed across tiers (the live bug)

Pre-fix, the DuckDB binder `parseDuckDBRawParam` (`internal/sqlgen/dualpath_sql_helpers.go:122-126`) parsed the literal with `ParseFloat`, and `ToDuckDBParam` rendered float64 with `%.15g` (`internal/sqlgen/duckdb_type_mapper.go:208-210`). Consequences:

[...]

Issue #276, opened 2026-07-17
test/cdc cleanup: deterministic clock ordering sweep for LWW probes + remaining bare-error/gofmt debt


iceboundrock:
Follow-up from PR #275 review round-1 (offered in https://github.com/Lychee-Technology/forma/pull/275#issuecomment-5005555435; user confirmed filing).

Small clock-ordering / cleanup debt, all pre-existing on main (none introduced by #275, which fixed only the surfaces inside its own diff):

- [ ] **`waitClockPast` sweep over the older LWW suites.** The pre-existing `assertStrictlyNewer` call sites share the same same-millisecond race #275's review flagged in the new probe: `changed_at` comes from independent in-process `time.Now().UnixMilli()` calls, so two writes landing in one millisecond fail the guard before the LWW assertion runs. #275 added `waitClockPast` (`internal/e2e_harness/production/filter_lww_helpers_test.go`) — it blocks until the process clock is strictly past the prior event's `ChangedAt`, which is deterministic because the repository stamps `changed_at` from the same process clock (`nowMillis`). Apply it before the subsequent write at each remaining call site:
  - `filter_lww_e2e_test.go:59,107`
  - `filter_lww_matrix_e2e_test.go:93`
  - `filter_lww_tombstone_e2e_test.go:32,86`
  - `keyset_lww_e2e_test.go:75,132,222`
- [ ] **`TestConcurrentFlushSnapshot/UpdateBeforeExport` snapshot race** (`internal/e2e_harness/production/cdc_concurrent_flush_e2e_test.go:228`). Same class, different anchor: the assertion `update.ChangedAt <= hookSnapshot` fails when the update applied inside `BeforeExportHook` lands in the same millisecond as the flush selection snapshot (`SelectBatchRowIDs` -> `snapshot := time.Now().UnixMilli()`, the same process clock the repository stamps `changed_at` from). Observed flaking CI on an unrelated PR (#280, actions run 29606517056); passes 5/5 locally. Fix: in the hook, spin until `time.Now().UnixMilli() > snapshot` before `env.ApplyEvents(update)` (a `snapshot`-taking variant of `waitClockPast`), so `update.ChangedAt > snapshot` is guaranteed. Check the sibling `assertStrictlyNewer` call sites in the same file (lines 288, 341) while there.
[...]

PR #284, opened 2026-07-17
fix(sqlgen,transform): #205 bound bigint carries full int64 — typed EAV pivot + exact write-path sidecar

### LINKED ISSUE #205: bound bigint precision: write path marshals int64 through float64, and federated reads pass BIGINT through DOUBLE (crash at MaxInt64, silent rounding above 2^53)
## Summary

Surfaced by the #174 boundary suite (`internal/e2e_harness/production/boundary_roundtrip_e2e_test.go`): a main-column-bound **bigint** attribute cannot faithfully carry the full int64 range. There are two independent float64 hops — one on the write path, one in the federated read projection — and together they mean:

- int64 payloads above 2^53 are silently rounded on write;
- `+MaxInt64` "survives" only by int64→float64→int64 saturation (a false positive), while `-MaxInt64` comes back as `MinInt64` (off by one);
- a stored `MaxInt64` (e.g. written by external tooling directly to Postgres) makes every federated read of that schema **throw**:

```
Conversion Error: Type DOUBLE with value 9.223372036854776e+18
    can't be cast because the value is out of range for INT64
    LINE 143: CAST(amount AS BIGINT) AS bigint_01
```

The empirically faithful ceiling today is **±2^62** (a power of two, float64-exact), documented in the probe `TestBoundBigintAcceptsInt64Probe` and the `maxBoundBigint` doc comment.

## Hop 1 — write path marshals numerics through float64

- `transform.populateTypedValue` (`internal/transform/transformer.go:351-356`) converts every numeric-family attribute (`numeric`, `bigint`, `integer`, `smallint`) with `numutil.Float64(value)` into `model.EAVRecord.ValueNumeric *float64` — including an exact `int64` payload (`internal/numutil/convert.go:24`). Column-bound attributes travel through this representation too.
- `transform.storeInMainColumn` (`internal/transform/persistent_record.go:209-212`, `MainColumnTypeBigint` case) casts back: `record.Int64Items[columnName] = int64(*attr.ValueNumeric)`. The same hop applies to unix_ms dates (`persistent_record.go:160`).

This is where `+MaxInt64` saturates and `-MaxInt64` lands on `MinInt64`.

## Hop 2 — federated read passes bound BIGINT columns through DOUBLE (all tiers)

- `buildPGProjection` (`internal/sqlgen/duckdb_schema_projection.go:197-198`) projects a bound attr as `COALESCE(ANY_VALUE(hot_vals.amount), m.bigint_01) AS amount`. `hot_vals.amount` comes from the EAV pivot `MAX(CASE WHEN attr_id = N THEN value_numeric END)` (`duckdb_schema_projection.go:226-228`), and `eav_data.value_numeric` is unconstrained PG `NUMERIC`, which DuckDB's postgres scan surfaces as DOUBLE. `COALESCE` type-unifies DOUBLE+BIGINT → DOUBLE for **every** hot row — even though a column-bound attr never has an `eav_data` row and the pivot side is always NULL.
- `unified` (`internal/sqlgen/advanced_query_template_duckdb.go:70-74`) is `SELECT * FROM s3_source UNION ALL SELECT * FROM pg_source`; `s3_source` projects the parquet BIGINT bare (`duckdb_schema_projection.go:150`), and UNION ALL type unification of BIGINT with DOUBLE yields DOUBLE — so warm/cold parquet values also pass through DOUBLE before the outer `CAST(amount AS BIGINT)` (`duckdb_schema_projection.go:405-406`, placed at `:296-297`) converts back.

Consequence: any federated read using the EAV-pivot PG projection silently rounds bound bigint values in (2^53, 2^63) that aren't float64-exact — hot, warm, and cold tiers alike — and crashes outright at `MaxInt64`. (The UNION-unification leg is read from the template/DuckDB semantics; the observed crash directly proves the pg_source leg.)

Two exact-path escapes exist today:
- `PreferHot` reads bypass DuckDB entirely (direct PG path) — which is why hot `Int64Items` assertions in the #174 suite are exact;
- `BuildPGSelectNoEAV` (`duckdb_schema_projection.go:344-365`) selects `m.bigint_01` bare with no DOUBLE hop, but is only used when all filter/sort attributes are column-bound.

## Impact

- The practical bound-bigint contract is ±2^62 (float64-exact powers of two survive; other values above 2^53 round silently).
- A legitimately stored near-max BIGINT turns federated queries into hard failures (operator-visible read-path error).
- Any future fix to Hop 1 alone (e.g. accepting `json.Number`/int64 end to end) would *increase* exposure to Hop 2's silent rounding, so the read-path projection should be fixed first or together.

## Acceptance criteria

- Either the float64 write-path bound is documented as the public API contract for numeric attributes, or the write path preserves int64 exactly for bound bigint columns (and unix_ms dates).
- The federated projection does not pass bound BIGINT main columns through DOUBLE: cast the EAV-pivot side per attribute type (or restructure the COALESCE) so BIGINT survives type unification, including through the `UNION ALL`.
- `CAST(... AS BIGINT)` on the read path cannot crash on values a legal Postgres BIGINT can hold.
- The #174 probe's `maxBoundBigint` (`internal/e2e_harness/production/boundary_roundtrip_e2e_test.go`) can be advanced from `1<<62` to `math.MaxInt64` with the test passing across all tiers.

Epic #172. Surfaced by the #174 round-trip suite (same pattern as #192–#198).

iceboundrock:
Closes #205.

## Root cause (two independent float64 hops)

- **Read (Hop 2)**: the hot-tier EAV pivot emitted bare `MAX(CASE WHEN attr_id = N THEN value_numeric END)` — DOUBLE via postgres_scan — so `COALESCE(pivot, m.bigint_01)` type-unified to DOUBLE and the `UNION ALL` dragged the parquet legs' exact BIGINT with it. The outer `CAST(attr AS BIGINT)` then crashed on a stored `MaxInt64` (`Conversion Error … out of range for INT64`) and silently rounded everything in (2^53, 2^63) on **all three tiers**.
- **Write (Hop 1)**: the transform chain bounces JSON → EAVRecord → EntityAttribute → EAVRecord → main columns, riding `float64` twice — `-MaxInt64` deterministically landed as `MinInt64`; `+MaxInt64` "survived" only by platform-dependent saturation.

## Fix

**Read side — typed EAV pivot** (`sqlgen.eavPivotExpr`): per-type `TRY_CAST` (BIGINT for bigint/date/datetime, INTEGER/SMALLINT for the smaller ints; numeric stays DOUBLE, bool keeps its `<> 0` wrap), mirroring the CDC export side `cdc.castEAVValue` — this is tier-parity restoration, not a new convention; the two sides now cross-reference each other. `TRY_CAST` (not `CAST`) matches export semantics: an out-of-range NUMERIC in eav_data becomes NULL on every tier alike instead of a read-path crash (acceptance criterion 3). SELECT-fragment change only — placeholders, bindings, GROUP BY, dirty-set barrier all untouched.

**Write side — exact int64 sidecar** (`model.EAVRecord.ValueInt64 *int64` + `numutil.Int64Exact`): set at every chain entry for bigint and unix-ms dates, preferred by `storeInMainColumn`/`extractValueFromEAVRecord`, with `ValueNumeric` always still set as the fallback and the **unchanged eav_data persistence contract** (EAV-only attributes deliberately keep the documented 2^53 float64 ceiling; the sidecar is cleared before records land in `OtherAttributes` so the create echo matches what eav_data persists — final-review finding I-1). Non-integral / out-of-range inputs fall back to byte-identical pre-#205 behavior.

## Tripwires flipped (as their own doc comments prescribed)

- `TestFederatedReadRejectsStoredMaxInt64` → `TestFederatedReadReturnsStoredMaxInt64` (exact round-trip + `Plan.Routing.UseDuckDB` proof)
- `TestBoundBigintMaxInt64WriteCorrupts` → `TestBoundBigintPreservesNegMaxInt64` (deterministic on every platform)
- `maxBoundBigint`: `1<<62` → `math.MaxInt64` — the boundary matrix now writes ±MaxInt64 through the API across hot/warm/cold

## New tests

- `internal/sqlgen/duckdb_bigint_pivot_test.go`: pivot-shape pins + real-DuckDB engine semantics (old shape's crash mechanism, typed shape's exactness, UNION unification rules, iso8601 COALESCE parity guard)
- `internal/transform/bigint_precision_test.go`: write + read-back exactness, table-driven with deterministic all-platform red cases (the MaxInt64-only legs saturate on arm64 — implementation-defined — so ±MaxInt64 and 2^53+1 legs carry the portable pin), plus the EAV-only sidecar-clearing regression
- `numutil.Int64Exact`: 13-case table

## Evidence

- Full production harness e2e green (100.5s, includes the boundary matrix at ±MaxInt64 across all tiers, oracle parity intact); federated `-short` suite green; `make test` + `make lint` (pinned v1.64.8) clean; no #276 flake observed.
- Every task went through spec+quality review; final whole-branch review (adversarial pass over the 9 commits) verdict: Ready to merge, with I-1 fixed in `e679324` and re-verified (BatchCreate flows through the same cleared site).

## Review notes / accepted observations

- `design_doc_sql_test.go` byte-pins only the `s3_source` projection, not the `pg_source` pivot; §5's example schema has no numeric EAV attr today so there is nothing to pin — but a future numeric-EAV line in §5 would not be text-guarded (pre-existing, informational).
- `TestFederatedReadReturnsStoredMaxInt64` keeps its ExecSQL injection deliberately: it proves the read-path fix independently of the write-path fix, matching the issue's external-writer framing.

## Follow-ups filed

- #282 — HTTP JSON decoding rides float64 (`UseNumber` decision; HTTP clients still see the 2^53 ceiling at the door — #205 AC1's "document as public contract" branch)
- #281 — bigint predicate + keyset-cursor binding exactness above 2^53 (comparison side, out of #205's projection scope)
- #283 — split `attribute_converter.go` (583 lines, pre-existing over-cap +25 here)

### DIFF
--- a/internal/cdc/duckdb_exporter.go
+++ b/internal/cdc/duckdb_exporter.go
@@ -400,7 +400,7 @@ func castDateMainValue(col string, meta forma.AttributeMetadata) string {
 // uuid live in value_text. The output types mirror the federated reader's
 // EAV pivot (sqlgen.BuildSchemaProjection): BOOLEAN for bool, epoch-ms
 // BIGINT for dates, native numeric types otherwise (#173).
-// 热端 pivot(sqlgen.eavPivotExpr)必须与本映射保持一致(#205)。
+// 热端 pivot(sqlgen.buildEAVPivotExpr)必须与本映射保持一致(#205)。
 func castEAVValue(meta forma.AttributeMetadata) string {
 	switch meta.ValueType {
 	case forma.ValueTypeBool:
--- a/internal/cdc/duckdb_exporter.go
+++ b/internal/cdc/duckdb_exporter.go
@@ -400,6 +400,7 @@ func castDateMainValue(col string, meta forma.AttributeMetadata) string {
 // uuid live in value_text. The output types mirror the federated reader's
 // EAV pivot (sqlgen.BuildSchemaProjection): BOOLEAN for bool, epoch-ms
 // BIGINT for dates, native numeric types otherwise (#173).
+// 热端 pivot(sqlgen.eavPivotExpr)必须与本映射保持一致(#205)。
 func castEAVValue(meta forma.AttributeMetadata) string {
 	switch meta.ValueType {
 	case forma.ValueTypeBool:
--- a/internal/e2e_harness/production/boundary_roundtrip_e2e_test.go
+++ b/internal/e2e_harness/production/boundary_roundtrip_e2e_test.go
@@ -5,7 +5,6 @@ package production
 import (
 	"context"
 	"math"
-	"strings"
 	"testing"
 
 	"github.com/google/uuid"
@@ -116,25 +115,14 @@ func TestBoundBigintMaxInt64WriteCorrupts(t *testing.T) {
 	}
 }
 
-// TestFederatedReadRejectsStoredMaxInt64 pins the federated crash on a stored
-// math.MaxInt64 (#205). A stored MaxInt64 is a legal Postgres BIGINT state an
-// external writer can produce, but the forma write path cannot produce it
-// portably (float64(MaxInt64) is out of int64 range and the conversion is
-// implementation-dependent per the Go spec), so it is injected with ExecSQL —
-// the harness's documented escape hatch — rather than written through the API.
-//
-// The corruption is NOT in storage: the direct hot read returns MaxInt64
-// exactly, proving Postgres holds it losslessly. It is the federated DuckDB
-// projection that fails, because duckdb_schema_projection unifies the hot leg
-// and the parquet main column with
-// COALESCE(ANY_VALUE(hot_vals.amount), m.bigint_01) into DOUBLE
-// (internal/sqlgen/duckdb_schema_projection.go:197-198), and casting DOUBLE
-// 2^63 back to INT64 overflows.
-//
-// When #205 fixes the projection to preserve int64 the federated read will
-// succeed and the guard below fails the test on purpose: flip it to assert the
-// exact round-trip and raise maxBoundBigint to math.MaxInt64.
-func TestFederatedReadRejectsStoredMaxInt64(t *testing.T) {
+// TestFederatedReadReturnsStoredMaxInt64 proves the #205 Hop-2 fix: a stored
+// math.MaxInt64 — a legal Postgres BIGINT state an external writer can produce,
+// injected via ExecSQL because the pre-#205 write path could not produce it
+// portably — now survives the federated DuckDB projection exactly. Pre-#205
+// the COALESCE(DOUBLE pivot, m.bigint_01) unification made CAST-back overflow
+// ("Conversion Error ... out of range for INT64") on every federated read of
+// the schema.
+func TestFederatedReadReturnsStoredMaxInt64(t *testing.T) {
 	cluster := SharedCluster(t)
 	env := NewEnv(t, cluster)
 	ctx := context.Background()
@@ -144,18 +132,16 @@ func TestFederatedReadRejectsStoredMaxInt64(t *testing.T) {
 	if err := env.ApplyEvents(ctx, ev); err != nil {
 		t.Fatalf("apply create: %v", err)
 	}
-
 	// Base file so the federated query routes to DuckDB; the row stays
-	// unflushed in change_log, so it is still served from the hot pg_source leg.
+	// unflushed in change_log, so it is served from the hot pg_source leg.
 	if _, err := env.RunInit(ctx, wide); err != nil {
 		t.Fatalf("run init: %v", err)
 	}
-	// Inject a legal-but-API-unreachable stored MaxInt64.
 	env.ExecSQL(ctx,
 		"UPDATE entity_main SET bigint_01 = $1 WHERE ltbase_schema_id = $2",
 		int64(math.MaxInt64), wide.ID)
 
-	// Direct path is exact: Postgres holds MaxInt64 losslessly.
+	// Direct hot path: storage holds MaxInt64 losslessly (unchanged pre/post fix).
 	hot, err := env.Query(ctx, Query{Schema: wide, PreferHot: true, Limit: 10})
 	if err != nil {
 		t.Fatalf("hot query: %v", err)
@@ -164,20 +150,23 @@ func TestFederatedReadRejectsStoredMaxInt64(t *testing.T) {
 		t.Fatal("hot query returned no records")
 	}
 	if got := hot.Records[0].Int64Items["bigint_01"]; got != math.MaxInt64 {
-		t.Fatalf("stored bound bigint via hot read = %d, want %d — storage must hold MaxInt64 exactly",
-			got, int64(math.MaxInt64))
+		t.Fatalf("stored bound bigint via hot read = %d, want %d", got, int64(math.MaxInt64))
 	}
 
-	// Federated path crashes: the DOUBLE 2^63 → INT64 cast overflows.
-	_, ferr := env.Query(ctx, Query{Schema: wide, Limit: 10})
-	if ferr == nil {
-		t.Fatalf("federated read of a stored MaxInt64 succeeded — #205 appears fixed; " +
-			"flip this test to assert the exact round-trip and raise maxBoundBigint to math.MaxInt64")
+	// Federated path: must succeed and be exact (#205).
+	fed, ferr := env.Query(ctx, Query{Schema: wide, Limit: 10})
+	if ferr != nil {
+		t.Fatalf("federated read of stored MaxInt64 must succeed post-#205: %v", ferr)
 	}
-	if !strings.Contains(ferr.Error(), "out of range") {
-		t.Fatalf("federated error = %q, want the DOUBLE→INT64 out-of-range cast failure (#205)", ferr.Error())
+	if !fed.Plan.Routing.UseDuckDB {
+		t.Errorf("federated query did not route to duckdb: %+v", fed.Plan.Routing)
 	}
-	t.Logf("OBSERVED federated read failure on stored MaxInt64: %v", ferr)
+	if len(fed.Records) == 0 {
+		t.Fatal("federated query returned no records")
+	}
+	if got := fed.Records[0].Int64Items["bigint_01"]; got != math.MaxInt64 {
+		t.Fatalf("federated bound bigint = %d, want %d (exact MaxInt64)", got, int64(math.MaxInt64))
+	}
 }
 
 const zeroUUID = "00000000-0000-0000-0000-000000000000"
--- a/internal/e2e_harness/production/boundary_roundtrip_e2e_test.go
+++ b/internal/e2e_harness/production/boundary_roundtrip_e2e_test.go
@@ -196,9 +196,10 @@ func buildBoundaryEvents(wide SchemaRef) (nullRow, zeroRow, maxRow, minRow *Even
 // TestNullAndBoundaryRoundTripAcrossTiers proves that NULL, empty-string,
 // zero, and extreme boundary values survive write → CDC export → parquet →
 // federated merge-on-read intact across all three tiers. Null+max land
-// cold-only, zero+min stay warm, and fresh null+max copies stay hot. The
-// physical layer reuses Task 4's per-file value assertion; the logical layer
-// adds float64-proof exact typed assertions the oracle cannot make.
+// cold-only, zero+min plus a fresh warm max stay warm, and fresh null+max
+// copies stay hot — so +MaxInt64 is proven on all three tiers. The physical
+// layer reuses Task 4's per-file value assertion; the logical layer adds
+// float64-proof exact typed assertions the oracle cannot make.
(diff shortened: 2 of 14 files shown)


iceboundrock:
## Findings

### P1: The HTTP write API still corrupts full-range bound bigints

Issue #205 requires an exact bound-bigint write path **or** a documented public float64 bound. This PR only preserves exactness after an `int64` or `json.Number` reaches the transformer. The public create/update handlers still call `readJSONBody` into `any` / `map[string]any`, whose default `encoding/json.Decoder` turns JSON integer literals into `float64` (`internal/httpapi/server.go:81-102`, `269-283`, `616-620`).

Consequently, a normal request containing `9223372036854775807` reaches `populateTypedValue` as `float64(2^63)`. `Int64Exact` rejects that value and the existing `int64(*ValueNumeric)` fallback remains (`internal/transform/typed_value.go:35-45`, `internal/transform/persistent_record.go:213-220`). `-9223372036854775807` is already rounded to `-2^63` and is accepted as the wrong exact sidecar value. The PR therefore does not deliver its claimed full-range bound-bigint contract to HTTP clients.

Use `Decoder.UseNumber()` for entity input decoding and add create/update HTTP regressions for both signs, or explicitly document and retain the float64 public API contract.

iceboundrock:
## Spec

Issue #205 acceptance criteria assessment:

- **AC1, exact write path: not met.** See the P1 finding: default HTTP JSON decoding introduces the very `float64` hop this PR is meant to remove. No alternative public float64-bound contract is documented.
- **AC2, typed federated projection: implemented.** `eavPivotExpr` casts bigint/date/datetime to `BIGINT`, integer to `INTEGER`, and smallint to `SMALLINT` before `COALESCE` / `UNION ALL` type unification (`internal/sqlgen/duckdb_schema_projection.go:227-252`).
- **AC3, legal PostgreSQL BIGINT no longer crashes federated reads: implemented and E2E-covered.** `TestFederatedReadReturnsStoredMaxInt64` injects and federated-reads `math.MaxInt64` (`internal/e2e_harness/production/boundary_roundtrip_e2e_test.go:89-141`).
- **AC4, advance `maxBoundBigint` and prove all tiers: partial.** The value is advanced to `math.MaxInt64`, but `TestNullAndBoundaryRoundTripAcrossTiers` sends the positive maximum only to cold and hot; the warm/delta row contains `-math.MaxInt64` (`internal/e2e_harness/production/boundary_roundtrip_e2e_test.go:171-192`, `218-255`). Add a positive `math.MaxInt64` warm-tier assertion.

No scope creep found.

iceboundrock:
## Standards

- **Hard violation:** `internal/transform/attribute_converter.go` is 583 lines at `e679324`, up from 558 lines before this PR. The PR adds the `int64ExactForEAV` helper there (`:190-200`) even though `coding-standard.md` section 1 rejects source files over 500 lines; AGENTS.md also identifies that limit as load-bearing. Extract the conversion helper to the new typed-value file or another focused file before merge.
- **Naming observation:** `int64ExactForEAV` and `eavPivotExpr` are private helpers but do not use the standard's verb-first convention (`coding-standard.md` section 4). This is non-blocking relative to the size violation.

`gofmt -d` produced no diff, and `make lint` exited successfully.

iceboundrock:
## Assessment

The typed EAV pivot and the in-memory `ValueInt64` sidecar correctly address the two internal precision hops for callers that supply exact Go values. The DuckDB semantics tests and the targeted production E2E give good evidence for that core path.

Do not approve yet: the P1 HTTP decoding gap means the public write path still fails issue #205's primary contract. The warm-tier positive-MaxInt64 gap should be closed in the same regression suite.

iceboundrock:
## Observations

- The sidecar is deliberately cleared for EAV-only attributes before persistence, preventing create-response data from differing from the stored `eav_data` representation; the dedicated regression test covers this.
- The read-side change uses `TRY_CAST`, which preserves the CDC exporter's null-on-out-of-range behavior instead of introducing a new read-path crash.
- No unrelated behavior change or scope creep was identified.
- Existing tests do not cover HTTP decoding of entity numeric payloads at the int64 boundary.

iceboundrock:
## Verification

Reviewed fixed PR head `e679324b3906a238dd7d78087f80dfe04f31c6fe` in an isolated, clean detached worktree. Commands run successfully:

```text
GOCACHE=<temp> GOFLAGS=-buildvcs=false go test ./internal/numutil ./internal/transform ./internal/sqlgen
GOCACHE=<temp> GOFLAGS=-buildvcs=false go test ./internal/httpapi
GOCACHE=<temp> GOFLAGS=-buildvcs=false go test -v ./internal/e2e_harness/production -tags=e2e -run 'Test(BoundBigintAcceptsInt64Probe|BoundBigintPreservesNegMaxInt64|FederatedReadReturnsStoredMaxInt64|NullAndBoundaryRoundTripAcrossTiers)$' -timeout=10m
make lint
gofmt -d <all changed Go files>
git diff --check 1eee04d...e679324
```

All listed test commands passed; lint exited successfully; `gofmt -d` and `git diff --check` produced no output. The test evidence does not cover the HTTP boundary defect described in Findings.

iceboundrock:
## Summary

- **Findings:** 1 blocking P1, HTTP JSON decoding still routes full-range bound bigint values through `float64`.
- **Spec:** AC2 and AC3 are met; AC1 is not met for the public API; AC4 lacks positive-MaxInt64 warm-tier coverage.
- **Standards:** 1 hard file-size violation (`attribute_converter.go`, 583 lines); 2 non-blocking private-helper naming observations.
- **Verification:** targeted unit, HTTP-package, and production E2E tests plus lint and formatting checks passed at the PR head.

Worst issue: HTTP clients cannot safely write the advertised full int64 bound-bigint range.

iceboundrock:
## Review response (3 fix commits pushed: `b472476`, `cd67796`, `cef2afc`)

### P1 — HTTP write path float64 hop: **fixed** in `cd67796`

- New `readEntityJSONBody` decodes with `Decoder.UseNumber()`, switched in at exactly the two entity-payload sites: `handleCreate` and `handleUpdate`. JSON integer literals now reach the transform chain as `json.Number`, which `numutil.Int64Exact` already handles exactly — both `9223372036854775807` and `-9223372036854775807` survive end to end.
- Scope deliberately excludes the batch-delete body (`[]string`, no numbers) and `handleAdvancedQuery` (typed `forma.QueryRequest`): switching query-condition values to `json.Number` changes the comparison-side binding surface, which is exactly #281's scope. This boundary is noted in the helper's doc comment.
- Behavior-preservation audit: the only previously-accepted numeric payload shape that would have regressed under `UseNumber` was numeric-truthiness bools (`"active": 1`) — `toBool` gained a `json.Number` case preserving `!= 0` semantics. `numutil.Float64`/`Int64Exact` already accept `json.Number`; `toTime` rejected numeric types before and still does (same rejection class); `toString` on a numeric value now yields the literal digits (`json.Number` is a `fmt.Stringer`) instead of Go float formatting — strictly better for the text-attr edge; `ValidateAgainstSchema` has zero production callers; the `.(float64)` assertions in `model/record_parsing.go`/`schemameta` sit on server-generated JSON and schema documents, not entity payloads.
- Regression added: `TestCreateAndUpdateDecodeEntityNumbersExactly` — create + update handlers × both signs, asserting the exact literal reaches the manager boundary as `json.Number`.

### Spec / AC4 — warm-tier positive MaxInt64: **fixed** in `cef2afc`

`TestNullAndBoundaryRoundTripAcrossTiers` now applies a fifth max-shape row after `RunInit` and before `RunFlush`, so it lands in the delta parquet only (absent from base, `flushed_at` set after flush): `+math.MaxInt64` is now proven on **warm** both physically (per-file parquet value assertion on the delta file) and logically (federated read; `assertBoundaryRecords` keys by RowID, so all three max-titled rows — cold, warm, hot — are asserted independently). Full production suite re-run green (101s, no #276 flake) on top of all three fix commits.

### Standards — `attribute_converter.go`: **helper extracted** in `b472476`

`toInt64ExactForEAV` (the helper this PR added) moved to `typed_value.go`; `attribute_converter.go` is now 571 lines. The residual +13 over the 558 pre-PR baseline are the in-place sidecar edits inside the existing `ToEAVRecord`/`extractValueFromEAVRecord` bodies, which can't relocate without splitting the file itself — #283 (filed from the earlier review round) owns that full converter/extract split and will bring the file under 500.

### Naming observation: **addressed** in `b472476` (though non-blocking)

Both helpers renamed verb-first: `int64ExactForEAV` → `toInt64ExactForEAV` (matching the `toFloat64ForEAV`/`toTimeForEAV`/`toBoolForEAV` siblings) and `eavPivotExpr` → `buildEAVPivotExpr` (matching `buildEAVPivot`/`buildPGProjection`).

### Observations / Verification

Acknowledged — and the previously-uncovered gap ("existing tests do not cover HTTP decoding of entity numeric payloads at the int64 boundary") is now closed by the P1 regression above. Verification at the new head: httpapi + transform + numutil + sqlgen unit suites, `make test`, `make lint` (pinned v1.64.8), targeted + full production e2e all green.

With these, AC1 is met for the public HTTP API (exact path, not the documented-bound alternative), and AC4's tier matrix is complete: ±MaxInt64 proven on hot, warm, and cold.

Issue #285, opened 2026-07-17
NewDuckExporter applies session-scoped DuckDB init through the pool — same gap as #245, with an unbounded pool


iceboundrock:
## Summary

`internal/cdc/duckdb_exporter.go` (`NewDuckExporter`, :43+) has the same class of bug that #245 fixed in `internal/federated/duckdb_conn.go`: it opens the DuckDB pool with `sql.Open("duckdb", dsn)` and then issues all session-scoped initialization through `db.ExecContext`, so the statements land on one arbitrary pooled connection:

- `PRAGMA memory_limit` / `PRAGMA threads` (:54, :59)
- `INSTALL`/`LOAD` for `postgres_scanner`, `httpfs`, `parquet` (:64-69)
- `SET s3_access_key_id` / `s3_secret_access_key` / `s3_session_token` / `s3_region` / `s3_endpoint` / `s3_use_ssl` / `s3_url_style` (:80-138)

Worse than the federated client, the exporter never calls `SetMaxOpenConns`, so the `database/sql` pool defaults to **unlimited** connections. Any concurrent use of the exporter's DB can lazily open connections that carry none of the S3 session settings and fail exactly like the #245 reproduction (`in region ''` HTTP 404).

The per-query `ATTACH ... (TYPE postgres, READ_ONLY)` statements (:270, :286) are also connection-local state to keep in mind when reworking this path.

## Current exposure

Latent: the exporter today runs its statements sequentially over a (typically) single warm connection, so the gap is not known to bite in production. It becomes real the moment anything queries the exporter's DB concurrently or the pool cycles a connection.

## Fix direction

Apply the #245 pattern: pre-build the init statement list (fail-fast credential validation) and run it from a `duckdb.NewConnector(dsn, connInitFn)` hook via `sql.OpenDB`, so every physical connection self-configures. See `buildInitStatements` / `makeConnInit` in `internal/federated/duckdb_conn.go` (PR for #245). Decide `SetMaxOpenConns` policy for the exporter at the same time.

## References

- #245 (federated client fix; this issue is the CDC-side follow-up called out in its PR)
- `internal/cdc/duckdb_exporter.go:43-140` (init), `:270`/`:286` (ATTACH)

iceboundrock:
[...]

PR #286, opened 2026-07-17
fix(federated): #245 run DuckDB session init on every pooled connection

### LINKED ISSUE #245: configureS3 applies session-scoped S3 settings to a single pooled connection; MaxConnections>1 over :memory: yields unconfigured connections

iceboundrock:
Closes #245.

## Root cause

`NewDuckDBClientContext` issued `INSTALL`/`LOAD`/`SET`/`PRAGMA` through `db.ExecContext` after `sql.Open`, landing all session-scoped configuration on one arbitrary pooled connection. With `MaxConnections > 1`, lazily opened connections had no S3 settings, and concurrent federated queries failed with `in region ''` HTTP 404 (deterministically reproduced by `TestCircuitBreaker_ConcurrentTransitions` at pool size 2). Production default `MaxConnections: 1` made this latent; the e2e harness pinned 11 call sites to `WithDuckMaxConnections(1)` as a workaround.

One correction to the issue text, verified against driver source (`duckdb-go/v2 v2.5.6`): with `:memory:`, `NewConnector` opens the database once and all pooled connections share that instance — connections are not independent databases. The bug is purely that session-scoped statements reached a single connection, which also means fix direction 1 (per-connection init) is strictly better than direction 2 (`SET GLOBAL`): it covers `LOAD` and any future session-scoped statement as well.

## Fix

- Init statements are pre-built once per client (`buildInitStatements` — S3 credential denylist validation still fails fast at construction) and executed by a `duckdb.NewConnector` init hook (`makeConnInit`) that the driver invokes for **every** new physical connection; the pool is opened with `sql.OpenDB`.
- Failed init statements keep the previous log-and-skip policy, so init never blocks a connection — zero behavior change beyond the fix. `PingContext` still opens, initializes, and verifies the first connection at construction.
- Public signatures (`NewDuckDBClient`/`NewDuckDBClientContext`) unchanged; the `factory` seam is untouched. Both validators and the file-backed "database locked" warning are deliberately unchanged; no `:memory:` warning is needed since connections now self-configure.
- All 11 `WithDuckMaxConnections(1)` e2e pins removed — the production harness runs at its default pool of 2 again, making the whole suite a regression gate. The option itself stays as a plain pool-sizing knob.

## Commits (TDD)

1. `a7f748e` red — `TestNewDuckDBClientContext_AllConnectionsConfigured` holds one pooled connection while checking out a second and proves the lazily opened connection has a NULL `s3_region`; `_ConcurrentConnectionsConfigured` mirrors the real concurrent failure shape. Both failed before the fix (connection 1 NULL; 7/8 workers NULL).
2. `261df85` green — the connector refactor. `_ExtensionStepsUseCallerCtx` is superseded by `_FirstConnectionConfigured` (init no longer runs as construction-time pool calls); `_InvalidS3CredentialFailsFast` locks the fail-fast contract.
3. `f5f865f` — e2e pin removal + `env.go`/`doc.go`/README workaround-comment cleanup.

## Verification

- `go test ./internal/federated ./factory` — green (red tests confirmed failing first).
- `make lint`, `make test` — green.
- Focused e2e: `TestCircuitBreaker_ConcurrentTransitions` at default pool 2 — **PASS** (previously 5/5 failing per the #245 A/B evidence).
- Full `go test -v ./internal/e2e_harness/production/ -tags=e2e` — all green (122s), including every de-pinned scenario.

## Out of scope

`internal/cdc/duckdb_exporter.go` has the same pattern (worse: no `SetMaxOpenConns`, so an unbounded pool) — filed as #285 rather than widening this PR's blast radius into the CDC path.

### DIFF
--- a/internal/e2e_harness/production/README.md
+++ b/internal/e2e_harness/production/README.md
@@ -79,7 +79,7 @@ needs schema IDs beyond the fixtures can register its own via
 `RegisterSchemas` + `WithSchemaDir`.
 
 Options: `WithSeed`, `WithSchemaDir`, `WithFlushThresholds` (#179),
-`WithDuckMemoryMB`, `WithBreaker` (#185), `WithDuckMaxConnections` (#185),
+`WithDuckMemoryMB`, `WithBreaker` (#185), `WithDuckMaxConnections`,
 `WithRoutingStrategy`, `WithoutManifest`.
 
 ## Fixture schemas (`schemas/`)
@@ -182,10 +182,9 @@ where cdc-init bootstraps base files before federated reads. Use
   observe the open-to-closed transition against a healthy DuckDB.
 - `WithBreaker(maxFailures, cooldown)` (#185) enables the federated engine's
   circuit breaker; `WithDuckMaxConnections(n)` overrides the per-test DuckDB
-  pool size (default 2). Pin it to `1` for tests that issue **concurrent**
-  DuckDB queries: `NewDuckDBClient` applies the session-scoped S3 `SET`
-  statements on a single pooled connection (`duckdb_conn.go`), so over
-  `:memory:` a second connection opened under load lacks S3 config and 404s.
+  pool size (default 2). Since #245 every pooled connection self-configures
+  via the driver's per-connection init hook, so this is purely a pool-sizing
+  knob — concurrent DuckDB queries need no pinning.
   Set `Query.AllowPartialDegradedMode` to forward the public degraded-mode
   flag so a tier outage falls back to a postgres-only result (complete in
   today's PG-retains-all model) instead of erroring.
@@ -227,5 +226,5 @@ On failure (or `KEEP_E2E_ENV=1`) the Env writes
 | #179 flush failure matrix | `FaultInjectingS3` + `RunFlushWith` + `ExecSQL` |
 | #180 dry-run semantics | `RunFlushDry` |
 | #181 failure injection | `ExecSQL` |
-| #185 degraded mode & circuit breaker | `WithBreaker`, `Query.AllowPartialDegradedMode`, `Cluster.HaltS3`, `Env.ReopenDuckDB`, `WithDuckMaxConnections` |
+| #185 degraded mode & circuit breaker | `WithBreaker`, `Query.AllowPartialDegradedMode`, `Cluster.HaltS3`, `Env.ReopenDuckDB` |
 | #186 multi-schema isolation | `e2e_simple`/`e2e_second` + `FaultInjectingS3`/`PausingS3` (`multi_schema_isolation_e2e_test.go`, `concurrent_flush_multi_schema_e2e_test.go`) |
--- a/internal/e2e_harness/production/circuit_breaker_e2e_test.go
+++ b/internal/e2e_harness/production/circuit_breaker_e2e_test.go
@@ -95,11 +95,10 @@ func TestCircuitBreaker_OpensAtThresholdAndRecovers(t *testing.T) {
 // calls go through eng.Query directly: Env.Query records results without a
 // lock and is not safe for concurrent use.
 //
-// DuckDB MaxConnections is pinned to 1: NewDuckDBClient applies the
-// session-scoped S3 SET statements to a single pooled connection, so a
-// second :memory: connection opened under concurrency lacks S3 config and
-// 404s — a harness/client gap unrelated to breaker behavior (production
-// defaults to MaxConnections=1 and is unaffected; follow-up issue drafted).
+// Runs on the harness-default DuckDB pool (2 connections): this test is the
+// regression gate for #245 — per-connection init must configure every pooled
+// connection, or one of the concurrent recovery queries 404s with an empty
+// s3_region.
 func TestCircuitBreaker_ConcurrentTransitions(t *testing.T) {
 	const threshold = 3
 	const cooldown = 3 * time.Second
@@ -107,7 +106,7 @@ func TestCircuitBreaker_ConcurrentTransitions(t *testing.T) {
 	const queriesPerWorker = 5
 
 	ctx := context.Background()
-	env := NewEnv(t, SharedCluster(t), WithBreaker(threshold, cooldown), WithDuckMaxConnections(1))
+	env := NewEnv(t, SharedCluster(t), WithBreaker(threshold, cooldown))
 	wide := DefaultSchemaFixtures()[1]
 
 	seedTwoTiers(ctx, t, env, wide)
--- a/internal/e2e_harness/production/doc.go
+++ b/internal/e2e_harness/production/doc.go
@@ -32,7 +32,7 @@
 //	#180 dry-run semantics         -> RunFlushDry, RunInitWith(InitOverrides{DryRun})
 //	#181 failure-state injection   -> Env.ExecSQL escape hatch
 //	#182 mid-flush concurrency     -> PausingS3, RunFlushWith(FlushOverrides{S3})
-//	#185 degraded/breaker          -> WithBreaker, WithDuckMaxConnections, HaltS3, ReopenDuckDB, Query.AllowPartialDegradedMode
+//	#185 degraded/breaker          -> WithBreaker, HaltS3, ReopenDuckDB, Query.AllowPartialDegradedMode
 //	#186 multi-schema isolation    -> schemas/e2e_simple + e2e_second, per-test DB
 //	#260 nested dotted attributes  -> schemas/e2e_nested, nested_attrs_e2e_test.go
 //
--- a/internal/e2e_harness/production/env.go
+++ b/internal/e2e_harness/production/env.go
@@ -108,11 +108,9 @@ func WithDuckMemoryMB(mb int) EnvOption {
 }
 
 // WithDuckMaxConnections overrides the per-test DuckDB connection pool size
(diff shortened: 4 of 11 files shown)


iceboundrock (commented):
## Standards

**Hard violation**

- `internal/federated/duckdb_conn.go:141,166,204` and `internal/federated/duckdb_conn_test.go:156`: new private helpers use noun-led names (`extensionStmts`, `s3Stmts`, `pragmaStmts`, `s3ConfiguredDuckDBConfig`). `coding-standard.md` §4, line 56 requires private helpers to be verb-first. Rename to forms such as `buildExtensionStatements`, `buildS3Statements`, `buildPragmaStatements`, and `newS3ConfiguredDuckDBConfig`.

**Judgment call**

- `internal/federated/duckdb_conn.go:172-200`: the four credential branches repeat the same validate-and-append sequence for more than 15 similar lines. `coding-standard.md` §2, line 15 requires checking for this duplication. A small descriptor-driven helper would remove it, though the current explicit branches remain readable.

No size-limit, nesting, error-wrapping, or exported-Godoc violations found in the diff. Tooling-enforced matters were not reviewed.

iceboundrock:
## Findings

- **P1 - preserve INSTALL/LOAD dependency:** [`extensionStmts`](https://github.com/Lychee-Technology/forma/blob/f5f865f933e2d5e97e7d03ea9b93bd7a4955f484/internal/federated/duckdb_conn.go#L141-L163) flattens each `INSTALL`/`LOAD` pair, while `makeConnInit` logs every failure and continues. Consequently an `INSTALL` failure is now followed by `LOAD`; previously user extensions used `continue`, and `httpfs`/`parquet` only loaded after a successful install. A previously installed local extension can therefore become loaded after a failed install, changing the documented log-and-skip behavior. Preserve the per-pair dependency in the init hook and add a regression test that forces `INSTALL` to fail and verifies `LOAD` is not attempted.

## Spec

- The core #245 requirement is met: `duckdb.NewConnector` plus `sql.OpenDB` invokes initialization for every physical pooled connection, and the new deterministic two-connection and concurrent tests assert the session-scoped `s3_region` setting.
- The PR correctly leaves the CDC exporter out of scope and references follow-up #285, as required by the plan.
- The constructor comment explains per-connection initialization. The plan also asked it to state that failed initialization is logged and skipped; that part is documented on `makeConnInit` instead. This is a minor documentation deviation, not a functional concern.

## Standards

- `coding-standard.md:56` requires verb-first private helpers. `extensionStmts`, `s3Stmts`, and `pragmaStmts` are noun-led. This is non-blocking; names such as `buildExtensionStatements`, `buildS3Statements`, and `buildPragmaStatements` would conform.
- No file/function size, error-wrapping, nesting, or exported-documentation violations found.

## Assessment

- **Request changes** for the P1 behavior regression. Apart from it, the connector-based design is a focused and appropriate fix for session-scoped DuckDB configuration.

## Observations

- Removing the 11 `WithDuckMaxConnections(1)` workarounds restores the harness default of two connections and makes the concurrent circuit-breaker scenario a useful regression gate.
- No unrelated CDC changes or other scope creep found.

## Verification

- `git diff --check main...origin/pr-286` passed.
- `GOCACHE=$PWD/.gocache-review GOFLAGS=-buildvcs=false go test ./internal/federated ./factory` passed.
- `make lint` passed.
- `GOCACHE=$PWD/.gocache-e2e-review GOFLAGS=-buildvcs=false go test -v ./internal/e2e_harness/production/ -tags=e2e -run TestCircuitBreaker_ConcurrentTransitions -timeout=5m` passed.
- GitHub checks: Lint, Test, Benchmark Smoke, and k6 Smoke passed; E2E Smoke Tests remains pending at review time.

## Summary

- **1 blocking finding:** preserve the old rule that a failed extension installation prevents its load attempt. The primary per-connection S3-configuration fix and its focused regression coverage otherwise satisfy #245.

iceboundrock:
Review items addressed in 6abc4b3:

**P1 — INSTALL→LOAD dependency (fixed).** Confirmed against the pre-PR code: user extensions skipped LOAD via `continue`, and httpfs/parquet only LOADed inside the INSTALL-success branch; the flattened list lost that. Init statements are now grouped into `initStep`s — a failed statement skips the rest of its step while later steps still run — so a failed INSTALL can no longer be followed by a LOAD of a stale on-disk extension. `TestMakeConnInit_FailedInstallSkipsLoad` locks the contract with a recording `driver.ExecerContext` fake: it forces `INSTALL bad_ext;` to fail and asserts `LOAD bad_ext;` is never attempted while `LOAD httpfs;` and the S3 SETs still execute.

**Standards — verb-first naming (fixed).** `buildExtensionSteps` / `buildS3Steps` / `buildPragmaSteps` / `buildInitSteps` / `makeSingleStmtStep` in source, `newS3ConfiguredDuckDBConfig` in tests.

**Standards judgment call — credential branch duplication (implemented).** The three identical validate-and-append branches (access key, secret key, region) collapsed into a descriptor loop. The endpoint branch deliberately stays separate: it trims the `http://` scheme before validation and carries the dependent `s3_use_ssl`/`s3_url_style` statements, so folding it into the loop would need special-casing that costs more than the duplication it removes.

**Spec observation — constructor doc (fixed).** `NewDuckDBClientContext`'s comment now states the log-and-skip policy directly ("Failed init statements are logged and skipped, never blocking the connection — construction fails only on credential validation or ping"), in addition to the `makeConnInit` doc.

Verification after the changes: `go test ./internal/federated ./factory`, `make lint`, `make test` all green; focused e2e gate `TestCircuitBreaker_ConcurrentTransitions` re-run at default pool 2 — PASS.

PR #288, opened 2026-07-18
feat: #204 list attributes round-trip end to end — items_type metadata, LIST parquet export, positional array_indices reconstruction

### LINKED ISSUE #204: list attributes are unsupported end to end: write path rejects them (transform), and CDC export would MAX-collapse them (array_indices dropped)
## Summary

`list` (array) attributes do not round-trip. Surfaced by the #174 full-type
round-trip suite (same pattern as #192–#198), the #174 probe
(`internal/e2e_harness/production/list_roundtrip_e2e_test.go`,
`TestListAttributeRoundTrip`) establishes empirically that there are **two
independent blockers**, the first of which is reached today:

### Blocker 1 (reached today): the write path rejects lists

`internal.EntityManager.Create` → `transform` rejects any `list`-typed
attribute before persistence. `transform.flattenToAttributes`
(`internal/transform/transformer.go:287-300`) correctly decomposes an array
into indexed elements, but each element is then typed against the attribute's
declared value type `list` in `populateTypedValue`
(`internal/transform/transformer.go:337-372`), whose switch has no `list` arm
and falls through to `default`:

```
default:
    return handleConversionError(fmt.Errorf("unsupported value type '%s'", meta.ValueType))
```

Observed error (probe output):

```
apply event seq=1 kind=create schema=e2e_wide: entity manager create:
failed to transform data to persistent record: failed to convert to attributes:
convert value for attribute 'tags': invalid value for attribute 'tags' (attrID=18):
invalid input: unsupported value type 'list'
```

No `eav_data` row is written; the hot read returns zero records. So lists never
enter the CDC pipeline at all.

### Blocker 2 (latent, only reached once Blocker 1 is fixed): CDC export collapses lists

Even if the write path is fixed to persist one `eav_data` row per element (with
distinct `array_indices`), the EAV export would collapse them:

- The EAV export select drops `array_indices` entirely — it selects only
  `schema_id, row_id, attr_id, value_text, value_numeric`
  (`internal/cdc/export_sql_builder.go:79-88`).
- The pivot aggregates each attribute with
  `MAX(CASE WHEN attr_id = N THEN <value> END) AS <alias>`
  (`internal/cdc/export_sql_builder.go:135`), collapsing all element rows of a
  multi-element attribute into a single scalar parquet column. Confirmed
  physically: with `tags` added to the fixture, the export produces a single
  `tags VARCHAR` column (not a DuckDB LIST), observed via the parquet
  `DESCRIBE` in the #174 suite.
- The federated (hot-side) reader hardcodes `'array_indices': ''` when building
  `attributes_json` (`internal/sqlgen/duckdb_schema_projection.go:328`, and the
  parallel path at `:691`), so it cannot represent or reconstruct list elements
  either.

## Impact

`forma.ValueTypeList` is declared (`schema_registry.go:20`) and documented as
uncovered (`internal/e2e_harness/production/README.md`: "`forma.ValueTypeList`
is not covered yet"), but is unusable: writing a list-typed attribute errors,
and even if that were relaxed, warm/cold/federated reads would collapse it to a
single element.

## Acceptance criteria

- The write path persists a list attribute as one `eav_data` row per element
  with distinct `array_indices` (element type derived from the array's `items`
  type, not the container type `list`).
- The CDC EAV export preserves `array_indices` and does NOT `MAX`-collapse
  multi-element attributes; parquet stores the elements (e.g. a DuckDB LIST or
  one row per element).
- The federated reader reconstructs `array_indices` rather than hardcoding `''`.
- The executable spec — the currently-skipped
  `TestListAttributeRoundTrip/round-trip across tiers` subtest — passes with the
  skip removed: 3 EAV rows on write, 3 elements on hot read, ≥3 values per
  parquet tier, 3 elements on federated merge-on-read.

## Notes / test-harness gotcha

Because the CDC export projection is schema-driven, adding an EAV attribute to
the shared `e2e_wide` fixture adds an unconditional column to every `e2e_wide`
parquet export, which breaks the exhaustive column assertion in
`TestFullTypeRoundTripAcrossTiers` (`assertWideParquetSchema`). The #174 probe
therefore injects `tags` into a private per-test schema dir via `WithSchemaDir`
rather than the shared fixture. Any fix that adds real list coverage to the
shared `e2e_wide` fixture must also extend `wideParquetTypes`
(`internal/e2e_harness/production/full_type_parquet_e2e_test.go`) with the
`tags` physical type.

Epic #172. Surfaced by the #174 round-trip suite.

iceboundrock:
Closes #204

## Summary

`list` (array) attributes now round-trip across all three tiers. The three blockers named in the issue are fixed in dependency order:

1. **Write path (blocker 1)** — `transform.populateTypedValue` gains a `list` arm: each already-decomposed element is typed by the attribute's declared `items_type` (new `AttributeMetadata.ItemsType`, default `text`), one `eav_data` row per element with distinct `array_indices`. Guards reject scalar payloads and multi-dimensional arrays with `forma.ErrInvalidInput`. `FromEAVRecords` substitutes the element type for the container so the whole EntityAttribute round trip (including `ToEAVRecord`, which also rejected `list`) sees scalar types.
2. **CDC export (blocker 2)** — `buildEAVQuery` carries `array_indices`; list attributes aggregate via `list(<elemCast> ORDER BY TRY_CAST(array_indices AS BIGINT)) FILTER (WHERE attr_id = N)` instead of the `MAX(CASE…)` scalar collapse. Parquet stores a DuckDB LIST (`VARCHAR[]` for text items).
3. **Federated reader** — the hot-tier pivot emits the same LIST aggregate (legs type-unify through COALESCE/UNION ALL), and `attributes_json` reconstructs positional `array_indices` via `list_transform((x, i) -> …)` + `flatten`, replacing the hardcoded `''`. Scalar-only schemas render byte-identical SQL. The downstream Go side (`model.ParseEAVAttribute`, merge key `(AttrID, ArrayIndices)`) was already index-aware.

Also per review decisions: the benchmark projection's `array_indices: ''` hardcode is aligned (list-aware, scalar output byte-identical); `generate-attributes` now emits `valueType: list` + `items_type` for primitive arrays; `validate-schema-consistency` classifies list attrs by items type; `tags` (attrID 18, list of string) joins the shared `e2e_wide` fixture with `wideParquetTypes` extended to `VARCHAR[]`.

## Acceptance

The formerly-skipped executable spec `TestListAttributeRoundTrip` now runs un-skipped and passes: 3 `eav_data` rows (`array_indices` "0","1","2") → hot read 3 elements in order → base **and** delta parquet both `VARCHAR[] [alpha, beta, gamma]` → federated merge-on-read returns both records with 3 ordered elements each.

## Verification

- `make lint` clean (golangci-lint v1.64.8)
- `make test` — 26 packages, no failures
- `go test ./internal/e2e_harness/production/... -tags=e2e -run 'TestListAttributeRoundTrip|TestFullTypeRoundTripAcrossTiers'` — both PASS (the exhaustive parquet-schema assertion accepts the new `tags VARCHAR[]` column)
- `go test ./internal/e2e_harness/federated/... -tags=e2e -short` — PASS (CI parity)
- infra smoke `go test ./internal/e2e_harness/...` — PASS
- New `TestDuckDBListSyntaxSupport` pins the DuckDB constructs this path relies on (aggregate `ORDER BY` + `FILTER`, two-arg indexed lambda, `flatten` of mixed sublists) against duckdb-go v2.5.6

## Migration note (operators)

Shipped `cmd/server/schemas/*_attributes.json` are **intentionally not regenerated** here. Regenerating flips legacy scalar-typed array attributes (`text`/`numeric`) to `list` + `items_type`, which changes their parquet column type scalar → LIST and conflicts with already-exported parquet under `union_by_name`. Existing deployments must re-export (or switch S3 prefix) before regenerating attributes. Until then their arrays keep the pre-existing (silently collapsing) behavior — unchanged by this PR.

## Scope explicitly left out

- List-containment predicates (the normalizer already rejects list-typed predicates cleanly).
- Nested/multi-dimensional arrays inside a list attribute (rejected at write with `ErrInvalidInput`; parquet LIST reconstruction is positional).
- Regeneration of shipped schemas + data migration (candidate follow-up issue).

Plan: `docs/superpowers/plans/2026-07-18-issue-204-list-attributes-e2e.md`


iceboundrock:
## Findings

- **P1 | Empty lists are silently converted to an absent attribute.** `internal/transform/transformer.go:285-298` iterates an empty `[]any` zero times, so the new `ValueTypeList` arm in `internal/transform/typed_value.go:63-79` is never reached and no EAV record is persisted. The federated reconstruction likewise emits no attribute when the LIST column is `NULL` (`internal/sqlgen/duckdb_schema_projection_json.go:101-104`). Therefore `{ "tags": [] }` cannot round-trip and is indistinguishable from omitting `tags`; an update can also erase prior list elements and return an absent attribute. Add an explicit empty-list representation or reject empty lists with `ErrInvalidInput`, then cover create/update and hot/parquet/federated reads.

## Spec

The explicit #204 acceptance path for a non-empty, three-element list is implemented: per-element EAV writes, ordered DuckDB LIST CDC export, positional federated reconstruction, shared fixture integration, and the previously skipped E2E test. The acceptance test does not exercise empty lists, leaving the P1 general list-round-trip gap above.

## Standards

`coding-standard.md` section 1 rejects changes that leave Go source files over 500 lines. This PR extends already-oversized files instead of extracting the touched area:

- `internal/transform/attribute_converter.go` is 585 lines.
- `internal/transform/attribute_converter_test.go` is 690 lines.
- `cmd/tools/generate_attributes_test.go` is 1,367 lines.

Also, `git diff --check main...review-pr-288` reports a new blank line at EOF in `internal/e2e_harness/production/list_roundtrip_e2e_test.go:331`.

## Assessment

Do not merge until the empty-list behavior is made explicit and lossless (or rejected), with regression coverage. The file-size and EOF issues should be addressed in the same update.

## Observations

No additional correctness regressions were found in the changed non-empty-list write, CDC, hot-tier pivot, or positional JSON reconstruction paths.

## Verification

- `make test` passed.
- `make lint` completed without diagnostics.
- The #204 production E2E selection for `TestListAttributeRoundTrip` and `TestFullTypeRoundTripAcrossTiers` passed.
- `git diff --check main...review-pr-288` reported the EOF blank line above.

## Summary

**1 P1 functional data-loss finding; 3 documented file-size violations; 1 whitespace issue.** The PR satisfies #204 for non-empty lists, but does not safely support an explicit empty list.

iceboundrock:
Review round-1 findings addressed in cd6938b + f662473. Per-finding response:

## P1 — empty lists silently become an absent attribute: **fixed (lossless representation)**

Confirmed, with one additional fact that shaped the fix: **Update is merge-semantics** (`mergeMaps`, `internal/entity_crud_service.go:175` — omitted keys preserve current values, and `null` is rejected by the pre-existing contract "omit the key to preserve its current value"). That makes `"tags": []` the **only** gesture a client has to clear a list, so the "reject with ErrInvalidInput" option would have removed clear-list entirely. Implemented the explicit-representation option instead (cd6938b):

- **Write**: an explicit `[]` persists a single **marker row** — `array_indices ''`, both value columns NULL (`transformer.go` empty-array arm). Create and clear-via-update both produce exactly this row; the EAV full-replace update makes clearing atomic.
- **CDC export / hot pivot**: the list aggregate becomes `CASE WHEN count(*) FILTER (WHERE attr_id = N) > 0 THEN coalesce(list(<elemCast> ORDER BY ...) FILTER (WHERE attr_id = N AND array_indices <> ''), []) END` — parquet now distinguishes `[]` (empty non-NULL LIST) from NULL (absent). Both sides stay mirrored.
- **Federated reconstruction**: an empty LIST column emits the marker object (`array_indices ''`, both values NULL) in `attributes_json`; NULL emits nothing.
- **Read materialization**: `FromAttributes` turns the marker back into a real `[]` (guarded so a stale marker never clobbers element values that were placed first).
- **Coverage**: new `TestEmptyListRoundTrip` e2e pins `[]`-vs-absent at every hop — create-empty, clear-via-update (prior elements erased by the explicit clear, marker remains), and an absent control row; observed delta parquet: `[[] [] <NULL>]`. Unit level: `TestPersistentRecord_EmptyListRoundTrip` (write echo + JSON rebuild), extended engine-executed `attributes_json` round trip (`[]` → marker → parser; NULL → nothing), and new DuckDB syntax pins for the presence-count CASE, untyped-`[]` coalesce unification, and all-NULL-field struct retention in `list_filter`.

## Standards — oversized files: **fixed (pure-move splits, f662473)**

- `attribute_converter.go` 585 → 479 + `attribute_value_extraction.go` (119)
- `attribute_converter_test.go` 738 → 361 + `attribute_value_extraction_test.go` (238) + `attribute_converter_list_test.go` (159, carries the #204 list/empty-list tests)
- `generate_attributes_test.go` 1367 → 240 + `generate_attributes_traverse_test.go` (334) + `generate_attributes_files_test.go` (400) + `generate_attributes_merge_test.go` (424)

All moves are mechanical; no test was modified in the move (54 generator tests still pass, transform suite green).

## Whitespace — EOF blank line: **fixed** (in cd6938b); `git diff --check main...` is clean.

One note from the split: `gofmt -l` also flags several files in this area (including ones untouched by this PR, e.g. `boundary_roundtrip_e2e_test.go` on main) — that is Go 1.19+ doc-comment quote normalization (`''` → `”`), pre-existing repo-wide and not enforced by the pinned golangci-lint; left untouched to avoid unrelated churn.

## Verification after fixes

- `make lint` exit 0; `make test` all packages pass
- `TestListAttributeRoundTrip`, `TestEmptyListRoundTrip`, `TestFullTypeRoundTripAcrossTiers` all PASS (`-tags=e2e`)
- `git diff --check` clean

Issue #292, opened 2026-07-20
manifest-reconcile: coverage-verified --repair promotion for init-shaped base orphans


iceboundrock:
Follow-up from #290 (deliberately excluded from its PR scope).

## Background

#290 made `cdc-init` hold the per-schema advisory lock, which let `manifest-reconcile --gc` safely delete init-shaped `{min}_{max}.parquet` orphans (failed manifest publish output, or files superseded by a later init run) through the existing two-phase sighting + grace machinery. Recovery for a failed publish is currently: re-run `cdc-init`.

## Why --repair was excluded

Auto-promoting init-shaped orphans into the manifest base tier is unsafe without a completeness proof: a run that failed **mid-export** leaves only a partial batch set behind, and from the object level alone the tool cannot distinguish "complete export, failed publish" from "partial export, failed mid-run". Rebuilding the base tier via `ReplaceTierFiles` with a partial set would drop rows from the base tier while the manifest still pointed at a consistent older inventory.

## Proposal

`--repair` for `ClassBaseInit` orphans must verify coverage before promoting:

1. Under the schema advisory lock, compute the orphan set's row coverage from parquet stats (DuckDB, `compaction.SingleFileStats`-style) and compare against live `entity_main` rows (PG, `reconcile.PGLiveRows`-style).
2. Only if the orphan set provably covers **all** live rows: promote via `ReplaceTierFiles` semantics (matching init's own `updateSchemaManifest`).
3. Otherwise: leave them as GC candidates (current behavior) and report why promotion was refused.

The repair machinery introduced in #289 (stats reader, live-row checker, etag-retried save) should be reusable here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

iceboundrock:
> *This was generated by AI during triage.*

## Dependency Update

PR #293 (`7fa77b1`) has merged the prerequisite lock coordination and safe two-phase GC for init-shaped base orphans. This follow-up remains open: it intentionally did not implement `--repair` promotion, because a partially exported init file set cannot be safely distinguished from a complete failed publication using object shape alone.

iceboundrock:
Implemented and merged in PR #352 (merge commit 432ec78a87069ceefaa1b3ee9210fe617fcfd7be). The issue was automatically closed by the PR's Closes #292 reference; all required CI checks passed before merge.

PR #293, opened 2026-07-20
feat(cdc): #290 cdc-init holds per-schema advisory lock; manifest-reconcile --gc handles init-shaped orphans

### LINKED ISSUE #290: cdc-init: hold the per-schema advisory lock so manifest-reconcile can safely handle init-shaped orphans
Follow-up from PR #289 review (spec finding 1) on #203.

## Problem

`manifest-reconcile` deliberately never repairs or GCs init-shaped `{min}_{max}.parquet` orphans: an in-flight `cdc-init` promotes each batch to its final base key immediately (`CopyTmpToFinal` per batch) but publishes the manifest only after ALL batches (`updateSchemaManifest` at the end of `initSchema`), and `cdc-init` does not take the per-schema advisory lock (`cdc.AcquireSchemaLock` is only called from the flusher). From the object level, a long-running init and a failed manifest publish are indistinguishable, so the tool must treat every unlisted init-shaped file as potentially in-flight.

Consequence (accepted scope limitation in #203): a successful init export whose `ReplaceTierFiles` publish failed has no automated recovery path — the exported base files stay report-only orphans until an operator re-runs `cdc-init` or hand-edits the manifest.

## Proposal

1. `cdc-init` acquires `pg_try_advisory_lock(schemaID, schemaID)` for the duration of a schema's export + manifest publish (skip-or-fail when contended, mirroring the flusher), pinned to one connection like `reconcile.PGAdvisoryLocker`.
2. Once init holds the lock, `manifest-reconcile` (which reconciles under the same lock) can no longer race an in-flight init, and init-shaped orphans under the lock are provably from a failed publish. The tool can then:
   - `--repair`: rebuild the base tier via `ReplaceTierFiles` semantics (or re-run init's `updateSchemaManifest`), and/or
   - `--gc`: delete superseded init-shaped leftovers through the existing two-phase sighting state.
3. Revisit `internal/e2e_harness/production/init_concurrent_e2e_test.go` — concurrent init behavior changes by design.

## Notes

- The advisory-lock convention `(schemaID, schemaID)` now has three call sites (`internal/cdc/helpers.go`, `internal/reconcile/db.go`, `internal/e2e_harness/federated/cdc.go`); consolidating on one helper would keep the key derivation single-sourced.
- Related known cleanup: the keyword/value Postgres DSN is hand-built in `internal/cdc/flusher.go:206`, `internal/cdc/init.go`, and the e2e harness without value quoting; `cmd/tools` gained a quoted variant (`quotePGConnValue`) in #203 — extract one shared DSN builder.

iceboundrock:
Closes #290

## What

Four blocks, in dependency order:

1. **Shared quoted PG DSN builder** (`cdc.BuildPGDSN`, 2ec6f78) — replaces three hand-built raw-interpolation DSN copies (`internal/cdc/init.go`, `internal/cdc/flusher.go` ×2 sites, `cmd/tools` `buildToolSQLDB`); values quoted per libpq rules (`\`→`\\` then `'`→`\'`), so passwords with spaces/quotes survive pgx and DuckDB postgres_scanner parsing. The flusher IAM branch is untouched (token resolves into the password before formatting).
2. **Single pinned-connection `cdc.TrySchemaLock`** (bd81ba1) — the canonical `pg_try_advisory_lock(schemaID, schemaID)` try-lock, pinned to one physical connection with a closure unlock (background ctx + conn close). The flusher migrates off the pool-based `AcquireSchemaLock`/`ReleaseSchemaLock` pair — **fixing the latent bug where acquire and release could land on different pooled connections, silently leaking the session lock**. `reconcile.PGAdvisoryLocker` and the federated e2e harness now delegate to it; contended-flusher semantics unchanged (skip + nil).
3. **cdc-init holds the lock** (e69a25b) — each schema's entire export + manifest publish (including 0-row and dry-run paths) runs under the advisory lock via `initSchemaUnderLock`. Contended is an **error, not a skip**: the schema joins `schemaErrors` (`errors.Is`-matchable `ErrSchemaLockContended`), other schemas continue, the run returns non-zero — init is operator-initiated, so a silently skipped schema must not go unnoticed.
4. **`manifest-reconcile --gc` now collects init-shaped orphans** (9c1bc37) — with init under the same lock reconcile holds, an init-shaped `{min}_{max}.parquet` orphan observed under the lock is provably not from an in-flight init: it is either output of a failed manifest publish or a file superseded by a later init run. Both flow through the existing two-phase sighting + grace machinery. `classifyObjectKey`/`diffSchema`/`report.go` are zero-change; residual/exit-code semantics hold automatically. `--repair` auto-promotion is deliberately excluded (partial mid-export runs are indistinguishable at the object level; promoting a partial set would drop base-tier rows) → follow-up #292.

Review fix wave (ed9840a, b18685c): the whole-branch review caught a **Critical regression introduced by block 1** — the quoted DSN, after `escapeLiteral` quote-doubling, defeated `redactConnStr`, leaking the PG password in the Info-level `sql_preview` log on every flush/init export. Fixed in `redactConnStr` (ordered alternation handling doubled-quote/quoted/legacy-unquoted forms), pinned by `redact_test.go` composing the real pipeline `redactConnStr(escapeLiteral(BuildPGDSN(...)))`; `flusher.go`'s hand-built unquoted log preview replaced with the redacted real DSN.

## Contended matrix (after this PR)

| caller | on contention |
|---|---|
| cdc-init | per-schema error `ErrSchemaLockContended`, run exits non-zero |
| flusher | skip + log, returns nil (unchanged; next tick retries) |
| manifest-reconcile | schema reported Skipped (unchanged; exit 2 on residual) |

## Operational notes

- **Flusher starvation window**: while init exports a schema (potentially long), the flusher skips that schema every tick — no incremental flush progress until init finishes. Per-batch lock cycling was rejected: it would reopen the init↔reconcile race this PR closes.
- This PR closes only the init↔reconcile race. Init's `ReplaceTierFiles` vs the compactor's manifest writes still rely on etag optimistic concurrency (compactor takes no advisory lock) — unchanged, per issue scope.
- The federated e2e harness's 2s unlock timeout is replaced by the shared background-ctx unlock (strictly stronger).
- `internal/cdc/init.go` sits at exactly the 500-line ceiling; the next edit there should split the file.
- `internal/federated/engine.go:440` still hand-builds a raw DSN — deliberately out of scope (different shape); the new redact pin test protects it when it eventually migrates.

## Verification

- `make test` (27 packages) + `make lint` green at HEAD.
- Unit: new `TestProcessInitSchemas_ContendedLockRecordsErrorAndContinues` (real `processInitSchemas` path, stubbed seams), inverted `TestGC_CollectsInitShapedBaseOrphans`, 4 flusher lock tests migrated to the single `tryLock` seam preserving original assertions, `TestBuildPGDSN_QuotesValues` + migrated quote cases, 3 redaction pin tests (red run reproduced the leak verbatim).
- E2E (Docker, run at HEAD): production `TestManifestReconcile*` — past-grace `--gc` now deletes {staleBase, staleTmp, initShaped}, deltaOrphan survives, residual clears; production `TestInit*` including `TestInitUnderConcurrentMutation` (mutator takes no advisory lock — race window and assertions unchanged, comment documents the new per-caller matrix); federated `-short` suite + benchmark green (harness lock helper migrated).
- Subagent-driven execution with per-task spec+quality review and a final whole-branch review (findings fixed and re-verified); plan at `docs/superpowers/plans/2026-07-20-issue-290-init-advisory-lock.md`.

### DIFF
--- a/cmd/tools/main.go
+++ b/cmd/tools/main.go
@@ -92,5 +92,5 @@ func printUsage(out io.Writer) {
 	fmt.Fprintln(out, "  cdc-flush             Run CDC change_log flush to S3 parquet files")
 	fmt.Fprintln(out, "  cdc-init              Initialize S3 parquet base files from existing data")
 	fmt.Fprintln(out, "  compactor             Run compaction on parquet files for a schema")
-	fmt.Fprintln(out, "  manifest-reconcile    Diff S3 parquet objects against manifests; report, --repair orphaned deltas, --gc compaction leftovers")
+	fmt.Fprintln(out, "  manifest-reconcile    Diff S3 parquet objects against manifests; report, --repair orphaned deltas, --gc stale base/tmp leftovers (compaction + init-shaped)")
 }
--- a/cmd/tools/manifest_reconcile.go
+++ b/cmd/tools/manifest_reconcile.go
@@ -257,25 +257,18 @@ func openReconcileStatsEngine(ctx context.Context, opts *reconcileOptions, logge
 // buildToolSQLDB opens a database/sql handle (pgx driver) for tools that
 // need session-scoped Postgres features — the reconcile advisory lock pins
 // a single connection, which pgxpool does not expose. Values are quoted so
-// passwords with spaces or quotes survive DSN parsing (the sibling copies
-// in internal/cdc/flusher.go and init.go still interpolate raw — follow-up:
-// extract one shared DSN builder).
 func buildToolSQLDB(pg postgresFlags) (*sql.DB, error) {
-	connStr := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=%s",
-		quotePGConnValue(pg.host), pg.port, quotePGConnValue(pg.user),
-		quotePGConnValue(pg.resolvedPassword("PGPASSWORD")),
-		quotePGConnValue(pg.database), quotePGConnValue(pg.sslMode))
+	connStr := cdc.BuildPGDSN(cdc.PGDSNParams{
+		Host:     pg.host,
+		Port:     pg.port,
+		User:     pg.user,
+		Password: pg.resolvedPassword("PGPASSWORD"),
+		DB:       pg.database,
+		SSLMode:  pg.sslMode,
+	})
 	db, err := sql.Open("pgx", connStr)
 	if err != nil {
 		return nil, fmt.Errorf("open pg: %w", err)
 	}
 	return db, nil
-}
-
-// quotePGConnValue quotes one keyword/value DSN value per libpq rules:
-// wrapped in single quotes with backslash and single-quote escaped.
-func quotePGConnValue(v string) string {
-	v = strings.ReplaceAll(v, `\`, `\\`)
-	v = strings.ReplaceAll(v, `'`, `\'`)
-	return "'" + v + "'"
 }
--- a/cmd/tools/manifest_reconcile_test.go
+++ b/cmd/tools/manifest_reconcile_test.go
@@ -86,21 +86,6 @@ func TestReconcileExitError_ToolFailureBeatsDiscrepancy(t *testing.T) {
 	}
 }
 
-func TestQuotePGConnValue(t *testing.T) {
-	tests := []struct{ in, want string }{
-		{"plain", "'plain'"},
-		{"pa ss", "'pa ss'"},
-		{"it's", `'it\'s'`},
-		{`back\slash`, `'back\\slash'`},
-		{"", "''"},
-	}
-	for _, tt := range tests {
-		if got := quotePGConnValue(tt.in); got != tt.want {
-			t.Fatalf("quotePGConnValue(%q) = %s, want %s", tt.in, got, tt.want)
-		}
-	}
-}
-
 func TestParseReconcileFlags_Defaults(t *testing.T) {
 	opts, err := parseReconcileFlags([]string{
 		"--s3-bucket", "bkt",
--- a/docs/manifest-reconcile.md
+++ b/docs/manifest-reconcile.md
@@ -25,7 +25,7 @@ manifest 条目，双向报告差异，并可选修复。它是两条既有故�
 |------|------|------|------|
 | delta 孤儿 | `{prefix}/{schemaID}/{uuid}.parquet` | #197 或 #188 删除失败的 merged source | `--repair` 经守卫判定后补录或转残留（见下） |
 | merged base 孤儿 | `base-{uuid}.parquet` | #188 | `--gc` 两阶段删除（见下） |
-| init base 孤儿 | `{min}_{max}.parquet` | cdc-init 导出 | 只报告，**永不自动删除**：in-flight cdc-init 先落对象、跑完才发布 manifest，且不持 advisory lock（自动处置需 init 先持锁，见 follow-up） |
+| init base 孤儿 | `{min}_{max}.parquet` | cdc-init 导出 | `--gc` 两阶段删除（见下）。#290 起 cdc-init 与 reconcile 持同一把 per-schema advisory lock，故此锁下的 init 形态孤儿必非 in-flight init——只可能是发布失败的 manifest 或被后续 init 覆盖的旧文件（`--repair` 自动重建留 follow-up） |
 | `_tmp` 孤儿 | `{prefix}/{schemaID}/_tmp/*` | staging 残留（#188 崩溃 / #226 swallowed-delete） | `--gc` 两阶段删除（见下） |
 
 形态匹配是严格的：`base-` 后缀与 `{min}_{max}` 两段都必须是合法 UUID，否则归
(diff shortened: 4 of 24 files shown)


iceboundrock:
## Summary

3 findings: **1 P1 credential leak**, 1 P2 error-context standards violation, and 1 P3 operator-documentation omission. The advisory-lock and GC design otherwise aligns with #290. Do not merge until the P1 redaction regression is fixed and tested with apostrophe-plus-backslash passwords.

iceboundrock:
## Spec

The core #290 behavior is implemented: cdc-init takes the shared per-schema lock over export and manifest publication, contention is surfaced as `ErrSchemaLockContended` while other schemas continue, and reconcile includes init-shaped base orphans in the two-phase GC path. The shared DSN builder and lock consolidation also match the issue notes.

Spec compliance is partial: the PR claims quoted password values and quote-safe redaction, but Finding P1 leaks quote-containing passwords in Info-level export logs. The implementation plan also explicitly calls for updating `cmd/tools/main.go` help text; Finding P3 remains incomplete. No unrequested scope creep found.

iceboundrock:
## Standards

Finding P2 violates `coding-standard.md` section 3 and `AGENTS.md:69`: errors must carry context via `fmt.Errorf(...: %w)` rather than be returned directly.

No other documented-standard violations found. Changed Go files remain within the 500-line file limit; changed functions remain within the 100-line limit, retain early-return control flow, and exported additions have GoDoc comments.

iceboundrock:
## Findings

1. **P1 - quote-containing PostgreSQL passwords still leak to Info logs.** [`internal/cdc/helpers.go:25`](https://github.com/Lychee-Technology/forma/blob/b18685ca0808d73d664623fbde6699c36a9b5cb4/internal/cdc/helpers.go#L25) treats an escaped apostrophe as the closing delimiter. With password `p\'w\\d`, `BuildPGDSN` produces `password='p\\'w\\\\d'`; the `'[^']*'` alternative redacts only `password='p\\'` and leaves `w\\d'`. The doubled SQL-literal form fails analogously. Both [`flusher.go:364`](https://github.com/Lychee-Technology/forma/blob/b18685ca0808d73d664623fbde6699c36a9b5cb4/internal/cdc/flusher.go#L364) and [`duckdb_exporter.go:180`](https://github.com/Lychee-Technology/forma/blob/b18685ca0808d73d664623fbde6699c36a9b5cb4/internal/cdc/duckdb_exporter.go#L180) log this value. The new tests use only `secret123`, so they do not exercise the advertised quote/backslash handling. Add raw and `escapeLiteral` redaction cases containing both an apostrophe and a backslash, then make the parser consume libpq escapes correctly.

2. **P2 - missing error context.** [`internal/cdc/init.go:268`](https://github.com/Lychee-Technology/forma/blob/b18685ca0808d73d664623fbde6699c36a9b5cb4/internal/cdc/init.go#L268) directly returns `initSchemaFn` errors. Wrap the error with the schema operation/context before returning it.

3. **P3 - CLI help omits the new capability.** [`cmd/tools/main.go:95`](https://github.com/Lychee-Technology/forma/blob/b18685ca0808d73d664623fbde6699c36a9b5cb4/cmd/tools/main.go#L95) still says `--gc compaction leftovers`, but `--gc` now also handles init-shaped base orphans.

iceboundrock:
## Verification

Reviewed the complete `main...feat/290-init-advisory-lock` PR diff, all seven commits, issue #290, the checked-in implementation plan, and the documented coding standards.

Current GitHub checks report passing: Benchmark Smoke, E2E Smoke Tests, Lint, Test, and k6 Smoke. I did not run the suite locally because this workspace remains on `main`, not the PR head; the P1 redaction failure was established by tracing the submitted regex against the libpq-escaped DSN format and its SQL-literal form.

iceboundrock:
## Assessment

**Changes requested.** Finding P1 is a credential-disclosure regression: valid PostgreSQL passwords containing an apostrophe can be partially emitted in `sql_preview` / `pgConnForDuck` Info logs. Fix and cover it before merge.

Findings P2 and P3 are non-blocking in isolation, but should be resolved in the same correction because both are small and directly related to this PR.

iceboundrock:
## Observations

- The pinned-connection helper correctly centralizes the advisory-lock key and avoids acquire/release crossing pooled sessions.
- Init holds the lock for zero-row and dry-run paths as intended; a contended init is reported while a contended flusher retains its skip-and-retry behavior.
- Long init runs can starve flush progress for the schema, but this is an intentional and documented tradeoff to close the init/reconcile race.
- No scope-creep concerns found.

iceboundrock:
Review round 1 addressed in 3eaceab — all three findings were real; fixes below.

**P1 (credential leak, fixed).** Confirmed by trace and by red tests: the quoted branch `'[^']*'` treated the libpq escape `\'` as the closing quote, and the doubled branch's non-greedy `*?` closed early at `\''` — with password `p'w\d` both log sites leaked the tail (`password=***REDACTED***w\\d'` reproduced in the red run). `redactConnStr` now uses an escape-aware alternation: doubled form `''(?:\\''|\\[^']|[^'\\])*''` (a bare `'` can never be consumed by a content atom, so the early-close bug is structurally gone and greedy matching is safe), raw quoted form `'(?:\\.|[^'\\])*'`, legacy unquoted unchanged. New tests pin hostile (`p'w\d` — apostrophe + backslash) and empty passwords through both the raw `BuildPGDSN` form and the `escapeLiteral`'d ATTACH form, composed through the real pipeline; trailing `dbname=`/`sslmode=` keys are asserted to survive un-mangled. One note vs. the finding text: after `quotePGConnValue` doubles backslashes the true leaked tail is `w\\d`, not `w\d` — the tests assert the actual post-quoting byte sequence (verified red against the old regex).

**P2 (bare error passthrough, fixed).** `initSchemaUnderLock` now wraps: `fmt.Errorf("init export+publish under schema lock: %w", err)`. The schema ID is deliberately not in the message — the caller `processInitSchemas` already wraps `schema %d: %w`, and double-naming was rejected in an earlier review round. Since `init.go` sat at exactly the 500-line ceiling, the wrapper moved to a new `internal/cdc/init_lock.go` (init.go is now 475 lines); `init_lock_test.go` passes unchanged.

**P3 (stale help text, fixed).** `cmd/tools/main.go` now reads `--gc stale base/tmp leftovers (compaction + init-shaped)`. The `--gc` flag usage string in `manifest_reconcile.go` was checked in the same pass and updated where stale.

Verification at 3eaceab: `go test ./internal/cdc ./cmd/tools` green (incl. 4 new hostile/empty-password redaction tests), `go vet -tags=e2e` on the production harness clean, `make lint` clean. CI re-running on the push.

Issue #294, opened 2026-07-20
OLTP update rejects rows carrying a dropped attribute's EAV data after schema evolution


iceboundrock:
Discovered while building the #257 e2e (mixed-generation compaction equivalence).

## Symptom

After a schema evolution that REMOVES an attribute, updating any row that still holds that attribute's EAV data fails loudly:

```
entity manager update: failed to transform existing record: failed to convert EAVRecords to EntityAttributes: unknown attribute id 3 for schema 20 (attribute not in metadata cache)
```

## Code path

`internal/entity_crud_service.go` `Update` loads the existing record and runs `s.transformer.FromPersistentRecord` over ALL of its EAV records under the CURRENT metadata cache (`internal/transform/attribute_converter.go` `FromEAVRecords`), which hard-errors on any attrID no longer present in the schema. The update never reaches the merge step, so the row becomes effectively read-only for partial updates until its stale EAV rows are removed out of band.

`Delete` is unaffected (no transform of the existing record), and the read path is unaffected (the projection simply never selects the dropped attribute — pinned by `TestSchemaEvolutionRemovedColumn`, #189).

## Repro

1. Schema v1 with EAV attribute `old_col` (attrID 3); create a row writing `old_col`.
2. Evolve to v2 with `old_col` removed (same registry reload as `Env.EvolveSchema`).
3. `Update` that row with any v2-valid attrs → error above.

The #257 fixture routes around this deliberately (`evoV1Profile` in `internal/e2e_harness/production/compaction_evolution_e2e_test.go` only writes `old_col` on rows that are never updated); when this issue is fixed, that fixture comment should be revisited.

## Open question (needs a design decision)

Is this a bug or a guardrail? Options include: (a) tolerate-and-drop unknown attrIDs on the update read-back (the stale EAV rows are dead weight the schema can no longer address), (b) tolerate-and-preserve them untouched, (c) keep rejecting but document that attribute removal requires an EAV cleanup migration first. The current behavior makes attribute removal unsafe for live rows without any documented cleanup path.

iceboundrock:
Follow-up from merged PR #295: its E2E fixture now uses a test-only cleanup-migration path to create a non-vacuous row-level LWW case. PR #295 contains no production update-path fix, so this issue remains open pending the design decision and production remediation described above.

iceboundrock:
## Direction adjudicated: tolerate-and-preserve

[...]

PR #295, opened 2026-07-20
test(e2e): #257 mixed-generation compaction equivalence

### LINKED ISSUE #257: [E2E] Mixed-generation compaction equivalence
Spun out of #189 (PR #254).

#189 added `union_by_name=true` to the compaction merge SQL (`internal/compaction/merge_sql.go`), so a merge spanning schema generations physically materializes the column union (NULLs where a generation lacked a column) instead of failing to unify — evolution heals monotonically through compaction. Unit-level SQL pins were updated, and #188's same-generation equivalence suites stay green, but there is no e2e that runs the real compactor over a mixed-generation base+delta set and asserts before/after query equivalence.

Scope: extend the #188 equivalence harness with an evolution fixture (the #189 `Env.EvolveSchema` + `writeSimpleSchemaDir` machinery is reusable): seed base under v1, delta under v2 (added + removed + retyped columns), `RunCompaction`, then assert (a) query results identical before/after, (b) the merged base's physical schema is the union with the widened types, (c) a subsequent v2 flush + query over (merged base + new delta) still resolves.

iceboundrock:
Closes #257.

Characterization e2e for the union_by_name mixed-generation merge (#189, PR #254): seeds base under v1 and delta under v2 — one evolution step carrying all three mutation kinds (removed `old_col` + retyped `score` integer→numeric + added `new_col`) — runs the real compactor, and asserts:

- **(a) bit-for-bit before/after query equivalence** over a 4-query set (full page, `value` sort, cross-generation `score gte` filter, v2-only `new_col` filter), with oracle equality on both sides; rewrite is triggered naturally by v2 updates/delete over v1 base rows (dirty ratio 60%), so the merge must fold cross-generation versions of the same `row_id`. RowsIn=12 / RowsOut=8 pinned.
- **(b) the merged base is the monotonic column union with widened types** — `old_col` VARCHAR retained, `new_col` INTEGER present, `score` DOUBLE (INTEGER widened) — and holds exactly the row-level LWW winners: v2 winners carry fractional scores and `new_col`, untouched v1 rows keep `old_col` values and integer-valued scores, the tombstoned row is physically gone, zero tombstones, `deleted_at` normalized.
- **(c) the schema stays evolvable**: a post-compaction v2 flush layers a fresh delta on the union-shaped base, all queries still resolve via DuckDB, and a SECOND compaction pass rewrites again (RowsIn=13/RowsOut=12) keeping the union shape and the v1 legacy column data (monotonic healing).

**Harness change:** `assertCompactionEquivalence` now takes the query set as a parameter (the previous hard-coded sort on `count` was e2e_wide-only); 8 existing call sites updated mechanically, no behavior change.

**Mutation evidence** (test bites): removing `union_by_name=true` from `internal/compaction/merge_sql.go` makes the suite fail loudly at the compaction step with `Invalid Input Error: … schema mismatch in glob: column "old_col" was read from the original file …, but could not be found in file …` — a hard error, not silent corruption. The flag was restored; this PR carries zero production-code changes.

**Found along the way → #294:** after an attribute is dropped, the OLTP update path rejects any row still carrying that attribute's EAV data (`Update` → `FromPersistentRecord` transforms all existing EAV records under current metadata). The fixture routes around it (`evoV1Profile` writes `old_col` only on never-updated rows) and documents the constraint in a comment.

**Verification:** `make test` green; `make test-e2e-production` green (104 tests, includes the new suite and all re-signed call sites); `make lint` clean; new file 358 lines, all functions <100 lines.

Spec: `docs/superpowers/specs/2026-07-20-issue-257-mixed-generation-compaction-equivalence-design.md` · Plan: `docs/superpowers/plans/2026-07-20-issue-257-mixed-generation-compaction-e2e.md`

### DIFF
--- a/docs/superpowers/plans/2026-07-20-issue-257-mixed-generation-compaction-e2e.md
+++ b/docs/superpowers/plans/2026-07-20-issue-257-mixed-generation-compaction-e2e.md
@@ -1,0 +1,648 @@
+# Mixed-Generation Compaction Equivalence E2E (#257) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Characterization e2e proving the real compactor's mixed-generation merge (union_by_name, #189) keeps federated queries bit-for-bit equivalent, materializes the widened column union, and stays mergeable afterward.
+
+**Architecture:** One new e2e file in `internal/e2e_harness/production` composes the existing #189 evolution machinery (`writeSimpleSchemaDir`/`EvolveSchema`) with the #188 equivalence helper (`assertCompactionEquivalence`, parametrized here to accept a query set). Single combined v1→v2 fixture: removed `old_col` + retyped `score` (integer→numeric) + added `new_col`. Rewrite is triggered naturally by v2 updates/delete on v1 base rows (dirty ratio 3/5 = 60%).
+
+**Tech Stack:** Go, testcontainers (Docker: Postgres + RustFS), DuckDB. Zero production-code changes — expected green on first run; a mutation check (Task 4) proves the test bites.
+
+**Context:** Issue https://github.com/Lychee-Technology/forma/issues/257, spun out of #189 (PR #254). Spec (user-approved): `docs/superpowers/specs/2026-07-20-issue-257-mixed-generation-compaction-equivalence-design.md`. #189 added `union_by_name=true` to the compaction merge SQL (`internal/compaction/merge_sql.go:69`), but no e2e runs the real compactor over a mixed-generation base+delta set.
+
+## Global Constraints
+
+- Source files ≤500 lines, functions ≤100 lines (coding-standard.md).
+- Always wrap errors with context; test failure messages name the attribute/column and the expected state.
+- All test commands use the Makefile env: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false` (avoids sandbox cache/VCS errors).
+- E2E requires Docker. Do NOT run e2e and unit suites concurrently (known avalanche false-failure trap).
+- 请不要自动合并 PR；PR body 关联 #257。
+- Production code must not change. If the test surfaces a real defect, STOP and report — file an issue, don't fix inline.
+- Known-flaky, NOT caused by this branch: `TestConcurrentFlushSnapshot`/`UpdateBeforeExport` same-millisecond `changed_at` flake (#276), `TestPerformance_ConcurrentQueries` ~89% hit-rate jitter.
+
+---
+
+### Task 0: Worktree + plan doc
+
+**Files:**
+- Create: worktree at `../forma-257` on branch `e2e/257-mixed-gen-compaction-equivalence`
+- Create: `docs/superpowers/plans/2026-07-20-issue-257-mixed-generation-compaction-e2e.md` (this plan, copied into the worktree)
+
+- [ ] **Step 1: Clean up merged branches, fast-forward main** (workspace AGENTS.md worktree rule)
+
+```bash
+cd /Users/ruoshi/code/Lychee/LTBase/forma
+git fetch origin --prune
+git branch --merged main | grep -v 'main' | xargs -r git branch -d
+git checkout main && git merge --ff-only origin/main
+```
+
+- [ ] **Step 2: Create worktree** (use explicit `git worktree add` + explicit paths in all later commands — EnterWorktree has a known session-failure trap)
+
+```bash
+git worktree add ../forma-257 -b e2e/257-mixed-gen-compaction-equivalence main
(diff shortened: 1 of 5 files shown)


iceboundrock (commented):
## Findings

1. **P1 - row-level LWW is not actually covered.** `evoV1Profile` writes `old_col` only for ordinals 0-2, while the v2 updates target base rows 3 and 4 (`compaction_evolution_e2e_test.go:67,124-127`). Those v1 source rows already have `old_col = NULL`, so the assertion that their merged v2 winners have NULL passes even if compaction incorrectly column-merges and preserves the base value. The approved design requires the two updated rows to demonstrate that row-level LWW replaces the whole row, including `old_col` becoming NULL (`docs/superpowers/specs/2026-07-20-issue-257-mixed-generation-compaction-equivalence-design.md:122-125`). Please use a fixture path that makes the pre-update v1 values non-NULL while still producing the v2 update, or revise the approved design and PR claim to state this behavior is not covered.

2. **P2 - query set differs from the approved design.** The score predicate is `score gte 15` (`compaction_evolution_e2e_test.go:91`), but the design specifies `score gte 20` (`docs/superpowers/specs/2026-07-20-issue-257-mixed-generation-compaction-equivalence-design.md:90-94`). The current seed makes the count identical, but the implementation should use the documented threshold or update the design with the reason for the change.

## Spec

The real compactor, mixed v1/v2 parquet shapes, union-column/type checks, hot-tier guard, query equivalence, and second compaction align with #257 and the design. The two deviations above leave the row-level LWW requirement unproven and one approved query shape unimplemented.

## Standards

- `coding-standard.md:52-56` requires verb-first private helpers. `evoV1Profile`, `evoV2Profile`, and `evolutionEquivalenceQueries` should be renamed, for example, to `buildEvoV1Profile`, `buildEvoV2Profile`, and `buildEvolutionEquivalenceQueries`.
- `TestCompactionMixedGenerationEquivalence` is 85 lines, below the required 100-line maximum but above the documented >80-line refactoring trigger (`coding-standard.md:7-11,82-88`). Extracting the manifest/hot-tier block would improve cohesion.

## Assessment

Not approving yet: P1 removes a central correctness guarantee from this characterization test, and P2 is a direct spec mismatch. The change remains appropriately scoped to tests and documentation.

## Observations

The fixture clearly documents the existing dropped-attribute OLTP update constraint tracked in #294. That constraint explains the current seed selection, but it does not make a vacuous LWW assertion valid; coverage or the approved design needs to change.

## Verification

- Reviewed PR #295 against `main` (commits `5fd6cf5..65b42fb`) and issue #257.
- GitHub Actions currently reports success for Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- Static review: new source file is 358 lines; the largest scenario function is 85 lines.
- I did not rerun the Docker-backed production E2E suite locally.

## Summary

Two spec findings: one P1 coverage gap and one P2 query mismatch. One naming-standard violation and one non-blocking refactoring trigger. Please address the findings before approval.

iceboundrock:
Review round-1 addressed in 807b931. Per finding:

**P1 (row-level LWW vacuous) — confirmed and fixed.** The conditional-`old_col` seed made the winner's NULL assertion unfalsifiable, exactly as stated. Fix takes the fixture through the #294 option-(c) path instead of routing around it: every v1 row now carries `old_col`; after `EvolveSchema` the two update targets get the documented stale-EAV cleanup (`DELETE FROM eav_data … attr_id = 3` for those row_ids), so the OLTP update succeeds while their `old_col` values remain exported in the base parquet. A new `requireBaseOldCol` positive control pins those base values as non-NULL (`old-0003`/`old-0004`) before compaction — the merged winner showing NULL now proves whole-row replacement; a column merge would leak the base values and fail. The tombstoned row keeps its stale EAV untouched (delete does no transform), so the fixture still documents both halves of the #294 constraint. Mutation check re-run against the strengthened fixture: still fails loudly (`schema mismatch in glob: column "old_col" … could not be found`). Spec seed recipe updated with the detour and positive control.

**P2 (threshold 15 vs 20) — fixed to the design's `gte 20`.** The 15 was an unforced deviation left over from re-picking the update targets during planning; with untouched rows at scores 10 and 20, the approved threshold discriminates identically (excludes the 10, boundary-includes the 20) — the boundary inclusion is a small bonus the design value buys. Expected count stays 10.

**Standards (verb-first) — renamed:** `buildEvoV1Profile`, `buildEvoV2Profile`, `buildEvolutionEquivalenceQueries` (call sites and the `assertCompactionEquivalence` doc comment updated). Note the pre-existing `compactionEquivalenceQueries` shares the old pattern; left untouched as out of scope for this PR — can follow up if you want it aligned.

**Standards (>80-line trigger) — extracted.** Manifest + hot-tier surfaces moved to `assertMixedGenRewriteSurfaces`; the scenario function is now 74 lines. Also deduplicated the three s3-URI constructions into `buildParquetS3Path`.

Verification after the changes: targeted test green, full `make test-e2e-production` green (104 tests), `make lint` clean, `merge_sql.go` diff-clean after the mutation re-check.

PR #298, opened 2026-07-21
fix(cdc/federated): #252 flush ordering — close the mark→append transient-invisibility window

### LINKED ISSUE #252: Flush ordering: mark-flushed precedes manifest-append — transient invisibility window under manifest-driven reads
Follow-up from #187 (PR #249). Related: #203, #226.

The flush batch pipeline commits in this order (`internal/cdc/flush_batch.go`, `executeBatch`):

1. `CopyTmpToFinal` — the delta parquet exists in storage
2. `MarkFlushedIDsAtSnapshot` — rows leave the hot tier (`flushed_at != 0`)
3. `updateManifest` — the delta is appended to the manifest

Under the glob-based reads #187 replaced, step order didn't matter for visibility: the object was globbable the moment step 1 finished. Under manifest-driven reads (#187), a federated query landing **between steps 2 and 3** sees the batch's rows in *neither* tier: they are no longer hot (marked flushed, excluded from the PG anchor and the dirty set) and not yet in the listed delta (manifest not appended). A sub-second transient-invisibility window per batch.

Not a #187 test-fault vector — the harness flush is synchronous and no test queries mid-flush — but it is a real production window once manifest-driven reads are wired server-side (#250).

## Constraints on a fix

Naively swapping steps 2↔3 (append manifest before mark-flushed) reintroduces a different hazard: a crash between them leaves rows unmarked → the retry re-exports them into a **second** listed delta (duplicate live versions are LWW-safe but violate the #179-pinned "no row exported twice" expectations, and the #197 contract — error embeds the orphaned final key; retry is a no-op — was adjudicated around the current order).

Options to evaluate:
- append manifest first + make re-export idempotent per batch (deterministic delta key per snapshot?)
- two-phase manifest entry (pending → sealed), readers ignore pending
- read-side grace: dirty-set anti-join keyed on `flushed_at > query snapshot` rather than `= 0` so just-flushed rows stay hot-readable until the manifest catches up

Any change must keep the #187 direction contract (manifest ⊆ live objects) and re-run the #179 failure matrix.

iceboundrock:
Closes #252.

## Problem

`executeBatch` committed **copy → mark-flushed → manifest-append**. Under manifest-driven reads (#187), a federated query landing between mark and append saw the batch's rows in *neither* tier: no longer hot (`flushed_at != 0` leaves the dirty set and the pg_source gate), not yet in the listed delta. Reproduced red by `TestFlushManifestWindowVisibility` (4 of 8 rows visible mid-flush).

## Fix (adjudicated: (a)+(c) belt-and-braces)

**(a) Write-side reorder (primary):** `copy → manifest-append → mark-flushed`. The middle state — delta listed, rows still `flushed_at = 0` — is safe by construction: the dirty anti-join discards the S3 copies and hot pg_source serves them (the exact scenario the anti-join was designed for). The manifest entry is built from `batchIDs` (mark hasn't run); RowCount/bounds may overcount concurrently-changed rows, feeding only compaction's promotion heuristic (reconcile recomputes from parquet).

**(c) Read-side grace (backstop):** the residual reader race — path set resolved before the append, dirty scan after the mark — is covered by widening the dirty barrier to `(flushed_at = 0 OR flushed_at > cutoff)`, where **cutoff = the instant this query resolved its parquet path set** minus a configurable clock-skew margin (`DuckDBConfig.FlushVisibilityGraceMs`, default 0 = exact anchor, negative = disabled). Because (a) appends before marking, rows flushed before path resolution already have their delta listed → **the steady state is never widened** (zero semantic drift); only rows racing the in-flight query stay hot-readable. The widening renders only in the HasHot form (hot-excluded shapes keep the strict barrier — no hot server exists for discarded rows). The cutoff splices via a compile-sentinel mirroring `dirtyIDsSentinel`, keeping the plan-cache key cutoff-independent; every render path defaults an absent cutoff to `MaxInt64` (never `0`, which would bypass the parquet tiers; never empty, which is invalid SQL).

> Design note: the first-cut wall-clock grace (`now − 60s`) failed full-suite verification — it flipped `created_at` semantics for recently-flushed rows (s3 `changed_at` proxy vs PG creation time; broke the #212 keyset pins and the oracle), dropped rows in hot-excluded tier forms, and defeated the reconcile e2e's stripped-manifest premise. The path-resolution anchor eliminates all three while covering the actual race.

## Contract re-adjudications (superseding prior pins)

- **#197 orphan contract superseded:** manifest-append failure now leaves rows dirty → retry self-heals with a fresh UUIDv7 key; the old copied final is an unlisted orphan for `manifest-reconcile --gc` (ClassDelta covered leftover). Recovery improves from "data loss until operator `--repair`" to "self-heal + junk". The final key stays in the error for observability.
- **#179 "no row exported twice" relaxes to crash-free runs:** mark failing after append → retry produces a second *listed* delta with identical `(row_id, ver_ts)` rows; LWW rn=1 + #183 tie-break dedups, compaction collapses later.
- **#187 direction contract unchanged:** copy still precedes append (manifest ⊆ live objects).
- `updatedIDs == 0` batches leave an LWW-inert junk delta listed — accepted, not rolled back.
- **Preserved invariants:** keys stay random UUIDv7 (bare-UUID → ClassDelta), so #226 tmp self-heal reasoning, the #290 advisory-lock/init-GC proofs, and the reconcile classifier are untouched.

## Tests

- Red-first e2e `TestFlushManifestWindowVisibility` (pauses the flush at its manifest PUT; grace disabled so it pins the reorder in isolation): red pre-fix, green post-fix.
- `TestFlushGraceBackstopServesStaleManifestSnapshot`: margin emulates a pre-flush path snapshot → rows hot-served; exact-anchor default → steady state not widened (zero drift pin).
- #179 matrix re-adjudicated: `TestFlushFaultManifestLoad/Save` (rows stay dirty, retry self-heals, old final = unlisted orphan), `TestFlushFaultMarkFlushed` (first delta LISTED, retry adds second listed delta, LWW dedup), multi-schema `ManifestSaveFault` (faulted schema self-heals, sibling untouched).
- Unit: flipped `TestExecuteBatch_ReturnsErrorWhenManifestLoad/SaveFails`, new `TestExecuteBatch_MarkFailsAfterManifestAppend` (via new `markFlushed` seam); sqlgen pins for both widened render sites, HasHot gating, sentinel splice, skeleton cutoff-independence, defensive defaults; `TestFlushVisibilityGraceResolution` for knob semantics. §5 executable-doc updated and still executes.

## Verification

- `make lint` clean; `make test` (full unit) green.
- Full `internal/e2e_harness/production` suite green (was 6 failures under the wall-clock design — all resolved by the anchor revision, no oracle/#212 pin changes needed).
- Infra smoke + `-tags=e2e -short` federated suite + benchmark package green.
- Docs updated: `docs/federated-query/design.md` (DirtySet definition, §5 template + params), `docs/manifest-reconcile.md` (delta-orphan provenance), production harness `doc.go` consumer map.

### DIFF
--- a/config_duckdb.go
+++ b/config_duckdb.go
@@ -33,13 +33,14 @@ type DuckDBConfig struct {
 	// tripping. Zero means use the built-in default.
 	CircuitBreakerOpenDuration time.Duration `json:"circuitBreakerOpenDuration"`
 
-	// FlushVisibilityGraceMs widens the federated dirty barrier (#252): rows
-	// whose flushed_at is within this many milliseconds of query start stay
-	// hot-readable, covering the reader-side staleness between loading the
-	// manifest and scanning change_log while a flush publishes. Zero means the
-	// built-in default (60s); a negative value disables the grace and restores
-	// the exact flushed_at = 0 barrier. Widening is a safe over-approximation:
-	// affected rows are served from Postgres, which always holds current state.
+	// FlushVisibilityGraceMs is the #252 clock-skew margin for the federated
+	// dirty barrier. Each query anchors its cutoff at the instant it resolved
+	// its parquet path set: rows marked flushed after that instant (their
+	// delta may be missing from the resolved set) stay hot-readable, minus
+	// this margin to absorb CDC-host vs query-host clock skew. Zero (the
+	// default) is the exact anchor — the steady state is never widened; a
+	// positive value hot-serves rows flushed up to that long before the
+	// query; a negative value disables the widening (the pre-#252 barrier).
 	FlushVisibilityGraceMs int64 `json:"flushVisibilityGraceMs"`
 
 	Routing RoutingPolicy `json:"routing"` // routing policy for federated queries
--- a/config_duckdb.go
+++ b/config_duckdb.go
@@ -33,6 +33,15 @@ type DuckDBConfig struct {
 	// tripping. Zero means use the built-in default.
 	CircuitBreakerOpenDuration time.Duration `json:"circuitBreakerOpenDuration"`
 
+	// FlushVisibilityGraceMs widens the federated dirty barrier (#252): rows
+	// whose flushed_at is within this many milliseconds of query start stay
+	// hot-readable, covering the reader-side staleness between loading the
+	// manifest and scanning change_log while a flush publishes. Zero means the
+	// built-in default (60s); a negative value disables the grace and restores
+	// the exact flushed_at = 0 barrier. Widening is a safe over-approximation:
+	// affected rows are served from Postgres, which always holds current state.
+	FlushVisibilityGraceMs int64 `json:"flushVisibilityGraceMs"`
+
 	Routing RoutingPolicy `json:"routing"` // routing policy for federated queries
 }
 
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -47,7 +47,7 @@ Formula:
 
 $$Result = (S3_{Data} \notin DirtySet) \cup PG_{HotData}$$
 
-* **DirtySet:** The set of row_ids currently present in the PostgreSQL change_log with flushed_at = 0, widened by the **flush-visibility grace** (#252): rows marked flushed after the instant this query resolved its parquet path set (minus the `FlushVisibilityGraceMs` clock-skew margin, default 0) also count as dirty. Because the flush appends the manifest before marking, a row flushed before path resolution already has its delta listed — the widening therefore catches exactly the rows racing this query (their delta may be missing from the resolved set) and keeps them hot-readable, while the steady state is never widened. The widening renders only in the HasHot tier form: it is safe solely because pg_source serves the discarded rows, so hot-excluded queries keep the strict flushed_at = 0 barrier.
+* **DirtySet:** The set of row_ids currently present in the PostgreSQL change_log with flushed_at = 0, widened by the **flush-visibility grace** (#252): rows marked flushed at or after the instant this query resolved its parquet path set (minus the `FlushVisibilityGraceMs` clock-skew margin, default 0) also count as dirty — inclusive, because millisecond stamps cannot order a mark and a path resolution landing in the same tick. Because the flush appends the manifest before marking, a row flushed before path resolution already has its delta listed — the widening therefore catches exactly the rows racing this query (their delta may be missing from the resolved set) and keeps them hot-readable, while the steady state is never widened. The widening renders only in the HasHot tier form: it is safe solely because pg_source serves the discarded rows, so hot-excluded queries keep the strict flushed_at = 0 barrier.
 * Any record found in S3 that also exists in the DirtySet is **discarded** immediately during the read phase, regardless of its timestamp.
 * **Flush ordering (#252):** the CDC flush publishes `copy tmp→final → manifest-append → mark-flushed`, so a batch is never simultaneously absent from the hot tier and from the manifest-listed delta set. The listed-but-unmarked middle state is resolved by this anti-join: the S3 copies of still-dirty rows are discarded and the hot versions serve.
 
@@ -172,7 +172,7 @@ dirty_ids AS (
     SELECT row_id   
     FROM postgres_scan($PG_CONN, 'public', 'change_log')  
     WHERE schema_id = $SCHEMA_ID  
-        AND (flushed_at = 0 OR flushed_at > $FLUSH_GRACE_CUTOFF_MS)  
+        AND (flushed_at = 0 OR flushed_at >= $FLUSH_GRACE_CUTOFF_MS)
 ),
 
 -- =========================================================================  
@@ -242,7 +242,7 @@ pg_source AS (
       ON cl.schema_id = e.schema_id AND cl.row_id = e.row_id  
       
     WHERE cl.schema_id = $SCHEMA_ID  
-        AND (cl.flushed_at = 0 OR cl.flushed_at > $FLUSH_GRACE_CUTOFF_MS)  
+        AND (cl.flushed_at = 0 OR cl.flushed_at >= $FLUSH_GRACE_CUTOFF_MS)
         AND m.ltbase_schema_id = $SCHEMA_ID  
         AND ($PG_WHERE_CLAUSE)  
     GROUP BY m.ltbase_row_id, m.ltbase_created_at, cl.changed_at, cl.deleted_at, m.text_01, m.integer_01  
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -47,8 +47,9 @@ Formula:
 
 $$Result = (S3_{Data} \notin DirtySet) \cup PG_{HotData}$$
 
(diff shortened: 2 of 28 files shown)


iceboundrock (commented):
## Findings

1. **P0 - CI is failing and the new plan-cache tests are flaky.** `TestEngineCompiledPlanCache` and `TestEnginePlanCachePathParity` compare fully rendered SQL from separate queries. Each query now embeds a fresh millisecond cutoff, so crossing a millisecond changes only the two `flushed_at > ...` literals. GitHub Actions already fails for this reason, and this review reproduced it with `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/federated -run 'TestEngine(CompiledPlanCache|PlanCachePathParity)' -count=50`. Control time in these tests or compare the cached skeleton / normalize the per-request cutoff; the PR cannot merge with `make test` failing.

2. **P1 - `flushed_at` is sampled before manifest append, which recreates the invisibility race.** In `internal/cdc/flush_batch.go:103-114`, let query Q stamp and resolve its paths after line 103 but before the manifest append finishes. Q has no new delta path, while the later mark writes the already-captured value. Since `flushed_at <= Q.cutoff`, the widened dirty predicate and `pg_source` both exclude the row, so it is in neither source. Sample the marking timestamp after the append and make the comparison safe at millisecond precision (the strict `>` boundary can still lose equality).

3. **P2 - the read-side E2E does not exercise the required exact-anchor in-flight race.** `cdc_flush_grace_e2e_test.go:76-100` completes both flushes before querying and uses a 60s margin to emulate staleness. It cannot catch P1. Pause a query after path resolution, then append and mark a flush; assert default exact-anchor returns the row while disabled grace exposes the gap.

## Spec

The write ordering meets #252 directionally, but P1 violates the design requirement that a row marked after a query path snapshot remains in that query's dirty set. P2 leaves the mandated in-flight visibility proof unpinned. No material scope creep found.

## Standards

- `StreamDuckDBFederatedQuery` is now 118 lines (`internal/federated/duckdb_query.go:70-187`), exceeding the 100-line maximum in `AGENTS.md:68` and `coding-standard.md:7-11`.
- The modified immediate area still returns bare errors at `duckdb_query.go:135-141,176`; the project requires contextual wrapping (`AGENTS.md:69`, `coding-standard.md:22-33,93-98`).
- `git diff --check` reports trailing whitespace in `docs/federated-query/design.md:175,245`.

## Assessment

**Changes requested.** The append-before-mark direction and sentinel-based cache design are reasonable, but P0 and P1 block correctness and mergeability.

## Observations

The PR documents the revised manifest-failure contracts and preserves the manifest-is-subset-of-live-objects direction. Hot-excluded gating is appropriately considered.

## Verification

- GitHub Actions: `Lint`, `Benchmark Smoke`, `E2E Smoke Tests`, and `k6 Smoke` pass; `Test` fails.
- Fresh PR worktree: the targeted cache tests fail under `-count=50`, with the cutoff literal differing by 1ms.
- `git diff --check base...HEAD` fails on the two documented trailing-whitespace lines.
- Full unit/E2E reruns were not continued after the blocking unit-test failure.

## Summary

3 findings: 2 blockers (P0, P1) and 1 missing regression proof (P2). Address those plus the listed standards issues, then rerun the full verification set.

iceboundrock:
All three findings verified and addressed in 06c5ad0. Per-finding disposition:

## P0 — flaky plan-cache tests: fixed (normalization)

Confirmed via the CI log (the two bound SQLs differed only in the cutoff literal, `...326` vs `...327`; local `-count=500` did not reproduce, the ms boundary crossing is machine-dependent). Under the anchor design, two separate requests *legitimately* differ in exactly the per-request cutoff literal — it is the same class of per-request value as the dirty CSV. `TestEngineCompiledPlanCache` and `TestEnginePlanCachePathParity` now normalize `flushed_at >= \d+` before comparing, which keeps the pins exact for every other byte: any second divergence still fails. Controlling time instead would have required a clock seam through `resolveScanSources` for no additional coverage — the skeleton/splice determinism is already pinned with fixed cutoffs in `internal/sqlgen/duckdb_flush_grace_test.go`.

## P1 — stamp sampled before append: fixed (both halves)

The timeline you describe was real: Q stamps its cutoff and resolves paths after the flush captured `flushedAt` but before the append lands → `flushed_at <= Q.cutoff` → excluded from both the widened dirty set and pg_source. Two changes:

1. `executeBatch` now samples the mark stamp **after** `updateManifest` returns. The manifest entry's `CreatedMin/Max` keep their own pre-append stamp — they only feed compaction's promotion ordering, so the two stamps diverging by the append latency is harmless (decoupling noted in the comment).
2. Both template predicates are now **inclusive** (`flushed_at >= cutoff`): millisecond stamps cannot order a mark and a path resolution landing in the same tick, and the ambiguous tick must resolve toward visibility.

Deterministic unit pin: `TestExecuteBatch_MarkStampSampledAfterManifestAppend` (mock store's `Save` slowed 5ms, asserts the written `flushed_at >=` the append-completion time). Mutation-verified: re-introducing the pre-append sampling turns it red.

One consequence of the inclusive boundary: a steady-state query whose path stamp lands in the *same millisecond* as a mark now widens (correct values, hot-served). The steady-state probe (`ExactAnchorNeverWidensSteadyState`) waits out that tick so it pins the steady state, not the boundary.

## P2 — in-flight race e2e: added

`TestFlushGraceExactAnchorCoversInFlightFlush` drives the literal race: a `pausingParquetSource` (via a new `Env.ParquetSourceWrap` seam — the race sits between path resolution and execution, where no S3-level pause can reach) suspends the query after it has stamped its cutoff and resolved a path set *without* the second delta; a full flush then appends and marks that generation; the query resumes. Default exact anchor → all 8 rows served (the raced generation comes from hot); `FlushVisibilityGraceMs = -1` → 4 rows (the uncovered gap, proving the anchor is the serving mechanism). One trap worth recording: the wrapper must install before ANY event — seeding goes through the EntityManager, which assembles the engine on first use.

## Standards

- `StreamDuckDBFederatedQuery` → 94 lines: the storage-facing pre-flight (cutoff stamp → path resolution → parquet schema validation) extracted to `resolveScanSources`.
- Bare returns at the flagged sites wrapped (`fetch dirty row ids`, `build duckdb federated query`, `stream duckdb federated rows`); `%w` chains preserved so `errors.Is` classification is unaffected.
- Trailing whitespace on `design.md:175,245` stripped; `git diff --check` clean over the branch range.

## Verification

- `make lint` clean; `make test` (full unit, incl. flipped/new cdc pins) green.
- `go test ./internal/federated -run 'TestEngine(CompiledPlanCache|PlanCachePathParity)' -count=100` ×5 green.
- Full `-tags=e2e` production suite: two non-cached runs. Round 1 had a single failure in `TestPostgresRestartRoundtrip` (container-restart timing under full parallel load); it passes in isolation, ×3 repeats, and in the round-2 full run — I did not capture round 1's message before it passed everywhere, so flagging it here for transparency rather than claiming a diagnosis.
- `-tags=e2e -short` federated suite + benchmark package green; §5 executable-doc tests green against the updated `>=` doc SQL.

PR #303, opened 2026-07-24
feat(factory/server): #250 wire manifest-driven parquet reads into forma.Config

iceboundrock:
Closes #250. Follow-up of #187 / PR #249.

## What

#187 landed manifest-driven parquet reads at the engine seam (`federated.WithParquetSource` + `manifest.QuerySource`), but only the production e2e harness wired it — `forma.Config` had no manifest surface, so in server mode the silent-loss fix and `ErrParquetSetInconsistent` classification were **inert**. This PR makes them production-real for `factory.NewEntityManagerWithConfig*` users and `cmd/server`:

- **Config surface** (user-adjudicated: extend `DuckDBConfig`, no new section): `s3Bucket`, `s3DataPrefix`, `manifestPrefix`, `manifestTemplate`. Gate = non-empty `manifestTemplate` (matches the CDC writer). Strict validation (`ValidateManifestRead`, single source of truth for `Config.Validate` and the factory): template-without-bucket, inert prefix-without-template, unparsable template ⇒ `ConfigError`, **server fails at startup** before any I/O.
- **Assembly** in `internal/manifest.NewS3QuerySource`: HeadObject exists-probe (`*types.NotFound` + strict HTTP-404 fallback ⇒ absent; anything else propagates — a misclassified probe would fabricate data-loss reports), fallback glob byte-identical to the harness (`s3://{bucket}/{dataPrefix}/{schemaID}/*.parquet`), no empty-template default leakage.
- **S3 client** in `internal/bootstrap.NewS3Client`: credential precedence mirrors `cdc setupAWSClient`; endpoint ⇒ path-style, same addressing as the DuckDB httpfs session. Factory itself stays AWS-import-free.
- **Factory**: one deps seam (`newParquetSource`), fail-fast config validation before `collectTables`, `WithParquetSource` appended only when the source is non-nil (literal-nil discipline — typed-nil would pass the engine's interface-nil guards and panic on first cold query; pinned by mutation-verified tests).
- **Server env**: `DUCKDB_S3_BUCKET→S3_BUCKET`, `DUCKDB_S3_PREFIX→S3_PREFIX`, `DUCKDB_MANIFEST_PREFIX→MANIFEST_PREFIX`, `DUCKDB_MANIFEST_TEMPLATE→MANIFEST_TEMPLATE`. **All-or-nothing semantics** (final-review adjudication, see below).
[...]

Issue #255, opened 2026-07-15
Cold-only reads fail on attributes added before their first flush (never-flushed column projection)


iceboundrock:
Spun out of #189 (PR #254), user-adjudicated out of scope there.

After #189, `read_parquet(..., union_by_name=true)` resolves the schema union across parquet generations — but the union can only contain columns that exist in **some** file. An attribute added to the schema **before any flush under the new generation** is absent from the entire cold set, so a cold-inclusive query that projects or filters it still fails: the explicit projection (`BuildSchemaProjection`) references the column, and the semijoin's `LOGICAL_WHERE_CLAUSE` may too.

Current behavior (acceptable, but worth fixing): a loud, classified (`ErrFederatedReadFailed`), degradable failure. Hot-inclusive queries under `AllowPartialDegradedMode` still serve the attribute from Postgres.

Fix sketch and its hazard:
- Projection half: feed the cold union column set (the #189 validator already probes every file's footer and could return it) into `BuildSchemaProjection`, emitting `NULL::<type> AS attr` for current-schema attributes absent from the entire cold set.
- **Plan-cache poisoning hazard (must address):** the rendered SQL would then depend on the cold column set, but the plan-cache shape hash does not include it — a plan cached while the column was cold-absent would keep projecting NULL after the first flush adds the column, making new data invisible until cache invalidation. The shape hash (or the projection cache key) must incorporate the cold union set, or the NULL-projection must be re-validated per execution.
- Semijoin half: filters on the never-flushed attribute inside the pushdown subquery need NULL-normalization or pre-render pruning.

iceboundrock:
Closed by PR #346 (squash-merged as b236227). Wrap-up:

**Delivered** — never-flushed attributes no longer fail cold-inclusive reads: both template scan sites (s3_source + pushdown semijoin) consume a per-query scan source that projects each cold-absent column as a typed NULL (`(SELECT *, NULL::<type> AS <col> … FROM read_parquet(…)) AS cold_scan`), with exact SQL NULL semantics for filters/sorts/keysets. Empty missing set renders byte-identical SQL to pre-#255.

**Design deviation from the issue sketch, deliberate** — the projection-mutation approach (feeding the cold union into `BuildSchemaProjection`) was replaced by the scan-source wrap: it fixes the projection half *and* the semijoin half in one mechanism, needs no predicate rewriting, and leaves `SchemaProjection`/`ProjectionCache` (#142) untouched.

[...]

Issue #250, opened 2026-07-14
Wire manifest-driven parquet reads into the server (public config surface for ParquetSource)


iceboundrock:
Follow-up from #187 (PR #249).

#187 landed manifest-driven parquet reads at the engine seam: `federated.WithParquetSource` resolves the manifest-listed object set (explicit `read_parquet` list instead of a glob) and classifies listed-but-absent objects as `ErrParquetSetInconsistent`. The production e2e harness wires it from its CDC config.

**The server does not.** `factory.NewEntityManager*` cannot construct a `manifest.QuerySource` because `forma.Config` has no manifest surface at all:

- no manifest location (prefix / path template — CDC-side these live in `cdc.CDCConfig`, a tool config the server never sees)
- no S3 bucket (the parquet path template arrives **per-request** via `QueryRequest.Federated.S3ParquetPathTemplate`, types.go:157)
- no S3 client (DuckDB httpfs credentials exist in `DuckDBConfig`, but the manifest GET and the `MissingKeys` existence probe need a Go-side client)

So in server mode, reads still resolve paths from the per-request glob hint and the silent-loss fix + inconsistency classification are inert.

## Ask

Design the public config surface and wire the factory:

1. Where does manifest config live? Options: extend `forma.DuckDBConfig` (it already carries S3 endpoint/creds/region for httpfs), a new `forma.LakeConfig`, or reuse-by-reference of the CDC manifest settings.
2. Factory constructs the S3 client + `manifest.QuerySource` (with the legacy-glob fallback for never-flushed schemas) and passes `WithParquetSource`.
3. Decide the interaction with the per-request `S3ParquetPathTemplate`: #187 pinned that an explicit hint wins (#184 contract) — confirm that remains the public semantics once the server has a manifest default.
4. Config absent ⇒ nil source ⇒ today's behavior (backward compatible).

Until this lands, the #187 fix is production-real only for embedders that construct the engine directly with `WithParquetSource`.

iceboundrock:
## Closeout

Delivered by PR #303, squash-merged as `3ec288d` (2026-07-25).

### What landed

- **`DuckDBConfig` public read surface** (`config_duckdb.go`): `manifestTemplate` is the single enable gate; `s3Bucket` / `s3DataPrefix` / `manifestTemplate` validated before any connection is opened.
[...]

PR #305, opened 2026-07-25
fix(federated,config): #299 zero parquet paths fail fast non-degradably; #300 probe-execute the manifest template

### LINKED ISSUE #299: Zero resolved parquet paths render read_parquet(<no value>) — guard the s3_source CTE or fail fast
Found empirically while building #250's factory-path e2e (and independently by static review).

When a federated query reaches the DuckDB engine with **zero resolved parquet paths** (no per-request hint, manifest reads off — or on, with an empty manifest and no \`s3DataPrefix\` fallback), the SQL template renders \`read_parquet(<no value>, union_by_name=true)\`:

- \`internal/sqlgen/advanced_query_template_duckdb.go:57,67\` — \`read_parquet({{.S3_PATHS}}, …)\` renders **unconditionally** (only \`pg_source\` has a \`{{if .HasHot}}\` guard)
- \`internal/sqlgen/duckdb_template_renderer.go:242-245\` — \`S3_PATHS\` is bound only when \`len(paths) > 0\`
- the template renderer sets no \`missingkey\` option, so Go renders \`<no value>\`

Result: a DuckDB parser error classified as degradable \`ErrFederatedReadFailed\` — indistinguishable from a transient S3 outage to every programmatic discriminator. With \`allow_partial_degraded_mode\` the query falls back to Postgres-only; without it the query fails outright.

Candidate fixes (pick one):
1. Guard the \`s3_source\` CTE (and its semijoin) on a \`HasS3Paths\` template flag — zero paths ⇒ hot/PG-only query, honestly.
2. Fail fast in the engine when the resolved path set is empty and warm/cold tiers were requested — a distinct, non-mysterious error.

Also worth doing alongside: strengthen \`TestFactoryWiring_ManifestDrivenReads\` Phase 4 (currently a near-vacuous negative control because the manifest-off hint-less query dies at parse time — issue-hinted query would prove object health with real discriminating power). Characterization of current behavior lives in \`factory_manifest_e2e_test.go\` Phase 4 comments and \`docs/federated-query/design.md\` §4.3.1 level 4 — both must be updated when this lands.

Origin: #250 final review (T6 follow-up).

### LINKED ISSUE #300: Manifest template config validation: Execute-probe to catch field typos (and whitespace edge)
Follow-up from #250 reviews (Task 1 minors #1/#5, confirmed by final review).

\`DuckDBConfig.ValidateManifestRead\` only \`text/template\`-parses \`manifestTemplate\`. A field-name typo like \`{{.SchemaId}}\` (wrong case) parses fine but renders \`<no value>\` at resolve time — every schema then resolves to the same bogus manifest key, every manifest loads empty, and reads silently degrade to the fallback glob. That is exactly the silent-loss shape #250 exists to kill, one config typo away.

Proposed hardening in \`ValidateManifestRead\` (config_duckdb.go):
- Execute the parsed template against a probe \`{SchemaID: <n>}\` and reject output containing \`<no value>\` (or set \`missingkey=error\`).
- While there: close the degenerate whitespace edge — a whitespace-only \`manifestTemplate\` currently escapes both the enable gate (trimmed) and the inert-config check (untrimmed \`!= ""\`), yielding silence instead of an error. Reject \`ManifestTemplate != "" && TrimSpace(ManifestTemplate) == ""\` as \`duckdb.manifestTemplate\`.

Keep parity in mind: \`manifest.PathResolver.Resolve\` uses a plain \`Parse\`+\`Execute\` with no Funcs, so anything the probe accepts must render there identically.

Origin: #250 (PR pending) review ledger T1-1/T1-5.

iceboundrock:
Closes #299
Closes #300

Epic #304 Phase 1, first item. Both issues share one contract: **a manifest misconfiguration is never silent, and never masquerades as a transient outage.**

## #299 — zero resolved parquet paths

`S3_PATHS` is bound only when `len(paths) > 0` (`duckdb_template_renderer.go:242`), but the template renders `read_parquet({{.S3_PATHS}}, union_by_name=true)` unconditionally — only `pg_source` has a `{{if .HasHot}}` guard. An empty path set therefore reached DuckDB as `read_parquet(<no value>)`, which failed as a parser error classified `ErrFederatedReadFailed` — **degradable**. So under `allow_partial_degraded_mode` a misconfigured read surface was answered from Postgres alone: silently short exactly where the cold tier was requested, which is the silent-loss shape #250 exists to eliminate.

Now the engine rejects an empty path set at resolution with `ErrNoParquetPaths` / `NoParquetPathsError{SchemaID, SourceConfigured}`, excluded from `degradableFederatedError` alongside `ErrParquetSetInconsistent`. The message distinguishes the two causes, which have different remedies: no source configured vs. source consulted with an empty manifest and the fallback glob disabled.

### Only candidate 2 is implemented — candidate 1 is unreachable

The issue offers two candidates and asks to pick one. Candidate 1 (a `HasS3Paths` template guard rendering an honest hot-only query) **cannot fire in production**: its guard condition is "warm/cold tiers were requested", and *every* query reaching the DuckDB engine satisfies that — `PreferHot` and `[hot]`-only short-circuit to Postgres at `engine.go:171`, and empty `PreferredTiers` means all three tiers (`FederatedQueryHasHot`). Nor is an empty set a legitimate steady state: a never-flushed schema takes the level-3 fallback glob, which is why the design doc says to leave `s3DataPrefix` set. Reaching zero paths is a misconfiguration.

Skipping candidate 1 also avoided a trap worth recording: pruning `s3_source` removes the **first** of the three `DuckArgs` groups in the advanced template's `[DuckArgs, PgMainArgs, DuckArgs]` interleave, so `BuildDuckDBQuery`, `CompileDuckDBQuery` and `Bind` would all have needed a parallel `HasS3Paths` field — `Bind` already gates `PgMainArgs` on `HasHot` for exactly this reason — or every later placeholder mis-binds.

And a hazard that turned out **not** to apply: `parquetPaths` already participates in the plan-cache **scope** hash (`duckdb_query_build.go:209`), so a zero-path render and a with-paths render can never share a cache entry. No shape-hash change was needed. (This is *not* the #255 plan-cache poisoning hazard.)

### Second layer: the bad render is now unreachable

`sqlgen.requireS3Paths` fails an advanced-template render with `S3_PATHS` unbound, at all three render sites — direct precedent: `requireProjectionParams` from #222. After the engine fix production cannot reach it; this closes the door for every other caller.

## #300 — probe-execute the manifest template

`ValidateManifestRead` only **parsed** the template. `PathResolver.Resolve` executes it against a `map[string]any` keyed `"SchemaID"` (`resolver.go:27`), so a case-typo like `{{.SchemaId}}` parses fine and renders `manifest/<no value>.json`: every schema resolves to the same bogus key, every manifest loads empty, reads degrade silently.

The probe renders against **two** schema IDs and rejects an execute error, empty output, `<no value>`, or **two equal renders**. Two probes because one cannot distinguish a collapsed template from a working one — `{{.SchemaId}}`, a constant path and a printf-style `%d` all render *something*; what condemns them is rendering the *same* something for different schemas.

Probe data shape mirrors `Resolve` exactly (map, not struct — a struct would reject unknown fields by a different mechanism and break parity), and `missingkey=error` makes the validator strictly stricter, never differently strict. New parity tests in `internal/manifest` are the lockstep guard and also demonstrate what `Resolve` silently does with each rejected shape.

The whitespace half of #300 was already closed in #250/PR #303; only the Execute-probe remained.

## Two things for the reviewer to weigh

1. **Behavior change (intended, but visible on upgrade):** a degraded-mode deployment whose hint-less queries were being absorbed into Postgres-only answers now fails loudly. That is the bug #299 reports, but it will surface as a change. Deployments whose queries currently *succeed* are unaffected — the pre-#299 path failed at parse time regardless.
2. **New rejection class:** rejecting a schema-invariant template means a hypothetical single-schema deployment using a constant manifest path would stop booting. Recommended (Forma is multi-schema by design and the collapse is catastrophic and silent), but it is separable from the rest of the probe if you disagree. Every template configured in this repo is the standard `manifest/{{.SchemaID}}.json`, so nothing here is affected.

## Verification

| Check | Result |
| :---- | :---- |
| `make lint` (golangci-lint v1.64.8) | pass |
| `make test` | pass |
| `make test-e2e-production` | pass (131s, full suite) |
| `go test -tags=e2e -short ./internal/e2e_harness/federated/...` | pass (123s + 55s) |
| `TestFactoryWiring_ManifestDrivenReads` | all 4 phases pass |

Red-first throughout: each behavior had a failing test before its fix.

**Mutation-verified:** making `ErrNoParquetPaths` degradable again fails Phase 4 on its degraded-mode assertion, so the new e2e coverage genuinely discriminates.

`factoryPhaseManifestOff` (Phase 4) was a near-vacuous control — it tolerated any error and asserted only the *absence* of `ErrParquetSetInconsistent`, its own comment conceding the query died at parse time. It now asserts the positive classification, non-degradability, and — the part that gives it discriminating power — that the **same** manager reads all seeded rows when given an explicit hint, proving the objects and engine are healthy so Phase 3's error can only come from the manifest wiring.

## Notes / adjacent findings

- **Test churn explained.** Engine and sqlgen tests previously relied on an empty path set being tolerated. `withTestParquetPath()` (federated) also pre-marks the path validated — a non-empty set activates the #189 pre-read footer probe, which issues its own queries through the very executor those tests script and count. `injectTestRenderParams` (sqlgen) is the new canonical renderable-params fixture; `injectTestProjection` stays projection-only so the guard tests can still exercise the missing case.
- **Adjacent hole, not fixed here:** `PG_CONN` and the `postgres_scan` table names render `<no value>` the same way when unbound. Production always binds them from `StorageTables`, so the new assertion is scoped to `read_parquet` rather than the whole statement. Say the word and I'll file it.
- **Near the cap:** `internal/federated/duckdb_query.go` is now 491 lines (was 477). The next change to it will need the split.
- Docs corrected where they documented the removed behavior: `design.md` §4.3.1 level 4 and its writer-mismatch bullet, `error-handling.md` (new `ErrNoParquetPaths` section), plus the `S3DataPrefix` and `ParquetSource.Paths` comments.


iceboundrock:
## Findings

### P1: half-open breaker 会把后续 zero-path 请求重新静默降级

- **Location:** `internal/federated/duckdb_query.go:97-105,183-195`，降级发生于 `internal/federated/engine.go:192-195`。
- **Why it matters:** 当 breaker 已 open 且进入 half-open，`Allow()` 会把本次请求保留为唯一 probe。第一次 zero-path 请求在 190 行返回 `ErrNoParquetPaths`，但没有 `RecordSuccess`/`RecordFailure`，因此 probe 保持占用。紧接着的第二次请求在 97 行得到 `ErrDuckDBUnavailable`；该错误仍可 degradable，于是 `AllowPartialDegradedMode` 返回 Postgres-only 结果。这直接违反本 PR 的核心契约：同一 zero-path 配置错误必须始终 loud、不可降级。
- **Suggested fix:** 让 breaker admission 返回可归属到调用方的 permit/token，并在 pre-execution classified error 上中性释放 half-open probe；后续请求仍应重新解析并得到 `ErrNoParquetPaths`，而不是被改写成 degradable `ErrDuckDBUnavailable`。
- **Test impact:** threshold=1 打开 breaker，等待进入 half-open，source 返回空路径；连续两次 degraded-mode 查询都必须 `errors.Is(..., ErrNoParquetPaths)`，且 Postgres fallback 调用数保持 0。

### P1: 第二个 template probe 未检查空结果或 `<no value>`

- **Location:** `config_duckdb.go:203-223`
- **Why it matters:** 只有 `first` 经过 empty/`<no value>` 检查；`second` 仅检查执行错误和是否等于 `first`。例如 `{{if eq .SchemaID 1}}manifest/1.json{{end}}` 对 schema 1 非空、schema 2 为空，二者不相等，因此校验通过。`{{if eq .SchemaID 1}}{{.SchemaID}}{{else}}{{index . "SchemaId"}}{{end}}` 还能通过 `index` 绕过 `missingkey=error` 并在第二次渲染 `<no value>`。这会让部分 schema 在运行期落回 glob/zero-path，重新引入 #300 要消除的静默配置错误。
- **Suggested fix:** 对每个 probe 输出应用同一个 non-empty/no-value validator。
- **Test impact:** 将上述两个 template 加入拒绝用例，并断言 `ConfigError.Field == "duckdb.manifestTemplate"`。

### P1: 两点采样无法兑现 per-schema 唯一性，可能读取并重标记另一 schema 的行

- **Location:** `config_duckdb.go:171-182,190,218-220`
- **Why it matters:** 本 PR 选择了“manifest path 必须随 schema 变化”的更强契约，但只比较 ID 1/2。`{{if eq .SchemaID 1}}m/1.json{{else}}m/shared.json{{end}}` 会通过，schema 2/3 却共用一个 manifest。`internal/manifest/query_source.go` 不校验载入的 `Manifest.SchemaID`；S3 CTE 也不按 schema 过滤，且 schema projection 会把扫描出的行标记为请求 schema。因此碰撞可能把另一 schema 的 parquet 行作为当前 schema 返回，而不只是少读数据。
- **Suggested fix:** 不要用两个样本声称唯一性。可限制允许的 template AST；或至少在 `QuerySource.load` 强校验 `m.SchemaID == requested schemaID`，使任何碰撞在扫描前 loud fail。若保留任意 Go template，则需要覆盖完整支持域或另一个可证明的唯一性约束。
- **Test impact:** 增加 schema 2/3 碰撞模板；再构造请求 schema 与 manifest `SchemaID` 不一致的查询，断言扫描器从未被调用。

### P2: 新分类无法由仓库外公共 API 调用方做 `errors.Is/As`

- **Location:** `internal/federated/errors.go:43-77`
- **Why it matters:** #299 和 PR body 都把“distinct programmatic discriminator”作为修复目标，但 `factory.NewEntityManagerWithConfig*` 返回公共 `forma.EntityManager`，仓库外调用方不能 import Go 的 `internal/federated`，所以无法引用 `ErrNoParquetPaths` 或 `NoParquetPathsError`。从公共 API 看仍只能比较错误文本，与根包 `errors.go` 规定的 `errors.Is/As` 契约冲突。
- **Suggested fix:** 将 sentinel 和 detail type 放到根包 `forma`，internal/federated 使用 alias 或直接返回根包类型。
- **Test impact:** 从外部测试包经 factory/public `EntityManager.Query` 触发该错误，断言 `errors.Is(err, forma.ErrNoParquetPaths)` 和 `errors.As` 到公共 detail type。

## Spec

- **#299 implemented:** empty path set 在 render/validator/DuckDB 前 fail-fast；`ErrNoParquetPaths` 从普通 degraded predicate 排除；advanced template 的三个 render site 都增加了 `requireS3Paths`；Phase 4 已具备 positive hint control；文档已更新。
- **#299 partial:** half-open breaker 路径会把同一配置错误转换为下一请求的 degradable `ErrDuckDBUnavailable`；公共调用方也无法使用承诺的 programmatic discriminator。
- **#300 partial:** parse+execute parity、`missingkey=error` 和两次 probe 已实现，但第二次输出未做完整检查；更强的 per-schema 唯一性策略只验证两个样本，无法兑现代码注释和文档宣称的保证。
- **Adjudication accepted:** updated epic #304 已记录 `HasS3Paths` guard 在当前路由下不可达，本 review 不把未实现 candidate 1 视为缺项。
- **Compatibility decision:** 接受拒绝 constant template 的升级行为，因为 Forma 是 multi-schema 且共享 manifest 风险高；但既然承担该兼容性成本，唯一性 enforcement 必须可靠，而不能只覆盖 ID 1/2。

## Standards

- **P2 hard violation:** `config_duckdb_test.go:64-261` 的 `TestValidateDuckDBConfig_ManifestRead` 为 198 行，超过 `coding-standard.md:8-11` 的 100 行硬上限；PR 在原有超限函数中又增加 49 行。请把 Execute-probe cases 拆成独立 table-driven test。
- **P2 hard violation:** `internal/sqlgen/duckdb_query_edge_cases_test.go` 为 523 行。该文件在 base 已超限，但本 PR 修改其 490/517 行附近；`coding-standard.md:93-98` 要求触及不合规区域时做增量修复。可把受影响的 non-keyset ordering tests 移到专用文件，使原文件降到 500 行以内。
- 其余 changed Go files 未超过硬限制；但 `internal/federated/duckdb_query.go` 491 行、`engine.go` 490 行已接近上限，下一次修改必须先拆分。
- `gofmt`、lint、`git diff --check` 均未发现问题。

## Assessment

**Request changes before merge.** 主路径方向正确，但 3 个 P1 会破坏“manifest misconfiguration is never silent”的核心目标，其中 breaker 路径可在实际运行状态下重新产生 Postgres-only 静默结果，模板碰撞还存在跨 schema 数据混淆风险。

## Observations

- PR 对 candidate 1 不可达性的调用链分析成立：hot-only/PreferHot 在 `engine.Query` 提前走 Postgres，进入 DuckDB 的请求均需要 warm/cold。
- `parquetPaths` 已进入 plan-cache scope hash，本 diff 不存在 #255 类 plan poisoning。
- HTTP 500 继续暴露详细 remediation 文本，但这是现有错误响应机制，且相邻 #301 已排期；本 review 不把它重复列为 blocker。
- probe abandonment 是 breaker 的既有通用行为；之所以在此成为 finding，是本 PR 新增并承诺了一个“任何 degraded mode 下都必须保持 loud”的 pre-execution error，而当前状态转换没有维护该承诺。
- 当前测试没有覆盖 half-open + zero-path 的连续请求、第二 probe 为空/`index` no-value、schema 2/3 collision、仓库外公共错误匹配。

## Verification

Reviewed head `557fd3f43e572fa930a2f84e69c03977c8f7bb8d` against base `3ec288d0fc85efc32e2b811eadfa346e92d0dcb0`.

- `git diff --check 3ec288d...HEAD`：pass
- `make test`：pass
- `make lint`：pass，golangci-lint v1.64.8
- `make test-e2e-production`：pass，130.366s
- `go test -tags=e2e -short ./internal/e2e_harness/federated/... -timeout=30m`：pass，118.419s + 51.898s
- GitHub checks：Lint / Test / E2E Smoke Tests / k6 Smoke / Benchmark Smoke 全部 pass

## Summary

共 4 个行为 finding（P1 x3、P2 x1）和 2 个 standards hard violations（P2 x2）；最高风险是 half-open breaker 将第一次 loud 的 `ErrNoParquetPaths` 转换为后续可降级的 `ErrDuckDBUnavailable`，重新产生本 PR 明确要消除的静默 Postgres-only 结果。未合并 PR。

iceboundrock:
## Review round-1 response — all 6 items addressed in `b53b1c4`

All four behavioral findings verified as real before implementing. Two of them I could reproduce as a failing test that returned **`nil`** — a silently successful degraded answer — which is the exact failure mode the PR claims to eliminate. Details per item.

### P1 — half-open breaker re-silences zero-path errors → **fixed**

Confirmed exactly as described. `Allow()` reserves the probe at `duckdb_query.go:97`, well before `resolveScanSources`, and the existing code comment already documented the abandonment ("early-error returns between here and duck.Query abandon the probe"). The red test made the consequence concrete: request 2 returned `nil` error with a Postgres-only page.

Fixed **generally** rather than for this error alone, because the mechanism is not specific to it: `CircuitBreaker.ReleaseProbe()` returns the slot without recording evidence, and `StreamDuckDBFederatedQuery` defers it until `duck.Query` is reached. So *every* pre-execution early return now hands the probe back — including the two pre-existing cases with the same shape (unrenderable hint template → `ErrInvalidInput`, missing schema metadata cache), which were equally non-degradable but had no explicit promise attached. Neutral by design: the breaker stays open, only the reservation returns; recording success would close it on no evidence, recording failure would extend an outage for a dependency never consulted.

Test is exactly your specified shape, with `openDuration` set to **one hour** so it cannot pass by waiting out the reservation: `TestHalfOpenZeroPathStaysLoudForConsecutiveRequests`. Both consecutive degraded-mode requests report `ErrNoParquetPaths`, neither reports `ErrDuckDBUnavailable`, and `pg.queryCalls == 0`. All existing #246 probe tests still pass.

### P1 — second probe output unchecked → **fixed**

Correct by inspection, and both of your templates passed before the fix. All probe outputs now run through one validator (non-empty, no `<no value>`), with the failing schema ID named in the message. Both templates added as reject cases.

### P1 — two-point sampling cannot prove uniqueness → **fixed, taking your fallback option**

The cross-schema consequence checks out: `QuerySource.load` did not verify the loaded manifest's `schema_id`, the `s3_source` CTE does not filter rows by schema, and the projection stamps scanned rows as the *requested* schema — so a collision serves another schema's rows under this identity rather than merely under-reading.

Implemented the enforcement you named: `load` now rejects a manifest whose stamp disagrees with the schema being read, before any path reaches a scan. I did **not** restrict the template AST — the runtime invariant is provable where the config check can only sample, so it seemed the better place to spend the constraint.

New `forma.ErrManifestSchemaMismatch` / `ManifestSchemaMismatchError{RequestedSchemaID, ManifestSchemaID, Path}`, non-degradable. It also required fixing an adjacent relabelling bug: `resolveParquetPaths` wrapped *all* source errors in `ErrFederatedReadFailed`, which would have handed this straight to the degraded fallback.

One deliberate carve-out: `schema_id == 0` is treated as **unstamped**, not as schema 0. Schema IDs are always positive (`ValidateFixtureSchemaID` rejects `<= 0`; the file registry starts at 100), and rejecting zero would break reads for any deployment still holding manifests written before the field existed. I verified this is safe to rely on going forward: `LoadOrCreate` is the *only* `Manifest` construction site in the repo and every `Save` path loads first, so all four writers stamp it.

Also softened the config-check comment: it samples for collapse, it does not prove injectivity.

### P2 — classification unreachable from the public API → **fixed**

Correct, and it undercut the PR's stated purpose. `ErrNoParquetPaths` / `NoParquetPathsError` and the new mismatch pair now live in the root `forma` package; `internal/federated` re-exports them as **type aliases**, so `errors.Is`/`errors.As` behave identically through either name and no call site churned. Phase 4 now asserts through `forma.ErrNoParquetPaths` and `errors.As` to `*forma.NoParquetPathsError`, checking `SchemaID` and `SourceConfigured`.

Noted but not changed: `ErrParquetSetInconsistent` has the same limitation. It predates this PR and moving it is a public-surface change for an error the PR doesn't introduce — happy to file it as a follow-up if you want the family made consistent.

### Standards — both hard violations fixed, plus one I introduced

- `TestValidateDuckDBConfig_ManifestRead` 198 → **83** lines, split into per-concern tests (`_ManifestTemplateProbe`, `_ManifestReadWhitespace`) over a shared `runManifestReadCases` helper that keeps the `ConfigError.Field` attribution.
- `duckdb_query_edge_cases_test.go` 523 → **465**, non-keyset sort tests moved to `duckdb_nonkeyset_sort_test.go`.
- **Self-reported:** extracting `executeAndStreamDuckDB` (needed to keep `StreamDuckDBFederatedQuery` under the 100-line function cap after the probe-release defer) pushed `duckdb_query.go` to 535 lines — past the file cap you warned was one change away. Split the execute/stream half into `duckdb_query_execute.go` (421 + 132), following the `duckdb_query_build.go` seam from #220.
- `gofmt -w` on the two touched files carrying pre-existing misalignment. The other four `gofmt -l` hits in `internal/sqlgen` are pre-existing on `main` and outside this diff — left alone rather than adding unrelated churn.

Every file this branch touches now satisfies both caps; the only remaining >500-line files repo-wide are the three pre-existing ones tracked in #220.

### Adjudications accepted, no action

- **HTTP 500 exposing remediation text** — agreed it belongs to #301, and #301 is next in epic #304 Phase 1.
- **Probe abandonment as pre-existing generic behavior** — agreed that framing is what makes it a finding here, and the fix is written generally for that reason.

### Verification after the fixes

| Check | Result |
| :---- | :---- |
| `make lint` (v1.64.8) | pass |
| `make test` | pass |
| `make test-e2e-production` | pass, 121.6s |
| `go test -tags=e2e -short ./internal/e2e_harness/federated/...` | pass, 114.3s + 51.0s |
| Caps + gofmt across all branch files | clean |

Red-first on every behavioral fix; the breaker and mismatch tests both started from an observed `nil` error.

Issue #315, opened 2026-07-26
generate-attributes silently collapses object $ref and shifts every attributeID


iceboundrock:
Found while implementing #314.

## The bug

`cmd/tools/generate_attributes.go` never resolves `$ref`. `traverseSchemaWithRequiredState` (`:272`) checks for `schema["properties"]`; a `$ref`-only node has none, so it falls through to `getSchemaType` and is emitted as a single scalar attribute.

For a `$ref` to a **scalar** this is accidentally correct — the fallthrough defaults to text, which is what the target declares. For a `$ref` to an **object** it silently collapses the entire subtree.

## Why it matters: attributeIDs are physical EAV keys

Measured by running the generator against `lead.json` before and after converting `properties.contact` from an inline object into `{"$ref": "#/$defs/contact"}`:

| | keys | `contact.*` | `id` attributeID |
| --- | --- | --- | --- |
| committed `lead_attributes.json` | 82 | 24 | 26 |
| regenerated, `contact` inline | 82 | 24 | 26 |
| regenerated, `contact` as `$ref` | **59** | **0** | **7** |

The 24 `contact.*` attributes collapse into one scalar `contact`, and **every subsequent attributeID shifts** — `id` moves from 26 to 7. `attributeID` is the `attr_id` column of the EAV table, so regenerating an attributes file after such a schema edit silently remaps every attribute of every stored row. Nothing gates it.

Reproduce:

```bash
go run ./cmd/tools generate-attributes -schema-file cmd/server/schemas/lead.json -out /tmp/x.json
diff /tmp/x.json cmd/server/schemas/lead_attributes.json   # clean today
# now make properties.contact a $ref to an equivalent $defs entry, re-run: 59 keys
```

## Current exposure

**None today** — no shipped schema uses an object-valued `$ref` in `properties`. #314 nearly introduced one: its plan repaired a dangling `visit.json` reference by hoisting `lead.json`'s `contact` into `$defs`. That was rejected in favour of repointing the reference (`lead.json#/properties/contact`) specifically to keep `properties.contact` inline, so #314 ships with the generator untouched.

So this is a **trap for the next schema author**, not a live defect. Anyone who reasonably refactors a shared object into `$defs` gets a silent database remap.

## Direction

[...]

Issue #296, opened 2026-07-20
Unknown sort attribute surfaces as HTTP 500 instead of 400 (both SortBy and Sort surfaces)


iceboundrock:
Spun out of #240's final review (PR pending).

`buildAttributeOrders` (now in `internal/entity_query_sort.go`) returns a plain error for an unknown sort attribute:

```
cannot sort by unknown attribute '<attr>' in schema '<schema>'
```

It does not wrap `forma.ErrInvalidInput`, so `classifyManagerError` maps it to HTTP 500. A caller typo in `sort_by` — or, since #240, in the structured `sort` field — is user input validation and should surface as 4xx per `docs/error-handling.md` (write-path validation wraps `ErrInvalidInput`).

#240 deliberately kept the new `Sort` path byte-compatible with the legacy message/classification so the two surfaces don't fork their contract (`TestBuildAttributeOrdersUnknownSortAttribute` pins the exact message). Fix should reclassify both surfaces in one coordinated change:

- wrap the unknown-attribute error with `forma.ErrInvalidInput` (keep the message text; keep naming attr + schema)
- update the pin test to assert `errors.Is(err, forma.ErrInvalidInput)` alongside the message
- audit for any operator dashboard/alerting that keys on the current 500

Related pre-existing gap noted in #240's PR body: legacy JSON-body `sort_order` values other than lowercase `asc`/`desc` are silently treated as non-desc (no validation) — worth deciding in the same pass whether to normalize or reject.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

iceboundrock:
PR #297 has merged. Its public structured sort surface now reaches this same unknown-attribute path; the 400-classification correction remains deliberately out of scope and should be handled here for both SortBy and Sort together.

iceboundrock:
## 本 issue 正文的前提已过时，且分类半边已在 #307 顺手完成

**前提更正。** 正文称 `cannot sort by unknown attribute '<attr>'` "maps it to HTTP 500"。实测并非如此：`classifyManagerError` 的子串启发式列表在本 issue 创建前就已包含 `"cannot sort"` 与 `"unknown attribute"`（merge-base `7efb34f` 的 `server.go:738-739`），所以它**一直返回 400**，只是靠错误文本匹配而非 sentinel。接手前请勿按"修 500→400"去改，那个分类已经是对的。

[...]

PR #316, opened 2026-07-26
feat(schemavalidate,entity): #314 enforce JSON Schema constraints on the write path

### LINKED ISSUE #314: JSON Schema constraints (enum, pattern, format, min/max) are never enforced on write
Found while investigating #312.

## The gap

Forma's headline is "entities defined by JSON Schema, no DDL migrations". But **the write path never validates a request body against the JSON Schema document.**

`transformer.ValidateAgainstSchema` (`internal/transform/transformer.go:201`) exists and works. It has **zero non-test callers** — the only references are `internal/transform/transformer_test.go:262,265`. It is not on any interface, so nothing can be wired to it by substitution either.

What the write path *does* validate, via the metadata cache rather than the schema document:
- attribute exists in the cache → otherwise `ErrInvalidInput` (`transformer.go:327-330`)
- required-ness → `validateRequiredAttributesFromInput` (`:360`)
- type coercion → `populateTypedValue`

What it does **not** validate: everything expressible only in JSON Schema.

## This is not theoretical — every shipped schema relies on it

`enum`, `pattern`, `format`, `minimum`, `maximum` appear in **all** of `cmd/server/schemas/*.json` and `cmd/sample/schemas/*.json`.

Concretely, from `cmd/server/schemas/lead.json`:

| property | declared constraint |
| --- | --- |
| `pipeline` | `enum: [buy, rent, sell, landlord]` |
| `stage` | `enum: [new, contacted, need_defined, viewing, offer, contract, closed]` |
| `status` | `enum: [open, won, lost, junk]` (and `required_always`) |
| `temperature` | `enum: [Hot, Warm, New, Qualified]` |
| `rating` | `pattern: ^[A-Z]$` |

`status` is `attributeID 71`, `valueType: text` (`cmd/server/schemas/lead_attributes.json:312-315`). So `POST /api/v1/lead` with `"status": "banana"` is accepted and persisted: it is a known attribute, it is present so `required_always` is satisfied, and it coerces to text. The enum is decorative.

Same for `rating: "zz"` against `^[A-Z]$`, and for any `minimum`/`maximum` on a numeric.

## Why it matters beyond validation

This is the root cause of the bug fixed in #312. A literal dotted key like `"contact.email"` is accepted because the cache contains `contact.email` as a flattened name — even though it is **not a legal property** of the schema, which declares a nested `contact` object. Schema validation would have rejected it at the boundary instead of letting it collide with its own nested form at the database primary key.

Expect other latent issues of the same shape: the cache's flattened namespace is strictly more permissive than the schema it was generated from.

## Decide first, then implement

This may be deliberate — the schema may be intended purely as the source for generating attribute metadata, with runtime validation left to callers. If so, the fix is **documentation**: say plainly that declared constraints are not enforced on write, so nobody ships a schema believing `enum` is a guarantee.

If it is not deliberate, wiring it needs care:

- **Cost.** Validating every write against a schema document is not free; it needs the compiled schema cached alongside the metadata cache, not recompiled per request.
- **Error class.** Failures are caller input → must wrap `forma.ErrInvalidInput` so they surface as 4xx (`AGENTS.md`, `docs/error-handling.md`). After #307 an unwrapped validator would answer 500 with a redacted body.
- **Update semantics.** Merge-update means the *merged* document is what should validate, not the request fragment — otherwise a partial update fails `required`.
- **Existing data.** Rows written before enforcement may violate the constraints now being enforced; an update that merely touches one field would start failing on unrelated pre-existing violations.
- **Blast radius.** Every e2e fixture and k6 payload becomes subject to constraints they may currently violate.

Given the last two points, a staged rollout (warn-then-enforce, or enforce-on-create-only) is probably the realistic path.

iceboundrock:
Closes #314.

## The problem

Forma's headline is "entities defined by JSON Schema, no DDL migrations". The write path never validated a request body against that schema.

`transformer.ValidateAgainstSchema` existed, worked, and had **zero non-test callers**. Validation was only ever: does the attribute exist in the metadata cache, does `required_policy` allow it, does the value coerce. Everything expressible only in JSON Schema was decorative — and `enum`, `pattern`, `format`, `minimum`, `maximum` appear in **every** shipped schema.

`cmd/server/schemas/lead.json` declares `status` as `enum: [open, won, lost, junk]`. `POST /api/v1/lead` with `"status": "banana"` was a known attribute, satisfied `required_always`, coerced to text, and persisted.

## What now happens

**Enforced on create** — `enum`, `pattern`, `type`, `minimum`/`maximum`, and the schema's `required`, *in addition to* the metadata's `required_policy`, which remains a separate mechanism. Violations wrap `forma.ErrInvalidInput` → 400 naming the constraint.

**Reported on update** — the *merged* document is validated and violations are logged with schema name and row id, but accepted. `Entity.ValidateUpdatesStrict` / `VALIDATE_UPDATES_STRICT` flips this to enforcing.

Report-only is the default deliberately: rows written before enforcement may already violate their schema, and rejecting on update would make them un-updatable over an unrelated field — the failure shape this cycle already rejected in #294.

**Startup fails closed.** Every registered schema resolves once at construction; any that cannot aborts the boot naming the schema. Resolving costs ~1.58 ms against ~5.5 µs to validate, so the resolved form is cached — uncached, a 500-record batch create would spend ~0.7 s re-parsing schemas.

## ⚠️ Upgrade-time failure mode

A `schema_registry` row that has `<name>_attributes.json` on disk but **no `<name>.json`** — a configuration `internal/schemameta/file_registry.go` deliberately tolerates today — will now **refuse server startup**, because construction resolves every name `ListSchemas()` returns.

**Operator preflight: every registered schema name needs a resolvable `<name>.json` in `SCHEMA_DIR`.** No database column is involved on this path.

## Other behaviour changes

- `watch` payloads without `id`, `name`, or `brand` are now rejected — `watch.json` requires them while `watch_attributes.json` requires nothing. The bundled CSV importer already supplies all three.
- A dangling `$ref` is fixed: `visit.json` referenced `lead.json#/$defs/contact`, which never existed. `visit` and `log` could not be resolved at all — invisible because nothing resolved them. Repaired by repointing one line at `lead.json#/properties/contact`.

**That repair is deliberately not the obvious one.** Hoisting `contact` into `$defs` was rejected: `cmd/tools/generate_attributes.go` never follows `$ref`, so regenerating `lead_attributes.json` afterwards collapses 24 `contact.*` attributes into one and shifts every subsequent `attributeID` — measured, 82 keys → 59, `id` from 26 → 7. Those are EAV lookup keys. Tracked as #315.

## What is *not* enforced, and why

- **`format` is annotation-only** in `github.com/google/jsonschema-go`, with no assertion option. Every `date-time`, `uuid`, and `email` format is inert. `contact.json`'s email is a real `pattern` and *is* enforced. Documented so nobody ships a schema trusting a `format`.
- **Unknown properties are accepted** — no shipped schema sets `additionalProperties`.

## Two structural validation gaps, documented

Both persist values exactly as before; they are simply not validated.

1. **A dotted literal naming an attribute under a schema array, written at a level *above* that array.** JSON cannot express "a member of this array" as one dotted key, so nesting it would place an object where the schema declares an array. A dotted key written *inside* the element **is** validated — the discriminator is where the caller writes the key, not which attribute it names. Exactly 15 shipped attributes: `requirement.areas.{city,note,prefecture,ward}` and `propertyInterests.{propertyId,status,isPrimary,notes,firstSeenAt,lastActivityAt,snapshot.{address,code,price,rent,title}}`.
2. **Anything under a relation root.** `StripComputedFields` removes relation roots by exact key, so a dotted descendant survives; expanding it would rebuild the stripped root and produce a 400 the caller cannot fix by sending the field. Only `visit.contactSnapshot` carries `x-relation` today.

**Latent, not currently reachable:** an array behind a `$ref` or in `$defs`, or under `if`/`then`/`prefixItems`/`patternProperties`/`dependentSchemas`/`contains`, is invisible to the array-path derivation and yields a false 400. No shipped schema uses any of these.

## Design decisions worth not re-deriving

**Normalization feeds the validator only; the writer receives the caller's original map.** Literal dotted keys are expanded so their values are inspected — `{"contact.email": 99999}` would otherwise pass as an unknown property with its value never examined. The first design fed the normalized document to the writer too, and that produced four separate data-loss bugs, because `internal/transform/eav_dedupe.go` resolves precedence by *spelling* and pre-merging spellings destroys that information. The split makes the whole class structurally impossible, and is pinned by pointer-identity guards at all four write sites.

**Array paths are derived from the schema, not the metadata.** No cache-only signal exists: `forma.AttributeMetadata.IsInsideArray()` answers exactly this question, has zero production callers, and no shipped attributes file contains the bracket its convention needs — `generate_attributes.go` computes array-ness at `:300` and discards it.

## Verification

`make lint`, `make test`, `go test -count=1 ./internal/... .`, `go vet -tags=e2e ./internal/e2e_harness/...`, **`make test-e2e-production` (109 PASS / 0 FAIL)**, **`go test ./internal/e2e_harness/federated/... -tags=e2e` (263 PASS / 0 FAIL / 0 SKIP)**, `gofmt -l`, `git diff --check`. No fixture broke under enforcement.

The production harness previously bypassed the factory with a `nil` validator, so "the shipped schemas pass over the real factory path" would have verified nothing. Fixed harness-side only — no production code changed to make a test pass — with a red anchor (`rank: 40000` against `maximum: 32767`) asserting rejection happens before the write.

Every fix round was mutation-verified: the fix was reverted and the intended tests confirmed to fail.

## Known residuals (parked, not blocking)

- The dot boundary in `RelationRoots.Covers` has no test. Behaviour verified correct by construction (`contactSnapshotExtra.x` still expands), but a naive `HasPrefix` mutant passes the suite.
- `docs/schema-consistency-migration.md:205` says "the remaining gap" singular where there are now two.

## Follow-ups to file

- No tripwire test for `format` inertness — a `jsonschema-go` bump that started asserting formats would reject shipped `uuid`/`email` values on the **create** path, which has no escape hatch. 14 documents use `format`.
- Report-only mode has no counter or metric, only per-write `Warn` — and report-only is the shipped default, so every deployment ships in the mode with no aggregate signal.
- `cmd/lambda/main_test.go`'s entry-point guard is hardcoded to two files, so it cannot catch the third it advertises.
- `funlen` is absent from the default golangci set, so the repo's documented 100-line function rule is unenforced everywhere.
- `factory/factory_test.go` is at 985 lines, roughly double the file cap.

## Note before merge

CI's `k6-smoke` job runs `bun run gen-data -- --schema all --count 2000` through the real HTTP write path — the only place shipped-schema creates cross real HTTP, and the one suite not runnable locally here. A static probe of those payloads against a real validator passes for `lead`, `visit`, and `log`.


iceboundrock:
## Findings

Reviewed `f4017db...b8794df` (latest head). Three actionable defects remain.

### High: validation can discard a value that the writer still persists

`internal/transform/normalize_keys.go:249-258` replaces two arrays wholesale in the validator-only view, while `internal/entity_write_validation.go:79-80` validates that view and the write sites still transform the original map (`internal/entity_crud_service.go:108,236`; `internal/entity_batch_service.go:234,324`).

Reproducer against shipped `lead_full`:

```json
{
  "requirement": {"areas": [{"note": 123}]},
  "requirement.areas": [{"city": "Tokyo"}]
}
```

`requirement.areas.note` is declared `string`. The normalized view replaces the first array with the literal array, so validation sees only `city` and succeeds. The writer flattens the original spellings to two different EAV attributes and persists both `note` and `city`; the numeric `note` is coerced to text. A temporary regression test confirmed `ToAttributes` emitted both attributes while `Validator.Validate` returned `nil`.

The validation view must retain every surviving EAV attribute (or reject this ambiguous mixed spelling) before create/strict-update can claim full enforcement.

### High: configuration faults in unregistered `$ref` targets bypass fail-closed startup

`internal/schemavalidate/validator.go:36-43` loads sibling schemas without running `checkSchemaSupported`; `New` checks only registry roots at `:176-185`. A registered root referencing `common.json` containing `{"type":"strig"}` constructs successfully. At runtime `:224-225` blanket-wraps the validator failure as `forma.ErrInvalidInput`; creates return a misleading 400, and default report-only updates absorb the operator fault and write anyway.

A temporary constructor regression test reproduced `New(...) == nil error`. Preflight every loaded reference (with URI caching/cycle handling), not only registered roots.

### Medium: typed-nil maps/slices are rewritten from JSON `null`

`internal/transform/normalize_keys.go:134-153` recurses into a typed nil `map[string]any`/`[]any` and materializes `{}`/`[]`. For a Go embedder sending a required object with no nested required fields, validation therefore passes, while the original map sent to the writer emits no attributes. A temporary regression test reproduced both `Validate == nil` and an empty persisted attribute set.

Preserve typed nil containers as nil before recursion, so the JSON round-trip validates them as `null`.

iceboundrock:
## Spec

**Status: partial compliance.** The implementation covers the main rollout shape: cached resolution, the schema repair, all four write seams, create enforcement, merged-document update validation, report-only/strict update modes, server/Lambda config wiring, and `ErrInvalidInput` separation.

Remaining spec mismatches:

- The verified array-merge and typed-nil findings violate the core requirement at design spec lines 103-106 that create validation close the dotted-key bypass and enforce the schema document.
- Design lines 105 and 118-121 require fail-closed construction. Unchecked faults in unregistered sibling `$ref` targets remain runtime 400s instead of startup failures.
- Design lines 149-153 say `normalize -> validate -> transform` with normalized data flowing onward. The final implementation deliberately validates a copy and writes the original map to preserve spelling provenance. That evolved design is well justified in the PR/plan, but the governing design spec still states the rejected architecture and should be updated.
- Design lines 173-180 require a real lead HTTP enum case, watch behavior coverage, an inert-`format` tripwire, and caching verification. Enum/caching have lower-level substitutes, but there is no `watch` test and no `format` tripwire; the plan self-review at lines 1217-1220 explicitly acknowledges both omissions.
- Design line 106 says dotted normalization “closes the bypass,” while the final contract intentionally exempts dotted attributes written above arrays and beneath relation roots. These are disclosed in the PR and `docs/error-handling.md`, but the design spec should record the accepted deviations.

No unrelated feature scope creep was found; most expansion beyond the original package shape is defensive work required by the discovered schema/data-shape constraints.

iceboundrock:
## Standards

Compared the diff with `AGENTS.md`, `coding-standard.md`, and the write/read error contract in `docs/error-handling.md`.

### Hard non-conformance

- The documented 500-line cap is still exceeded in touched files: `factory/factory_test.go` is 985 lines and grows in this PR; `internal/entity_manager_test.go` is 1,196 lines; `internal/entity_manager_update_batch_test.go` is 508 lines. `AGENTS.md:68-71` and `coding-standard.md:7-11,93-98` require refactoring the immediate touched area rather than extending oversized files.
- Introduced code still propagates errors bare, notably `factory/factory.go:164` and loader/walker paths such as `internal/schemavalidate/validator.go:40,42` and `internal/schemavalidate/schema_check.go:50-53,116-127`. The repository rule is explicit: always wrap with operation context, never bare `return err` (`AGENTS.md:69`, `coding-standard.md:22-33`). Some outer callers add context, but the local functions do not meet the stated rule.
- `internal/schemavalidate/schema_check.go:89-92` reports an invalid schema `type` value but not the property/subschema path containing it. The wrapper names the entity schema, but not the offending attribute. This falls short of `AGENTS.md:80` / `docs/error-handling.md:872-878` requiring the offending column/attribute and expected state.

### Checked and compliant

- No changed/new function exceeds 100 lines.
- New Go source files stay below 500 lines.
- Package naming, exported-item documentation, layer placement, nesting depth, error matching via `errors.Is`, formatting, and lint checks are otherwise consistent.

iceboundrock:
## Assessment

**Recommendation: changes requested; do not merge this head yet.**

The architecture and rollout policy are directionally sound, but the central guarantee is not yet reliable: a create can validate one document and persist a materially different set of attributes. The array-spelling reproducer demonstrates this with a shipped schema, not only a synthetic edge case. The unchecked referenced-schema fault also breaks the advertised fail-closed/error-class boundary.

Merge gate suggested for this PR:

1. Fix and regression-test the validator-view/persisted-EAV mismatch.
2. Preflight every loaded `$ref` document so schema configuration faults cannot become `ErrInvalidInput`.
3. Preserve typed nil JSON semantics for direct Go callers.
4. Reconcile the design spec with the final validator-only normalization architecture and explicitly accepted gaps.
5. Resolve or consciously waive the documented standards violations before approval.

The intentionally documented top-level array and relation-root gaps are not counted again in the three findings above; the findings are additional failure modes.

iceboundrock:
## Observations

- All four production write seams are wired, including the atomic batch path used by HTTP create traffic. The pointer/immutability guards make the final “validator copy, writer original” decision explicit.
- Update validation correctly uses the merged document and report-only mode absorbs only errors carrying `forma.ErrInvalidInput`; plain operator errors continue to fail the write.
- `VALIDATE_UPDATES_STRICT` is applied by both server and Lambda entry points, and report-only warnings include schema name, schema ID, and row ID without logging payload data.
- Validator construction is correctly moved before DuckDB/parquet read-surface allocation, avoiding the resource-leak path on ordinary schema resolution failures.
- The `visit.json` repair to `lead.json#/properties/contact` avoids the attribute-ID churn that a `$defs` hoist would trigger in the current generator.
- The PR transparently documents two accepted validation gaps: dotted keys written above schema arrays and descendants of relation roots. It also documents the required-relation-root hazard. These disclosures are useful, but they reduce the strength of “create always enforces” and should remain visible in release/operator notes.
- Report-only mode still has no aggregate metric, and inert `format` behavior has no tripwire test. Both are reasonable follow-ups if not included here.

iceboundrock:
## Verification

Reviewed latest head `b8794dfb4ec593943eadcfccd000642372d7204f` against base `f4017db8b67cb810498f4e6bba51539a5881f30d` in an isolated clean worktree.

Local commands:

- `make lint` -> pass
- `make test` -> pass
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -count=1 ./internal/schemavalidate ./internal/transform ./internal ./factory` -> pass (uncached)
- `git diff --check f4017db...HEAD` -> clean
- `gofmt -l <all changed Go files>` -> no output

Fresh GitHub checks at review time:

- Lint -> success
- Test -> success
- E2E Smoke Tests -> success
- k6 Smoke -> success
- Benchmark Smoke -> success

Targeted negative verification used temporary, uncommitted regression tests and then removed them:

- mixed nested/literal array spelling: writer produced both EAV attributes, validator returned `nil` -> reproduced
- typed nil required object: validator returned `nil`, writer produced no attributes -> reproduced
- invalid `type` in an unregistered sibling `$ref`: `schemavalidate.New` returned `nil` error -> reproduced

The worktree was clean again after removing those temporary tests. I did not independently rerun the full Docker production/federated E2E suites; the PR description reports them, while the current CI E2E smoke and k6 jobs are green.

iceboundrock:
## Summary

- **Findings:** 3 total: 2 High, 1 Medium.
- **Spec:** partial; core rollout is implemented, but fail-closed coverage and validation/persistence equivalence are incomplete, and the design document has not caught up with the final architecture/accepted gaps.
- **Standards:** function limits and tooling gates pass; touched-file caps, bare error propagation, and one schema-error diagnostic do not meet documented rules.
- **Verification:** local lint/unit/targeted tests and all current CI checks pass, but focused negative tests reproduce all three findings.
- **Worst issue:** the validator can omit a schema-invalid value that the write path still persists.
- **Recommendation:** changes requested; do not merge `b8794df` as-is.

iceboundrock:
## 回应评审：3 项 finding 全部为真、已修（`89ea832`、`25cf5a7`、`9bd5bcd`、`1944e2f`、`3f8c824`）

两项 High 的根因都在我这边——一个是我在 Task 3 下令删除的东西，一个是我把先前评审的 Minor 误判为 latent。

### [High] validator view 丢掉了 writer 仍会持久化的值 — 已修

成立，且这是**我下令删除的直接后果**。Task 3 中我告诉实施者 element-wise array merge "只为保证 writer 正确" 并让其删除——**错了**：它同时维持 validator view 的完整性。我那个错误论据甚至活进了代码注释（"two arrays included — is replaced, because the later spelling replaces the whole attribute the writer stores"），这正是它长期隐形的原因。

已恢复 element-wise merge。在当下是**无歧义安全**的（当时不是）：规范化不再喂 writer，合并只能**增加**被检查的值。

### [High] 被引用 schema 中的配置故障绕过 fail-closed — 已修

成立。**这是我先前误判的 Minor**：早先评审指出 `schema_check.go` 的注释夸大了"只有真正的 violation 能到达 wrap"，我以"所有跨文件目标本身都已注册"为由 park 为 latent。你用未注册目标给出了可达后果——构造干净通过，运行时 blanket wrap 成 `ErrInvalidInput`，create 返回误导性 400，而**默认的 report-only update 会吸收该运维故障并照写**。

现在 loader 载入的**每一个**文档都跑同一个 `checkSchemaSupported`，按解析路径记忆化。实施者的测试还加了 root → middle → leaf 链，超出要求。

### [Medium] typed-nil 容器被物化 — 已修

成立。已在递归前保留 typed nil，让 JSON round-trip 呈现 `null` 由 schema 的 `type` 裁决。

**两处实施者的独立发现值得记录**：(1) 我建议的测试 fixture **不具判别力**——`lead_full` 的 `contact` 自带 `required:["name"]`，故 `{}` 无论有无该 bug 都被拒；改用 `requirement` 才真正判别。(2) 保留 typed nil 使 nil map 成为文档的真实居民，`setNestedValue` 与 `mergeAny` 都会写入它 → `panic: assignment to entry in nil map`，已加守卫并各自钉测试。

---

## 修复本身引入的一个 Important，已裁决为**接受并明示**

element-wise merge 会产生**假 400**：writer 的 dedupe 对落败拼写替换**整个逻辑属性（所有 index）**，而 `mergeSlices` 只按 index 替换。当嵌套拼写更长时，其多余元素进入 view 但从不进入存储：

```json
{"requirement": {"areas": [{"city":"A"},{"city":12345}]},
 "requirement.areas": [{"city":"NEW"}]}
```
writer 只存 `city[0]="NEW"`；view 保留 `city[1]=12345` → 400，对一个永不落库的值。

**两种误差不对称**：replace 是**欠近似**（持久化的值不被校验——正是 #314 要关闭的 bypass，即上面的 High）；merge 是**过近似**（被丢弃的值被校验）。裁决为保留过近似——它是更安全的一侧，且该拒绝还额外要求存量多余数据**本就**违反 schema：默认 report-only 下只是一条 `Warn`，仅在 strict 模式才拒绝，而那恰是分阶段上线要暴露的存量违规。

**未采用精确的 per-leaf-key 规则**：EAV 按 (属性, index) 存而 JSON 按 index 存全部属性，本任务每一次细化该 merge 都引入过新缺陷（先 4 个 parity bug，再是这一个）。

已按裁决处理：改掉两处宣称"这不可能发生"的表述（`normalize_keys.go` 与 spec）、修掉声称匹配 writer last-spelling-wins 的过时包注释、用你的复现体钉住行为并在测试点写明为何接受而非修复、并在 spec 的 accepted-deviations 与 `docs/error-handling.md` 各记一处（后者标注为 **live, not latent**，与其下方的 latent 小节区分）。

---

## Spec / Standards

- **Spec 已对齐**：149-153 行描述的被否决架构改为已发布设计（规范化只喂 validator；writer 收调用方原图，因为 `eav_dedupe.go` 拥有拼写优先级），106 行的"closes the bypass"已用两处豁免限定。被取代的决策保留为 superseded 而非改写。
- **`format` tripwire 已加**。`watch` 专项测试**豁免**（行为变化已文档化、内置 CSV importer 提供全部三个必填字段、`TestShippedSchemasPassConstructionGuards` 覆盖构造），理由已记入 spec。
- **bare `return err`**（`factory.go:164`、`validator.go:40,42`、`schema_check.go:50-53,116-127`）已全部包装；**invalid `type` 错误现在指名属性路径**，7 个表格用例。
- **`factory/factory_test.go` 985→887**，`mockSchemaRegistry` 抽到独立文件。
- **豁免两个超限文件**：`internal/entity_manager_test.go`(1196) 与 `internal/entity_manager_update_batch_test.go`(508)。本 PR 对它们只是给既有调用点追加 `, nil`，属机械参数传播而非实质扩写；拆分会把本 PR 的 diff 埋进无关 churn。实施者另指出 `federated/benchmark/execute.go` 属同类（唯二超 100 行函数均为既有，本 PR 对该文件仅 1 增 1 删），同样豁免。

## 我自己的一处失误，已修

`.superpowers/`（评审草稿：内部修复报告 + 75KB 评审 diff）在本 worktree **未被忽略**，被两个 docs commit 裹带提交并推送。已 `git rm --cached` 移除并本地排除（`36a2358`）。文件不属于交付物；本仓库 squash 合并，故不会进入 main 历史。

## 验证

`make lint`、`make test`、`go test -count=1 ./internal/... .`、`go vet -tags=e2e`、`make test-e2e-production`（125s、0 FAIL）、`git diff --check f4017db...HEAD` 均通过。

三项 finding 全部先红后绿；过近似的钉测试以变异证明非空洞（把 `mergeSlices` 回退为 replace → 该测试转红）。

**另**：你对实施者第三条自述的更正也已核实并写进代码注释——annotations 在 `Resolved.resolvedInfos`（per-`Resolved` 侧表），真正的风险是 `resolve.go:550` 把**引用方**的 `$schema` 写进被载入文档，故共享指针会把一次 resolve 的 draft 带进另一次。结论对、机制原先说错了，现已在 `fileLoader` 与 `checkLoadedSchema` 处写明正确机制。

未自动合并。

iceboundrock:
## `.superpowers/` 已加入 `.gitignore`；follow-up issues 已开

**`.gitignore`**（`f758ce9`）：`.superpowers/` 现由仓库 `.gitignore:78` 忽略，而非仅我本地的 `.git/info/exclude`。同批修掉一处一词不准确——`docs/schema-consistency-migration.md:205` 的 "the remaining gap" 改为复数，现在确实有三处（数组之上的 dotted key、relation root 之下、validator view 的数组过近似）。

| # | 内容 | 标签 |
|---|---|---|
| #317 | report-only 校验无聚合信号，而它是**默认模式** | enhancement |
| #318 | `StripComputedFields` 按精确键剥离 relation root，故 dotted 拼写被持久化而嵌套拼写不被 | bug |
| #319 | 已文档化的 100 行函数上限无强制——默认 golangci 集不含 `funlen` | enhancement |
| #320 | 三个测试文件超 500 行上限（1196 / 887 / 508） | architecture, priority:low |
| #321 | 两个守卫无法因其注释所声称的回归而失败（`Covers` 点边界、lambda entry-point 扫描） | enhancement, priority:low |
| #322 | embedder payload 中的 NaN/Inf 被分类为运维故障（5xx）而非 invalid input | bug, priority:low |

### 两项**未**开（开前逐一核实，避免开出过时 issue）

- **`format` inertness tripwire** —— 本轮修复已加（`internal/schemavalidate/format_inertness_test.go`），无需 issue。
- **`newValidationHarness` 的死参数 `validate bool`** —— 已在 standards 那批修掉。

这两项都出现在我此前的 PR 正文清单里；开 issue 前实测发现已解决。本会话已两次因未核实前提而开出/差点开出过时 issue（#296 正文陈旧、#315 早期措辞），故现在一律先验证。

### 分歧点，供你裁决

#318 我刻意**不**在 #314 里修，因为两种修法都改变**持久化行为**：要么 dotted 拼写停止落库（今天会落库），要么嵌套剥离变得不那么激进。这超出"给写路径加 schema 校验"的范围。issue 里把方向摆成一个选择而非既定结论——若你认为其中一侧显然正确，直接在 issue 上裁决即可。

#317 是我认为唯一可能影响上线节奏的一项：分阶段上线依赖运维方判断"何时可以翻 strict"，而默认模式目前只给出逐次写入的 `Warn`，无计数、无按 schema 分解。翻 strict 前值得先有它。

PR #316 未自动合并，仍待你的复审结论。

iceboundrock:
## Findings

Re-reviewed `f758ce9` (current head) against the three defects flagged in round 1. All are **resolved**.

### [Resolved] High: validation can discard a value that the writer still persists

`internal/transform/normalize_keys.go:292-347` now merges arrays element-wise by index (`mergeSlices`), matching the flattener's EAV primary key alignment. The normalizer's view retains every surviving EAV attribute from both spellings, so the schema validator sees them all. `TestNormalizeKeepsArrayOnPathValidatable` confirms the merging behaviour.

A **known over-approximation** is documented at `:317-331` and in `docs/error-handling.md`: when the nested spelling is longer than the literal, surplus elements persist in the view but are discarded by the writer's dedupe. This can produce a false 400 on strict updates where the extra elements violate the schema. Deliberately accepted — replacing would cause under-approximation instead (the bypass #314 exists to close). `TestNormalizeArrayMergeOverApproximatesShrinkingList` pins the decision.

### [Resolved] High: configuration faults in unregistered `$ref` targets bypass fail-closed startup

`internal/schemavalidate/validator.go:53-68` now runs `checkLoadedSchema` on every document the loader retrieves, memoised by resolved path. A root referencing `common.json` with `{"type":"strig"}` now fails construction. `TestNewRejectsUnknownTypeInReferencedSchema` and `TestCheckLoadedSchemaMemoisesPerPath` cover the guard.

### [Resolved] Medium: typed-nil maps/slices are rewritten from JSON `null`

`internal/transform/normalize_keys.go:149-164` preserves typed nil `map[string]any` and `[]any` without recursing. `setNestedValue` at `:250-261` and `mergeAny` at `:283-291` guard against writing into nil maps. The JSON round-trip in `Validate` presents them as `null`, letting the schema's `type` decide. `TestNormalizePreservesTypedNilContainers` covers both map and slice.

No new findings discovered in this round.

iceboundrock:
## Spec

**Status: fully aligned with the shipped implementation.**

The design spec (`docs/superpowers/specs/2026-07-25-issue-314-enforce-json-schema-on-write-design.md`) has been updated (`8b67988`) to match the final architecture:

- Lines 149-153 now state the validator-only normalization architecture, with the original map preserved for the writer. The rejected "normalize → validate → transform with normalized data flowing onward" design is kept as a superseded block.
- Line 106 ("closes the bypass") is qualified with the two accepted exemptions: dotted keys above schema arrays, and dotted keys beneath relation roots.
- The accepted deviations section records the array merge over-approximation as a live trade-off (not latent) and the relation-root gap.
- The `format` inertness tripwire has been added.
- The `watch` behavior change is documented, with the CSV importer guard as rationale for waiving a dedicated test.

All five round-1 spec mismatches are resolved. The spec and the code now describe the same contract.

iceboundrock:
## Standards

Compared the diff against `AGENTS.md`, `coding-standard.md`, and `docs/error-handling.md`.

### Resolved from round 1

- Bare `return err` at `factory.go:164`, `validator.go:40,42`, `schema_check.go:50-53,116-127` — all now wrapped with operation context.
- Invalid `type` error now names the property path (`schema_check.go:96-110`), pinning the offending attribute as required by `AGENTS.md:80`/`docs/error-handling.md:872-878`. 7 table-driven test cases.
- `factory/factory_test.go` reduced from 985 to 887 lines via extracting `mockSchemaRegistry` to `factory/mock_schema_registry_test.go`.

### Still non-conforming, consciously waived

- `factory/factory_test.go` (887) remains over the 500-line cap but is reduced; `internal/entity_manager_test.go` (1196) and `internal/entity_manager_update_batch_test.go` (508) remain over cap. The PR author's comment #316#issuecomment-5085590553 waives them because the changes in those files are mechanical `, nil` parameter propagation, not substantive additions, and splitting would bury the review diff in unrelated churn.

### Checked and compliant

- No changed/new function exceeds 100 lines. New files ≤500 lines. Package naming, exported-item documentation, layer placement, nesting depth, `errors.Is` matching, formatting, and lint checks pass.

iceboundrock:
## Assessment

**Recommendation: approve.** The three round-1 defects are fixed, the design spec is reconciled, and standards violations are either resolved or consciously waived.

The array merge is now a deliberate over-approximation — it can reject a strict update for a value the writer discards, but never accepts a value the writer persists. That is the safer error direction for enforcement, and the trade-off is documented in code, tests, and operator-facing docs. Default report-only updates are unaffected.

The one operational concern that remains is report-only mode's lack of aggregate signal (no counter/metric, only per-write `Warn`). This is the shipped default and the staged rollout's decision point — the operator needs to know when logs are clean before flipping `VALIDATE_UPDATES_STRICT`. The PR author filed #317 to track this independently. It does not block merge.

iceboundrock:
## Observations

- Element-wise array merge is restored and explicitly scoped to the validator's view, with over-approximation documented in code, tests, and operator-facing docs. The earlier reasoning that it was unnecessary has been corrected in the code comments.
- The preflight guard now applies to every document the loader retrieves, not just registered roots, with per-path memoisation. A referenced file with a typo'd `type` or `$schema` fails construction and names both the file and the offending path.
- Typed nil containers are preserved, and both `setNestedValue` and `mergeAny` guard against nil-map writes. The implementer noted the real test fixture required `requirement` (no nested `required`) rather than `contact` (which has `required: ["name"]`, making `{}` indistinguishable from `nil`).
- The design spec now describes the actual shipped architecture. Superseded decisions are preserved rather than rewritten.
- A `format` inertness tripwire is now in `internal/schemavalidate/format_inertness_test.go`.
- `.superpowers/` scratch files that were accidentally committed in the earlier round have been removed and `.gitignore` updated.
- Follow-up issues #317–#322 were filed for residuals, including the report-only signal gap and oversized test files.
- The `docs/error-handling.md` section on the array merge gap correctly distinguishes it as "live, not latent" from the latent gaps below it.

iceboundrock:
## Verification

Reviewed head `f758ce9aae175aec1819f96769273e74fed283f6` against base `f4017db8b67cb810498f4e6bba51539a5881f30d` in an isolated worktree (`.claude/worktrees/314-review-r2`).

Local commands, all passing:

- `make lint` — pass
- `make test` — pass (all 33 packages)
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -count=1 ./internal/schemavalidate ./internal/transform ./internal ./factory` — pass
- `git diff --check f4017db...HEAD` — clean
- `gofmt -l` on all changed Go files — no output

CI checks at review time: Lint/Test/E2E-Smoke/k6-Smoke/Benchmark-Smoke all **success**.

Round-1 negative tests that reproduced the three findings are now covered by committed regression tests:
- Array sibling persistence: `TestNormalizeKeepsArrayOnPathValidatable`, `TestNormalizeArrayMergeOverApproximatesShrinkingList`
- Referenced schema preflight: `TestNewRejectsUnknownTypeInReferencedSchema`
- Typed nil containers: `TestNormalizePreservesTypedNilContainers`

Worktree is dirty-free after review.

iceboundrock:
## Summary

- **Findings:** Round 1 — 3 total (2 High, 1 Medium). Round 2 — all 3 resolved, 0 new.
- **Spec:** Now fully aligned with the shipped implementation; superseded architecture and accepted deviations are recorded.
- **Standards:** 3 round-1 violations resolved; 3 file-size caps consciously waived for mechanical parameter-propagation changes.
- **Verification:** Local lint/unit/all-targeted tests, CI checks, and static checks all pass. Regression tests pin each fix.
- **Recommendation:** Approve.

PR #325, opened 2026-07-27
fix(federated): #306 scrub PG_CONN credentials from engine error strings

### LINKED ISSUE #306: Scrub PG_CONN credentials from federated engine error strings
Follow-up from #301.

#301 redacted the public HTTP 5xx surface, so API clients no longer see the
Postgres password that DuckDB quotes back on a `postgres_scan` attach failure:

```
IO Error: Unable to connect to Postgres at "host=… user=… password=… dbname=…": …
```

The credential still reaches the error string itself. A Go embedder using
`factory.NewEntityManager*` will capture it in its own logs, and so will any
future transport that does not repeat #301's redaction.

Direction: scrub at the source rather than at each consumer — wrap
`duck.Query` failures through a helper that rewrites `password=…` (and the
whole `PG_CONN` substring) before the text enters an error chain. Candidate
sites: `internal/federated/duckdb_query_execute.go:49,68` and
`internal/federated/engine.go:382`. `DuckDBPostgresConnStringFromPool`
(`internal/federated/engine.go:477`) is where the string is built, so it can
also expose the exact substring to redact.

Verify with the #301 canary shape: assert the *error string returned from the
engine* — not just the HTTP body — carries no `password=`.

iceboundrock:
Closes #306.

Scrubs the Postgres password at the source — the federated engine's two DuckDB
driver-error emergence points (`executeAndStreamDuckDB`, execute + mid-stream
branches) — instead of at each consumer, plus the internal execution plan's
failure Notes and rendered SQL (the same embedder-facing surface reached via
`attachExecutionPlan`). `redact.Error` rewrites only the composed message and
Unwraps to the original chain, so `errors.Is`/`errors.As` classification
(`ErrFederatedReadFailed`, degradation gating, httpapi carriers) is untouched.

Commits:
- `redact.Error`: message-scrubbing wrapper on the shared #290/#301 matcher (no new regex); identity path for clean errors.
- Engine scrub at both emergence points + `TestCredentialScrub_*` canary suite (#301 canary shape: `;` in the value so truncation regressions surface; assignment/value/head/tail asserted; every NotContains paired with a positive precondition).
- Plan rendered-SQL scrub in `recordTranslation`, with a mutation-verified guard that the engine still **executes** the unscrubbed SQL (`duck.lastSQL` keeps the raw DSN) — only the plan copy is redacted.
- Refresh of the two plan-projection SECURITY comments (`types.go`, `entity_query_service.go`): password now redacted at source, but host/user/dbname, Params, and note internals remain — the public-projection restriction stands.

Deliberately not changed:
- `queryPostgresOnly` (the issue's `engine.go:382` candidate): pgx error prose carries no password and the DuckDB DSN never enters that path.
- `parquet_schema_validation.go` probes: their SQL (`glob`/`DESCRIBE read_parquet`) never embeds the DSN, and both callers discard the errors.
- The public HTTP plan projection already omits SQL/Notes (#301); this PR closes the internal/embedder surface the projection cannot reach. httpapi's boundary redaction stays as belt-and-braces.

Audit notes (full disposition tables in the branch workspace reports):
- All 6 `pgConnString`/`PG_CONN` flows in `internal/federated` either never leave the process (plan-cache scope key is a digest) or pass through a scrub.
- `recordDegradedFallbackPlan` — a third output surface — is clean transitively with zero edits, evidence the source placement is right.
- Plan `Params` cannot carry the DSN: it renders as a SQL literal, never a bind arg.
- Scrub-before-`recordQueryFailure` ordering is pinned by `TestCredentialScrub_ExecutionPlanFailureNotes` (a reorder leaks the note and fails it).

Verified with the #301 canary shape at the engine boundary: the error string
returned from `engine.Query` — not just the HTTP body — carries no
`password=` value, no bare value, and no truncation tail.

Gate: `go build ./... && go vet ./... && go test ./...` green (30 pkgs), plus
the docker e2e suite explicitly (`go test ./internal/e2e_harness/production/
-tags=e2e`, 127.8s, real containers — note plain `go test ./...` does NOT run
it), and pinned golangci-lint v1.64.8 clean on the touched packages.

Known deferred minor: `TestCredentialScrub_PlanRenderedSQL` asserts the last
duckdb plan source only — equivalent today since exactly one exists per query.

### DIFF
--- a/internal/entity_query_service.go
+++ b/internal/entity_query_service.go
@@ -141,10 +141,12 @@ func toExecutionPlan(plan *model.ExecutionPlan) *forma.ExecutionPlan {
 	}
 
 	// SECURITY: do not project src.SQL / src.Params or plan.Notes / merge.Notes.
-	// The DuckDB source SQL embeds the postgres_scan connection string (with the
-	// DB password) and notes can echo raw engine errors — surfacing them on the
-	// HTTP response would leak credentials to any advanced_query caller. Only
-	// safe routing/tier metadata crosses into the public projection.
+	// Since #306, the database password is redacted at the source (plan SQL and
+	// failure notes carry password=***REDACTED***), but the rendered DuckDB SQL
+	// still embeds host/user/dbname, table internals; Params carry query arguments;
+	// notes carry storage keys and engine internals — surfacing them on the HTTP
+	// response would leak internals to any advanced_query caller. Only safe
+	// routing/tier metadata crosses into the public projection.
 	out := &forma.ExecutionPlan{
 		Routing: forma.ExecutionRouting{
 			UsedDuckDB: plan.Routing.UseDuckDB,
--- a/internal/federated/credential_scrub_test.go
+++ b/internal/federated/credential_scrub_test.go
@@ -1,0 +1,102 @@
+package federated
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/lychee-technology/forma/internal/model"
+	"github.com/lychee-technology/forma/internal/redact"
+	"github.com/stretchr/testify/require"
+)
+
+// #306: the credential DuckDB quotes back on a postgres_scan attach failure
+// must be scrubbed before the text enters the engine's error chain — asserted
+// on the error string returned from the engine, not on any HTTP body (#301
+// already gates that boundary). Canary construction follows
+// internal/httpapi/canaries_test.go: the ';' makes truncation regressions
+// visible as a surviving tail, so all fragments are asserted separately.
+const (
+	scrubCanaryHead = "SUPERSECRET"
+	scrubCanaryTail = "CANARY-TAIL"
+	scrubCanary     = scrubCanaryHead + ";" + scrubCanaryTail
+)
+
+// attachFailureErr reproduces the DuckDB attach-failure prose shape (verified
+// against duckdb-go in #301): the whole given conn string echoed back, quoted.
+func attachFailureErr() error {
+	return fmt.Errorf(
+		`IO Error: Unable to connect to Postgres at "host=h port=5432 user=u password='%s' dbname=d": connection refused`,
+		scrubCanary)
+}
+
+func requireNoCanary(t *testing.T, s, surface string) {
+	t.Helper()
+	for _, frag := range []string{"password='" + scrubCanary + "'", scrubCanary, scrubCanaryHead, scrubCanaryTail} {
+		require.NotContains(t, s, frag, "credential fragment leaked into %s", surface)
+	}
+}
+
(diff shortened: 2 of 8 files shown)


iceboundrock:
## Assessment

No blocking or non-blocking changes requested. The source-side message scrub covers the documented embedder-facing error and execution-plan surfaces without altering query execution or federated error classification. Review assessment: approve.

iceboundrock:
## Observations

`redactedError.Unwrap()` deliberately retains the original credential-bearing leaf. This is not a finding: the accepted plan records it as an explicit residual needed to preserve sentinel/type classification, while requiring all normal composed error text to remain scrubbed. Any future change to that security boundary should add a separate policy for chain-walking loggers rather than silently changing this contract.

iceboundrock:
## Standards

Pass. Reviewed against `AGENTS.md` and `coding-standard.md`. The changed files/functions remain within the documented size limits, exported `redact.Error` is documented, error identity uses `errors.Is`/`errors.As`-compatible wrapping, and the tests pair credential-negative assertions with positive preconditions.

iceboundrock:
## Findings

No actionable findings.

I specifically examined the apparent raw-leaf exposure through `redactedError.Unwrap()`. The implementation plan explicitly accepts that residual in exchange for preserving `errors.Is`/`errors.As` and requires only that no composed error message contain the password. The returned engine error, execution-plan failure note, and rendered plan SQL meet that defined boundary.

iceboundrock:
## Verification

Fresh local verification at PR head `d2fc1e3`:

```text
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go build ./...
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go vet ./...
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./...
```

All commands passed. GitHub checks are also green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

iceboundrock:
## Spec

Pass. The PR satisfies issue #306 and its implementation plan:

- Scrubs DuckDB execute and mid-stream failure messages before they reach the returned composed engine error and plan failure notes.
- Preserves sentinel/type traversal, as required by the plan.
- Redacts the password in internal execution-plan rendered SQL while retaining the raw SQL for execution.
- Uses the established shared matcher and the required semicolon canary at the engine-error boundary.
- Correctly leaves `queryPostgresOnly` and parquet-schema probes unchanged because they do not carry the DuckDB DSN.

iceboundrock:
## Summary

Reviewed PR #325 against issue #306, the checked-in implementation plan, and repository standards. Findings: 0. Spec: pass. Standards: pass. Verification: pass. No merge action was performed.

Issue #326, opened 2026-07-27
cdc/runner.go: env credential half-pair + unconditional region clobber on the memoized Runner S3 runtime (third site deferred from #302)


iceboundrock:
Deferred from #302 (PR review ledger): the two AWS S3 client sites named by #302 (`internal/bootstrap/s3.go`, `internal/cdc/flusher.go setupAWSClient`) now (a) adopt env static credentials only when BOTH `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are non-empty (half-pair falls to the default chain) and (b) pass region to `LoadDefaultConfig` via `WithRegion`. A third site was found during review and deliberately left out of #302's scope; it carries **both** bugs plus two aggravators:

**`internal/cdc/runner.go` (the long-running Runner daemon path):**
- `resolveS3Credentials` (runner.go:273-283) falls back to env **per variable**, so a half-pair still builds an empty-secret static provider (`getOrCreateS3Runtime`, runner.go:211-213) — the exact opaque signing failure #302 eliminated elsewhere. It can also cross-mix a config key with an env secret.
- Region: runner.go:186-189 hardcodes a `"us-east-1"` default and line 210 assigns it post-load **unconditionally**, clobbering the chain-resolved region (worse than the pre-#302 sites, which only overwrote when configured).
- Aggravator: `secretAccessKey` is part of `s3RuntimeKey`, so the broken runtime is **memoized in `r.s3Runtimes` for the Runner's lifetime**.
- Note: the production e2e harness flush path (`internal/e2e_harness/production/cdc.go:83`) rides this unfixed path.

**Bundled follow-ups from the same review ledger (small, same file family):**
- Consolidate the two confusable cdc seams: pre-existing `loadAWSConfigFn` (runner.go:19) vs the new `loadAWSConfig` (flusher.go) — the old signature cannot carry `WithRegion`; 2 swap sites in runner_test.go. A future test stubbing the wrong seam gets a silent no-op.
- Intra-`RunOnce` divergence: the DuckDB httpfs credential plumbing (flusher.go:97-104) still reads the two env vars per-variable, so under a half-pair the SDK client now falls to the default chain while DuckDB gets key-with-empty-secret (before #302 both failed identically). Decide whether the both-halves rule should govern all four env-resolution points.
[...]

PR #327, opened 2026-07-27
feat: #302 harness parquetSource dedup, EntityManager Close hook, AWS config parity, TS e2e manifest reads

### LINKED ISSUE #302: Dedup harness parquetSource() with manifest.NewS3QuerySource; factory Close hook for DuckDB client
Two structural follow-ups deliberately deferred from #250:

1. **Harness dedup**: \`internal/e2e_harness/production/engine.go parquetSource()\` remains a second copy of the assembly now centralized in \`manifest.NewS3QuerySource\`. It was intentionally NOT deduped in #250 (plan decision D2): its closures read \`e.Cluster.S3\` at call time so \`RestartS3\` is picked up without re-wiring; naive dedup would freeze the client at construction and silently break the restart e2e family. Dedup requires a thin forwarding wrapper (a type whose GetObject/PutObject/HeadObject delegate to the current \`e.Cluster.S3\`).

2. **Factory Close hook**: \`factory.NewEntityManagerWithConfig*\` constructs a DuckDB client but \`forma.EntityManager\` exposes no Close; embedders (and the #250 factory-path e2e) leak an in-memory DuckDB instance per manager until process exit. Mitigated in tests via \`MaxConnections=1\`, but a real lifecycle hook is the fix.

Also parked here from #250 reviews (same file family, change together):
- \`internal/bootstrap/s3.go\` + \`internal/cdc/flusher.go setupAWSClient\`: empty \`AWS_SECRET_ACCESS_KEY\` with non-empty \`AWS_ACCESS_KEY_ID\` builds an empty-secret static provider (opaque signing failure instead of default-chain fallback); and region is applied post-load rather than via \`WithRegion\` (matters only for region-sensitive default-chain resolution: STS/SSO). Both sites must change together to preserve parity.
- Optional: enable manifest reads in the TS black-box e2e stack (tests/e2e) for HTTP-layer evidence — changes CI behavior, needs its own decision.

Origin: #250 plan follow-up list + review ledger T2/T3/T4 minors.

iceboundrock:
Closes #302

Five commits, one per follow-up deferred from #250:

1. **`b95eb48` harness dedup** — `internal/e2e_harness/production/engine.go parquetSource()` now delegates to `manifest.NewS3QuerySource` through a `clusterS3Client` forwarder whose `GetObject`/`PutObject`/`HeadObject` read `Cluster.S3` at call time, satisfying #250 plan decision D2's condition (a client rebuilt by `RestartS3` is picked up without re-wiring). Strict improvement over the old inline copy: the manifest `S3Store` client was previously frozen at construction, and the harness now inherits `S3ExistsProbe`'s bodyless-HEAD-404 classification (RustFS) and `S3FallbackGlob`'s writer-parity trim.
2. **`bf12ee9` EntityManager Close** — `forma.EntityManager` gains `Close() error`. **Deliberate breaking change for external interface implementers, adjudicated 2026-07-26**: compile-time breakage was chosen over embedders silently leaking a DuckDB instance per manager. Resources register via a new variadic `internal.WithCloser` option (variadic specifically to avoid the #314 add-a-positional-param trap); Close runs every closer (`errors.Join`), second call is a no-op, no-resources managers return nil.
3. **`5265b3b` factory wiring** — the factory registers the DuckDB client it constructs (typed-nil guarded at the call site, since a boxed typed-nil passes `WithCloser`'s interface-nil check); `cmd/server` closes the manager on graceful shutdown before the pool; `cmd/sample` defers Close; the factory e2e closes both managers via `t.Cleanup` and **keeps `MaxConnections=1`** — no longer leak mitigation, still pins #245 single-connection S3 session config. **`cmd/lambda` is deliberately unchanged**: its manager is process-lifetime (Lambda freezes/kills the sandbox; there is no shutdown hook to run Close from).
4. **`90f0ab5` AWS config parity (both sites in one commit, as #302 mandates)** — a non-empty `AWS_ACCESS_KEY_ID` with an empty `AWS_SECRET_ACCESS_KEY` used to build an empty-secret static provider whose only observable behavior is an opaque signing failure ("static credentials are empty" — captured verbatim in the red test run); it now falls through to the default chain. Region reaches `LoadDefaultConfig` via `WithRegion` so region-sensitive chain resolution (STS/SSO) sees it; final `Config.Region` is unchanged, pinned by tests. Explicit config-credential branches untouched. New cdc test seam mirrors bootstrap's.
5. **`5e643ee` TS smoke manifest reads** — CI's server now starts with `DUCKDB_MANIFEST_TEMPLATE`/`S3_BUCKET`/`S3_PREFIX`, and `federated-check` gains `--manifest-reads`, which omits the client-side `s3_parquet_path_template` hint. The flag is load-bearing: an explicit hint always wins over the manifest source (`internal/federated/duckdb_query_build.go:251-284`), so only a hint-free request proves the manifest path at the HTTP layer. The existing hinted run is kept as hint-path coverage.

## Honest-scope note (Task 5)

With `S3_PREFIX` set (writer parity), a schema whose manifest went missing would still be served by the legacy fallback glob — the new CI check proves the manifest read surface is **wired and healthy**, not that the fallback never fires. Sharp missing-key failure modes remain pinned by the Go e2e (`TestFactoryWiring_ManifestDrivenReads`). Local falsification controls during implementation showed the check is non-vacuous: with a bogus `S3_PREFIX` (fallback can't match) the hint-free run still passed (paths came from the manifest), and with the template unset it failed loudly (HTTP 500, 0/2000).

## Third site deferred → #326

Review found `internal/cdc/runner.go` carries **both** bugs #302 fixed elsewhere, plus aggravators: `resolveS3Credentials` falls back to env per-variable (half-pair → empty-secret static provider, config-key/env-secret cross-mix), region is hardcoded-defaulted to `us-east-1` and clobbers the chain-resolved region unconditionally, and the broken runtime is memoized in `s3Runtimes` for the Runner's lifetime. Deliberately out of this PR's scope (#302 names two sites and mandates they move together); tracked with the seam-consolidation and doc follow-ups in #326.

## Verification

- Per-commit: `make lint` + `make test` green; TDD red/green evidence in the branch history for Tasks 2-4.
- Locally (Docker): `TestS3Restart_CleanRecovery` + `TestFactoryWiring_ManifestDrivenReads` per task, and a **full `-tags=e2e` production sweep at the head: PASS, 0 failures (121s; the known #276 flakes were green this run)**.
- Task 5 runtime proof in CI is the smoke job itself; the full stack was also run locally (register → gen-data 2000 → cdc-flush → both federated-check modes) — all green.
- Final whole-branch review: zero Critical/Important findings; all deferred minors triaged (doc/polish items folded into #326).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,14 @@ jobs:
           S3_ENDPOINT: http://localhost:9000
           S3_ACCESS_KEY: minio
           S3_SECRET_KEY: minio_password
+          # Manifest-driven reads (#302): the flush step below writes
+          # manifests (cdc-flush passes --manifest-template); these vars make
+          # the server resolve parquet paths from them instead of relying on
+          # per-request hints. S3_BUCKET/S3_PREFIX mirror the tests/e2e
+          # defaults (forma-cdc / delta) the writers use.
+          DUCKDB_MANIFEST_TEMPLATE: "manifest/{{.SchemaID}}.json"
+          S3_BUCKET: forma-cdc
+          S3_PREFIX: delta
         run: |
           ./build/server &
           for i in $(seq 1 30); do
@@ -177,6 +185,10 @@ jobs:
           # flat schema; lead/visit have nested dotted attributes that currently
           # hit a federated-projection binder bug (tracked separately).
           bun run federated-check -- --schema log --require-duckdb
+          # Same route proof without the client-side path hint: an explicit
+          # hint always wins over the manifest source, so only a hint-free
+          # request exercises the server's manifest-driven read path (#302).
+          bun run federated-check -- --schema log --require-duckdb --manifest-reads
 
       # OLTP-path load smoke. The DuckDB route is proven above by
       # federated-check --require-duckdb; the k6 latency SLA deliberately does
--- a/cmd/lambda/main.go
+++ b/cmd/lambda/main.go
@@ -104,7 +104,12 @@ func bootstrapLambda(ctx context.Context, sugar *zap.SugaredLogger) (*lambdaRunt
 
 	formaConfig := lambdaFormaConfig(registry, schemaDir, tableNames)
 
-	// Initialize EntityManager using factory
+	// Initialize EntityManager using factory. Deliberately never Closed (#302):
+	// the manager is process-lifetime — bootstrapLambda runs once per execution
+	// environment — and the Go Lambda runtime has no shutdown hook to run
+	// cleanup from (SIGTERM reaches the function process only when an external
+	// extension is registered; otherwise the frozen sandbox is reclaimed with
+	// the process, DuckDB instance included).
 	manager, err := factory.NewEntityManagerWithConfigContext(startupCtx, formaConfig, dbPool)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create entity manager: %w", err)
--- a/cmd/sample/main.go
+++ b/cmd/sample/main.go
@@ -137,7 +137,11 @@ func main() {
 		sugar.Fatalf("Failed to create entity manager: %v", err)
 	}
 	// Defers run LIFO, so the manager is released before the pool (#302).
-	defer func() { _ = entityManager.Close() }()
+	defer func() {
+		if err := entityManager.Close(); err != nil {
+			sugar.Warnf("Failed to close entity manager: %v", err)
+		}
+	}()
 
 	// Create importer
 	importer := NewCSVImporter(entityManager, mapper, *batchSize)
--- a/cmd/sample/main.go
+++ b/cmd/sample/main.go
@@ -136,6 +136,8 @@ func main() {
 	if err != nil {
 		sugar.Fatalf("Failed to create entity manager: %v", err)
 	}
+	// Defers run LIFO, so the manager is released before the pool (#302).
+	defer func() { _ = entityManager.Close() }()
 
 	// Create importer
 	importer := NewCSVImporter(entityManager, mapper, *batchSize)
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -22,8 +22,9 @@ import (
 )
 
 type serverRuntime struct {
-	pool   *pgxpool.Pool
-	server *httpapi.Server
+	pool    *pgxpool.Pool
+	manager forma.EntityManager
+	server  *httpapi.Server
 }
 
 func main() {
@@ -44,6 +45,13 @@ func main() {
 		sugar.Fatalf("failed to bootstrap server: %v", err)
 	}
 	defer runtime.pool.Close()
+	// Defers run LIFO, so the manager — whose DuckDB holds a Postgres
+	// attachment DSN — is released before the pool it points at (#302).
+	defer func() {
+		if err := runtime.manager.Close(); err != nil {
+			sugar.Warnw("failed to close entity manager", "err", err)
+		}
+	}()
 
 	port := bootstrap.Env("PORT", "8080")
 	zap.S().Infow("starting server", "port", port)
@@ -198,8 +206,9 @@ func bootstrapServer(ctx context.Context, sugar *zap.SugaredLogger) (*serverRunt
 	}
 
 	return &serverRuntime{
-		pool:   pool,
-		server: httpapi.NewServer(manager, httpapi.Options{EnableHealth: true}),
+		pool:    pool,
+		manager: manager,
+		server:  httpapi.NewServer(manager, httpapi.Options{EnableHealth: true}),
 	}, nil
 }
 
--- a/factory/factory.go
+++ b/factory/factory.go
@@ -184,9 +184,17 @@ func newEntityManagerWithConfigContext(ctx context.Context, config *forma.Config
 		federated.DuckDBPostgresConnStringFromPool(pool),
 		engineOpts...,
 	)
+	// The DuckDB client has no owner other than the manager being built:
+	// register it so EntityManager.Close releases it (#302). Guarded on nil
+	// because a typed-nil *DuckDBClient boxed into io.Closer would defeat
+	// WithCloser's nil check.
+	var managerOpts []internal.EntityManagerOption
+	if duckClient != nil {
+		managerOpts = append(managerOpts, internal.WithCloser(duckClient))
+	}
 	// Create and return entity manager
 	return internal.NewEntityManager(
-		transformer, repository, federatedEngine, registry, effectiveConfig, schemaValidator), nil
+		transformer, repository, federatedEngine, registry, effectiveConfig, schemaValidator, managerOpts...), nil
 }
 
 // newFederatedReadSurface opens the DuckDB client and the manifest parquet source
--- a/factory/factory_close_test.go
+++ b/factory/factory_close_test.go
@@ -1,0 +1,53 @@
+package factory
+
+import (
+	"context"
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/lychee-technology/forma/internal/federated"
+	"github.com/lychee-technology/forma/internal/schemameta"
+)
+
+// TestNewEntityManagerWithConfig_CloseReleasesDuckDBClient pins the #302
+// lifecycle fix: the DuckDB client the factory constructs is owned by the
+// returned manager and released by Close — previously it leaked until
+// process exit (mitigated in e2e only via MaxConnections=1).
+func TestNewEntityManagerWithConfig_CloseReleasesDuckDBClient(t *testing.T) {
+	t.Parallel()
+
+	// Base fixture: same stubbed deps + config as
+	// TestNewEntityManagerWithConfig_Unit_Success, with DuckDB enabled and
+	// the client constructor capturing the real client it builds.
+	var captured *federated.DuckDBClient
+	deps := unitEntityManagerDeps(schemameta.NewMetadataCache())
+	deps.newDuckDBClient = func(ctx context.Context, cfg forma.DuckDBConfig) (*federated.DuckDBClient, error) {
+		c, err := federated.NewDuckDBClientContext(ctx, cfg)
+		captured = c
+		return c, err
+	}
+	config := unitEntityManagerConfig(t)
+	config.DuckDB.Enabled = true
+	config.DuckDB.MaxConnections = 1
+
+	em, err := newEntityManagerWithConfigContext(context.Background(), config, nil, deps)
+	if err != nil {
+		t.Fatalf("build entity manager: %v", err)
+	}
+	if captured == nil || captured.DB == nil {
+		t.Fatal("test wiring: DuckDB client was not constructed; the assertion below would be vacuous")
+	}
+	if err := captured.DB.Ping(); err != nil {
+		t.Fatalf("DuckDB client must be live before Close: %v", err)
+	}
+
+	if err := em.Close(); err != nil {
+		t.Fatalf("Close: %v", err)
+	}
+	if err := captured.DB.Ping(); err == nil {
+		t.Fatal("DuckDB client still answers Ping after manager Close; the factory did not register it")
+	}
+	if err := em.Close(); err != nil {
+		t.Fatalf("second Close must be a no-op: %v", err)
+	}
+}
--- a/factory/factory_test.go
+++ b/factory/factory_test.go
@@ -222,6 +222,21 @@ func unitEntityManagerDeps(cache *schemameta.MetadataCache) entityManagerDepende
 	return deps
 }
 
+// unitEntityManagerConfig is the config half of the successful-construction
+// fixture: the table names unitEntityManagerDeps' collector reports, plus an
+// empty schema directory.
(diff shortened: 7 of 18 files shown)


iceboundrock (commented):
## Summary

PR #327 addresses #302 across five well-scoped commits: e2e harness parquetSource dedup (using `manifest.NewS3QuerySource` via `clusterS3Client`), `EntityManager.Close()` API, factory wiring of DuckDB into Close, AWS credential half-pair fix (both bootstrap and CDC sites), and TS smoke test manifest-driven reads. The architecture choices are sound — the `clusterS3Client` forwarding pattern correctly handles `RestartS3` without freezing the client at construction, and the `WithRegion` at load time properly governs STS/SSO resolution.

## Spec

The code matches issue #302 requirements. All five listed items are covered, with #326 deferred for CDC site 3 (`internal/cdc/runner.go`) as noted. The `Close()` idempotency contract (second Close is a no-op) is correctly implemented and tested.

## Standards

- Error wrapping: all new error paths use `fmt.Errorf("...: %w", ...)` with context — compliant.
- Function/file sizes: no new function or file exceeds the 100/500 line limits.
- Early returns and flat structure: maintained throughout.
- Godoc: exported items (`Close`, `WithCloser`, `EntityManagerOption`, `clusterS3Client`, `loadAWSConfig`) have complete doc comments.

## Findings

### 1. Lambda entrypoint does not close the manager

`cmd/lambda/main.go:108` creates a manager via `factory.NewEntityManagerWithConfigContext` but never calls `Close()`. Both `cmd/server/main.go` and `cmd/sample/main.go` were updated with Close deferrals; the Lambda was missed. Severity is low (Lambda invocations are short-lived), but it's an inconsistency worth noting.

### 2. Sample silently discards Close error

`cmd/sample/main.go` uses `defer func() { _ = entityManager.Close() }()` while `cmd/server/main.go` logs it with `sugar.Warnw`. The sample tool is indeed less critical, but consider at least logging the error at Warn level for operational visibility.

### 3. DuckDB credential env half-pair remains in RunOnce

`internal/cdc/flusher.go:92-98` (DuckDB S3 credentials for the exporter, not the S3 client) independently reads `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` with empty-check on each individually. This is the same half-pair pattern that was fixed in `setupAWSClient` and `applyCredentials`. This code path is pre-existing, not introduced by this PR, but it should be tracked as a follow-up since DuckDB's httpfs would hit the same opaque signing failure.

### 4. Process-global seam variables — correct but fragile

Both `internal/bootstrap/s3.go` and `internal/cdc/flusher.go` use package-level `var loadAWSConfig` as a test seam. Tests correctly avoid `t.Parallel()` and restore via `t.Cleanup`. This pattern is safe as implemented, but the indirect coupling (a parallel test in the same package would race) is worth flagging for future maintainers.

## Observations

- `factory/factory.go:196-200` typed-nil guard on `duckClient != nil` before wrapping in `WithCloser` is correct — a `(*DuckDBClient)(nil)` boxed into `io.Closer` would pass `WithCloser`'s `if c != nil` check. The comment explains this well. Future callers of `WithCloser` need to replicate this discipline; consider adding a typed-nil check inside `WithCloser` itself as a defense-in-depth measure.
- `clusterS3Client` reads `Cluster.S3` at call time — this is the intended design for `RestartS3` support, but note that a racing `RestartS3` during an active read could technically observe a partially-stored pointer. In practice this is harmless since `RestartS3` is called between test phases.
- The `mockEntityManager.Close()` addition in `server_test.go` is the only mock update needed — no other `EntityManager` interface implementor exists outside of tests.

## Verification

- `factory/factory_close_test.go`: verifies factory-registered DuckDB client is released by Close (Ping fails after), idempotent re-close, and nil-closer vacuum guard.
- `internal/entity_manager_close_test.go`: verifies idempotency, errors.Join for failing closers, and no-resource case returns nil.
- `internal/bootstrap/s3_test.go`: `TestNewS3Client_EnvHalfPairPreservesDefaultChain` and `TestNewS3Client_RegionPassedAtLoad` — both pin the new behavior.
- `internal/cdc/flusher_aws_test.go`: mirrors all three bootstrap tests plus `TestSetupAWSClient_ConfigCredentialsStillWin` regression guard.
- `tests/e2e/scripts/federated-check.ts`: `--manifest-reads` flag omits the per-request path hint to exercise the server's manifest-driven read path.
- CI: two `federated-check` invocations in `.github/workflows/ci.yml` — one with hint (existing), one without (new manifest-reads path).

## Assessment

The PR is well-structured and each commit is self-contained. The lifecycle fix (Close) is correct and idempotent. The AWS config parity fixes are mirrored faithfully across both S3 client sites. Test coverage is thorough for all new behaviors. The Lambda Close gap is the only actionable item; everything else is observation-level.

Please address the Lambda gap (either add Close, or document why it's not needed) before merging.

iceboundrock:
Review response — fixes in `ef62064`, item-by-item:

**Finding 1 (Lambda Close gap) — documented in code (`ef62064`).** One correction to the finding's framing: the Lambda was not missed — it was a deliberate decision recorded in the PR body ("cmd/lambda is deliberately unchanged"). Taking the review's second option, the rationale now lives at the construction site (`cmd/lambda/main.go:107-112`): the manager is process-lifetime (`bootstrapLambda` runs once per execution environment), and the Go Lambda runtime offers no shutdown hook to run a Close from — SIGTERM reaches the function process only when an external extension is registered; otherwise the frozen sandbox is reclaimed with the process, DuckDB instance included. Adding signal-based cleanup for a hook that doesn't fire in this deployment shape would be speculative.

**Finding 2 (sample discards Close error) — fixed (`ef62064`).** `cmd/sample/main.go` now logs a failed Close via `sugar.Warnf`, matching the file's formatted-sugar idiom and mirroring `cmd/server`'s Warn-level treatment.

**Finding 3 (DuckDB env half-pair in RunOnce) — already tracked, no change.** This is the "intra-RunOnce divergence" bullet in #326: `flusher.go:97-104` still resolves the pair per-variable for DuckDB httpfs, so under a half-pair the SDK client now falls to the default chain while DuckDB gets key-with-empty-secret. #326 frames the open decision as whether the both-halves rule should govern all four env-resolution points (the two fixed SDK sites, runner.go's `resolveS3Credentials`, and this DuckDB path) — fixing it piecemeal here would add a fifth semantic mid-flight.

**Finding 4 (process-global seam fragility) — flag already in place, no change.** Both stub helpers carry the warning for future maintainers: `internal/bootstrap/s3_test.go:14-16` ("Tests using it must NOT call t.Parallel(): the seam is process-global state") and its mirror in `internal/cdc/flusher_aws_test.go`. `internal/cdc` currently has zero `t.Parallel()` calls package-wide (checked during the task review).

**Observation (typed-nil check inside `WithCloser`) — pushing back, with an offer.** A robust in-`WithCloser` guard requires reflection (`reflect.ValueOf(c).IsNil()` gated on kind), which buys defense against a caller mistake that is currently harmless twice over: `(*DuckDBClient).Close` is nil-receiver safe, and `WithCloser` has exactly one caller (`factory`), which guards. The reflect dependency and its kind-gating subtlety outweigh the risk today. If a second `WithCloser` caller ever lands, that's the moment to revisit — noted as a fold-in candidate for #326.

**Observation (racing RestartS3 vs in-flight forwarder read) — agreed, harmless today.** Same conclusion as the task-1 review: pre-existing pattern (the old `Exists` closure read `c.S3` identically), restart tests are sequential, and it only becomes real if a future test restarts S3 concurrently with an in-flight query under `-race`. On the ledger as a known-latent item.

Gates on `ef62064`: `go build ./cmd/...`, `go vet ./cmd/...`, `make test`, `make lint` — all clean.

iceboundrock (commented):
## Summary

Second review of the current head `ef62064` (including the response to the prior review). The Lambda lifecycle rationale is now documented at the construction site, and the sample command now logs a failed manager close. The remaining introduced correctness issue is concurrent `EntityManager.Close()`.

## Findings

### P2: Concurrent Close races and can close owned resources more than once

`internal/entity_manager.go:61-69` implements sequential idempotency by clearing `em.closers` only after iteration, but it has no synchronization. Two goroutines can both read the slice, call the same `io.Closer`, and race while assigning it to nil. This is reachable through the newly public `forma.EntityManager.Close()` API and is particularly unsafe for future closers that do not tolerate a double close.

I reproduced this against the PR head with a temporary 32-goroutine test harness: `go test -race ./internal -run TestReviewConcurrentClose -count=1` reports races at `entity_manager.go:63` and `:68`.

Guard teardown with `sync.Once` (or a mutex), cache the joined result so every caller receives the same error, and add a concurrent-close regression test run under `-race`.

### P2: Close returns raw closer errors without manager-level context

`internal/entity_manager.go:64-65` appends the raw error from each owned closer. For the actual DuckDB path, `DuckDBClient.Close()` returns `c.DB.Close()` directly (`internal/federated/duckdb_conn.go:245-250`), so an API caller cannot tell that the failure occurred while closing an EntityManager resource. This violates `coding-standard.md` §3, which requires errors to be wrapped with contextual `fmt.Errorf(...: %w, err)`. Wrap each resource-close error before joining while preserving `errors.Is`.

## Spec

Pass. The current diff fulfills issue #302:

- `parquetSource()` now uses `manifest.NewS3QuerySource` through a call-time `Cluster.S3` forwarder, preserving `RestartS3`.
- Factory-built managers own and close their DuckDB client; normal server shutdown closes the manager before the Postgres pool.
- Both requested SDK sites apply `WithRegion` at load time and retain the default credential chain for an environment half-pair.
- CI exercises the hint-free HTTP manifest-read route.

The DuckDB-exporter half-pair path in `RunOnce` and `internal/cdc/runner.go` remain deferred to #326 as explicitly documented, so they are not spec misses in this PR.

## Standards

- **Violation:** raw close-error propagation above violates `coding-standard.md` §3.
- **Follow-up debt:** `factory/factory_test.go` is 896 lines after this change, above the 500-line limit in `coding-standard.md` §1. It was already oversized, so this is non-blocking, but the incremental-improvement rule calls for a follow-up refactor when touching it.
- Exported API documentation, function limits, flat control flow, and API-break justification are otherwise compliant.

## Observations

- The prior Lambda finding is resolved: `cmd/lambda/main.go:107-112` documents intentional process-lifetime ownership, and `cmd/sample/main.go` now warns on close failure.
- The process-global AWS load seam remains test-safe as written because its tests do not run in parallel and restore it with `t.Cleanup`.
- The hint-free manifest smoke test confirms the read surface is wired, but, as the PR description candidly notes, its configured fallback glob means it cannot prove a manifest object was consumed when the manifest is absent. The production Go e2e remains the coverage for missing-manifest/missing-key failure semantics.

## Verification

- Passed locally at PR head:
  - `go test ./internal -run "^TestEntityManagerClose_" -count=1`
  - `go test ./factory -run "^TestNewEntityManagerWithConfig_CloseReleasesDuckDBClient$" -count=1`
  - `go test ./internal/bootstrap -run "^(TestNewS3Client_EnvHalfPairPreservesDefaultChain|TestNewS3Client_RegionPassedAtLoad)$" -count=1`
  - `go test ./internal/cdc -run "^TestSetupAWSClient_" -count=1`
- Failed as expected: temporary concurrent-close harness under `go test -race` (race at `internal/entity_manager.go:63,68`).
- GitHub CI currently passes: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

## Assessment

The prior actionable review feedback is addressed and #302 itself is implemented. Do not merge until `EntityManager.Close()` is concurrency-safe and close failures carry manager-level context.

iceboundrock:
Round-2 response — both P2s fixed in `4de33fe`:

**P2-1 (concurrent Close race) — fixed, race confirmed first.** Reproduced your finding red-first: a 32-goroutine regression test (`TestEntityManagerClose_ConcurrentCallsCloseOnce`) under `go test -race` failed exactly as you reported — race at the old `entity_manager.go:63/68` plus the resource closed twice. Fix is the shape you suggested: teardown guarded by `sync.Once`, joined result cached in `closeErr` so every caller — first, repeated, or racing — receives the same outcome (the `Once` happens-before edge makes the cached read safe). One deliberate behavior change to flag: previously a *sequential* second Close after a **failed** first close returned nil; it now returns the same cached error, so the failure stays observable to late callers. The interface doc (`storage.go`) and method doc now state the concurrency contract. Green under `-race` for the whole close family plus the factory lifecycle test.

**P2-2 (raw close errors) — fixed.** Each resource-close failure is wrapped as `close entity manager resource %T: %w` before joining — an API caller now sees where the failure occurred, and `errors.Is` still reaches the underlying error (pinned: `TestEntityManagerClose_JoinsResourceErrors` now asserts both the sentinel via `errors.Is` and the manager-level context string).

**Standards note (factory_test.go 896 lines) — recorded on #326** (issuecomment-5087390692) with the natural split seams, per the incremental-improvement rule. Not expanded in this PR to keep the round scoped to the two P2s.

Gates on `4de33fe`: `go test -race ./internal -run TestEntityManagerClose -count=1` 4/4, `go test -race ./factory -run CloseReleasesDuckDBClient` ok, `make test` no failures, `make lint` clean, `gofmt` clean.

PR #328, opened 2026-07-27
fix(cdc): #285 per-connection DuckDB init and bounded pool for DuckExporter

### LINKED ISSUE #285: NewDuckExporter applies session-scoped DuckDB init through the pool — same gap as #245, with an unbounded pool

iceboundrock:
Closes #285

## What

`NewDuckExporter` had the same defect class #245 fixed on the federated side: all session-scoped DuckDB initialization (PRAGMA / INSTALL+LOAD / `SET s3_*`) was issued through the pool via `db.ExecContext`, landing on one arbitrary pooled connection — and the pool was **unbounded** (`sql.Open` default, no `SetMaxOpenConns`). Any concurrent use (the Runner's cached exporters serving concurrent flush passes, `compaction.DuckMerger` reusing `exporter.DB`) could lazily open connections carrying none of the S3 session settings and fail exactly like the #245 reproduction (`in region ''` HTTP 404).

## How

- **New shared package `internal/duckdbinit`** (user decision: extract, not duplicate): `Stmt`/`Step` types, `SingleStmtStep`, `ExtensionStep` (INSTALL+LOAD paired so a failed INSTALL skips its LOAD), `ValidateS3Credential`, and `MakeConnInit` — the connector init hook that runs the statement list on every new physical connection (failures are logged and skipped, never blocking the connection).
- **federated migrated to the shared package** — pure mechanical type/function substitution, zero behavior change; the existing federated suite (incl. the three #245 behavior tests) is the regression gate. The package-local `validateS3Credential` duplicates in both `internal/federated` and `internal/cdc` are gone.
- **`NewDuckExporter` rewired**: pre-build init statements (fail-fast credential validation before any connection opens) → `duckdb.NewConnector(dsn, connInit)` → `sql.OpenDB` → `SetMaxOpenConns(1)` → `PingContext` (opens and thereby initializes the first connection). Statement set and order are preserved exactly, including the `AWS_SESSION_TOKEN` env fallback and the `https://`/`http://` endpoint trimming.
- **Pool policy (user decision)**: hard-coded `SetMaxOpenConns(1)`, no new `CDCConfig` knob (YAGNI — all six call sites use the exporter sequentially; per-connection init makes a future larger pool safe when one is actually needed). The now-redundant manual `SetMaxOpenConns(1)` pin in `cmd/tools/manifest_reconcile.go` is removed.

## Deliberate tightening

Construction now fails fast (`PingContext`, 5s) if the DuckDB path is unusable. Previously `sql.Open` + warn-and-skip exec meant a broken `DuckDBPath` produced a "successful" constructor and only blew up at first export. This aligns the exporter with the federated client's contract.

## Notes for review

- The per-export SQL still embeds `PRAGMA memory_limit; ATTACH IF NOT EXISTS ... AS pg_db` (the `:270`/`:286` sites named in the issue): those run in the **same `ExecContext` batch** as the `COPY`, i.e. on the same connection, so their connection-locality is unaffected by this change.
- Red-first TDD: the new `TestNewDuckExporter_FreshConnectionsConfigured` demonstrated the defect pre-fix (fresh physical connection returns NULL for `current_setting('s3_region')`), and `TestNewDuckExporter_PoolBoundedToSingleConnection` pinned the unbounded pool (`MaxOpenConnections == 0`).
- `Runner.duckExporters` caching exporters for process lifetime is tracked separately in #326 — out of scope here.

## Verification

- `make test` — all 31 packages green; `make lint` clean (pinned golangci-lint v1.64.8).
- Focused packages: `go test ./internal/cdc ./internal/federated ./internal/duckdbinit ./internal/compaction ./factory` green.
- `make test-e2e-production` — full suite green (122s, 0 failures), covering real CDC flush/init, compaction, fault matrix, and circuit-breaker concurrency paths on the rewired exporter.

### DIFF
--- a/cmd/tools/manifest_reconcile.go
+++ b/cmd/tools/manifest_reconcile.go
@@ -250,7 +250,6 @@ func openReconcileStatsEngine(ctx context.Context, opts *reconcileOptions, logge
 	if err != nil {
 		return nil, fmt.Errorf("open stats duckdb: %w", err)
 	}
-	exporter.DB.SetMaxOpenConns(1)
 	return exporter, nil
 }
 
--- a/docs/superpowers/plans/2026-07-26-issue-285-cdc-exporter-per-connection-init.md
+++ b/docs/superpowers/plans/2026-07-26-issue-285-cdc-exporter-per-connection-init.md
@@ -1,0 +1,614 @@
+# Issue #285 — CDC DuckExporter per-connection init + 有界池 实施计划
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** 把 #245 的 per-connection init 模式应用到 `internal/cdc/duckdb_exporter.go` 的 `NewDuckExporter`，消除"session 级 DuckDB 初始化只落单条池连接 + 无界池"缺陷；同时按用户裁决把 init 机制抽成共享包 `internal/duckdbinit`，federated 同步迁移。
+
+**Architecture:** 新建 `internal/duckdbinit` 承载 Step/Stmt 类型、`MakeConnInit` 钩子和 `ValidateS3Credential`（两包现有副本删除）。federated 的语句构建器保留在原包但改用共享类型（纯机械迁移，零行为变化）。cdc 侧把 `NewDuckExporter` 里 ~100 行的逐语句 ExecContext 改为：构建语句列表（fail-fast 凭据校验）→ `duckdb.NewConnector(dsn, connInit)` → `sql.OpenDB` → `SetMaxOpenConns(1)` → `PingContext`。
+
+**Tech Stack:** Go, github.com/duckdb/duckdb-go/v2（`NewConnector` 每条新物理连接调用 connInitFn）, database/sql, zap。
+
+## Context（为什么做）
+
+`NewDuckExporter`（internal/cdc/duckdb_exporter.go:43-144）用 `sql.Open` + `db.ExecContext` 做 PRAGMA/INSTALL/LOAD/SET s3_* 初始化——这些是 **session（连接）级**语句，只落到池中一条任意连接；且从不调 `SetMaxOpenConns`，池默认**无界**。任何并发使用（`Runner.getOrCreateDuckExporter` 缓存 exporter 供并发 flush 复用，runner.go:234-271；`compaction.DuckMerger` 直接复用 `exporter.DB`）都可能懒开出**没有任何 S3 配置**的连接 → 与 #245 相同的 `in region ''` HTTP 404。`cmd/tools/manifest_reconcile.go:253` 已经手工 `SetMaxOpenConns(1)` workaround。#245 已在 federated 侧修完（PR #286, 437efc9），本 issue 是它 PR 里点名的 CDC 侧 follow-up。
+
+**用户裁决（2026-07-26）：**
+1. init 机制抽共享包（非 cdc 本地复制一份）。
+2. exporter 池策略 = 硬编码 `SetMaxOpenConns(1)`，不加 CDCConfig 字段（YAGNI；6 个调用方全顺序使用；per-connection init 已让未来放开变安全）。
+
+**已核实的关键事实：**
+- federated 参考实现：`internal/federated/duckdb_conn.go` — `buildInitSteps`(:143) / `initStep`/`initStmt`(:123-137) / `makeConnInit`(:227) / `validateS3Credential`(:48)。step 语义：step 内一条失败跳过该 step 余下语句（INSTALL 失败不 LOAD），后续 step 照跑；所有执行失败 log+skip 不阻断连接。
+- cdc 现状语句集（顺序）：PRAGMA memory_limit（若 DuckMemLimit≠""）→ PRAGMA threads（若 >0）→ INSTALL/LOAD postgres_scanner, httpfs, parquet → SET s3_access_key_id / s3_secret_access_key（非空时，先校验）→ SET s3_session_token（cfg.S3SessionToken 或 env AWS_SESSION_TOKEN 回落）→ SET s3_region → SET s3_endpoint（trim https:// 和 http:// 两种前缀）→ SET s3_use_ssl=true/false（**恒设**）→ SET s3_url_style='path'（仅 S3UsePath）。执行失败全部 Warnw+skip；凭据非法则 `db.Close()` + 返回错误。
+- `validateS3Credential` 两包字面相同副本：`internal/cdc/helpers.go:196`、`internal/federated/duckdb_conn.go:48`（禁字符集 `'";\ ` 含空格）；repo 内无直接单测。
+- cdc 包内 `validateS3Credential` 唯一调用方就是 duckdb_exporter.go（5 处）；federated 内调用在 buildS3Steps(:194,:201)。
+- federated 测试引用点：`duckdb_conn_test.go:262-266`（`TestMakeConnInit_FailedInstallSkipsLoad`，用 `recordingExecer` fake）；#245 三个行为测试（First/All/ConcurrentConnectionsConfigured）以 `SELECT current_setting('s3_region')` 为断言锚点。
+- `NewDuckExporter` 调用方 6 处，签名不变故均无需改动：flusher.go:105、init.go:106、runner.go:254（经 `newDuckExporterFn` seam）、cmd/tools/compactor.go:123、cmd/tools/manifest_reconcile.go:249（**:253 的手工 SetMaxOpenConns(1) 变冗余，须删**）、internal/e2e_harness/production/cdc.go:199。
+- `sql.Open("duckdb", ...)` 生产代码仅此一处；修复后全仓生产代码无裸 sql.Open duckdb。
+- 每次导出 SQL 内嵌的 `PRAGMA memory_limit; ATTACH IF NOT EXISTS ... AS pg_db` 与 COPY 在**同一次 ExecContext**（单连接内的多语句批）执行，连接本地性天然满足，不改。
+- 现有 cdc 单测对 NewDuckExporter 初始化路径**零覆盖**（仅 SQL 构建测试）。
+
+**刻意的行为收紧（PR 里要点名）：** 现状下 DuckDBPath 不可用时构造照样成功（Exec 失败只 Warnw），首次导出才炸；新实现加 `PingContext`（5s）后构造期 fail-fast。与 federated 行为对齐。
+
+## Global Constraints
+
+- 源文件 ≤500 行、函数 ≤100 行（coding-standard.md）。
+- 错误必须带上下文包装 `fmt.Errorf("...: %w", err)`，禁止裸 `return err`。
+- lint 用仓库钉住的 golangci-lint v1.64.8（`make lint`），不得升级。
+- 单测命令模板：`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test <pkg> -run <Name> -v`。
+- 子代理禁用 Sonnet 5；不能用 Haiku 的任务用 Opus 4.8。
+- PR 不自动合并；PR 引用 `Closes #285`。
+- federated 迁移必须**零行为变化**——现有 federated 测试套件是回归门。
+
+## File Structure
+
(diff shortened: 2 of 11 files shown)


iceboundrock:
## Findings

### High: connection initialization discards the caller deadline

`internal/duckdbinit/duckdbinit.go:69` calls `ExecContext(context.Background(), ...)`. `duckdb.Connector.Connect(ctx)` installs the supplied connect context before invoking this callback, but this call replaces it with an uncancelable context. Therefore the constructor's five-second `PingContext` no longer bounds a blocking `INSTALL` or `LOAD`.

On a host where the extension repository or network stalls, CDC flush/init, compaction, or manifest reconcile can hang during exporter construction rather than returning the intended fail-fast error. Preserve a bounded context for the init executions, then add a regression test that proves a blocked init statement observes cancellation.

iceboundrock:
## Spec

One requirement remains partial. Issue #285 requires per-physical-connection initialization, while this PR deliberately tightens construction to fail fast through a five-second `PingContext`. The connector hook receives that context, but `MakeConnInit` executes each statement using `context.Background()`, so the init phase is not actually bounded by the five-second deadline.

Apart from that gap, the PR follows the issue and implementation plan: it uses `duckdb.NewConnector` plus `sql.OpenDB`, validates credentials before opening a connection, initializes every physical connection, pins the exporter pool to one connection, and retains connection-local `ATTACH` work in the export batch. No unrequested scope creep found.

iceboundrock:
## Standards

One documented-standard violation: `internal/federated/duckdb_conn_test.go` is now 531 lines. `coding-standard.md` §1 rejects source files over 500 lines. The file already exceeded the limit before this PR, but this PR modifies it and `coding-standard.md` Incremental Improvement Mode requires fixing the immediate standards issue or suggesting follow-up refactoring.

Move the connector-init test helpers/tests into a focused test file, or record an explicit follow-up. No other documented-standard violations found.

iceboundrock:
## Assessment

**Request changes.** The per-connection initialization and single-connection pool address the primary defect, but the high-severity timeout/cancellation regression can hang process startup under extension/network failure. Resolve that before merge.

The test-file size issue is lower priority but should be addressed or tracked explicitly.

iceboundrock:
## Observations

- Extracting `internal/duckdbinit` removes duplicated connection-hook logic and keeps `INSTALL`/`LOAD` failure semantics consistent between CDC and federated queries.
- `TestNewDuckExporter_FreshConnectionsConfigured` correctly forces fresh physical connections and verifies a session-scoped S3 setting, which directly protects the original failure mode.
- `SetMaxOpenConns(1)` matches the stated sequential-export policy while per-connection initialization keeps a future pool expansion safe.
- A cancellation/timeout test for the connector hook is still missing; current tests only cover statement failure and successful fresh connections.

iceboundrock:
## Verification

Executed against the PR checkout:

- `go test ./internal/duckdbinit ./internal/cdc ./internal/federated ./internal/compaction ./factory`
- `make test`
- `make lint`

All commands passed. The full Docker-backed production E2E suite was not re-run in this review.

iceboundrock:
## Summary

Found **1 high-severity correctness issue** and **1 standards issue**. The high issue is the lost init timeout/cancellation at `internal/duckdbinit/duckdbinit.go:69`; it blocks approval until fixed. The design otherwise implements issue #285 as intended, and the independently run unit and lint gates pass.

iceboundrock:
## Review round 1 — both findings addressed

### High: init discards the caller deadline → fixed in 71147b0

Verified the mechanism against the driver source (duckdb-go v2.5.6) before fixing; one precision note on the diagnosis: `Conn.setContext` goes through `ctxStore.store(id, ctx, replace=false)`, so the hook's `ExecContext(context.Background())` did not actually *replace* the stored connect context — the stored context is simply never consulted for cancellation. Cancellation rides entirely on the ctx **argument** threaded into `runWithCtxInterrupt` via `Stmt.ExecContext` (context.go:77, statement.go:589), and that argument was `Background`. The outcome is exactly as the finding states: a stalled `INSTALL`/`LOAD` blocks `Connect`, and since `database/sql`'s synchronous open path calls `connector.Connect(ctx)` inline, the constructor hangs past its intended 5s bound.

Fix shape: `MakeConnInit` now takes an `initTimeout` and runs each hook invocation under one shared deadline (`duckdbinit.DefaultInitTimeout` = 5s — matches both constructors' ping bound and the pre-#285 exporter's construction-time init deadline). I bounded per hook run rather than literally preserving the caller deadline for two reasons:

1. The hook signature `func(driver.ExecerContext) error` has no access to the Connect context (the store is unexported), so the caller deadline is unreachable through the public API.
2. Pool-cycled connections opened later by `database/sql`'s background `connectionOpener` carry the pool's lifetime context, not any caller deadline — a per-run bound is the only shape that also covers those.

Timed-out statements keep the existing log-and-skip contract (init never blocks the connection; the expired shared deadline short-circuits the remaining statements via `runWithCtxInterrupt`'s `ctx.Err()` check). This also closes the same latent gap on the federated side, present since #245.

Regression test: `TestMakeConnInit_BlockedStmtObservesCancellation` — a fake execer blocks its statement until ctx cancellation and records whether it saw a deadline. Mutation-verified: reverting the implementation to `context.Background()` drives the test into its 5s hang guard (FAIL), restoring the bound turns it green.

### Standards: 531-line test file → fixed in b42ffd2

Pure move (sed-extracted block, no content edits): the #245/#285 connection-init tests (`FirstConnection`/`AllConnections`/`ConcurrentConnectionsConfigured`, `recordingExecer`, `TestMakeConnInit_FailedInstallSkipsLoad`, and the shared `newS3ConfiguredDuckDBConfig` fixture) moved to a new focused `duckdb_conn_init_test.go`. Line counts now 372 + 173.

Commit note: 71147b0 updates the `MakeConnInit` call site inside the pre-split `duckdb_conn_test.go` so both commits compile independently; b42ffd2 is then a pure move on top.

### Verification after both fixes

- `make test` — no failures; `make lint` clean.
- Full `make test-e2e-production` re-run: green, 130s, 0 failures.

iceboundrock:
## Findings

No correctness issues found. The two issues flagged in the previous review round are resolved:

1. **Timeout/cancellation regression** (previous high): `MakeConnInit` now accepts an `initTimeout` parameter (`DefaultInitTimeout = 5s`) and wraps init statements in `context.WithTimeout`. Both constructors pass `duckdbinit.DefaultInitTimeout`, matching the 5s PingContext bound. The regression test `TestMakeConnInit_BlockedStmtObservesCancellation` proves a blocked statement observes cancellation (verified: PASS, 0.05s).

2. **Test file size** (previous standards issue): `duckdb_conn_test.go` reduced from 531 to 372 lines; connection-init tests moved to `duckdb_conn_init_test.go` (173 lines). Both under the 500-line limit.

iceboundrock:
## Spec

Issue #285 requirements are fully implemented:

- Per-physical-connection DuckDB initialization via `duckdb.NewConnector` + connector init hook: **done** (`MakeConnInit`, both callers).
- Exporter pool bounded (`SetMaxOpenConns(1)`): **done**, matching YAGNI decision.
- Credential validation fail-fast before connection open: **done** (`ValidateS3Credential` in `buildExporterInitSteps`).
- Construction-time PingContext fail-fast for broken DuckDBPath: **done**.
- Federated migration zero-behavior-change, federated suite as regression gate: **done**.
- `ATTACH`/`PRAGMA` batch within export `ExecContext` unaffected: **confirmed** (per-export batch unchanged).
- Init timeout now properly bounded (`DefaultInitTimeout` → `context.WithTimeout`): the one previously-missing requirement is closed.

iceboundrock:
## Standards

No violations. Previously noted `duckdb_conn_test.go` 531-line violation (`coding-standard.md` §1) is resolved: the file is now 372 lines; init tests are in the new 173-line `duckdb_conn_init_test.go`. All other source files in this diff are well under the 500-line limit (`duckdbinit.go`: 94, `duckdbinit_test.go`: 92, `duckdb_exporter_init.go`: 78, `duckdb_exporter_init_test.go`: 118, `duckdb_exporter.go`: net reduction). Error wrapping follows `coding-standard.md` §3 (context via `fmt.Errorf(...: %w)`). Early returns, flat structure, clear naming.

iceboundrock:
## Assessment

**Approve.** Both issues from the previous review are resolved:

- The high-severity timeout/cancellation regression is fixed with a bounded `DefaultInitTimeout` + regression test.
- The 500-line file size violation is resolved by splitting init tests to a dedicated file.

Verified clean: `make test` (34 packages), `make lint` (clean), focused packages, and the blocking-statement cancellation regression test pass.

iceboundrock:
## Observations

- `DefaultInitTimeout = 5s` is a sensible constant shared between both constructors; if a caller needs a different value, the `initTimeout` parameter on `MakeConnInit` supports non-positive to disable the bound.
- `TestMakeConnInit_BlockedStmtObservesCancellation` is well-written: uses a `blockingExecer` fake that blocks until context cancellation, runs the hook in a goroutine with a 50ms timeout, and asserts via a 5s guard channel that the hook returns promptly. Reverse mutation (reverting to `context.Background()`) would correctly hang into this guard, as documented in the commit message.
- `duckdb_conn_init_test.go` cleanly extracts the 5 init-focused tests + `recordingExecer` fake from the main test file; both files are now comfortably under 500 lines.
- No new dead code: `cmd/tools/manifest_reconcile.go:253` `SetMaxOpenConns(1)` pin correctly removed (constructor now self-sets), and the orphaned `validateS3Credential` in `helpers.go` is deleted.

iceboundrock:
## Verification

Executed against PR head (b42ffd2):

- `go test ./internal/duckdbinit ./internal/cdc ./internal/federated ./internal/compaction ./factory -count=1` — 5/5 packages pass
- `make test` — 34/34 packages pass
- `make lint` — clean (golangci-lint v1.64.8)
- `go test ./internal/duckdbinit -run TestMakeConnInit_BlockedStmtObservesCancellation -v -count=1` — PASS (0.05s)
- File size check: `duckdb_conn_test.go` 372 lines, `duckdb_conn_init_test.go` 173 lines, `duckdbinit.go` 94 lines, `duckdbinit_test.go` 92 lines — all under 500.

Docker-backed production E2E suite not re-run in this review round.

iceboundrock:
## Summary

**0 findings, 0 standards violations.** The two issues from the previous review (high: init timeout regression, standards: 500+ line test file) are both resolved. Full unit test suite (34 packages) and lint pass clean. This PR correctly implements issue #285: per-connection DuckDB initialization with bounded pool, shared `internal/duckdbinit` package, and construction-time fail-fast. Approved.

Issue #329, opened 2026-07-27
cdc: remaining credential-resolution stragglers after #326 (session token, compactor, exporter region key)


iceboundrock:
Deferred from the #326 branch review ledger — three small items adjacent to, but outside, the `resolveStaticS3Credentials` unification:

1. **`resolveExporterSessionToken` is an independent credential input** (`internal/cdc/duckdb_exporter_init.go:43,73`): the session token resolves separately from the key/secret pair, so a config static pair can still combine with an ambient `AWS_SESSION_TOKEN` from the environment — the same cross-source mixing class #326 eliminated for the pair itself. Decide whether the token should ride the same config-wins/both-halves discipline (likely: token adopted only alongside the pair source that supplied it).

2. **`cmd/tools/compactor.go:103` (`resolveMergeCredentials`)** still reads the three AWS env vars per-variable, and can return a half-set triple when the chain fails. This is a deliberate #302-era chain-first fallback for anonymous local stores and lives outside package cdc — but it is the last per-variable resolution site in the repo. If the single-rule ambition extends beyond package cdc, this is the remaining site; otherwise mark it as intentionally divergent with a pointer comment.

3. **`getOrCreateDuckExporter` keys on chain-resolved region but configures with raw config region** (`internal/cdc/runner.go:232-252`): `duckExporterKey.region` derives from `s3Runtime.region` (chain-resolved when `S3Region` is unset) while `NewDuckExporter` receives raw `cfg` whose empty `S3Region` suppresses the `SET s3_region` statement. Within one process the ambient chain region is stable so there is no observable divergence today; recorded so the mismatch does not surprise a future cache-key change.

Also carried from the #326 ledger (cosmetic): a `stubNewS3Client(t, fn)` companion test helper would unify the seam-restore idioms in `runner_test.go`; `factory_entity_manager_unit_test.go` (341 lines) hosts the 4 circuit-breaker tests under a `NewEntityManagerWithConfig` banner — loose fit, future seam.

Origin: #326 implementation (final whole-branch review + per-task review ledger).

iceboundrock:
Two more cosmetic items from PR #330 review (both declined in-PR for scope/pure-move reasons, tracked here):

[...]

PR #330, opened 2026-07-27
fix(cdc): #326 both-halves credential rule + WithRegion for the Runner S3 runtime (third site from #302)

### LINKED ISSUE #326: cdc/runner.go: env credential half-pair + unconditional region clobber on the memoized Runner S3 runtime (third site deferred from #302)

iceboundrock:
Closes #326

## Summary

The CDC Runner daemon (`internal/cdc/runner.go`) was the third AWS S3 client site carrying the two bugs #302 fixed in `internal/bootstrap/s3.go` and `internal/cdc/flusher.go setupAWSClient` — per-variable env credential fallback (half-pair → empty-secret static provider; config key could cross-mix with an env secret) and a `"us-east-1"` region default assigned unconditionally after load, clobbering the chain-resolved region — aggravated by the resolved pair being memoized in `r.s3Runtimes` for the Runner's lifetime.

- **Seam consolidation** (`65b600e`): retired the old fixed-signature `loadAWSConfigFn`; package cdc now has exactly one AWS-config seam (the variadic `loadAWSConfig`), so `WithRegion` can be carried and tests can no longer stub the wrong seam silently.
- **Single credential rule** (`fef8b03`): new `resolveStaticS3Credentials(cfg)` (`internal/cdc/aws_credentials.go`) — config pair as-is when `S3AccessKeyID` is set; else env pair only when BOTH halves are non-empty; else empty pair → SDK default chain. Runner wired to it; buggy `resolveS3Credentials` deleted. Because the resolved pair is part of both memoization keys, a fixed configuration can never be served a previously-poisoned cached runtime.
- **Region** (`0345d69`): `config.WithRegion(cfg.S3Region)` at load only when configured; hardcoded default and post-load clobber removed. `s3RuntimeKey` keys on raw `cfg.S3Region`; the cached runtime stores the chain-resolved region.
- **All five resolution points unified** (`f578fa5`, `bf5812c`): `setupAWSClient` (pure refactor, pinned by existing tests), `RunOnce`'s DuckDB httpfs plumbing, and — found by the final whole-branch review as a fifth, previously unenumerated site — `newInitRunContext` in `init.go` (cdc-init bulk export). `os.Getenv` of the AWS pair now exists in exactly one place in package cdc.
- **Doc fix** (`eec7cb2`): `applyCredentials` comment now says the chain survives "when no fully-set pair is available" (was "when neither is set" — wrong for half-pairs).
- **e2e script** (`b8db62b`): `federated-check.ts` warns when `--manifest-reads` is passed without `--require-duckdb` (the flag only takes effect on the forced-DuckDB route — previously a silent no-op that read as manifest coverage).
- **factory_test.go split** (`7aa8a2f`): 896 lines → 4 files (285/65/341/238) along the seams named in the PR #327 review comment; pure move — byte-identity and 23/23 test-name parity verified in review; `factory_close_test.go` untouched.

## Behavior changes (operator-visible)

1. **Env half-pair now falls to the default chain** on every cdc path (Runner, flusher, cdc-init). Previously a lone `AWS_ACCESS_KEY_ID` built an empty-secret static provider whose only symptom was an opaque signing failure. DuckDB httpfs under a half-pair now receives no `SET s3_access_key_id`/`s3_secret_access_key` at all (empty values are skipped) and inherits its own environment chain — SDK client and httpfs no longer diverge.
2. **Unset `S3Region` on the Runner path now uses the chain-resolved region** (previously always `us-east-1`), matching the `RunOnce`/flusher path. With `PG_USE_IAM`, token generation against a chain-less environment logs a WARN and falls back to `PG_PASSWORD` (`setupPostgresConnection` swallows token errors) — it does not fail loudly.
3. **Config key with empty config secret no longer cross-mixes with an ambient env secret.** The config pair passes through as-is, so DuckDB now gets `SET s3_access_key_id` with no secret (signs and fails) instead of accidentally borrowing `AWS_SECRET_ACCESS_KEY` — this mirrors the long-standing `setupAWSClient` static-provider behavior, which is the parity #326 exists to create.

## Adjudication record

- Both-halves rule unified across all resolution points: user-approved 2026-07-27 (originally scoped to the four points the issue enumerated; extended to the fifth point in `init.go` discovered during the final review, under the same single-truth-source principle — flagging explicitly for reviewer attention).
- `EntityManager.Close` doc nit from the issue body is **superseded**: merged PR #327 added `closeOnce`, making concurrent `Close` genuinely safe, and `storage.go` / `internal/entity_manager.go` already document exactly that. No change made.
- The "pointer comment at `resolveS3Credentials` marking the divergence known-unfixed" item dissolved — the function was deleted and the divergence fixed.
- Deferred stragglers (session-token independent input, `compactor.go` chain-first fallback, duck-exporter region key vs config mismatch, cosmetic test-helper items) are tracked in follow-up #329.

## Verification

- `make lint` clean (golangci-lint v1.64.8); `make test` green (31 packages).
- `make test-e2e-production` green twice: on `7aa8a2f` (129s) and on the final head `bf5812c` (135s) — the production harness flush path (`internal/e2e_harness/production/cdc.go`) rides the fixed Runner path.
- TDD red-first evidence recorded for the credential and region fixes (half-pair test failed with `accessKeyID = "env-key"`; region tests failed with `loadedRegion=""` / `runtime.region="us-east-1"` before the fix).
- factory split: `go test -list` parity before == after (44 names), moved blocks byte-identical vs `git show` of the pre-split file.


iceboundrock:
## Standards

Against AGENTS.md and coding-standard.md:

- All changed Go files are below the 500-line limit; inspected changed functions are below 100 lines.
- No new duplication, deep nesting, SRP, or test-coverage standards issue was found.
- Two documented-style exceptions remain: the unwrapped test-stub error in internal/cdc/runner_test.go:168-170 (AGENTS.md:69; coding-standard.md Error Handling) and the two non-verb-first private helpers in factory/factory_entity_manager_unit_test.go:33,47 (coding-standard.md Naming Conventions).

These are non-blocking cleanup items.

iceboundrock:
## Findings

1. [P3, non-blocking] internal/cdc/runner_test.go:168-170 returns an AWS load-option error without context. Wrap it (for example, `fmt.Errorf("apply AWS load option: %w", err)`) to comply with the project error-handling rule.
2. [P3, non-blocking] factory/factory_entity_manager_unit_test.go:33 and :47 introduce private helper names `unitEntityManagerDeps` and `unitEntityManagerConfig`. The documented private-helper convention is verb-first; rename to names such as `buildUnitEntityManagerDeps` and `newUnitEntityManagerConfig`.
3. [P3, scope] The factory test-file split is unrelated to #326’s Runner S3 behavior. It is a pure move and was motivated by prior review feedback, but it accounts for most of this PR’s diff and would be easier to review independently.

No blocking functional defect was found in the #326 credential or region change.

iceboundrock:
## Spec

#326 requirements are implemented:

- Static environment credentials are selected only when both AWS key and secret are non-empty; a half-pair returns an empty pair and leaves the SDK default chain intact.
- Explicit configuration still wins as a pair, so a configured key cannot combine with an ambient environment secret.
- The Runner supplies an explicit region through WithRegion during AWS config loading and preserves the chain-resolved region when none is configured.
- The resolved credential pair remains in both cache keys.
- The duplicate Runner AWS-config seam was retired, the shared credential resolver now serves SDK and DuckDB paths, the bootstrap comment was corrected, and the manifest-read warning was added.

No missing, partial, or incorrectly implemented #326 requirement was found. The factory test split is out of the issue scope; see Findings.

iceboundrock:
## Summary

Re-review of #330 against #326: no blocking functional issue found. The credential-resolution, region-loading, cache-key, shared-seam, documentation, and CLI-warning requirements are satisfied.

Three P3/non-blocking items were recorded: contextual wrapping in one test stub, verb-first names for two new private test helpers, and the unrelated factory test-file split expanding scope. Verification: diff check, lint, full unit suite, and focused CDC/factory tests passed.

iceboundrock:
## Observations

- Credential environment reads are centralized in internal/cdc/aws_credentials.go, avoiding per-call-site half-pair behavior.
- The Runner’s runtime cache key intentionally uses configured S3Region, while the cached runtime retains the region resolved by the AWS config chain. That preserves reuse for a stable configuration without reviving the former us-east-1 default.
- The new Runner tests cover fully-set environment credentials, half-pair fallback, configured-region load options, and an unset-region chain result.
- The factory test split appears byte-move-only from this review perspective, but it broadens this PR’s review surface without changing #326 behavior.

iceboundrock:
## Assessment

The operational regression described in #326 is addressed: the Runner no longer constructs a poisoned static provider from an environment half-pair, and an unset S3Region no longer overwrites the AWS chain’s resolved region with us-east-1. Cache identity tracks the resolved static pair, preventing reuse of a prior credential shape.

Assessment: functionally acceptable. The three findings are P3/non-blocking; two are small standards fixes and one is review-scope hygiene. No approval or merge action is taken by this comment.

iceboundrock:
## Verification

Reviewed main...bf5812c2bfe6e5a86941658a5f011a293367e1c4 (merge base dd7ec7bad7d47afe6aa9553946b1b4c996beb5ca).

Independently run on the PR head:

- git diff --check: clean.
- make test: passed across all listed unit-test packages.
- make lint: passed.
- go test ./internal/cdc ./factory: passed.

Production E2E was not independently rerun in this review; the PR description records two successful runs.

iceboundrock:

**Finding 1 (wrap AWS load-option error, `runner_test.go:168-170`)** — Fixed in `ca99945`: the stub now returns `fmt.Errorf("apply AWS load option: %w", err)`. Focused runner tests, `gofmt`, and `make lint` re-verified. Note: `flusher_aws_test.go`'s `TestSetupAWSClient_RegionPassedAtLoad` (pre-existing on main, untouched by this PR) carries the identical bare return; wrapping it here would add a file to the diff, so it is tracked in #329 to keep the mirrored stubs consistent in one change.

**Finding 2 (verb-first renames for `unitEntityManagerDeps` / `unitEntityManagerConfig`)** — Not applied in this PR; premise correction: these names are not introduced here. They exist on main at `factory_test.go:214,228` (dd7ec7b) and were moved byte-identically as part of the pure-move split — the same byte-identity this PR's per-task review relied on to verify no silent edits. Their call sites also span `schema_validator_test.go`, `parquet_source_test.go`, and `factory_close_test.go`, none of which this PR touches, so renaming would expand the diff into three unrelated files — directly against Finding 3's scope concern. Tracked in #329 with the suggested names for an all-sites-at-once rename.

**Finding 3 (factory split expands review surface)** — Acknowledged; no change. Inclusion was adjudicated up front: the split was parked on #326 by its own issue comment ("the next change touching it should carry a split"), and the pre-implementation plan review explicitly chose bundling it here over deferring. Mitigations for reviewability: it is an isolated pure-move commit (`7aa8a2f`) with verified byte-identity and 23/23 test-name parity, so it can be reviewed (or reverted) independently of the seven behavior commits. If independent review is still preferred, it can be dropped from this branch and re-raised as a standalone PR — say the word.

Verification on the new head `ca99945`: `make lint` clean, `go test ./internal/cdc -run TestRunner` all green, `gofmt -l internal/cdc` shows only the pre-existing `redact_test.go` offender.

Issue #331, opened 2026-07-27
cdc: Runner S3-runtime/exporter caches never evict — each credential rotation strands an open DuckDB instance


iceboundrock:
Deferred from the #329 final whole-branch review (O2 in the review ledger).

`Runner.s3Runtimes` and `Runner.duckExporters` (`internal/cdc/runner.go:33-34`) are append-only maps: entries are only drained by `Runner.Close()` at shutdown. Every change in a cache-key field mints a new entry while the old one stays alive for process lifetime.

#329 made the keys honest — both now include the resolved static credential triple (`accessKeyID`, `secretAccessKey`, and since #329 `sessionToken`). That is required for correctness (a rotated `AWS_SESSION_TOKEN` under an unchanged env pair must not hit a stale-token artifact), but it raises key cardinality under credential rotation:

- A long-lived `Runner` under hourly STS rotation strands ~24 open DuckDB instances (and 24 cached S3 runtimes/clients) per day.
- Each stranded `DuckExporter` holds an open DuckDB handle (`SetMaxOpenConns(1)` per #285), so the leak is real memory + file handles, not just map entries.

The #329 red-anchor test even names the failure mode in its comment ("building — and leaking — a second identical DuckDB instance"). #329 removed one *spurious* source of new keys (chain-resolved region, now raw `cfg.S3Region`) but the unbounded growth underneath is untouched — it predates #329.

Candidate directions (not adjudicated):
1. Evict-on-supersede: when a new key differs from an existing entry only in credential fields, close and replace the old entry (needs care re: in-flight use — refcount or drain).
2. Bounded LRU with close-on-evict.
3. Accept-and-document: production Lambda runners are short-lived; long-lived deployments are the only exposure. If chosen, record it in the Runner doc comment.

Origin: #329 implementation (final whole-branch review; per-task review of the cache-key task).

iceboundrock:
**Implementation plan** (produced 2026-08-18; working copy at `docs/superpowers/plans/2026-08-18-331-runner-cache-eviction.md`, gitignored). User adjudication recorded inside: direction 1 (evict-on-supersede + refcount) chosen over LRU and accept-and-document.

---

# #331 Runner Cache Evict-on-Supersede Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

[...]

PR #332, opened 2026-07-27
fix(cdc): #329 session token rides the pair source; unify cmd/tools onto the single credential rule; honest exporter cache keys

### LINKED ISSUE #329: cdc: remaining credential-resolution stragglers after #326 (session token, compactor, exporter region key)

iceboundrock:
Closes #329.

## What this does

Finishes the credential-resolution stragglers left after #326 (PR #330), per the issue's three items plus the ledger/comment cosmetics:

1. **Session token rides the pair source** (`internal/cdc/aws_credentials.go`). `ResolveStaticS3Credentials` (now exported) returns a `(key, secret, token)` triple: a config pair carries `cfg.S3SessionToken`, an env pair carries `AWS_SESSION_TOKEN`, and the token never crosses sources — a config pair no longer adopts an ambient env token (the cross-source mixing class #326 eliminated for the pair itself). A token with no pair is not a credential source. `resolveExporterSessionToken` is retired; the token is now an explicit `NewDuckExporter` parameter so the DuckDB `SET s3_session_token` can only ever come from the same resolution as the pair.
   - This also fixes a latent SDK-side bug: all static providers previously hard-coded an empty token (`NewStaticCredentialsProvider(key, secret, "")`), so env-pair STS temporary credentials signed without their session token. Both SDK sites (flusher `setupAWSClient`, runner `getOrCreateS3Runtime`) now carry the resolved token.

2. **`cmd/tools` fully unified onto the single rule** (user ruling). `resolveMergeCredentials` (the last per-variable env-var resolution site, able to return a half-set triple) is deleted; the compactor merge engine and the manifest-reconcile stats engine now call `cdc.ResolveStaticS3Credentials`.
   - ⚠️ **Deliberate behavior narrowing**: chain-only credential sources (IMDS/instance and container roles, SSO, assumed roles, web identity, shared profiles) no longer feed the engines' DuckDB httpfs `SET` statements. The tools expose no `--s3-access-key` flags, so the engines are env-pair-or-nothing in practice. Lambda-style env triples are unaffected — the token now rides the env pair. The failure shape on e.g. instance-role hosts is asymmetric: `buildToolS3Client` still walks the full chain, so manifest reads succeed while httpfs parquet IO goes unsigned (reads like an S3 permission error). A WARNING doc comment on `openMergeEngine` records all of this, and both engines now `logger.Warn` when they come up with no static credentials. Restoring chain support later means adding credential flags, not reinstating a second resolution rule.

3. **Honest exporter cache keys** (`internal/cdc/runner.go`). `duckExporterKey.region` now uses raw `cfg.S3Region` (what actually configures the exporter — empty suppresses `SET s3_region`) instead of the chain-resolved SDK region; the old key could produce a latent wrong cache hit (empty-region cfg colliding with an explicit-region cfg). Both `s3RuntimeKey` and `duckExporterKey` now include the resolved `sessionToken`, so an `AWS_SESSION_TOKEN` rotation under an unchanged pair no longer returns stale-token artifacts.

4. **Cosmetics from the #326 ledger / #330 review comments**: `stubNewS3Client(t, fn)` helper unifies 7 manual seam save/restores in `runner_test.go`; `unitEntityManagerDeps`/`unitEntityManagerConfig` → `buildUnitEntityManagerDeps`/`newUnitEntityManagerConfig` across all 4 factory test files (22 refs); the 4 `TestNewDuckDBCircuitBreaker*` tests moved byte-faithfully into `factory/factory_circuit_breaker_unit_test.go`; the flusher region-stub error now wraps identically to its runner twin (ca99945).

## Verification

- `make test` green across all packages; new/updated coverage: 14-case resolver table (mutation-verified — each token-discipline direction kills a distinct injected mutant), SDK-provider token tests (flusher + runner), exporter-builder discipline tests, `cmd/tools` chain-not-consulted + positive credential-wiring tests, cache-key red anchors (region + token, both s3Runtime and exporter).
- `go vet -tags e2e ./internal/e2e_harness/production/` clean — the only gate that compiles the e2e-tagged `NewDuckExporter` caller.
- `make lint` (golangci-lint v1.64.8) exit 0; `grep` shows zero residue of `resolveMergeCredentials`/`resolveExporterSessionToken`/the old unexported resolver.
- T9 move verified byte-faithful (`git show` range diff = zero).

## Follow-up

- #331 — Runner S3-runtime/exporter caches never evict; with the token now in the keys, credential rotation strands open DuckDB instances in long-lived runners (pre-existing class, aggravated by the honest keys).

### DIFF
--- a/cmd/tools/compactor.go
+++ b/cmd/tools/compactor.go
@@ -120,7 +120,7 @@ func openMergeEngine(ctx context.Context, opts *compactorOptions, logger *zap.Lo
 		S3UsePath:               opts.s3.usePath,
 		S3SessionToken:          token,
 	}
-	exporter, err := cdc.NewDuckExporter(ctx, duckCfg, key, secret, logger)
+	exporter, err := cdc.NewDuckExporter(ctx, duckCfg, key, secret, token, logger)
 	if err != nil {
 		return nil, fmt.Errorf("open merge duckdb: %w", err)
 	}
--- a/cmd/tools/compactor.go
+++ b/cmd/tools/compactor.go
@@ -5,10 +5,8 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"os"
 	"time"
 
-	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/lychee-technology/forma/internal/cdc"
 	"github.com/lychee-technology/forma/internal/compaction"
 	"go.uber.org/zap"
@@ -85,27 +83,28 @@ func parseCompactorFlags(args []string) (*compactorOptions, error) {
 	return opts, nil
 }
 
-// resolveMergeCredentials resolves the FULL default AWS credential chain
-// (env, shared profiles, assumed roles, web identity, IMDS) so DuckDB signs
-// with the same identity as the SDK S3 client — including the session token
-// temporary credentials require. Falls back to raw env vars when the chain
-// yields nothing (e.g. anonymous local object stores).
-func resolveMergeCredentials(ctx context.Context, region string, logger *zap.Logger) (key, secret, token string) {
-	awsCfg, err := toolLoadAWSConfigFn(ctx, awsconfig.WithRegion(region))
-	if err == nil {
-		if creds, retrieveErr := awsCfg.Credentials.Retrieve(ctx); retrieveErr == nil {
-			return creds.AccessKeyID, creds.SecretAccessKey, creds.SessionToken
-		} else {
-			err = retrieveErr
-		}
-	}
-	logger.Warn("could not resolve AWS credential chain for the merge engine; falling back to env vars", zap.Error(err))
-	return os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), os.Getenv("AWS_SESSION_TOKEN")
-}
-
 // openMergeEngine builds the CDC DuckDB exporter the rewrite merges through.
+// Credentials come from the single shared rule cdc.ResolveStaticS3Credentials
+// (#329): explicit config wins, otherwise the AWS_ACCESS_KEY_ID /
+// AWS_SECRET_ACCESS_KEY pair, with AWS_SESSION_TOKEN riding whichever source
+// supplied that pair.
+//
+// WARNING — this is a deliberate narrowing. The merge engine previously
+// resolved the FULL AWS default credential chain, so chain-only sources
+// (shared profiles, assumed roles, web identity, IMDS/container roles) reached
+// the DuckDB httpfs SET statements. They no longer do: an operator whose
+// credentials live only in ~/.aws/config or an instance role now gets an
+// engine with no S3 credentials at all, and DuckDB falls back to whatever its
+// own environment chain provides. The compactor exposes no credential flags,
+// so in practice the contract is env-pair-or-nothing.
+//
+// Lambda and container deployments that already export the environment triple
+// are unaffected — the environment is exactly what the shared rule reads. The
+// correct way to restore chain support is to add explicit credential flags
+// feeding cdc.CDCConfig.S3AccessKeyID/S3SecretAccessKey/S3SessionToken, not to
+// reintroduce a second private resolver: two credential rules signing the same
+// bucket differently is the failure mode #329 retires.
 func openMergeEngine(ctx context.Context, opts *compactorOptions, logger *zap.Logger) (*cdc.DuckExporter, error) {
-	key, secret, token := resolveMergeCredentials(ctx, opts.s3.region, logger)
 	duckCfg := cdc.CDCConfig{
 		DuckDBPath:              opts.duck.duckDBPath,
 		DuckThreads:             opts.duck.duckThreads,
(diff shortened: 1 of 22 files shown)


iceboundrock:
## Findings

No untracked blocking correctness findings.

- **P3, standards:** `internal/cdc/runner.go:157` returns the `getOrCreateDuckExporter` error unchanged. This preserves a single `new duck exporter` prefix, but conflicts with the strict no-bare-error rule in `AGENTS.md:69` and `coding-standard.md:22-32`; the same edited `RunOnce` also retains bare propagations at lines 124, 130, and 148. This is non-blocking because the callee already supplies context, but it remains a documented-standard deviation. Consider defining the single contextual boundary explicitly, or add context here while moving the lower-level prefix.

The apparent tool credential-chain regression is **not** a finding: issue #329 explicitly requires the deliberate env-pair-or-nothing narrowing, and the new `openMergeEngine` warning documents it.

## Spec

Pass. Issue #329 and the PR implementation plan require: same-source session-token resolution, `cmd/tools` use of `cdc.ResolveStaticS3Credentials`, raw `cfg.S3Region` for exporter cache identity, and the listed test cosmetics. The diff implements each requirement without identified scope creep.

## Standards

- File/function size limits: pass for changed Go code.
- Error context: one non-blocking deviation noted above.
- Naming/duplication: the requested verb-first fixture cleanup is applied; no mandatory duplication violation identified.

## Assessment

No correctness or specification blocker found in this review. I am not submitting an approval or merging the PR.

## Observations

- The config-wins/both-halves resolver now keeps the session token with the selected pair, and both AWS SDK static providers plus DuckDB receive that same token.
- Tool-side chain-only credentials are intentionally no longer configured in DuckDB httpfs; the inline warning names the affected sources and recovery path.
- Token-aware cache keys fix stale-token reuse. The resulting append-only cache growth on credential rotation is a real pre-existing lifecycle concern, transparently tracked in open issue #331; it is not an untracked #332 regression.

## Verification

- `git diff --check origin/main...origin/pr-332` completed with no whitespace errors.
- Local: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/cdc ./cmd/tools -count=1` passed.
- Local: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go vet -tags e2e ./internal/e2e_harness/production/` passed.
- GitHub checks currently pass: Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

No blocking findings; one non-blocking strict error-wrapping standards deviation in `Runner.RunOnce`. #332 satisfies #329 as specified, with the cache-lifecycle follow-up recorded in #331.

iceboundrock:
Response to the review:

**P3 (RunOnce bare error propagations) — fixed in 3bc2d26.** All four pass-through sites in `Runner.RunOnce` now wrap with distinct run-level context: `prepare s3 runtime:` / `resolve s3 clients:` / `setup postgres connection:` / `prepare duck exporter:`. Prefixes were chosen to never duplicate the callee-supplied ones (`load aws config:`, `new duck exporter:`, etc. — verified no collisions and no test asserts these strings), so each layer adds exactly one step of context, satisfying the strict rule without double-prefixing. The now-obsolete pass-through comment was replaced with one explaining the two-layer split. `gofmt` clean, `go test ./internal/cdc ./cdc -count=1` green, `go vet` clean.

Note: `resolveS3Clients`/`setupPostgresConnection` callers in the flusher path (`flusher.go` `runFlush`) have the same pre-existing pass-through shape but were not part of this branch's edits — left untouched per the fix-the-immediate-area rule; can sweep them in a follow-up if desired.

**Non-findings, acknowledged as read:**
- Tool credential-chain narrowing correctly treated as spec, not regression — that matches the #329 ruling and the `openMergeEngine` WARNING.
- Cache growth on rotation → tracked in #331, agreed it is pre-existing and not a #332 regression.

No other open items from the review.

PR #307, opened 2026-07-25
fix(httpapi): #301 redact operator detail from public HTTP error bodies

iceboundrock:
Closes #301.

## What this fixes, and why it is wider than the issue

#301 recorded that `ParquetSetInconsistentError.MissingKeys` — bucket-relative S3 object keys — reaches API clients through `internal/httpapi/server.go`, which echoes engine errors verbatim into 5xx bodies.

Auditing that path turned up something sharper. The federated template interpolates `postgres_scan('{{.PG_CONN}}', …)`, and `DuckDBPostgresConnStringFromPool` (`internal/federated/engine.go:477`) builds `PG_CONN` as `host=… user=… password=… dbname=…`. When DuckDB cannot attach, its own message quotes that string back:

```
IO Error: Unable to connect to Postgres at "host=db.internal port=5432
user=forma password=SUPERSECRET dbname=forma": could not translate host name …
```

Verified empirically against `duckdb-go v2.5.6`. That text reached `writeError(w, classifyManagerError(err), fmt.Sprintf("query failed: %v", err))` unmodified, so **the Postgres password was reachable on the public HTTP surface**, not just object keys. A missing parquet object likewise yielded the full `s3://bucket/prefix/key` and the resolved endpoint URL.

This is why redaction had to be an **allowlist**: the worst leak originates in driver-authored text, not in a Forma error type, so no blocklist of known-sensitive carriers could have caught it.

Two supporting findings shaped the design:

1. **There was no error logging to relocate detail into.** Handlers logged requests (`zap.S().Infow`) but every error path only wrote the body — zero `Errorw` in the package. Redacting without adding logging would have destroyed operator visibility rather than moved it.
2. **The success path was already fixed.** `toExecutionPlan` (`internal/entity_query_service.go:143`) carries a `// SECURITY:` allowlist dropping `SQL`/`Params`/`Notes` for exactly this reason, from #259's P0. This is its error-path counterpart.

## The contract

Disclosure is decided by the **error**, not by the status:

```go
if status < http.StatusInternalServerError && isClientError(err) {  // verbatim
```

`isClientError` is `errors.Is` against `forma.ErrInvalidInput` / `ErrNotFound` / `ErrConflict`. Everything else is redacted whatever its status, and **every redacted response logs at `Errorw`** under an `error_id` the client can quote back.

```json
{
  "success": false,
  "error": "internal read error",
  "error_class": "parquet_set_inconsistent",
[...]

PR #334, opened 2026-07-28
docs: #309 "message text unchanged" is imprecise — prefix preserved, sentinel suffix appended

### LINKED ISSUE #309: "message text unchanged" is imprecise: %w of ErrInvalidInput appends ": invalid input"
Follow-up from #307 (issue #301).

#307 wrapped six caller-input validators with `forma.ErrInvalidInput` so they keep answering 4xx after the substring-matching classifier was deleted. The selling point recorded in three places is that this happens **with the message text unchanged**:

- `internal/httpapi/error_response.go:118`
- `docs/error-handling.md:168`
- `internal/entity_query_sort.go:88` — "The message text is unchanged on purpose"

That is imprecise. Wrapping with `%w` of `forma.ErrInvalidInput` appends `: invalid input` to the rendered string, so the **prefix** is unchanged but the **full body text** is not. Observed during #307's implementation as `unsupported operator: equals: invalid input` in characterization tests.

Existing tests use `Contains`, so nothing broke — but a client asserting on the exact full string does. And `docs/error-handling.md` already shows the effect in its own worked example (`…: schema not found: nosuchschema: not found`), so the document contradicts itself.

**Fix:** correct the three claims to say the human-authored prefix is preserved and the sentinel suffix is appended. Decide separately whether the appended `: invalid input` is wanted in public bodies at all — if not, that is a message-construction change, not a doc change, and should be scoped as such.

iceboundrock:
Closes #309.

## What

#307 recorded in three places that the `forma.ErrInvalidInput` sweep left "message text unchanged". That is imprecise: `%w` wrapping appends the sentinel's own text, so the rendered string is the human-authored **prefix** plus `: invalid input` (observed as `unsupported operator: equals: invalid input` in #307's characterization tests). `docs/error-handling.md`'s own worked example (`…schema not found: nosuchschema: not found`) already showed the effect, so the document contradicted itself.

## Changes (doc/comment only, zero behavior change)

Five claim sites corrected — the issue lists three; a fourth with the same wording landed after it was filed (in #307's operator-whitelist underreach fix), and a fifth was caught by review round 1:

- `internal/httpapi/error_response.go` (`classifyManagerError` comment)
- `internal/entity_query_sort.go` (`buildAttributeOrders` comment)
- `docs/error-handling.md` — sweep-table intro
- `docs/error-handling.md` — operator-whitelist underreach paragraph
- `docs/superpowers/plans/2026-07-25-issue-301-redact-http-error-bodies.md` — the "Superseded by code review" preface, which describes shipped state, not plan history (review round 1, P2)

All now say the human-authored message prefix is preserved and the `%w` rendering appends the sentinel suffix.

## Decision recorded

The issue asked for a separate decision on whether the appended `: invalid input` is wanted in public bodies at all. **Decision: keep the suffix** (owner ruling, 2026-07-27). A new paragraph in `docs/error-handling.md` records it as part of the body contract: clients must not assert on the exact full message — match the prefix (this repo's tests use `Contains`) or key on the status code. Stripping it would be a message-construction change to every wrap site and is explicitly not pursued.

## Verification

- `grep -rn -i "message text unchanged\|text is unchanged" --include="*.go" --include="*.md" .` with **no path exclusions** → zero matches (round 1 of this check wrongly excluded `docs/superpowers/`, which is what let the plan-preface claim through)
- `make lint` clean; `make test` 31 packages ok, zero FAIL
- The new paragraph's cross-reference ("Accepted disclosures inside the allowlist") verified against the actual section heading above the `nosuchschema` example

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -441,7 +441,8 @@ errors that wrapped no sentinel. It was removed for two reasons:
 The consequence is that a genuine client error earns its 4xx only by wrapping a
 sentinel. Removing the heuristic therefore required a sweep of the sites that had
 been relying on it — every one of them now wraps `forma.ErrInvalidInput`, with
-message text unchanged:
+the human-authored message prefix unchanged (the rendered string gains a
+sentinel suffix — see below):
 
 | site | caller mistake |
 | --- | --- |
@@ -456,6 +457,19 @@ attribute would answer `500` with an opaque body instead of naming the attribute
 The `sqlgen`/`conditionexpr` group is reachable through `POST
 /api/v1/advanced_query`, whose `condition` payload is entirely caller-supplied.
 
+**The rendered body is the prefix plus a sentinel suffix, and the suffix is
+deliberate (#309).** Wrapping with `%w` renders the sentinel's own text, so a
+verbatim 4xx body ends in `: invalid input` (likewise `: not found` and
+`: conflict` for the other two sentinels) — e.g.
+`unsupported operator: equals: invalid input`, or the
+`…: schema not found: nosuchschema: not found` body in the worked example under
+"Accepted disclosures inside the allowlist" below. What the sweep preserved is
+the human-authored prefix, not the exact full string. The suffix is kept as part
+of the body contract: clients must not assert on the exact full message — match
+the prefix (this repo's tests use `Contains`) or, better, key on the status
+code. Stripping the suffix would be a message-construction change to every wrap
+site, and #309 decided against it.
+
 Since #314 there is a **second** write-path validator on the same footing,
 `internal/schemavalidate`'s `Validator.Validate`, which wraps
 `forma.ErrInvalidInput` for a JSON Schema violation — `enum`, `pattern`, `type`,
@@ -468,8 +482,8 @@ enforcement on write" for the split.
 `normalizePgEavPayload`'s operator whitelist — the two rejections that pair an
 operator with an attribute type it does not accept — so `starts_with` on a UUID
 column answered an opaque `500` while the "condition-DSL errors stay 400" claim
-below said otherwise. Both now wrap the sentinel, message text unchanged, and are
-pinned by the `clientError` column of
+below said otherwise. Both now wrap the sentinel — prefix unchanged, same
+appended suffix as the sweep above — and are pinned by the `clientError` column of
 `TestToDualClauses_Characterization_Errors` plus
 `TestAdvancedQueryOperatorWhitelistIs400AndVerbatim`.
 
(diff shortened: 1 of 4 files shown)


iceboundrock:
## Findings

- **[P2] Correct the retained #301 plan's current-state summary**: `docs/superpowers/plans/2026-07-25-issue-301-redact-http-error-bodies.md:27` still says all validators wrap `forma.ErrInvalidInput` "with their message text unchanged." Although the file is a historical plan, its prefatory summary explicitly describes what currently shipped, so this remains the same inaccurate full-message claim corrected elsewhere. Replace it with the prefix-preserved / `: invalid input` suffix wording. This also means the PR description's zero-match grep verification is not currently true.

## Spec

- **Partial.** The three locations named by #309, plus the additional operator-whitelist statement, now correctly distinguish the human-authored prefix from the rendered sentinel suffix. However, #309's goal is to eliminate the contradictory claim; the retained #301 current-state summary still makes it.

## Standards

- No documented coding-standard violations found. The Go changes are comments only and remain within the repository's file/function limits; no error-handling, layering, or API behavior changed.

## Assessment

- **Changes requested.** This is a small documentation-consistency fix. Once the remaining current-state assertion is corrected, the PR matches the intended decision to retain the suffix.

## Observations

- The diff is documentation/comment-only and does not change runtime behavior.
- The new public-contract paragraph is consistent with `%w` rendering and with the existing `Contains`-based test convention.

## Verification

- `GOCACHE=/var/folders/37/2bpjmt0129z96c4plfy52gkw0000gn/T/opencode/forma-334-review-gocache GOFLAGS=-buildvcs=false make test` passed across all listed packages.
- `rg -n -i "message text unchanged|text is unchanged" --glob "*.go" --glob "*.md" .` found the remaining assertion at `docs/superpowers/plans/2026-07-25-issue-301-redact-http-error-bodies.md:27`.

## Summary

- **1 P2 finding; 0 standards findings.** Please correct the remaining current-state documentation claim before merging.

iceboundrock:
## Review round 1 response

**[P2] Retained #301 plan's current-state summary** — confirmed and fixed in 4a085a7. The claim sits in the plan's "Superseded by code review" preface, which explicitly describes what shipped ("All now wrap `forma.ErrInvalidInput` …") rather than plan history, so it is the same live inaccuracy the PR exists to correct. Reworded to prefix-preserved / `: invalid input` suffix appended, with the #309 reference.

The root cause of the miss: my verification grep deliberately excluded `docs/superpowers/` on the assumption that plan files are frozen history. That assumption is wrong for this file's preface. Rerun with no exclusions:

```
grep -rn -i "message text unchanged|text is unchanged" --include="*.go" --include="*.md" .
```

→ zero matches. The PR body's Verification section is updated accordingly (it also now counts five corrected sites, not four).

PR #335, opened 2026-07-28
refactor: #310 consolidate the three SQL quote-doubler copies into sqlutil.EscapeLiteral

iceboundrock:
Closes #310.

## What

`strings.ReplaceAll(s, "'", "''")` — escaping a value for a single-quoted SQL literal — existed in three independent copies. This PR consolidates the rule into one tested helper, `sqlutil.EscapeLiteral` (`internal/sqlutil/literal.go`), consumed by all three former sites:

- `internal/cdc/duckdb_exporter.go` (was `escapeLiteral`; 5 call sites — ATTACH DSN, S3 path, subqueries)
- `internal/sqlgen/duckdb_template_renderer.go` (was `escapeSQLLiteral`, added by #307; the `postgres_scan('{{.PG_CONN}}', …)` credential path)
- `internal/e2e_harness/federated/query_build.go` (`benchmarkSQLLiteral` string case; the issue cites `query.go:616`, which #220's slicing moved here)

Plus a sweep: every comment/test-case mention of the deleted function names now says `sqlutil.EscapeLiteral` (internal/redact, internal/httpapi, and living doc `docs/error-handling.md` — the latter absent from the plan's file table but authorized by its living-docs clause).

## Decisions (the issue asked for explicit ones)

1. **Helper location**: `internal/sqlutil` — the issue itself pointed there; sits beside `SanitizeIdentifier`.
2. **The e2e harness consumes the shared helper** rather than keeping its copy. The issue allowed keeping it "if importing production code there is undesirable" — that concern is moot: `internal/e2e_harness/federated` already imports `internal/cdc`, `internal/model`, and `internal/federated`, so no new coupling precedent.
3. **#333's carefully-corrected comment survives**: the `escapeSQLLiteral` doc comment moved verbatim-in-substance to its only call site (the PG_CONN injection block), keeping the #301 quoted-DSN rationale, the pre-#301 hole description, the "deployment-configured, not caller-reachable" framing, and the #290 ATTACH cross-reference. Reviewed byte-for-byte against the plan's mandated text.
[...]

Issue #310, opened 2026-07-25
Consolidate the three copies of the SQL quote-doubler


iceboundrock:
Follow-up from #307 (issue #301).

`strings.ReplaceAll(s, "'", "''")` — escaping a value for a single-quoted SQL literal — now exists in three places:

- `internal/cdc/duckdb_exporter.go:159`
- `internal/sqlgen/duckdb_template_renderer.go:366` (added by #307)
- `internal/e2e_harness/federated/query.go:616`

This is mildly at odds with #307's own stated rationale for creating `internal/pgdsn`: that two copies of an escaping rule drift apart. The distinction drawn at the time — order-dependent libpq quoting is risky, a one-line `ReplaceAll` is not — is an argument about *risk*, not about duplication, and is not worth defending hard.

**Context that raises the stakes:** #307 established that this escaping is load-bearing. The `postgres_scan('{{.PG_CONN}}', …)` literal previously had **no** escaping at all, so a Postgres password containing `'` went raw into SQL structure. A shared, tested helper is cheaper to reason about than three independent `ReplaceAll` calls, one of which was missing entirely until #307.

**Suggested:** a single `sqlutil`-level helper (that package already exists) consumed by all three. The e2e harness copy can stay if importing production code there is undesirable — decide explicitly rather than by omission.

iceboundrock:
Closed by PR #335 (squash-merged as 9ffb747, 2026-07-30).

Outcome: the rule now lives solely in `internal/sqlutil.EscapeLiteral`; all three copies (cdc `escapeLiteral`, sqlgen `escapeSQLLiteral`, harness inline `ReplaceAll`) are deleted, stale name mentions swept (code, test names, living docs). The harness question this issue asked to "decide explicitly" was decided: it consumes the shared helper — `internal/e2e_harness/federated` already imports production internal packages, so no new precedent.

Beyond the consolidation, review surfaced that the credential-load-bearing CDC ATTACH path had zero real escaping assertions (its `TestEscapePgConnAndS3` never calls production code); PR #335 added `TestExportSQLEscapesQuotedPGConn` pinning both export builders with a quoted DSN, mutation-verified.

Note for archaeology: the issue's `internal/e2e_harness/federated/query.go:616` reference had been relocated by #220's file split to `query_build.go` before this work started.

Follow-up residue is tracked separately (tautological-test deletion, one comment-pointer nit, and a pre-existing gofmt quote-corruption trap in `internal/cdc/redact_test.go`).

PR #338, opened 2026-07-30
test/docs: #336 post-#310 residue — delete tautological escape test, fix EscapeLiteral comment location

### LINKED ISSUE #336: Post-#310 residue: delete tautological TestEscapePgConnAndS3; point EscapeLiteral comment parenthetical at its definition
Follow-up from PR #335 (issue #310). Two one-line items, both recorded in that PR's review and left out of scope there.

**1. `internal/cdc/duckdb_exporter_test.go` — `TestEscapePgConnAndS3` asserts nothing about production code.**

It escapes two local strings with its own `strings.ReplaceAll(…, "'", "''")` and asserts on that output — it would still pass if `buildExportSQL` stopped escaping entirely. #310's plan preserved it under an "independent test oracle" rationale that the final review showed to be false (there is no production output being compared against). The real pin now exists: `TestExportSQLEscapesQuotedPGConn` (added in #335) drives a quoted DSN through both export builders and is mutation-verified. With that in place, `TestEscapePgConnAndS3` is pure noise — delete it (or repoint it at real builder output if there is something left worth keeping, which is doubtful).

**2. `internal/federated/connstring.go:33` — the parenthetical reads as a location claim and names the wrong place.**

```go
// (postgres_scan('{{.PG_CONN}}', …)), so the renderer escapes it with
// sqlutil.EscapeLiteral (internal/sqlgen/duckdb_template_renderer.go).
```

`sqlutil.EscapeLiteral` lives in `internal/sqlutil/literal.go`; the renderer file is only the call site. This wording was mandated by #310's plan text (a plan defect, not an implementation one). Suggested:

```go
// (postgres_scan('{{.PG_CONN}}', …)), so the renderer escapes it with
// sqlutil.EscapeLiteral (internal/sqlutil/literal.go), called from the
// DuckDB renderer.
```

Both changes are test/comment-only; `make test` + `make lint` are the only gates needed.

iceboundrock:
Closes #336.

Two items of #310/#335 residue, both recorded in PR #335's review:

1. **Deleted `internal/cdc/duckdb_exporter_test.go` (`TestEscapePgConnAndS3`).** It escaped two local strings with its own `strings.ReplaceAll` and asserted on its own output — it never touched production code and would still pass if `buildExportSQL` stopped escaping. The DSN half was already pinned for real by #335's mutation-verified test. The S3-path half (`duckdb_exporter.go:130`) had **no** real-output coverage anywhere, so per the issue's "repoint at real builder output" alternative, the existing pin is extended with a quote-bearing S3 tmp path and renamed `TestExportSQLEscapesQuotedLiterals`. The new S3 assertion is mutation-verified (dropping the `EscapeLiteral` call on the S3 path fails both subtests).

2. **`internal/federated/connstring.go` comment** now points `sqlutil.EscapeLiteral` at its definition (`internal/sqlutil/literal.go`) instead of the renderer call site — the old wording was a #310 plan defect that read as a wrong location claim.

Gates: `make test` + `make lint` pass locally.


iceboundrock:
## Findings

- [P2] `internal/cdc/duckdb_exporter_sql_test.go:146-149` The new S3 checks only match a substring. A malformed target such as `TO '...pre''fix/...parquet'' (` still contains `wantCopyTarget` and does not contain the raw path, so both export-mode subtests pass despite a malformed or unterminated COPY literal. Require the literal boundary as well (at minimum `wantCopyTarget + " ("`) and preferably extract, undouble, and round-trip the COPY target, mirroring the existing DSN structural assertion.

## Spec

- Pass. PR #338 deletes the tautological test, moves the uncovered S3 concern to genuine output from both builders, and corrects the `EscapeLiteral` definition reference as required by #336 and its implementation plan. No scope creep found.

## Standards

- Pass. No violations found against `AGENTS.md`, `CLAUDE.md`, or `coding-standard.md`; the change remains test/comment-only, within size limits, and introduces no error-handling or layering concerns.

## Assessment

- Changes requested for the P2 test-strength gap above. The intended loss-of-`EscapeLiteral` mutation is caught, but the assertion should also prove the generated COPY literal closes at the expected boundary.

## Observations

- Both snapshot and base exports share `buildExportSQLPlan`, so the single table-driven test exercises both consumers of the escaped S3 path.
- The deleted test asserted only locally transformed strings and did not exercise production output.
- The revised `DuckDBPostgresConnStringFromPool` comment correctly names `internal/sqlutil/literal.go` as the `EscapeLiteral` definition and describes the DuckDB renderer as its caller.

## Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/cdc -run TestExportSQLEscapesQuotedLiterals -v` passed: snapshot and base subtests.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go build ./internal/federated` passed.
- `make test` passed.
- `make lint` passed.
- `git diff --check 9ffb747d9dedda3fb6cb144047dd40da0b936de1...HEAD` passed.
- Existing GitHub Actions checks are all passing: Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

- 1 P2 finding: strengthen the S3 COPY-target assertion to validate the literal boundary/round-trip. Spec and standards reviews otherwise pass.

iceboundrock:
## Review response — P2 addressed in 876e562

**[P2] S3 assertions only matched a substring — confirmed and fixed.** The claim verifies: with the template shape `) TO '%s' (%s);`, a malformed target like `TO '...parquet'' (` contains `wantCopyTarget` and omits the raw path, so both prior assertions stayed green. Both suggested remedies are now in place:

- `wantCopyTarget` is rebuilt as `TO '<escaped>' (` — the trailing `' (` anchors the literal's closing quote to the start of the COPY options list (the "at minimum" fix).
- A structural block mirroring the DSN assertion extracts the literal between `) TO '` and `' (`, requires it equals `wantS3LiteralBody`, and requires undoubling round-trips to the raw `hostileS3TmpPath` (the preferred fix).

**Mutation verification, both classes:**
1. `s3Esc := s3TmpPath` (drop the escape) → both subtests FAIL (as before).
2. Template tail `) TO '%s'' (%s);` — the exact malformed-boundary class from the finding → both subtests FAIL under the new assertions; the pre-fix assertions were blind to this class (verified conceptually: the old `Contains`/`NotContains` pair matches that output).

Both mutations manually reverted; `git diff` on `duckdb_exporter.go` empty; full `./internal/cdc` package green; `gofmt` clean.

No findings were pushed back — the single P2 was valid.

Issue #343, opened 2026-08-02
Demote/memoize the per-row Warn for skipped dropped-attr EAV records


iceboundrock:
Follow-up from #294 (tolerate-and-preserve). Surfaced by task-3 and final reviews.

`transform.FromEAVRecords` logs at Warn once per call when it skips attrIDs removed by schema evolution. The call runs once per ROW on the read path (`toDataRecord` → `FromPersistentRecord`), so a 1000-row page over a schema with a dropped attribute attempts 1000 Warn emissions — steady-state, forever, for a condition #294 defines as supported. zap's production sampler bounds the volume (100/s + 1-in-100), so this is not a flood, but Warn reads as an incident to an on-call operator.

Wanted: demote to Info, or memoize per (schemaID, skipped-attrID-set) so the tolerated condition is logged once per process per schema shape. Also parked from review: the warn-once-per-call behavior itself is untested (zaptest observer precedent exists in internal/entity_write_validation_test.go if coverage is wanted).

iceboundrock:
Context update: #342 shipped via #376. The per-row Warn this issue wants to demote/memoize still exists and is unaffected by the retired-marker change — retired entries are stripped from active caches, so reads skip their EAV rows exactly as #294 specified, and the Warn path is unchanged. The retired marker adds one new nuance the Warn may want to account for: with shipped ledgers now retiring `visit_full.logs` (29) and `visit.contactSnapshot` (25), the skip-Warn now fires for those ids on deployments holding such rows. No changes made to this issue; still open.

iceboundrock:
**Implementation plan** (working copy: `docs/superpowers/plans/2026-08-18-343-demote-memoize-skip-log.md`, gitignored local). Full text follows.

# #343 Demote/Memoize the Per-Row Skipped-Attr Warn — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Replace the per-row Warn in `transform.FromEAVRecords` for #294-tolerated skipped attrIDs with an Info log emitted once per process per (schemaID, skipped-attrID-set), and pin that contract with tests.

[...]

Issue #342, opened 2026-08-02
Guard against attributeID reuse after attribute removal (cross-generation)


iceboundrock:
Follow-up from #294 (tolerate-and-preserve). Surfaced by the final whole-branch review.

#294 makes attribute removal non-destructive: dropped-attr EAV rows are preserved so a schema re-add (same attributeID) restores the values. That contract makes reusing a freed attributeID for a *different* attribute a trap:

- **Same valueType** (drop `old_col` text/attrID 3, later add `nickname` text/attrID 3): preserved `old_col` values silently surface as `nickname` on every legacy row. No error, no log.
- **Different valueType** (later add `age` numeric/attrID 3): `extractValueFromEAVRecord` hits a storage type mismatch → `FromEAVRecords` hard-errors → the row becomes unreadable and un-updatable (500) — exactly the failure class #294 removed, displaced onto a new trigger.

Not a #294 regression (pre-#294 the same rows existed because updates hard-errored before any cleanup), but pre-#294 attribute removal was unsupported; now it is documented as safe, so the trap is invited. `validateSchemaAttributeCache` only checks duplicate IDs within one generation; there is no cross-generation guard and no persisted attribute-ID history to build one from.

Wanted: a guard (registration-time error or at minimum a loud warning) when a schema generation binds an attributeID to a name/type differing from what stored EAV rows for that schema carry — or a persisted retired-ID list (shared concern with #341's classification problem). Docs already warn (`docs/error-handling.md` #294 paragraph, `docs/schema-consistency-migration.md`); this issue is about enforcement.

iceboundrock:
# #342 Cross-Generation AttributeID Reuse Guard — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Make attribute removal ledger-preserving (`retired: true` marker in `_attributes.json`) and hard-error at registration/startup when an active attribute rebinds a retired attributeID — closing the #294 reuse trap (silent value bleed-through on same valueType, unreadable rows on different valueType).

[...]

Issue #341, opened 2026-08-02
validate-schema-consistency flags #294-preserved EAV rows as failures; migration doc coherence residue


iceboundrock:
Follow-up from #294 (tolerate-and-preserve, PR: fix/294-tolerate-preserve-dropped-attr-eav).

## Tool classification

`cmd/tools/validate_schema_consistency.go` `checkUnknownAttributeIDs` reports **every** EAV row whose attrID was removed by schema evolution as a `validationIssue`, and the subcommand exits non-zero. Post-#294 these rows are an expected, supported steady state (skipped on read, preserved on update, restored on re-add). Deployments that have removed attributes can never get a green pre-flight.

Wanted: classify preserved-by-removal rows as informational rather than failing. Hard part (noted by the fix-wave implementer): removed-attribute metadata is gone at runtime, so there is no signal distinguishing "was once a legitimate generation" from "never legitimate". Options include a persisted retired-attribute list per schema, or a tool-side allowlist argument.

## Doc coherence residue (parked at final review of the #294 branch)

`docs/schema-consistency-migration.md`:
- Case (i) in the unknown-attrID remediation section defines the rows as "never legitimate" but still carries the "restore the missing attribute metadata if the data is legitimate" option inherited from the pre-split text — the accidental-metadata-loss scenario (bad rollback / partial deploy) now matches neither case cleanly. Minimal fix: hoist "restore metadata" into the lead-in as a third possibility; leave case (i) with deletion only.
- Front matter (lines ~7/14) still claims EAV rows with unknown attrIDs become request-time errors — false post-#294 for the runtime path; contradicts the rewritten remediation section below it. One-clause qualifier fixes it.
- Rollback step 3 ("fix the reported metadata or EAV inconsistencies") is generic enough to be read as license to delete preserved rows.

iceboundrock:
Prerequisite shipped: #342 is merged via #376. The attributeID ledger now carries a `"retired": true` marker on schema-absent entries (parser + generator + validation all landed), which is exactly the signal this issue's classification work needs — the validator can distinguish #294-preserved rows belonging to a *retired* ledger entry from genuinely orphaned attr_ids. Shipped ledgers now retire `visit.contactSnapshot` (attrID 25) and `visit_full.logs` (attrID 29), so deployments holding rows under those ids will show them flagged on every validator run (docs/schema-consistency-migration.md already calls this out). No changes made to this issue; still open for the classification work itself.

iceboundrock:
[...]

Issue #271, opened 2026-07-16
refactor(test): split internal/entity_manager_test.go (1,196 lines) into per-service test files


iceboundrock:
## Problem

`internal/entity_manager_test.go` is 1,196 lines — far above the 500-line limit in `coding-standard.md`. It predates the limit and accretes because it hosts the shared `mockPersistentRecordRepository` plus tests for CRUD, query, batch, and relation behaviour that all belong to different services (`entity_*_service.go`).

PR #270's review flagged this; that PR resolved its immediate area by moving its new relation-enrichment test and mock extension into `internal/entity_relation_service_test.go` (197 lines), leaving a net +1 line in the oversized file. The full split remains.

## Suggested split (per Incremental Improvement Mode)

- `entity_manager_mocks_test.go` — `mockPersistentRecordRepository`, `mockFederatedQueryEngine`, shared builders (`buildPersistentRecord`, `createTestConfig`)
- `entity_crud_service_test.go` — Create/Get/Update/Delete tests
- `entity_query_service_test.go` — Query/attribute-order/pagination tests
- `entity_batch_service_test.go` — batch/atomic tests
- `entity_relation_service_test.go` — already exists (relation enrichment)

Mechanical moves only (same package, no behaviour change); follow the sed-based move discipline to avoid accidental edits.

## Lineage

Flagged in PR #270 review (Standards); rule: `coding-standard.md` §1 + Incremental Improvement Mode.

iceboundrock:
Execution plan for this split now lives on #320 (comment: https://github.com/Lychee-Technology/forma/issues/320#issuecomment-5384127891). #320, #271, and #401 describe the same piece of work; the plan's PR will close all three. The suggested file names from this issue's body are adopted as-is.

iceboundrock:
PR #450 implements this split (together with #320/#401; the PR closes all three). This issue's suggested file names were adopted as-is; final layout and line counts are in the PR body.

iceboundrock:
Merged via #450 @ dd58a69 (2026-08-23T23:17:52Z) together with #320/#401.

Suggested seams adopted: `entity_manager_mocks_test.go`, `entity_crud_service_test.go`, `entity_query_service_test.go`, `entity_batch_service_test.go` (plus `entity_cross_schema_search_test.go` / `entity_schema_registry_test.go` per #401 and `entity_manager_update_test.go` as separate update seam — 423 lines if folded into crud would leave 77 headroom under the new guard).

[...]

PR #344, opened 2026-08-02
fix: #294 tolerate-and-preserve EAV rows for attributes dropped by schema evolution

### LINKED ISSUE #294: OLTP update rejects rows carrying a dropped attribute's EAV data after schema evolution

iceboundrock:
Fixes #294.

## What

After a schema evolution removes an attribute, rows still carrying that attribute's EAV data previously became effectively read-only: `Update` (and `BatchUpdate`) hard-errored with `unknown attribute id N for schema M`, and the OLTP hot read paths (`Get`, hot-route query assembly, relation enrichment) failed the same way. This PR implements the adjudicated direction from #294 (issue comment, 2026-07-25 — **tolerate-and-preserve**, please don't re-litigate here):

Two choke points, zero service-layer changes:

1. **`transform.FromEAVRecords` tolerates unknown attrIDs** — skipped (one Warn per call, sorted IDs), never an error; skipped rows do not feed the required-attribute bookkeeping. This single fix covers every affected path: Update/BatchUpdate read-back, Get, hot-route query assembly (`eav_aggregated` has no attr filter), relation enrichment.
2. **`replaceEAVAttributes` scopes its DELETE to current-schema attrIDs** (`attr_id = ANY($3)`, sorted+deduped `[]int16`) — stale rows survive both `UpdatePersistentRecord` and `BatchUpdatePersistentRecords` physically untouched. `DeletePersistentRecord` deliberately keeps the unscoped delete (row removal purges stale rows). Missing metadata cache → plain fail-loud error (read-path consistency class, never `ErrInvalidInput`).

## E2E contract (red-first, `TestUpdateAfterAttributeRemoval`)

Written red before the fix (failed with the adjudicated message verbatim), green after — all four assertions verified against a live stack:

1. Update on a row carrying dropped-attr EAV data succeeds.
2. The stale EAV row (`attr_id=3`, `value_text='keep-me'`) is physically present after the update's replace cycle.
3. Under v2 the preserved value is invisible (no `old_col` key; updated `value` correct).
4. After re-adding the attribute (same attributeID), the preserved value reappears — the assertion that distinguishes preserve from drop — and further updates round-trip normally.

Also: the #257 fixture's test-only stale-EAV cleanup detour is retired (`seedMixedGenerationTiers` now updates the rows directly); `docs/error-handling.md` and `docs/schema-consistency-migration.md` updated to the new contract — notably the migration doc's unknown-attrID remediation previously instructed a DELETE that would have destroyed the preserved rows.

## Verification

- `make lint` (golangci-lint v1.64.8) clean; `make test` green.
- Full production e2e suite: 125s, zero failures — including the #189 pin `TestSchemaEvolutionRemovedColumn` and `TestCompactionMixedGenerationEquivalence` (oracle events unchanged; the v2 flush projection never exports dropped attrs, so parquet assertions are unaffected — independently verified against `export_sql_builder.go`'s `attr_id IN (…)` filter + LEFT JOIN).
- `attr_id = ANY($3)` with Go `[]int16` proven against live Postgres (int2[] codec; no cast needed).
- Reviews included mutation testing: sort/compact removal and a "skip swallows missing-required" mutation each fail exactly one pinned test.
- CDC/compaction/federated read paths verified unaffected (exporter and DuckDB pivots filter by known attrIDs; preserved rows contribute nothing).

## Review notes / known deferrals

- **Merge must be a squash**: commits 38f1d80..04c4922 have `Signed-off-by` glued to the subject line (no blank line).
- Parked at final review (tracked in #341): two coherence nits inside `docs/schema-consistency-migration.md` — the case-(i) remediation still carries the "restore metadata if legitimate" option its own definition excludes, and the front matter (~lines 7/14) still claims request-time errors for rows the runtime now tolerates. Both non-destructive; the destructive path is fenced.
- `validate-schema-consistency` still flags preserved rows as failures — classification follow-up: #341.
- attributeID reuse after removal is now an invited trap (silent rebind same-type / unreadable row different-type) — guard follow-up: #342; docs warn in the meantime.
- Per-row Warn on skipped attrIDs is bounded by zap's production sampler but is arguably the wrong level for a supported steady state — follow-up: #343.
- `postgres_persistent_repository_repo_test.go` is pre-existing over the 500-line cap (+51 here; same class as #271). `FromEAVRecords` is now 96/100 lines — next change forces a split. Bare `return err` at the `replaceEAVAttributes` call site matches its three sibling call sites; a 4-site wrap sweep is cleaner as a follow-up than a 1-site patch.
- Embedder-only notes (unreachable via shipped `fileSchemaRegistry`, which fail-closes at startup): a custom `SchemaRegistry` serving a schema absent from the metadata cache now fails loud (`no cache for schema id N`); one serving MORE attributes than the cache would produce inserts outside the delete scope → 23505 → 409 on the second update.


iceboundrock (commented):
## Review: #294 tolerate-and-preserve EAV rows for attributes dropped by schema evolution

### Summary

This PR implements the adjudicated direction from #294 — tolerate-and-preserve rather than hard-error-on-read — through two surgical choke points with zero service-layer changes. The approach is well-architected, thoroughly tested (red-first E2E + unit coverage of edge cases), and correctly scoped. The delete path appropriately remains unscoped. Documentation is updated to reflect the new contract. I recommend approval.

---

### Assessment

**Correctness**: The dual choke-point design (`FromEAVRecords` tolerates unknown attrIDs; `replaceEAVAttributes` scopes DELETE to current-schema attrIDs) is sound and covers all affected code paths (Update, BatchUpdate, Get, hot-route query assembly, relation enrichment) without modification to callers. The decision to keep delete operations unscoped (row removal should purge all EAV data) is intentional and correct.

**Design**: The `knownAttrIDs` function is correctly placed on the repository (where the metadata cache lives), with sort+compact ensuring a deterministic, deduplicated bind value. The `FromEAVRecords` skip correctly avoids feeding `presentAttrIndices`, so skipped rows cannot mask missing-required-attribute errors — verified by `TestAttributeConverterFromEAVRecords_SkippedUnknownDoesNotSatisfyRequired`.

**Quality**: Red-first E2E test (`TestUpdateAfterAttributeRemoval`) covers all four required assertions including the re-add-discriminator. Unit tests cover: scoped delete with sorted/deduplicated attrIDs, nil metadata cache, missing schema cache, unknown-attrID skip, and skip-does-not-satisfy-required. All mock/integration tests updated correctly.

---

### Verification

- [x] `make lint` clean (golangci-lint v1.64.8); `make test` green — confirmed by PR body
- [x] Production e2e suite: 125s, zero failures — confirmed by PR body
- [x] CDC/compaction/federated read paths unaffected — verified against `export_sql_builder.go` attrID filter + LEFT JOIN
- [x] `attr_id = ANY($3)` with Go `[]int16` works against live Postgres (int2[] codec; no cast needed) — confirmed
- [x] TestSchemaEvolutionRemovedColumn (#189 pin) passes with unchanged oracle events — confirmed
- [x] TestCompactionMixedGenerationEquivalence passes — confirmed
- [x] Old EAV cleanup detour in `seedMixedGenerationTiers` retired; rows updated directly — correct
- [x] Confirm that `BatchUpdatePersistentRecords` and `UpdatePersistentRecord` both go through the scoped `replaceEAVAttributes` — yes, both at `postgres_persistent_repository_batch.go:81` and `postgres_persistent_repository.go:168`
- [x] Confirm `DeletePersistentRecord` (line 206) and `BatchDeletePersistentRecords` (line 112) remain unscoped — yes, both use `WHERE schema_id = $1 AND row_id = $2` without attr_id filter

---

### Standards

**File sizes**:
- `attribute_required.go`: 87 lines — within limits
- `update_after_attr_removal_e2e_test.go`: 153 lines — within limits
- `postgres_persistent_repository_repo_test.go`: pre-existing over the 500-line cap (+48 lines here, same class as #271) — noted in PR body, acceptable
- `attribute_converter.go`: net reduction from 494 to (est.) ~442 lines — improvement

**Function sizes**:
- `FromEAVRecords` is now estimated at ~96 lines (noted in PR body as 96/100, next change forces split) — acceptable for this PR per the coding standard advice ("If a change would exceed limits, automatically suggest refactoring instead")
- All new functions well within 100 lines

**Error handling**: 
- `knownAttrIDs` wraps nil cache / missing schema errors with context — correct: `"no metadata cache configured: cannot scope eav delete for schema %d"` and `"no cache for schema id %d: cannot scope eav delete"`
- Missing-required attribute error message updated to new canonical form: `"missing required attribute %q (attrID=%d) in EAV records"` — per error-handling.md anchor
- `return err` at `replaceEAVAttributes` call sites in both `UpdatePersistentRecord` (line 168) and `BatchUpdatePersistentRecords` (line 81) matches existing sibling call sites — acknowledged in PR body as a 4-site sweep follow-up

**Naming conventions**: `knownAttrIDs` is clear, descriptive, and uses the project convention (verb-first private helper). `skippedAttrIDs` is equally clear.

**Imports**: `"slices"` added to two files — standard library package valid at Go 1.26.0 (`go.mod` confirms). No `golang.org/x/exp/slices` usage anywhere in the repo.

---

### Findings

1. **Performance: `knownAttrIDs` recomputed per record in batch loop** — In `BatchUpdatePersistentRecords`, each iteration calls `replaceEAVAttributes` → `knownAttrIDs(schemaID)`, which traverses the cache map, allocates a slice, sorts, and compacts. For a batch of N records sharing the same schemaID, this is O(N × A) where A is the number of attributes in the schema. Not a correctness problem and unlikely to matter for typical batch sizes, but a follow-up could cache the result per schemaID within the batch transaction. **Non-blocking, optimization follow-up.**

2. **`slices.Compact` comment is subtly imprecise** — The comment says "compact because distinct attribute names may share an attributeID." `slices.Compact` only removes *consecutive* duplicates; it works because `slices.Sort` is called first. The test correctly verifies this (`{3, 12, 21}` from the map `{alpha:3, zeta:21, mid:12, mid_aliased:12}`). The comment could be clearer but is not misleading. **Nit, non-blocking.**

3. **Warn log level for skipped attrs in steady state** — The PR body acknowledges that per-row `Warn` on skipped attrIDs is arguably wrong for a supported steady state (normal operation after schema evolution) and tracks it in #343. **Deferred to follow-up, acceptable.**

---

### Observations

- The `#294` comment in `FromEAVRecords` is thorough and self-documenting — it explains *why* the skip and *where* the preservation happens, reducing the need for readers to chase cross-file references. This is a positive pattern.
- The E2E test reads rows through `env.EntityManager()` resolved per call (not captured), ensuring the manager binds fresh metadata after `EvolveSchema`. The comment explains this is deliberate to avoid a vacuous re-add assertion — excellent defensive testing.
- The `slices.Sort(ids)` in the `FromEAVRecords` skip-warning block sorts the reported attrIDs for deterministic warning output. This is a thoughtful touch for production-observability.
- The error-handling.md docs update correctly reclassifies unknown-attrID from "read-path consistency error" to "preserved, not an error," with a clear warning about attrID reuse danger. Good alignment with the code change.
- The `schema-consistency-migration.md` update adds case (i) vs case (ii) distinction with explicit instructions not to delete preserved rows — this is critical to prevent operators from destroying data during remediation.

---

### Spec (#294)

The PR fulfills the adjudicated direction:
1. **Tolerate unknown attrIDs on read** — `FromEAVRecords` skips them
2. **Preserve stale EAV rows on update** — `replaceEAVAttributes` DELETE scoped to `attr_id = ANY($3)` with current-schema attrIDs
3. **Re-adding the attribute restores preserved values** — verified by E2E test
4. **Delete operations purge all EAV rows** — unscoped DELETE retained for both `DeletePersistentRecord` and `BatchDeletePersistentRecords`
5. **Missing required attributes still error** — skip does not feed `presentAttrIndices`; verified by `TestAttributeConverterFromEAVRecords_SkippedUnknownDoesNotSatisfyRequired`
6. **Read-path consistency class for nil/missing metadata cache** — plain error, not `ErrInvalidInput`; verified by `TestUpdatePersistentRecord_NoSchemaCache_FailsBeforeDelete`

**Known deferrals** (explicitly tracked):
- #341: `validate-schema-consistency` classification follow-up
- #342: attributeID reuse guard
- #343: warn log level follow-up

All non-blocking, tracked separately.

iceboundrock:

**Finding 1 (knownAttrIDs recomputed per record in batch loop)** — Fixed in 4a8671c rather than deferred: `BatchUpdatePersistentRecords` now resolves the delete scope once per schemaID via a `scopeBySchema` map and calls the extracted `replaceEAVAttributesScoped`; the single-record path is unchanged (`replaceEAVAttributes` computes then delegates). SQL and bind args are identical, so all existing mock expectations pass unmodified. `make test` green, `make lint` clean.

**Finding 2 (Compact comment imprecise)** — Fixed in the same commit. The comment now states explicitly that `slices.Compact` only removes consecutive duplicates and relies on the preceding `slices.Sort` to make duplicates adjacent.

**Finding 3 (Warn level for steady-state skips)** — No change on this branch, as the review accepts; tracked in #343. Interim behavior is bounded by zap's production sampler (100/s + 1-in-100 for identical messages), so the exposure is log noise/level semantics, not volume.

One correction to the review's Standards section for the record: the missing-required message was not changed in this PR — it remains the pre-existing `missing required attribute '%s' (attrID=%d) in EAV records` form (`internal/transform/attribute_converter.go`). The docs edit in fa61add cited that existing message as an example; no canonical-form migration to `%q` happened here.

iceboundrock:
## Re-review: #294 tolerate-and-preserve dropped-attribute EAV rows

### Findings

1. **P2 - Standards:** `internal/postgres_persistent_repository_repo_test.go` is 745 lines at this head (700 on the merge base). This PR adds 48 lines and removes 3, so it increases an already-over-limit file by 45 net lines. `coding-standard.md` rejects source files above 500 lines, and `AGENTS.md` requires improving the immediate area when modifying code that does not meet the standard. Please move the newly added focused coverage (at minimum `TestUpdatePersistentRecord_NoSchemaCache_FailsBeforeDelete`) to `postgres_persistent_repository_eav_test.go` or split the repository mock tests, rather than further expanding this file.

No new blocking correctness finding was identified in `7a60c78...4a8671c`.

### Spec

Pass. The implementation matches the adjudicated #294 contract:

- `FromEAVRecords` skips removed attrIDs before required-attribute bookkeeping.
- Update replacement deletes only attrIDs addressed by the current schema, preserving stale EAV rows.
- Delete remains deliberately unscoped.
- The new E2E proves update success, physical preservation, invisibility under v2, restoration after re-add, and subsequent normal updates.
- The #257 fixture no longer routes around the production path.

The known validator-classification limitation is correctly tracked in #341; it is not a regression introduced by this PR.

### Standards

The new and changed functions remain below the 100-line hard limit (`FromEAVRecords`: 96 lines). The transform split keeps `attribute_converter.go` within the 500-line limit. The repository test-file growth above is the outstanding standards violation.

The added tests match plain consistency-error text because that path intentionally has no `ErrInvalidInput` sentinel. This is a residual tension with the generic `errors.Is`/`errors.As` guidance, not a separate correctness finding.

### Assessment

The two-choke-point design is sound. The transform change prevents stale rows from making OLTP reads and update read-back fail; the scoped delete prevents the following update from silently converting the behavior into tolerate-and-drop. Batch updates reuse the scope per schema, and missing metadata fails within the transaction so the main-row update is rolled back.

### Observations

- The E2E resolves `env.EntityManager()` on every read, which correctly refreshes metadata after schema evolution and makes the re-add assertion meaningful.
- The per-call warning for skipped attrIDs and attributeID-reuse hazard remain explicit, separately tracked follow-ups (#343 and #342).
- There is no exact E2E for multi-record batch updates containing stale IDs across more than one schema. The repository implementation is covered by static tracing and unit coverage; this is a residual coverage gap, not a defect found here.

### Verification

- Reviewed exact range `7a60c781d57345c2a8b439cda34da40f99c17028...4a8671c91c0989450072779cb4b5f31314636400`; `git diff --check` is clean.
- Fresh local verification on head `4a8671c`: `GOCACHE=/var/folders/37/2bpjmt0129z96c4plfy52gkw0000gn/T/opencode/forma-pr344-gocache GOFLAGS=-buildvcs=false make test` passed.
- GitHub checks on this head all pass: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

### Summary

Functional/spec review passes with no blocking correctness issue. Requesting a standards follow-up for the expanded 745-line repository test file before approval.

Issue #345, opened 2026-08-02
Production e2e harness engine omits the plan cache that factory wires


iceboundrock:
Spun out of #255's final whole-branch review.

`internal/e2e_harness/production/engine.go` builds the federated engine with only `WithParquetSource(...)`, while the real composition root (`factory/factory.go`) also wires `WithPlanCache(...)`. Consequence: `serveFromPlanCache` returns `ok=false` on every harness query, so the **entire production e2e suite never exercises the compiled-plan serve path end-to-end**, despite `Engine()`'s intent to mirror the production assembly.

Found because #255's e2e plan-cache-poisoning probe was structurally vacuous for this reason (its comments now say so and point at the genuine unit-level proof, `federated.TestEngineColdMissingSetRekeysPlanCache`).

Wiring `WithPlanCache` into the harness is a suite-wide behavior change (every e2e query would start hitting cached skeletons) and deserves its own branch with a full sweep. When done, consider also:
- restoring the #255 e2e poisoning probe to a genuine end-to-end proof (glob-hint Env so the path-set component of the scope key stays constant across the flush);
- the harness comment in `schema_evolution_never_flushed_e2e_test.go` that documents the missing plan cache — it becomes stale the moment this lands.

iceboundrock:
**Implementation plan for this issue** (working copy: `docs/superpowers/plans/2026-08-14-345-harness-plan-cache.md`, gitignored — this comment is the record; posted at planning time, before execution):

# #345 Production e2e harness plan-cache wiring — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Wire the compiled-plan cache into the production e2e harness engine so the suite exercises `serveFromPlanCache` end-to-end the way `factory.NewEntityManager` does, guard the wiring with a delete-it-and-it-goes-red e2e test, and restore #255's plan-cache-poisoning probe as a genuine end-to-end proof.

[...]

PR #346, opened 2026-08-02
feat(federated): #255 never-flushed column projection — NULL-augmented scan source, complete-union gate, plan-cache re-keying

### LINKED ISSUE #255: Cold-only reads fail on attributes added before their first flush (never-flushed column projection)

iceboundrock:
Closes #255. Follow-up filed: #345.

## What this delivers

A federated query that projects, filters, sorts, or keysets on a schema attribute added **before its first flush** (physically absent from every parquet file in the resolved scan set) now succeeds with exact SQL NULL semantics, instead of dying as a classified DuckDB binder error. Red-first e2e `TestSchemaEvolutionNeverFlushedColumn` pins the contract: full-scan 8 rows, filter/sort on the never-flushed attribute, cold-only reads, and a pre/post-first-flush transition probe.

## Design — wrapped scan source, not projection mutation

The issue sketched feeding the cold union into `BuildSchemaProjection` (`NULL::<type> AS attr` in the projection). We landed a different shape: both template scan sites — the `s3_source` scan **and** its pushdown semijoin — consume one `{{.S3_SCAN_SOURCE}}` param that renders either the exact pre-#255 `read_parquet(..., union_by_name=true)` text (empty missing set → **byte-identical SQL**, keeping the #214 executable-doc guard and every rendered-SQL contract green untouched) or `(SELECT *, NULL::<type> AS <col> ... FROM read_parquet(...)) AS cold_scan`. Why:

- **One mechanism covers both halves of the issue.** The explicit projection binds, and the semijoin's `LOGICAL_WHERE_CLAUSE` evaluates against a real NULL column — exact SQL NULL semantics for every operator, no predicate pruning/rewriting.
- **`SchemaProjection` stays schema-pure**, so the per-schema `ProjectionCache` (#142) is untouched.
- Derivation lives only in `injectDuckDBTemplateParams`, the single funnel both `CompileDuckDBQuery` and the direct builder pass through — skeleton and direct render structurally cannot disagree (pinned by `TestCompiledQueryParity`, now including keyset/hot-excluded rows with non-empty missing sets and a non-vacuity guard).

**Cold union source:** the #189 pre-read validator already DESCRIBEs every new footer (write-once cache) and expands globs per query; it now caches each path's column map and returns `(union, complete)`. **Only a complete union may augment** — augmenting a column that exists in an unprobed file would collide with the real column — so inconclusive cases keep today's loud, classified, degradable failure. `DuckDBNullScanType` mirrors the hot-leg EAV pivot types case-for-case so the UNION ALL cannot widen (#205 stays closed). Benchmark schemas (100–102) are exempt end-to-end.

## Plan-cache poisoning (the issue's mandatory hazard)

The rendered SQL now depends on the missing set, so the missing set joins the compiled-plan **scope hash** (`scope-v2`, `"cold-missing"` marker, name-sorted). Manifest-authored path sets already re-key on flush (paths are in the scope), but **glob-hint path strings do not change when the first flush lands the column** — the per-query recomputation of the missing set is then the only re-keying signal. Proven at the integration seam by `TestEngineColdMissingSetRekeysPlanCache` (real `WithPlanCache` engine, constant path set, miss→hit→miss with only the footer union changing; mutation-verified: removing the cold-missing scope component makes the stale NULL-projecting skeleton serve — the exact poisoning) plus `TestDuckPlanScopePartsIncludeColdMissingSet` at the unit seam.

## Adjudications (user-ruled during review)

- **Glob listing skew — accepted and documented** (code comment at the augmentation site + design.md): the validator's glob expansion and the scan's `read_parquet` are separate S3 LISTs, so a flush landing between them can collide the NULL alias with the newly-landed real column — a one-time loud classified failure that self-heals on the next query. Strictly better than the status quo, where every such query fails.
- Final review found the e2e's plan-cache probe was structurally vacuous (the production harness engine wires no plan cache — factory does; suite-wide gap now filed as #345); the genuine re-key proof moved to the engine-level unit test above and the e2e comments state exactly what the probe does prove.

## Verification

- Red-first: e2e committed failing with `Binder Error: Referenced column "score" not found` at the full-scan probe; green after wiring.
- `go test -tags=e2e ./internal/e2e_harness/production/ -run 'TestSchemaEvolution|TestParquetCorruption|TestManifestConsistency'` — 15/15 (six evolution scenarios, six corruption incl. `_WrongSchemaFile`/`_WrongTypeFile`, two manifest, the new #255 test).
- `make test` — all packages green (`-count=1` forced on sqlgen/federated/queryplan).
- `make lint` — clean (pinned v1.64.8).
- Design-doc guard `TestDesignDoc*` — green with the §5 fence untouched; new §5-adjacent note documents the mechanism.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -331,6 +331,24 @@ a strict inequality on that key at the boundary, which silently skips every row
 tied there; the trailing `row_id` gives the composite key a unique tiebreak so
 each boundary tie is resolvable (#183).
 
+**Never-flushed columns (#255).** `union_by_name` can only union columns that
+exist in *some* file. An attribute added to the schema before its first flush is
+absent from the entire scan set, so the scan source (both `s3_source` and the
+semijoin) is wrapped as
+`(SELECT *, NULL::<type> AS <col> … FROM read_parquet(…)) AS cold_scan`,
+projecting each such column as a typed NULL — computed per query from the
+pre-read validator's footer-column union, and only when that union is complete
+(an incomplete union falls back to the unaugmented scan and today's loud
+classified failure). The missing-column set participates in the compiled-plan
+scope hash, so a skeleton compiled while a column was cold-absent is re-keyed the
+moment the first flush lands it (the plan-cache poisoning hazard the issue
+mandates addressing). With no missing columns the rendered SQL is byte-identical
+to this document's §5 sketch. One residual race is accepted: for glob-hint path
+sets the validator's listing and the scan's listing are separate S3 LISTs, so a
+flush landing between them can collide the NULL alias with the newly-landed real
+column — a one-time loud classified failure that self-heals on the next query,
+since the missing set is recomputed per query.
+
 ## **6. Optimization Strategies**
 
 ### **6.1 Predicate Pushdown (Critical)**
--- a/internal/cdc/duckdb_exporter_type_lockstep_test.go
+++ b/internal/cdc/duckdb_exporter_type_lockstep_test.go
@@ -1,0 +1,56 @@
+package cdc
+
+import (
+	"database/sql"
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/lychee-technology/forma/internal/sqlgen"
+	"github.com/stretchr/testify/require"
+)
+
+// eavExportFixture mimics the eav_data columns castEAVValue reads, at the
+// types the export query actually carries: value_numeric is DOUBLE (the
+// numeric-family column every scalar cast starts from) and value_text is
+// VARCHAR.
+const eavExportFixture = "(SELECT CAST(1 AS DOUBLE) AS value_numeric, CAST('x' AS VARCHAR) AS value_text)"
+
+// TestCastEAVValueMatchesNullScanTypeof is the #255 type-lockstep proof for
+// the CDC export leg, against a real DuckDB engine: the parquet column type
+// the exporter mints for an attribute must equal the type
+// sqlgen.DuckDBNullScanType names for the same attribute. The federated read
+// UNIONs a NULL-augmented parquet leg (for a column no generation carries
+// yet) with real parquet columns written by this exporter — divergence widens
+// the UNION ALL and re-opens #205.
+//
+// The comparison is on typeof() rather than on the rendered SQL text: bool
+// renders as `(value_numeric <> 0)`, which carries no type literal to
+// extract — only the engine knows the resulting type.
+func TestCastEAVValueMatchesNullScanTypeof(t *testing.T) {
+	db, err := sql.Open("duckdb", ":memory:")
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, db.Close()) })
+
+	scalars := []forma.ValueType{
+		forma.ValueTypeBool, forma.ValueTypeBigInt, forma.ValueTypeDate,
+		forma.ValueTypeDateTime, forma.ValueTypeInteger, forma.ValueTypeSmallInt,
+		forma.ValueTypeNumeric, forma.ValueTypeText, forma.ValueTypeUUID,
+	}
+	for _, vt := range scalars {
+		t.Run(string(vt), func(t *testing.T) {
+			exportExpr := castEAVValue(forma.AttributeMetadata{ValueType: vt})
+
+			var exported string
+			require.NoError(t,
+				db.QueryRow("SELECT typeof(("+exportExpr+")) FROM "+eavExportFixture).Scan(&exported),
+				"export cast %q must evaluate", exportExpr)
+
+			var scan string
+			require.NoError(t,
+				db.QueryRow("SELECT typeof(NULL::"+sqlgen.DuckDBNullScanType(vt, "")+")").Scan(&scan))
+
+			require.Equal(t, exported, scan,
+				"cold NULL scan type must equal the exported parquet column type for %s (#205 no-widening)", vt)
+		})
+	}
+}
--- a/internal/e2e_harness/production/schema_evolution_never_flushed_e2e_test.go
+++ b/internal/e2e_harness/production/schema_evolution_never_flushed_e2e_test.go
@@ -1,0 +1,114 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"testing"
+
+	"github.com/lychee-technology/forma/internal/model"
+)
+
+// queryTierRestricted runs a parquet-only (warm+cold) read through the engine
+// alone and asserts its total. The oracle is deliberately tier-blind — it
+// folds the event log into the full logical set with no notion of what has
+// been flushed — so a tier-restricted read taken while unflushed hot rows
+// exist cannot be compared against it (that divergence would be a fixture
+// artifact, not a product defect). These assertions therefore pin the
+// engine's physical answer directly, the same way assertTierMatrixS3Only
+// does for #184.
+func queryTierRestricted(ctx context.Context, t *testing.T, env *Env, q Query, want int64, why string) *QueryResult {
+	t.Helper()
+	res, err := env.Query(ctx, q)
+	if err != nil {
+		t.Fatalf("tier-restricted query (%s): %v", why, err)
+	}
+	if res.Total != want {
+		t.Fatalf("tier-restricted total (%s) = %d, want %d", why, res.Total, want)
+	}
+	return res
+}
+
+// TestSchemaEvolutionNeverFlushedColumn covers #255: v2 adds `score` BEFORE
+// its first flush, so the base parquet (v1 generation) is the ENTIRE cold
+// set and physically lacks the column — union_by_name (#189) cannot invent
+// it. Contract: cold-inclusive reads succeed with score as a typed NULL on
+// every parquet-tier row (exact SQL NULL semantics for filters and sorts),
+// hot rows serve real values, and the compiled-plan cache does not keep
+// projecting NULL after the first flush lands the column (the #255
+// plan-cache poisoning hazard).
+func TestSchemaEvolutionNeverFlushedColumn(t *testing.T) {
+	ctx := context.Background()
+	cluster := SharedCluster(t)
+	v1 := writeSimpleSchemaDir(t, coreProps, coreAttrs)
+	v2 := writeSimpleSchemaDir(t, scoreIntProps, scoreIntAttrs)
+	env := NewEnv(t, cluster, WithSchemaDir(v1))
+	simple := DefaultSchemaFixtures()[0]
+
+	// 5 base rows under v1 (ordinals 0-4), exported as base; evolve to v2;
+	// 3 hot rows under v2 (ordinals 5-7, score 50/60/70) left UNFLUSHED.
+	seedGeneration(ctx, t, env, simple, 5, buildEvolutionProfile(buildLabeledExtras(nil)))
+	baseKey := runInitBase(ctx, t, env, simple)
+	if err := env.EvolveSchema(ctx, v2); err != nil {
+		t.Fatalf("evolve schema to v2: %v", err)
+	}
+	seedGeneration(ctx, t, env, simple, 3, buildEvolutionProfile(buildLabeledExtras(func(ordinal int) map[string]any {
+		return map[string]any{"score": float64(ordinal * 10)}
+	})))
+
+	// Shape precondition: the whole cold set lacks score.
+	forbidParquetCols(t, "base (v1)", describeParquetCols(ctx, t, env, baseKey), "score")
+
+	// Full scan across all tiers: 8 rows (5 base + 3 hot), no binder error.
+	full := env.AssertQueryMatches(ctx, Query{Schema: simple, Limit: 20})
+	assertUsesDuckDB(t, full)
+	if full != nil && full.Total != 8 {
+		t.Fatalf("full scan total = %d, want 8 (5 v1 base + 3 hot)", full.Total)
+	}
+
+	// Filter on the never-flushed attribute: hot rows only (score 50/60/70).
+	filtered := env.AssertQueryMatches(ctx, Query{
+		Schema:  simple,
+		Filters: []Filter{{Attr: "score", Op: "gte", Value: "50"}},
+		Limit:   20,
+	})
+	if filtered != nil && filtered.Total != 3 {
+		t.Fatalf("score >= 50 total = %d, want 3 (hot rows only; NULL excludes every base row)", filtered.Total)
+	}
+
+	// Sort on it: base rows (score NULL) land NULLS LAST.
+	env.AssertQueryMatches(ctx, Query{
+		Schema: simple,
+		Sorts:  []Sort{{Attr: "score"}},
+		Limit:  20,
+	})
+
+	// Hot-excluded (cold-only) read — the issue's title case: must succeed,
+	// serving the 5 base rows with score wholly absent/NULL. This is the
+	// positive query that makes the zero-row assertion below trustworthy.
+	coldTiers := []model.DataTier{model.DataTierWarm, model.DataTierCold}
+	coldOnly := queryTierRestricted(ctx, t, env, Query{
+		Schema: simple, PreferredTiers: coldTiers, Limit: 20,
+	}, 5, "base rows; unflushed hot invisible")
+	assertUsesDuckDB(t, coldOnly)
+
+	// Plan-transition (cache-poisoning) regression: compile & cache the
+	// cold-only filter shape while score is cold-absent...
+	coldFilter := Query{
+		Schema: simple, PreferredTiers: coldTiers,
+		Filters: []Filter{{Attr: "score", Op: "gte", Value: "50"}},
+		Limit:   20,
+	}
+	queryTierRestricted(ctx, t, env, coldFilter, 0, "pre-flush cold-only score filter")
+
+	// ...then land the FIRST flush carrying score and re-run the same shape.
+	// A poisoned plan (cached NULL projection) would keep the flushed rows
+	// invisible and return 0. Post-flush nothing is hot, so the parquet-only
+	// read is the whole logical set again and the oracle is authoritative.
+	mustFlush(ctx, t, env)
+	postFlush := env.AssertQueryMatches(ctx, coldFilter)
+	assertUsesDuckDB(t, postFlush)
+	if postFlush != nil && postFlush.Total != 3 {
+		t.Fatalf("post-flush cold-only score filter total = %d, want 3 — plan-cache poisoning (#255): the skeleton compiled while score was cold-absent is still projecting NULL", postFlush.Total)
+	}
+}
--- a/internal/e2e_harness/production/schema_evolution_never_flushed_e2e_test.go
+++ b/internal/e2e_harness/production/schema_evolution_never_flushed_e2e_test.go
@@ -34,9 +34,8 @@ func queryTierRestricted(ctx context.Context, t *testing.T, env *Env, q Query, w
 // set and physically lacks the column — union_by_name (#189) cannot invent
 // it. Contract: cold-inclusive reads succeed with score as a typed NULL on
 // every parquet-tier row (exact SQL NULL semantics for filters and sorts),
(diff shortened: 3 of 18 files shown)


iceboundrock (commented):
## Review — #255 never-flushed column projection (NULL-augmented scan source, complete-union gate, plan-cache re-keying)

Reviewed the full branch diff vs `main` (6 commits, red-first) against issue #255, the follow-up #345, and `AGENTS.md`/`coding-standard.md`. Full unit suite, targeted new tests, `go vet`, and pinned `golangci-lint` v1.64.8 all run green in a fresh worktree; e2e compiles under `-tags=e2e` (not executed — needs Docker).

### Findings

1. **The `#205` type-lockstep is pinned by a hardcoded unit test, not a cross-assertion.** `TestDuckDBNullScanTypeMirrorsTierTypes` asserts literal strings (`INTEGER`, `BIGINT[]`, …), so it validates `DuckDBNullScanType` against itself. It never renders and compares `buildEAVPivotExpr` / `eavElementCastExpr` (hot leg) or `cdc.castEAVValue` (export leg) — the very mappings whose divergence this table exists to prevent. A future edit to the pivot or export type mapping (the exact drift that caused #205) would silently re-open the UNION ALL widening without failing this test. Consider a test that renders the pivot/export cast for a shared attribute set and asserts `DuckDBNullScanType` matches — cheap, and it makes the lockstep real.

2. **`TestEngineColdMissingSetRekeysPlanCache` proves the re-keying mechanism but not the production trigger.** The test grows the union by *overwriting* the validator's cached footer (`markValidated(coldPlanCachePath, …true)`), but `parquet_schema_validation.go` documents the cache as **write-once** ("Parquet objects are write-once: validated paths are never re-probed"), and that is the production invariant — a path's footer never changes. In production the union grows because a *new* file (new path) is probed, not because a cached footer is overwritten. The re-keying logic itself is valid and the test is a reasonable constant-path-set model of the glob-hint case (see Observations), but a reader could misread it as the write-once cache being mutable. Suggest a sibling case that adds a second path (the real trigger) or a comment clarifying the overwrite is a stand-in for "a new file's columns joined the union."

3. **Minor: `coldMissingColumns` augments *every* schema attribute absent from the union, including ones the query never touches.** Harmless (the extra columns stay inside `cold_scan`; `S3SourceSelect` selects explicitly, so no column leaks into the `UNION ALL`), and it keeps scope hashes deterministic per schema. But it means the plan-cache scope for a schema carries the full never-flushed set even for a single-column query, and a never-flushed column with an unusual type can force the wrap on otherwise-trivial reads. No action required — just confirming this is intentional and understood.

### Spec

Issue #255 asked for three things; all three are delivered, and the two "sketch" mechanisms were deliberately improved:

- **Projection half** (`NULL::<type> AS attr`): delivered as a scan-source wrap rather than feeding the union into `BuildSchemaProjection`. The deviation is justified — projection-only NULLs would not fix the semijoin's `LOGICAL_WHERE_CLAUSE`, which evaluates against raw `read_parquet` output. Wrapping both template scan sites (`s3_source` FROM and the semijoin inner FROM) fixes projection *and* pushdown filter in one mechanism, with exact SQL NULL semantics. `SchemaProjection`/`ProjectionCache` (#142) stay untouched.
- **Semijoin half** (NULL-normalization or pruning): the wrap makes the semijoin evaluate against a real typed-NULL column — correct and strictly better than pruning (no predicate rewriting, no query-dependent behavior).
- **Plan-cache poisoning hazard (the issue's mandatory point):** addressed by adding the name-sorted cold-missing set to the compiled-plan scope hash (`duckPlanScopeParts`, `scope-v2` + `"cold-missing"` marker). Correctly reasoned: manifest path sets already re-key on flush via the path component, glob-hint path strings do *not* change, so the per-query recomputed missing set is the only re-keying signal there. The unit proof is genuine (real `WithPlanCache` engine, miss→hit→miss with only the union changing).

The `complete`-flag gate also matches the issue's constraint: an incomplete union (unprobed file) must not augment, or the NULL alias could collide with a real column — the loud classified failure is preserved for inconclusive cases. Benchmark schemas (100–102) are exempt end-to-end, consistent with the pre-existing skip in `resolveScanSources`. No spec requirements are missing or partial.

### Standards

- File/function size limits respected (largest touched file 451 lines; no function near 100).
- Error wrapping with context everywhere (`fmt.Errorf("resolve parquet paths: %w", …)`); no bare `return err` introduced.
- Early returns, flat structure, ≤3 nesting levels maintained.
- Comments carry issue references and invariants (#255, #189, #205, #214, #260, #299), matching house style.
- `requireS3Paths` was correctly migrated to validate `S3_SCAN_SOURCE` (what the template actually consumes since #255) and its message updated; the old `TestCompileDuckDBQuery_MissingS3PathsRejected` still passes (substring `S3_PATHS` present in the new message).
- Lint (pinned v1.64.8) and `go vet` clean on the touched packages; `gofmt`-clean.

One judgement-call nit: `duckPlanScopeParts` hardcodes `"scope-v2"` — good (the skeleton format changed, one-time cache invalidation is correct), just calling out the bump is the intended invalidation mechanism.

### Assessment

Approve. The change is well-scoped, red-first, and the design genuinely covers the issue's hardest requirement (plan-cache poisoning) rather than paying it lip service. The two `Findings` above are robustness/test-fidelity improvements, not correctness defects. The accepted glob-listing-skew race is a strict improvement over the status quo (one-time loud self-healing failure vs. every such query failing) and is documented at the augmentation site, in `design.md`, and in the PR body.

### Observations

- **Empty-missing-set renders byte-identical SQL** to pre-#255 (`BuildParquetScanSource` returns the bare `read_parquet(…, union_by_name=true)` text), keeping every rendered-SQL contract and the #214 design-doc guard green — verified by running `TestDesignDocSQL_*`.
- **Glob re-listing per query** is what makes the poisoning guard work for glob hints: the validator re-expands the glob on every `Validate`, so a newly-landed file joins the union and shrinks the missing set, re-keying the scope even though the glob string in the path component is unchanged. This matches the PR's claim exactly.
- **`complete=false` is conservative** (any `'";` path, glob listing failure, or unreadable footer): the union may still contribute probed columns, but augmentation is blocked wholesale. Preserves today's failure classification for uncertain states.
- **`requireProjectionParams` / `requireS3Paths`** now both guard the compiled-plan entry point too (`TestCompileDuckDBQuery_Missing*`), so the skeleton path cannot render an unbound scan source.
- **#345 is well-filed**: the e2e poisoning probe is honestly documented as vacuous (harness engine has no `WithPlanCache`), and the genuine proof lives at the engine seam. Worth noting the harness's `serveFromPlanCache` returns `ok=false` for *every* e2e query — a latent suite-wide blind spot the issue captures.
- The `schemaCacheByID` nil-safe accessor is a clean dedup of the two pre-existing nil-guarded lookups.

### Verification

Run in a fresh worktree at `c1d659e` with the Makefile's env (`GOCACHE`, `GOFLAGS=-buildvcs=false`):

- `go build ./...` — clean.
- `go test . ./cdc ./cmd/... ./factory ./internal/... -count=1` — all packages pass.
- New tests pass individually: `TestBuildParquetScanSource*`, `TestDuckDBNullScanTypeMirrorsTierTypes`, `TestAdvancedTemplateRendersScanSourceAtBothSites`, `TestColdMissingColumns*`, `TestDuckPlanScopePartsIncludeColdMissingSet`, `TestEngineColdMissingSetRekeysPlanCache`, `TestCompiledQueryParity` (all 8 cases incl. cold-missing/keyset), `TestValidateReturnsCompleteColumnUnion`, `TestValidateUnionIncompleteOnProbeFailure`, `TestValidateCachedPathContributesColumnsWithoutReprobe`.
- `TestDesignDocSQL_*` — pass (executable-doc guard intact).
- `go vet` on `sqlgen`/`federated`/`queryplan`/`cdc` — clean.
- `golangci-lint run` (pinned v1.64.8) — clean.
- `go vet -tags=e2e ./internal/e2e_harness/production/` — compiles; e2e suite not executed here (requires Docker).

### Summary

6 findings-adjacent notes, 0 blocking. Spec fully delivered (all three of #255's requirements, with a justified design deviation and a genuine, unit-proven plan-cache guard). Standards clean. Recommended actions: (1) make the type-lockstep test a cross-assertion against the pivot/export casts, (2) optionally strengthen the plan-cache re-key test to model union growth via a new path, matching the write-once cache invariant. Merge-ready after those are considered.

iceboundrock:
Round-1 response — findings 1–2 fixed in `3e8cfab`, finding 3 confirmed intentional.

**Finding 1 (type-lockstep self-validation) — fixed.** String extraction can't carry the cross-assertion (bool renders `(value_numeric <> 0)`, text renders bare `value_text` — no type literal to compare), so the new tests execute both legs on an in-memory DuckDB and compare `typeof()`:
- `sqlgen.TestDuckDBNullScanTypeMatchesHotLegTypeof` — hot leg: `typeof(eavElementCastExpr(vt))` over a `value_numeric DOUBLE / value_text VARCHAR` fixture vs `typeof(NULL::DuckDBNullScanType(vt))`, all 9 scalars plus a `list(...)` case pinning the `elem+\"[]\"` construction against DuckDB's real list type name.
- `cdc.TestCastEAVValueMatchesNullScanTypeof` — export leg: same technique over `castEAVValue`.

Mutation-verified in both directions: `duckDBNullScalarType` bigint→DOUBLE fails the hot-leg test (BIGINT vs DOUBLE, BIGINT[] vs DOUBLE[]); `castEAVValue` date→DOUBLE fails the export-leg test. The literal table test stays as documentation of the intended map, now commented as self-validating.

**Finding 2 (overwrite vs write-once invariant) — fixed, both suggested forms.** The `markValidated` overwrite now carries a comment marking it as a test stand-in for \"a new file's columns joined the union\", and the new sibling `TestEngineColdMissingSetRekeysPlanCacheViaGlobExpansion` models the production trigger exactly: a glob hint (`s3://b/7/*.parquet`) whose un-expanded string is what `duckPlanScopeParts` receives — so the path component is byte-identical across runs — while the fake's `glob()` expansion gains `delta1.parquet` carrying `score` between run 1 and run 2 (base stays validator-cached, no re-probe). Run 2 asserts `plan_cache=miss` with no NULL alias; dropping `missing` from `duckPlanScopeParts` turns it into a hit — the poisoning. `fakeDuckDBExecutor` gained two nil-default fields (`globFiles`, `describeCols`); nil preserves existing behavior exactly, no existing test touched.

**Finding 3 (full-schema augmentation) — intentional, for a binding reason.** `S3SourceSelect` is schema-shaped, not query-shaped: it projects every schema attribute unconditionally (the outer SELECT must align positionally with `EntityMainColumnDescriptors` — the #147 positional-scan contract), so the augmentation domain must equal the projection domain or the projection itself fails to bind on any absent attribute the query never mentions. Narrowing to query-referenced columns would (a) still binder-fail on the untouched absent attributes, and (b) fragment the scope key by referenced-set for zero gain. The \"unusual type forces the wrap on trivial reads\" cost is one constant-projection subquery node — and any cold-inclusive read on that schema has to render the full projection regardless.

Also confirming the `scope-v2` nit: the bump is deliberate as the one-time whole-cache invalidation for the skeleton format change.

iceboundrock (commented):
## Review round 2 — `3e8cfab` (cross-leg type-lockstep proofs, glob-expansion re-key case)

Re-reviewed the incremental diff `c1d659e...3e8cfab` (4 test files, +232/−3, **no production code touched**) against issue #255 and my round-1 findings. Verified in a fresh worktree at `3e8cfab` with the Makefile env.

### Findings

**Round-1 finding 1 (type-lockstep was self-validating) — resolved, and genuinely.** The new cross-leg proofs execute both legs on an in-memory DuckDB and compare `typeof()`:
- `sqlgen.TestDuckDBNullScanTypeMatchesHotLegTypeof` — hot leg via `eavElementCastExpr` over a `value_numeric DOUBLE / value_text VARCHAR` fixture vs `NULL::DuckDBNullScanType(vt)`, all 9 scalars plus a `list(...)` case pinning the `elem+"[]"` construction to DuckDB's real list type name.
- `cdc.TestCastEAVValueMatchesNullScanTypeof` — export leg via `castEAVValue`, same technique.

The `typeof()` choice is correct: bool renders `(value_numeric <> 0)` and text renders bare `value_text`, neither of which carries a type literal, so string comparison could never have worked. I mutation-verified both are non-vacuous — flipping `duckDBNullScalarType`'s `INTEGER`→`BIGINT` fails **both** tests with `expected "INTEGER" actual "BIGINT"` (hot leg and export leg). This closes the #205 re-opening path this PR is the guard for.

**Round-1 finding 2 (overwrite vs write-once invariant) — resolved in both suggested forms.** The `markValidated` overwrite in the original test now carries an explicit comment marking it a test stand-in, and the new `TestEngineColdMissingSetRekeysPlanCacheViaGlobExpansion` models the real trigger exactly: a glob hint (`s3://b/7/*.parquet`) whose un-expanded string is byte-identical across both runs (that string is what `duckPlanScopeParts` receives), while the fake's glob expansion gains `delta1.parquet` carrying `score` between runs. The write-once `base.parquet` footer stays served from the validator cache (no re-probe — I confirmed via the cached-path path in `validateConcrete`), the union grows through the *new* file, the missing set empties, and run 2 is a `plan_cache=miss` with no NULL alias. The comment correctly states that dropping `missing` from `duckPlanScopeParts` turns run 2 into a poisoning hit. This is the production-faithful proof the issue's mandatory hazard deserved.

**Round-1 finding 3 (full-schema augmentation) — confirmed intentional, and the justification is right.** The author's reasoning is sound and worth recording: `S3SourceSelect` is schema-shaped, not query-shaped (every attribute is projected unconditionally for positional alignment with `EntityMainColumnDescriptors` under the #147 positional-scan contract), so the augmentation domain must equal the projection domain. Narrowing to query-referenced columns would binder-fail on the untouched absent attributes anyway and fragment the scope key for zero gain. The cost is one constant-projection subquery node. No action.

### Spec

No spec changes in this commit — it is verification-only for already-delivered behavior. Issue #255's three requirements remain fully met (projection, semijoin, plan-cache re-keying), and now each carries a production-faithful proof: the hot-leg and export-leg `typeof()` parity for #205, and the glob-expansion re-key for the poisoning guard. Benchmark exemption and the `complete`-gate fallback are unchanged.

### Standards

- Round-2 is test-only and follows established repo patterns: `sql.Open("duckdb", ":memory:")` is already used across `internal/sqlgen`, `internal/cdc`, `internal/federated`, `internal/compaction` (e.g. `duckdb_list_pivot_test.go`, `duckdb_bigint_pivot_test.go`), so the new tests add no new infrastructure or build-tag gating.
- `fakeDuckDBExecutor`'s two new fields (`globFiles`, `describeCols`) are nil-default and the existing default behavior (fixed system columns, unlistable globs) is preserved — verified: no existing test touched, full suite green.
- File sizes, error wrapping, comment/issue-reference style all within standard (largest touched file 175 lines).
- `go vet` and pinned `golangci-lint` v1.64.8 clean.

### Assessment

Approve. Both round-1 findings are fixed with real, mutation-verified proofs rather than stronger assertions or comments — this is exactly the response the review asked for. No new findings. The type-lockstep and plan-cache hazards are now guarded in a way that fails loud if either mapping drifts.

### Observations

- **Scalar lockstep is asserted against `eavElementCastExpr`, not `buildEAVPivotExpr`.** For all nine scalars the two produce identical types (both mirror `cdc.castEAVValue`), and the list case genuinely requires the element cast — so the test is valid. But the two are only *coincidentally* identical today: if a future change made `buildEAVPivotExpr` diverge from `eavElementCastExpr` for some scalar, this test would not catch it (nothing else asserts scalar pivot-vs-element type parity). A one-line note in the test comment that the scalar pivot and element cast must stay type-identical would close that residual gap.
- **List lockstep is exercised only for `bigint` items.** The `elem+"[]"` construction for other items types (`VARCHAR[]`, `BIGINT[]` from date/datetime, `DOUBLE[]` from numeric) is covered transitively via the scalar `typeof()` cases for the element type plus the one list construction — adequate, just noting it is not exhaustive.
- The glob re-key test drives the hint through `DuckDBHints.S3ParquetPathTemplate` (hint always wins over the source), which is the correct seam: it makes the scope-key path component provably constant, isolating the missing set as the sole re-keying input.
- Worth restating from round 1 (unchanged, still true): empty-missing-set renders byte-identical SQL, `TestDesignDocSQL_*` guards stay green, and #345 correctly tracks the harness's missing `WithPlanCache`.

### Verification

Run in a fresh worktree at `3e8cfab` with `GOCACHE` + `GOFLAGS=-buildvcs=false`:

- `go build ./...` — clean.
- `go test . ./cdc ./cmd/... ./factory ./internal/... -count=1` — all packages pass.
- New round-2 tests pass: `TestDuckDBNullScanTypeMatchesHotLegTypeof` (10 subtests incl. `list_bigint`), `TestCastEAVValueMatchesNullScanTypeof` (9), `TestEngineColdMissingSetRekeysPlanCacheViaGlobExpansion`, plus all round-1 tests still green.
- `go test -race ./internal/sqlgen ./internal/cdc ./internal/federated` — pass (matches CI's race run).
- **Mutation check (non-vacuity):** `INTEGER`→`BIGINT` in `duckDBNullScalarType` fails both cross-leg tests — the proofs genuinely guard #205.
- `go vet` and `golangci-lint` (pinned v1.64.8) — clean.
- `TestDesignDocSQL_*` — pass; `go vet -tags=e2e ./internal/e2e_harness/production/` — compiles.

### Summary

Round-1's two findings are resolved with genuine, mutation-verified proofs; finding 3 confirmed intentional with correct reasoning. 0 blocking, 2 minor observations (scalar pivot-vs-element type parity not cross-asserted; list lockstep non-exhaustive). Merge-ready.

iceboundrock:
Round-2 response — both observations addressed in `8df0a6c` (test-only, +76/−10 in `duckdb_cold_scan_test.go`).

**Observation 1 (scalar lockstep asserted against `eavElementCastExpr`, not `buildEAVPivotExpr`)** — went one step further than the suggested comment: new `TestDuckDBNullScanTypeMatchesPivotLegTypeof` runs the same `typeof()` cross-assertion against the **full pivot expression** (`buildEAVPivotExpr` over a fixture carrying `attr_id`/`array_indices`), for all 9 scalars *and* all 9 list items types. The pivot, not the bare element cast, is what `pg_source` actually projects into the UNION ALL, so the "coincidentally identical today" gap is now a red test instead of a comment. Mutation-verified: flipping the pivot's integer arm to BIGINT fails `pivot-leg/integer` while the element-leg test stays green — exactly the divergence the observation said nothing would catch.

**Observation 2 (list lockstep only exercised for bigint items)** — the element-leg list case now loops all 9 items types (`list_bool` … `list_uuid`), and the new pivot-leg test covers the same 9 through the full `CASE WHEN count(*) FILTER … list(...) FILTER` aggregate, so the `elem+"[]"` construction is exhaustively pinned on both legs.

`go test ./internal/sqlgen -count=1` and pinned lint green; production files untouched (verified empty diff after the mutation restore).

Issue #348, opened 2026-08-03
Post-#251 polish: Timings rewind on retry, two missing test pins, HTTP-visible partial marker decision


iceboundrock:
Post-#251 polish items from the final branch review of the partial-read-resilience PR (all adjudicated non-blocking; grouped here so they don't scatter):

1. **`rewind` does not rewind `Timings`** (`internal/federated/duckdb_query.go`, `markExecutionPlan`/`rewind`): `Timings` is a merged map, not an append log, so after a successful corrupt-exclusion retry a single request can publicly report **both** `plan_cache_miss` and `plan_cache_hit` (`Timings` IS projected through `toExecutionPlan`, unlike Notes). Either rewind timing keys written by the failed pass or namespace them per pass. The code comment documents this gap and references this issue.

2. **Pre-pass-1 plan survival is unpinned** (`internal/federated/corrupt_retry_seam_test.go`): replacing `markExecutionPlan`'s body with `return executionPlanMark{}` (rewind-to-empty, which would destroy a caller's pre-pass sources, e.g. the pagination-recorded source) leaves the whole unit package green — the seam test asserts only `Routing.UseDuckDB`, which `rewind` can never touch. Add an assertion that a caller-recorded pre-pass Source/Note survives the rewind.

3. **Half-open probe + confirmed-corruption accounting has no unit pin**: the trace is correct (corrupt branch `ReleaseProbe` frees the reserved slot; the retry's own `Allow` re-reserves; every post-Query path resolves the probe) but only the e2e exercises it. A unit test with an open breaker entering half-open, first query hitting confirmed corruption, should pin: probe released, retry admitted, `RecordSuccess` closes.

4. **(Product decision) HTTP callers get a shorter page with no partial marker**: the #251 exclusion note lives in plan `Notes`, which `toExecutionPlan` deliberately drops (#301/#306 adjudication). An API consumer sees fewer rows + smaller `total_records` with no explanation unless they are a Go embedder. If product wants an HTTP-visible degraded-partial marker, it needs a sanctioned public field (same class as the degraded-fallback `Routing.Reason`, which IS visible).

Context: #251, its PR, and `docs/federated-query/design.md` §7.3.

iceboundrock:
[...]

Issue #347, opened 2026-08-03
Silent parquet page corruption can mis-decode row_id without error (no page CRC verification)


iceboundrock:
## Summary

The #251 spike (`docs/superpowers/plans/2026-08-02-issue-251-spike-findings.md`, landing with PR for #251) proved that byte corruption inside a parquet data page can decode **silently into wrong values** — no query error, no stream error, full row count preserved:

- Fixture: 500-row parquet, 4 columns, SNAPPY, written by DuckDB v1.4.5 (repo-pinned `duckdb-go/v2 v2.5.6`).
- 64-byte XOR at 10% / 25% / 50% of the file (inside the `row_id` column chunk): **drain returns `OK rows=500`** — and a controlled-copy diff shows `row_ids_lost=5 row_ids_invented=5`. Five real UUIDs vanished, five fabricated ones appeared.
- The same damage at 75% (`changed_at`, compressible ints) raises `Snappy decompression failure`; at 90% (footer) raises `TProtocolException`. **Detection is a function of which bytes are hit, not of whether the file is corrupt.** Incompressible columns (UUID `row_id`) are the most likely to corrupt silently.
- Parquet's per-page CRC is optional; DuckDB v1.4.5 does not write or verify it, and `parquet_metadata()` exposes **no checksum column** — there is no integrity signal to fall back on at read time.

## Impact

`row_id` drives the federated dirty-set anti-join (the tiering design's consistency mechanism). A mis-decoded `row_id` means:
- a hot-tier row can fail to mask its stale S3 counterpart (lost id) → stale row surfaces;
- a fabricated id matches nothing → phantom row surfaces.

Silent wrong answers, not a loud failure.

## Scope

Out of scope for #251 (partial-read resilience): that mechanism reacts to **scan-detectable** failures only. Silent mis-decode is invisible to every reader — including today's all-or-nothing scan — so it is a pre-existing integrity gap independent of #251, documented in `docs/federated-query/design.md` §7.3 (Limitation).

## Candidate directions (to evaluate, not a design)

1. Writer-side parquet page CRCs — upstream/writer capability question first: confirm the DuckDB COPY writer (CDC exporter path) can emit page checksums at all.
2. Manifest-level content checksums (e.g. object ETag/SHA recorded at flush, verified opportunistically or by the #203 reconcile path).
3. Reader-side verification when DuckDB exposes CRC checking.
[...]

PR #349, opened 2026-08-03
feat(federated): #251 partial-read resilience — one corrupt parquet no longer fails the whole scan

### LINKED ISSUE #251: Partial-read resilience: one corrupt parquet fails the whole federated scan
Follow-up from #187 (PR #249), scenario 7 characterization.

`TestManifestConsistency_OneGoodOneBadFile` pins today's behavior: `read_parquet` over the resolved path set has no `ignore_errors`/`union_by_name`, so one corrupt delta fails the **entire** scan. Good parquet data is recovered only via the degraded Postgres-only fallback — which is complete today solely because flush never deletes PG rows, a retention coincidence the tiering design intends to break. Once PG retention is bounded, a single corrupt cold file makes all warm/cold data unreadable with no partial recovery.

## Ask

Evaluate per-file resilience for the DuckDB read path:

- `read_parquet(..., ignore_errors := true)` semantics — does it skip corrupt files wholesale or also mask row-level corruption in otherwise-good files? (The latter would silently drop rows — the #187 scenario-2 class of failure — so a blind flag is probably wrong.)
- per-file preflight (footer probe) on the resolved manifest list, excluding bad files and surfacing a degraded-partial marker on the execution plan (mirroring `recordDegradedFallbackPlan`), so partial results are *loud*
- interaction with `ErrFederatedReadFailed` classification and the circuit breaker (a permanently corrupt file must not trip the breaker forever)

Acceptance sketch: scenario-7 setup returns the good file's rows plus hot rows, with the plan recording the excluded object — instead of all-or-nothing failure. The #187 test flips from characterization to contract when this lands.

iceboundrock:
Closes #251.

## What

One corrupt parquet object in a manifest-authored scan set no longer fails the whole federated scan. The query answers from the readable remainder plus hot rows — loudly partial, correctly classified, and breaker-neutral. `TestManifestConsistency_OneGoodOneBadFile` (#187 scenario 7) flips from characterization to contract.

## Mechanism: verify-and-exclude, not a reader flag

Scan failure → `MissingIn` classification **first** (a manifest-listed-but-missing object stays non-degradable `ErrParquetSetInconsistent`, never retried) → per-file full-drain verification (`verifyParquetPaths`; only source-authored sets, ≥2 objects, and only when at least one object verifies readable — all-objects-unreadable is store sickness, not corruption) → confirmed-corrupt objects into an engine-level TTL cache (default 5 min, `WithCorruptPathRetention`) → resolve-time exclusion in `resolveParquetPaths` (all-corrupt guard scans the full set so total loss stays loud) → exactly one retry in `ExecuteDuckDBFederatedQuery`.

- **Loud**: the internal execution plan names every excluded object via `NotePartialParquetExclusion` (Notes stay internal per the #301/#306 adjudication; HTTP-visible marker is a product decision → #348 item 4). On a successful retry the plan is rewound to a pre-pass-1 mark so it describes only the answering pass (Routing preserved).
- **Breaker-neutral**: confirmed corruption releases the half-open probe instead of recording a failure — the verification pass just proved the engine healthy on the other objects. A permanently corrupt object cannot hold the breaker open (pinned by breaker-armed `WithBreaker(1, 1m)` e2e envs where the first post-corruption query is the discriminator). Store-wide unreadability still records.
- **Plan-cache safe for free**: the excluded set changes `parquetPaths`, so the scope-v2 hash re-keys — pinned by a seam test that reproduces the would-be poisoning under mutation.

## `ignore_errors` verdict (issue bullet 1)

Does not exist on `read_parquet` in the pinned DuckDB v1.4.5 (bind-time Binder Error; it's a `read_csv` parameter) — and remains disqualified in principle: wholesale skipping is silent (no marker, no classification, no attribution). Full spike findings: `docs/superpowers/plans/2026-08-02-issue-251-spike-findings.md`; verdict comment on #251.

## Spike discovery → #347

The spike also proved detection is **byte-position-dependent**: corruption inside an incompressible column chunk (UUID `row_id`) can decode silently into different valid values (5 row_ids lost + 5 fabricated, zero errors; v1.4.5 exposes no page CRC). That class is invisible to every reader including today's all-or-nothing scan — pre-existing, out of scope here, documented in design.md §7.3 and tracked as #347. #251 covers scan-detectable corruption; the scenario-7 harness corruption provably is (e2e green).

## Out of scope

- Hint-authored path sets (`S3ParquetPathTemplate`): operator-pinned, all-or-nothing preserved (pinned by `TestParquetCorruption_WrongSchemaFile_GlobHint` and a unit test).
- Glob path sets (manifest fallback): exclusion is inexpressible in a glob; behavior unchanged.
- HTTP-visible partial marker: #348.
- Silent mis-decode integrity gap: #347.
- One documented interaction: a corrupt-cached path escapes the `MissingIn` probe during its TTL, so a delete-inside-TTL surfaces as inconsistency only after expiry (design.md §7.3).

## Verification

- Flipped e2e green: `OneGoodOneBadFile` (mid-stream branch, breaker-armed, degraded-off first + cache path + 8-query neutrality loop + degraded-on) and new `OneGoodOneTruncatedFile` (Query-error branch, degraded-on first); scenario-2 `MissingListedParquet` regression untouched and green.
- Regression sweep: corruption+permission suites 7/7 (single-file sets keep loud all-or-nothing via the all-corrupt guard); `make test` 31 packages; `make lint` (pinned v1.64.8) clean; `make test-e2e-production` green (137.8s); federated `-tags=e2e -short` + benchmark green.
- Unit seams pinned by mutation-verified tests: MissingIn-before-verification ordering, drain-actually-drains + SQL shape, exact corrupt-set attribution, exclusion→SQL+scope-hash+note, retry boundedness + fresh-slice, plan rewind.
- Review process: per-task review + scoped re-reviews (2 e2e findings, 2 test-strength findings, 1 seam-coverage finding, 3 execute-path findings — all fixed and re-verified), final whole-branch review on the full 14-commit diff with ledger triage.


iceboundrock:
## Review — Findings

Scope: full 14-commit diff of `issue-251-partial-read-resilience` (3350 lines). Verified locally: `go test ./internal/federated/` (incl. all new seam tests), `go vet ./internal/federated/`, full `go test ./...` (31 packages), and `go vet -tags e2e ./internal/e2e_harness/production/`. e2e (Docker/testcontainers) and pinned `golangci-lint v1.64.8` were not runnable in this environment; see the PR's CI claims for those.

Findings are ordered by importance. None are blocking.

### F1 — `ReleaseProbe` is called *after* the dependency was touched, contradicting its own contract (doc drift)
`internal/federated/circuit_breaker.go` documents `ReleaseProbe` as: *"for callers admitted by Allow that then failed BEFORE touching the dependency — a misconfiguration caught during path resolution, invalid caller input — so they learned nothing about its health."* But `failDuckDBScan` (`duckdb_query_execute.go`) now calls it on the **corrupt-confirmed path**, i.e. after a real scan already failed against DuckDB/S3. The *behavior* is well-justified (verification drained every other object through the same engine, so the engine was proven healthy — the code comment says exactly this), but the breaker method's own contract no longer describes the call site. This is a documentation/contract drift, not a logic bug. Recommend updating `ReleaseProbe`'s doc comment to cover the second (post-verification) use, so future readers don't "fix" the call site back to `RecordFailure`.

### F2 — Verification cost amplification on the failure path (cost risk)
`verifyParquetPaths` drains **every** object in the failed set, one-by-one, sequentially on the single pooled connection (#285), then the retry issues a second full scan. Worst case: a permanently corrupt object inside a large cold tier (say 100 objects) costs the first post-corruption query **100 full-object drains + 2 scans**, and — because of the 5-minute TTL — **the same again on the first query after every TTL expiry**. The design explicitly bounds this to the failure path ("cost is bounded by failures, not by queries"), which is fair, but it is worth an operator-facing note: expect a bounded periodic failure blip (one per corrupt object per TTL window), and for very large tiers consider a footer-probe pre-filter before the drain (a cheap negative discriminator for the truncation class, per the spike's own Q2 data).

### F3 — "Loud" holds for internal plan consumers only; API callers get a silent partial answer
The issue asked for a degraded-partial marker "mirroring `recordDegradedFallbackPlan`" — and `recordDegradedFallbackPlan` rewrites `Routing` (which **is** HTTP-visible). This PR instead records `NotePartialParquetExclusion` in `Notes`, and `toExecutionPlan` (`internal/entity_query_service.go`) explicitly drops `Notes` at the HTTP boundary. So an HTTP `advanced_query` caller receives a **partial** page with **no `partial_result` flag and no routing/marker of any kind** — the partialness is invisible to the API consumer. This is acknowledged and deferred to #348 item 4, but I flag it because it is a genuine deviation from the issue's acceptance phrasing ("so partial results are *loud*"). If a partial answer landing on an API caller with zero signal is acceptable for this merge (pending #348), fine — otherwise this should gate. The e2e only proves loudness at the embedder level, which is exactly the boundary that drops the note.

### F4 — Stale corrupt-cache entries are never evicted unless the path is rescanned
`corruptParquetCache.Split` evicts expired entries **only for paths present in the input list**. A cached path that is retired by compaction (or a manifest reconcile that stops listing it) is never passed through `Split` again, so its entry lingers in `expires` until process restart. Bounded in practice (corrupt objects are rare) and harmless to correctness (the entry only matters if the path reappears, and then it's evicted on sight), but a TTL sweep or a size cap would make the map self-maintaining. Minor.

### F5 — Second-retryable-failure message names pass-1's corrupt set, not pass-2's
`ExecuteDuckDBFederatedQuery` wraps a second retryable failure as `"retry after excluding corrupt parquet %v: %w"` using `retry.Corrupt` from the **first** pass. If mid-flight corruption produces a *different* set on the retry, the surfaced message names the first set while the wrapped cause carries the second. Accuracy nit on an already-loud error path; consider using the error's own corrupt set when available.

iceboundrock:
## Review — Spec

Spec source: issue #251 "Partial-read resilience: one corrupt parquet fails the whole federated scan".

### Issue bullets → implementation

| Issue bullet | Status | Evidence |
|---|---|---|
| Evaluate `read_parquet(..., ignore_errors:=true)` semantics | **Verdict delivered** | Spike findings doc Q1: does not exist on `read_parquet` in pinned DuckDB v1.4.5 (bind-time Binder Error; `ignore_errors` is a `read_csv` parameter). Disqualified in principle on silence grounds. Matches the issue's own warning that "a blind flag is probably wrong". |
| Per-file preflight (footer probe) on the resolved manifest list, excluding bad files + degraded-partial marker | **Implemented with a justified deviation** | Footer probe is explicitly **not** used for preflight because the spike proved `DESCRIBE` succeeds on mid-file corruption (Q2) — preflight would miss exactly the scenario-7 class. Instead: failure-path per-file drain verification → TTL cache → resolve-time exclusion → one retry. The exclusion is surfaced via `NotePartialParquetExclusion` on the internal plan. |
| Interaction with `ErrFederatedReadFailed` classification and the circuit breaker (permanently corrupt file must not trip the breaker forever) | **Implemented** | `MissingIn` classification runs **first** (missing = `ErrParquetSetInconsistent`, non-degradable, breaker-worthy, never retried — the #187 scenario-2 contract is preserved). Confirmed per-file corruption is breaker-neutral (`ReleaseProbe`, not `RecordFailure`); all-objects-unreadable still records. Both orders are pinned by tests. |

### Acceptance sketch

> "scenario-7 setup returns the good file's rows plus hot rows, with the plan recording the excluded object — instead of all-or-nothing failure. The #187 test flips from characterization to contract."

- **Met.** `TestManifestConsistency_OneGoodOneBadFile` flips from characterization to contract: asserts the partial row set (good file + hot, exact set equality), `Total` agrees with the page, the plan names the excluded object via `assertCorruptExclusionNote`, DuckDB routing survives an 8-query loop under a breaker-armed (threshold 1) env, and degraded mode stays DuckDB-routed instead of falling back to Postgres. The new `OneGoodOneTruncatedFile` covers the footer-dead error branch with inverted degraded-mode ordering.

### Spec deviations / gaps

1. **Footer-probe preflight → failure-path verify-and-exclude.** The issue's sketch said "per-file preflight (footer probe)". The implementation verifies on the failure path instead, because the spike demonstrated the probe cannot detect scenario-7's mid-file corruption. This is a well-evidenced, documented deviation (design.md §7.3, spike Q2/Q3) and is the correct call given the data.
2. **"Loud" partialness is embedder-only.** The marker is a `Notes` entry, dropped at the HTTP boundary. The issue's acceptance says "the plan recording the excluded object" — the *internal* plan does record it, so the letter of the acceptance is met. The broader "so partial results are *loud*" intent is only partially met for HTTP consumers (see Findings F3). Deferred to #348 item 4.
3. **`ignore_errors` future-upgrade stance** is consistent with the issue's own caution: depending on a flag whose future appearance would silently change read semantics is rejected.

### Scope out-of-scope handling

- Hint-authored path sets (`S3ParquetPathTemplate`): all-or-nothing preserved, pinned by `TestResolveParquetPathsHintSetNeverFiltered` + `TestFailDuckDBScanHintPathsNeverVerified`. Matches the issue's "operator-pinned" rationale.
- Glob sets: exclusion inexpressible in a glob; behavior unchanged and documented.
- Silent mis-decode integrity gap: correctly carved out (invisible to every reader incl. the pre-#251 all-or-nothing scan) and tracked as #347; documented in design §7.3. Not a regression introduced here.

**Spec verdict:** all three issue bullets and the acceptance sketch are implemented. Two deviations exist (failure-path verification instead of footer-probe preflight; embedder-only loudness), both documented and justified, with the HTTP-marker gap explicitly tracked (#348). No missing requirement was found.

iceboundrock:
## Review — Standards

Authority: `coding-standard.md` + `AGENTS.md`.

### File & function size limits — PASS
- Source files all under the 500-line cap. `duckdb_query.go` (456), `duckdb_query_execute.go` (188), `duckdb_query_resolve.go` (109), `parquet_verify.go` (71), `corrupt_paths.go` (59). The retry logic is properly extracted into `ExecuteDuckDBFederatedQuery` / `collectDuckDBFederatedQuery` rather than bloating one function.
- Functions under 100 lines. The largest new one is `resolveScanSources` (75 lines) and `StreamDuckDBFederatedQuery` stays within the cap. `failDuckDBScan`/`confirmCorruptPaths`/`verifyParquetPaths` are small and flat.

### Error handling — PASS
- All new error paths wrap with context: `fmt.Errorf("%s: %w: %w", op, classified, cause)`, `fmt.Errorf("open parquet %s: %w", ...)`, `fmt.Errorf("retry after excluding corrupt parquet %v: %w", ...)`. No bare `return err` introduced.
- Sentinel/typed error semantics respected: `corruptParquetRetryError` implements `Unwrap` to keep the `ErrFederatedReadFailed` classification chain, so non-retrying callers (direct Stream consumers) degrade exactly as before; matching is via `errors.As`/`errors.Is`, never string comparison.

### Structure / early returns — PASS
- Early returns, flat structure, ≤2 levels of nesting in the hot paths (`executeAndStreamDuckDB`, `resolveScanSources`). The retry loop is a single guarded `if errors.As` block, not a loop — bounded retry is structurally enforced, not just commented.

### Naming & cohesion — PASS
- Verb-first helpers (`verifyParquetPaths`, `drainParquet`, `confirmCorruptPaths`), noun types (`corruptParquetCache`, `corruptParquetRetryError`, `scanSources`, `executionPlanMark`).
- The `scanSources` struct is a good cohesion win: it replaces a 5-value tuple return from `resolveScanSources`, making the "what was resolved" context explicit.

### New-file placement — PASS
- New files are cohesive: `corrupt_paths.go` (cache), `parquet_verify.go` (verification + retry error), `duckdb_query_resolve.go` (storage pre-flight). Keeps `duckdb_query.go` under the cap and mirrors the repo's existing split (`duckdb_query_build.go`, `duckdb_query_helpers.go`).

### Godoc / exported-surface discipline — PASS
- Every exported symbol carries godoc: `NotePartialParquetExclusion`, `WithCorruptPathRetention`, the constant's doc explains the internal-only boundary.
- `defaultCorruptPathRetention` and the `d <= 0` disable semantics are documented, including the fail-open behavior.

### Test conventions — PASS
- Mutation-verified seam tests (`TestCorruptExclusionReachesScanAndPlanNote`, `TestCorruptExclusionRekeysPlanCache`, `TestFailDuckDBScanMissingObjectStaysInconsistentAndBreakerWorthy`) explicitly document which single-line mutation would make them fail — the repo's strongest test pattern.
- e2e helpers (`readParquetRowIDs`, `setMinus`, `assertIDSetEqual`, `assertCorruptExclusionNote`) are small, reused, and cast UUID→VARCHAR to dodge the #147 raw-bytes issue (correctly referenced).

### Minor standards observations (non-blocking)
- `verifyParquetPaths` and the pre-existing `parquet_schema_validation.go` both re-implement the glob/quote-path skip predicate (`strings.ContainsAny(path, "\"'")` etc.). A tiny shared helper (e.g. `isVerifiableParquetPath`) would remove the duplication; not worth blocking.
- `executionPlanMark.rewind` truncates `Sources`/`Notes` but intentionally leaves `Timings` merged — the asymmetry is clearly documented as #348, but the plan's "truthful by construction" claim is slightly weakened by it (a retried request can report both `plan_cache_hit` and `plan_cache_miss` in Timings). Acceptable and tracked; noted here for the record.

**Standards verdict:** conforms. No hard violations of the load-bearing rules (size caps, error wrapping, early returns, exported-symbol documentation).

iceboundrock:
## Review — Assessment

**Overall: strong work — approve with minor findings (F1–F5), none blocking.**

The PR is a genuinely well-engineered answer to a hard problem. The design's two hardest calls are both backed by evidence rather than assumption:

1. **Verify-and-exclude instead of a reader flag or footer probe.** The spike finding (byte-position-dependent decode; `DESCRIBE` passes on mid-file corruption) is what made the footer-probe preflight impossible and `ignore_errors` moot — and the design correctly moved verification to the failure path where it is *provably* sound: the solo drain reads a superset of the projected columns, so any failure the set scan can detect is deterministically reproduced per-file. Zero healthy-path overhead follows as a free consequence.
2. **Breaker-neutrality for confirmed corruption.** Justified empirically (verification just read every other object through the same engine) and structurally sound: the all-objects-unreadable guard (`len(corrupt) >= len(paths)`) keeps store-wide sickness on the `RecordFailure` path, so the exception is narrow and principled, not a blanket forgiveness.

Strengths I would highlight:
- The **classification-order contract** (`MissingIn` first, then verification) is the right ordering and it is *pinned by a biting test* — `TestFailDuckDBScanMissingObjectStaysInconsistentAndBreakerWorthy` constructs a scenario where reordering would silently swallow an #187 scenario-2 inconsistency as corruption. That's exactly the kind of ordering contract that normally rots.
- The **plan-rewind design** (`markExecutionPlan`/`rewind`) is a careful resolution of a genuinely subtle diagnostic bug (the plan must describe only the answering pass, Notes never reach HTTP, so truthfulness-by-construction is the only option). The edge cases (routing survives, exclusion note re-recorded, single `duckdb template rendered` source) are all tested.
- **Plan-cache safety for free** via scope-v2 re-keying is correct and pinned by a real-cache reuse test (`TestCorruptExclusionRekeysPlanCache`), which is the class of regression that would otherwise only surface in production.
- The **e2e suite is meaningfully strengthened**: scenario 7 flips to contract, a new footer-dead sibling test covers the other error branch, and both deliberately arm the breaker at threshold 1 so neutrality is falsifiable rather than assumed.
- **Boundedness is structural**: one retry, not a loop; the fresh-slice-per-pass guarantee makes the retry safe after a mid-stream failure delivered partial rows.

Areas that would benefit from follow-up (all tracked or non-blocking):
- **HTTP-visible loudness** is the one real gap against the issue's spirit (Findings F3): API consumers receive a partial page with no marker. It is honestly documented and deferred to #348, but I'd treat it as the first follow-up rather than letting it lapse.
- The **failure-path verification cost** (F2) deserves an operator note; for large cold tiers the first query in each TTL window can be expensive.
- The **`ReleaseProbe` doc drift** (F1) is a small but real contract inconsistency worth a one-line fix in this PR or the immediate next.

**Do not auto-merge** (per repo rule). After merge: clean up the branch/worktree, fast-forward main, close #251 once #347/#348 follow-ups are filed, per the PR rules in AGENTS.md.

iceboundrock:
## Review — Observations

Non-blocking notes from reading the full diff and docs.

1. **`scanSources` struct is a nice refactor.** Replacing the 5-value tuple return of `resolveScanSources` with a struct not only reads better but makes the corrupt-exclusion leg of the flow (paths + provenance + excluded set travel together) structurally visible. Low-risk refactor, good side effect.

2. **The "cached object escapes `MissingIn` during TTL" interaction is the most subtle documented consequence.** A confirmed-corrupt object that is subsequently *deleted* from storage won't classify as manifest inconsistency until its TTL entry expires and it is rescanned. The design explicitly calls this out in §7.3 and the PR body. Correct and bounded, but it means the `MissingIn` probe is (by design) blind to a specific window of inconsistency — operators reading the logs should know the 5-minute default retention window.

3. **Timing asymmetry on retry.** Because `rewind` truncates `Sources`/`Notes` but leaves `Timings` merged, a retried request can carry both `plan_cache_hit` and `plan_cache_miss` in `Timings`. The plan's Notes will correctly show only the answering pass's cache outcome, so an HTTP consumer reading Timings could momentarily be confused. Tracked as #348; noted for the record.

4. **`errors.As` vs `errors.Is` for the retry signal is the right tool.** `corruptParquetRetryError` is wrapped in a cause chain (`%s: %w: %w`), so matching it requires `errors.As` — which the code uses — while `ErrFederatedReadFailed` classification survives via `Unwrap` for `errors.Is`. Correct pairing; no string matching anywhere.

5. **The single-connection (#285) ordering constraint is handled with the right trick and a test.** Explicitly closing the failed scan's `rows` *before* verification drains is the only way to avoid a pool deadlock with `SetMaxOpenConns(1)`, and `retryFakeDuck.drainsAtClose` pins the ordering so a future "optimization" can't reintroduce the deadlock.

6. **Benchmark-schema exemption is correctly scoped.** `isBenchmarkSchemaID` skipping the pre-read validator is pre-existing (#255) and the PR's `execFailQuery`/`coldTierQuery` tests stay within the production path. No cross-contamination observed.

7. **Documentation quality is high.** design.md §7.3 reads like a proper ADR: mechanism chain, why-a-drain-not-a-probe, why-not-`ignore_errors`, loudness contract, breaker contract, scope, the TTL/probe interaction, and the #347 known-limitation all in one place. The spike-findings doc preserves raw logs — the right way to record "we tested the thing and it doesn't work as assumed."

8. **Commit hygiene.** 14 commits, each scoped (`test(e2e):`, `feat(federated):`, `docs(federated):`), review-round fixes labeled, and the PR body's review-process ledger ("2 e2e findings, 2 test-strength findings, 1 seam-coverage finding, 3 execute-path findings") is consistent with what the code shows. `Co-Authored-By` on doc commits is fine.

iceboundrock:
## Review — Verification

What I ran in a clean worktree at `origin/issue-251-partial-read-resilience` (commit `6b4d1eb`):

```
# Full unit suite (mirrors `make test` package set, without e2e tags)
go test ./...                       → all 31 packages PASS (no failures)

# Federated package focused run — all new tests green:
go test ./internal/federated/ -run 'Corrupt|Verify|Retry|FailDuckDBScan|ResolveParquetPaths|RecordCorrupt' -v
  TestCorruptExclusionReachesScanAndPlanNote         PASS
  TestCorruptExclusionRekeysPlanCache                PASS
  TestCorruptExclusionNoteAbsentOnFullScan           PASS
  TestCorruptParquetCacheSplitExcludesUnexpired      PASS
  TestCorruptParquetCacheEntryExpires                PASS
  TestCorruptParquetCacheNilSafe                     PASS
  TestCorruptRetryAnswersFromReadableRemainder       PASS
  TestCorruptRetryPlanDescribesOnlyTheRetryPass      PASS
  TestCorruptRetryStopsAfterOneRetry                 PASS
  TestFailDuckDBScanConfirmedCorruptionSkipsBreakerAndRetries  PASS
  TestFailDuckDBScanMissingObjectStaysInconsistentAndBreakerWorthy PASS
  TestFailDuckDBScanAllPathsUnreadableIsEngineSickness PASS
  TestFailDuckDBScanHintPathsNeverVerified           PASS
  TestResolveParquetPathsExcludesCachedCorrupt       PASS
  TestResolveParquetPathsAllCorruptKeepsFullSet      PASS
  TestResolveParquetPathsHintSetNeverFiltered        PASS
  TestRecordCorruptExclusionNotesExcludedObjects     PASS
  TestRecordCorruptExclusionEmptyIsNoop              PASS
  TestVerifyParquetPathsFlagsOpenAndStreamFailures   PASS
  TestVerifyParquetPathsSkipsUnverifiableEntries     PASS
  TestVerifyParquetPathsCancelledContextVerifiesNothing PASS
  TestCorruptParquetRetryErrorChain                  PASS

# Static checks
go vet ./internal/federated/                        → clean
go build ./internal/federated/ ./internal/e2e_harness/production/ → OK
go vet -tags e2e ./internal/e2e_harness/production/ → clean
```

Environment used (per Makefile/AGENTS.md guidance to avoid sandbox cache/VCS errors):
`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false`

**Not executed in this environment:**
- The Docker/testcontainers e2e suites (`internal/e2e_harness/...`, `-tags=e2e`), incl. `TestManifestConsistency_OneGoodOneBadFile` and `OneGoodOneTruncatedFile`. The production harness compiles and vets clean; I could not spin up the cluster here. Relying on the PR's stated `make test-e2e-production` (137.8s) + federated `-tags=e2e -short` green.
- `golangci-lint v1.64.8` (not installed in this environment; CI runs it).
- `make benchmark-*` suites.

Claim consistency note: the PR body's verification claims (31 packages, new seam tests, 7/7 corruption+permission regression sweep, flipped e2e) are consistent with what the code and the local run show. I did not find a claimed test that doesn't exist or a test that fails locally.

iceboundrock:
## Review — Summary

**Verdict: approve with minor findings — no blocking issues.**

- **Standards:** conforms to `coding-standard.md` — size caps respected, error wrapping with context everywhere, early returns, exported symbols documented, mutation-verified tests. Minor DRY nit on the path-skip predicate (not blocking).
- **Spec:** all three issue #251 bullets implemented; the footer-probe-preflight suggestion was correctly replaced by failure-path verify-and-exclude (the spike proved the probe can't detect scenario-7 corruption); the scenario-7 e2e flips from characterization to contract. One gap against the issue's spirit: partialness is "loud" only on the internal plan, not the HTTP response (deferred to #348 item 4).
- **Findings (5, non-blocking):** F1 `ReleaseProbe` doc no longer matches its post-verification call site; F2 failure-path verification cost amplification for large tiers + periodic TTL-window blips; F3 HTTP-visible partial marker missing (tracked #348); F4 corrupt-cache entries evicted only on rescan; F5 retry-failure message names pass-1's corrupt set.
- **Verification:** local `go test ./...` all green (31 packages); all 22 new federated seam tests pass; `go vet`/`go build` (incl. `-tags e2e`) clean. Docker e2e and pinned lint not runnable here.

**Bottom line:** the mechanism (verify-and-exclude with one bounded retry, breaker-neutral confirmed corruption, plan-rewind, plan-cache re-keying) is sound and unusually well-tested. Merge after the PR author's usual self-merge rule, then file/close follow-ups: #347 (silent mis-decode) is already tracked; #348 item 4 (HTTP-visible partial marker) is the one I'd chase first.

iceboundrock:
## Response to review findings — fixes in `54a57a4`

**F1 (ReleaseProbe doc drift) — fixed.** The method's doc comment now covers both call classes: pre-execution callers that learned nothing, and the one post-execution caller whose failure indicts a specific object rather than the dependency — per-file verification having just drained every other object through the same engine is live proof of health, which makes `RecordFailure` dishonest while the failed query makes `RecordSuccess` equally so. The contract and the `failDuckDBScan` call site now describe each other, so a future reader has no basis to "fix" it back to `RecordFailure`.

**F2 (verification cost amplification) — operator note added; pre-filter pushed back.** design.md §7.3 now carries an explicit "Operator note — failure-path cost" paragraph: ~N solo drains + a second scan on the first query after a corruption, recurring once per TTL window per corrupt object until compaction/reconcile retires it. On the footer-probe pre-filter suggestion: it would not reduce this bill. Footer-dead (truncation-class) objects already fail the drain at open time — `drainParquet`'s `Query()` bind reads the footer, so for that class the drain *is* the footer probe at the same cost. The dominant term is draining the healthy objects, and no metadata-only probe can certify a healthy object readable (that is the spike's Q2/Q3 core result). The note documents this reasoning so the pre-filter doesn't get re-proposed later.

**F3 (HTTP-visible partialness) — deferred to #348 item 4, per this review's own verdict (non-blocking).** Two reasons not to bolt it into this PR: `recordDegradedFallbackPlan` rewrites `Routing` because the route *actually changed* — here it didn't (still DuckDB), so overloading `Routing.Reason` would misuse the one public field whose semantics are "why this route"; and any public partial-result surface is a new API contract (which field, what granularity, does it name counts or keys) that deserves its own review rather than riding in here. Agreed it is the first follow-up to chase — noted as such on #348.

**F4 (stale cache entries) — fixed.** `Add` now sweeps all expired entries before recording new ones (`corrupt_paths.go`), so an entry whose path is never rescanned (retired by compaction or reconcile) is evicted by the next corruption event instead of lingering until restart. Chose sweep-in-Add over a timer/goroutine or size cap: `Add` runs only on the failure path, so the full-map walk is free in practice and adds no lifecycle to manage. Pinned by `TestCorruptParquetCacheAddSweepsExpiredEntries`.

**F5 (second-failure message names pass-1's set) — no change, reasoning:** the `%v` in `"retry after excluding corrupt parquet %v: %w"` names the set whose exclusion *triggered* this retry — pass-1's — which is what "retry after excluding X" asserts, and it is accurate. If the retry fails with a second retryable error, that error's own `Error()` prints its own corrupt set (`corruptParquetRetryError.Error` includes `Corrupt`), so the surfaced chain reads "retry after excluding [pass-1 set]: … corrupt parquet objects excluded, retry available [pass-2 set]: …" — both sets present, each attributed to its pass. Rewriting the outer `%v` with pass-2's set would lose the pass-1 attribution and make the message claim an exclusion that never drove a retry.

Verification for `54a57a4`: `go test -race -count=1 ./internal/federated` green (incl. the new sweep test), `go build ./...`, gofmt, and pinned `golangci-lint` all clean. Doc/comment changes carry no behavior beyond the F4 sweep.

iceboundrock:
## Review — Findings (round 2)

Scope: current PR head `54a57a4`, reviewed against `main...PR head`. The previous F1/F2/F4 findings appear addressed by the latest commit; these are remaining or newly exposed issues.

### R2-1 — High: every per-file drain error is treated as permanent corruption

`verifyParquetPaths` appends a path for any `drainParquet` error (`internal/federated/parquet_verify.go:23-39`). There is no distinction between deterministic corruption and a transient object-level S3/network timeout. `failDuckDBScan` then caches that path for the retention window and calls `ReleaseProbe` (`internal/federated/duckdb_query_execute.go:107-112`).

Therefore a temporary failure affecting one object can be converted into a successful but silently short response for up to five minutes, while also avoiding `RecordFailure` and breaker protection. The issue requires permanent corruption to be breaker-neutral, not arbitrary per-file I/O failures. Add a deterministic error classification/retry policy, or keep uncertain verification failures on the ordinary `RecordFailure` path; add a regression test where one drain fails transiently and later succeeds.

### R2-2 — Medium: `ReleaseProbe` can clear another caller's half-open probe

`CircuitBreaker.ReleaseProbe` has no ownership token (`internal/federated/circuit_breaker.go:81-110`). A query admitted while the breaker was closed can remain in flight while another failure opens the breaker and a newer caller acquires the half-open probe. If the older query later takes the confirmed-corruption path, its `ReleaseProbe` clears the newer caller's `probing` flag. A third caller can then be admitted concurrently with the real probe.

The new post-execution use of `ReleaseProbe` makes this race reachable; the existing stale-caller note only describes `RecordFailure`. Return/validate a probe token, or otherwise make release conditional on the reservation owner, and cover the interleaving with a concurrency test.

### R2-3 — Medium, known spec gap: HTTP callers still cannot observe partialness

The exclusion note is recorded only in internal `ExecutionPlan.Notes` (`internal/federated/duckdb_query_helpers.go:447-461`), while `toExecutionPlan` deliberately drops notes at the public boundary (`internal/entity_query_service.go:143-149`). Thus an HTTP caller can receive a partial page with no `partial_result`, warning, or equivalent marker. This is documented as deferred to #348, so it may be accepted for this PR, but the issue's “surfacing a degraded-partial marker” requirement is not met for API consumers.

iceboundrock:
## Review — Standards

Authority: `AGENTS.md` and `coding-standard.md`.

### S2-1 — Context-free error returns remain in the changed path

`internal/federated/duckdb_query.go:76-77` returns `err` directly from `ExecuteDuckDBFederatedQuery`, and `:144-145` does the same from the newly introduced `collectDuckDBFederatedQuery`. `internal/federated/parquet_source.go:55-57` also returns the hint-render error without operation context after the function was changed in this PR.

This violates `AGENTS.md:68-70` and `coding-standard.md:20-33`, which require every error path to add context with `%w` rather than returning a bare error. These are non-blocking quality findings, but should be fixed while this path is being revised.

### Standards pass

No file/function size or nesting violations were found in the changed Go code. `make lint` and `go vet ./internal/federated` also pass; the finding above is the documented review standard rather than a linter failure.

iceboundrock:
## Review — Observations

1. The latest commit fixes the previous F1 documentation drift: `ReleaseProbe` now documents the confirmed-corruption call site.
2. The failure-path verification cost is now documented, including the N drains plus retry scan and TTL-window recurrence.
3. Expired cache entries are now swept on `Add`, with regression coverage for paths retired without another `Split`.
4. The spike's silent mid-file mis-decode result is correctly preserved as a known limitation and tracked as #347; it is not caused by this PR, but it means the partial-read guarantee is necessarily limited to detectable failures.
5. The documented TTL interaction still allows a deleted, previously-corrupt object to bypass `MissingIn` until expiry because exclusion happens before probing. This is bounded and documented, but it is an intentional exception to the manifest inconsistency contract in `docs/error-handling.md:314-327` and deserves follow-up.
6. `SELECT *` verification is conservative: corruption in a column not used by the federated projection can exclude an otherwise useful object. That tradeoff should remain explicit if verification is later narrowed to projected columns.

iceboundrock:
## Review — Assessment

**Request changes.** The verify-and-exclude design is strong for corruption classes that produce a DuckDB read error, and the latest commit correctly addresses the prior ReleaseProbe documentation drift, adds the operator cost note, and sweeps expired cache entries. The retry is bounded, uses a fresh result slice, preserves the plan cache key boundary, and the missing-object ordering is well pinned.

The remaining R2-1 issue is behavioral: it can turn a transient object read failure into an unmarked/under-counted answer and suppress breaker protection. R2-2 is a concurrency correctness issue in the new post-execution probe release. R2-3 is an acknowledged API-contract gap tracked by #348. I would not approve the PR until R2-1 and R2-2 are either fixed or explicitly redesigned with tests.

Per repository policy, no merge action was taken.

iceboundrock:
## Review — Verification

Verified against PR head `54a57a4` in a clean detached worktree:

```text
go test ./...                         PASS (31 packages)
go test -race ./internal/federated   PASS
go vet ./internal/federated          PASS
make lint                             PASS (pinned golangci-lint)
git diff --check                      PASS
```

The focused tests cover the new retry, cache, missing-object ordering, plan rewind, cache re-keying, and breaker behavior. Docker/testcontainers production e2e and tagged federated e2e were not independently run in this review pass; the PR's stated e2e results were not reproduced locally here.

iceboundrock:
## Review — Spec

Spec source: issue #251, “Partial-read resilience: one corrupt parquet fails the whole federated scan”.

### Requirement mapping

| Requirement | Status | Assessment |
|---|---|---|
| Evaluate `read_parquet(..., ignore_errors := true)` | Met | The spike records that the parameter is invalid on pinned DuckDB v1.4.5 and rejects a silent reader-level skip. |
| Per-file resilience, exclusion, and a loud marker | Partial | Failure-path verify-and-exclude is a justified deviation from the proposed footer preflight: the spike shows footer probes and full drains cannot detect silent mid-file corruption. For scan-detectable corruption, the internal plan names excluded objects and the scenario-7 contract is covered. However, R2-1 means transient per-file failures can be mislabeled as confirmed corruption, and the marker is dropped before HTTP responses. |
| Correct `ErrFederatedReadFailed` and breaker interaction | Partial | Missing-object classification remains first and confirmed corruption is breaker-neutral. But arbitrary verification errors also take `ReleaseProbe`, so the implementation does not limit neutrality to permanent corruption. R2-2 also weakens probe ownership under concurrency. |
| Acceptance: good-file rows plus hot rows, with excluded object recorded | Met for the internal/e2e contract | The new seam/e2e tests cover readable remainder, fresh retry buffering, plan rewind, cache re-keying, and breaker-neutrality. The public API loudness part remains deferred to #348. |

**Spec verdict:** the core acceptance path works for corruption that DuckDB actually reports, but the error classification is not yet safe enough to claim general partial-read resilience. The known silent mis-decode limitation is explicitly documented and tracked as #347; I am not treating that pre-existing limitation as a newly introduced regression.

iceboundrock:
## Review — Summary

**Verdict: request changes.**

- Findings: 3 behavioral/spec findings remain: transient per-file errors are cached as corruption and bypass the breaker; `ReleaseProbe` lacks ownership and can clear another half-open probe; HTTP consumers receive no partial-result marker.
- Standards: 1 non-blocking context-wrapping finding; size, structure, lint, and vet checks pass.
- Spec: the internal/e2e acceptance path is implemented for scan-detectable corruption, while the public loudness requirement remains deferred and the transient-error classification needs correction.
- Verification: unit, race, vet, lint, and diff checks pass; Docker e2e was not independently run.

The PR has a solid bounded retry and verification structure, but R2-1 can silently shorten results for a transient storage error, so this review does not recommend approval yet.

iceboundrock:
## Response to round-2 findings — fixes in `65b9b58`

**R2-1 (every drain error treated as permanent corruption) — fixed.** Confirmation now requires **two consecutive failed drains** per object (`verifyParquetPaths`): deterministic corruption fails every drain — same bytes, same decode — while a transient object-level fault that reads clean on the immediate re-drain is inconclusive and leaves the query on the ordinary `RecordFailure` path: nothing cached, nothing excluded, breaker accounting intact, non-retryable error surfaces. This is a deterministic policy with no error-text parsing (DuckDB error prose is version-dependent and the repo forbids string-matching errors). Regression tests at both layers: `TestVerifyParquetPathsTransientSingleFailureNotConfirmed` (+ its persistent-failure sibling) and `TestFailDuckDBScanTransientDrainErrorStaysOnFailurePath` (threshold-1 breaker opens, cache stays empty, no retryable error). Residual risk stated in §7.3: a fault persisting across both back-to-back drains of one object still confirms — no read-path policy can fully distinguish that from corruption; the TTL bounds it and store-wide faults keep hitting the all-unreadable guard. The operator cost note now includes the extra re-drain per failing object.

**R2-2 (ReleaseProbe ownership) — fixed.** `Allow` now returns a reservation token (`ProbeToken`; zero = admitted-while-closed, no reservation held) and `ReleaseProbe(token)` frees the slot only when the live reservation matches. The token threads through the `scan` context to the corrupt-confirmed release, and the pre-execution deferred release uses it too — closing the same race for the pre-existing #299 release path, which was equally unscoped. Your exact interleaving is pinned by `TestCircuitBreakerLapsedReservationReleaseCannotClearNewProbe` (reservation lapses → newer caller reclaims → old caller's late release must not clear the new probe) plus `TestCircuitBreakerReleaseProbeScopedToOwner` (zero-token and foreign-token releases are no-ops against a live probe); the suite runs under `-race`. `RecordFailure`'s stale-caller behavior is unchanged and its documented conservatism now explicitly contrasts with the token-scoped release in the type doc.

**R2-3 (HTTP-visible partialness) — unchanged, per this round's own framing** ("documented as deferred to #348, so it may be accepted for this PR"). Priority and the design constraint (don't overload `Routing.Reason`; a public partial-result surface is a new API contract) are recorded on #348.

**S2-1 (context-free error returns) — fixed.** All three cited sites now wrap: `ExecuteDuckDBFederatedQuery` → `"federated read: %w"`, `collectDuckDBFederatedQuery` → `"collect federated rows: %w"`, the hint-render error in `resolveParquetPaths` → `"render parquet path hint: %w"`. `errors.Is/As` contracts unaffected (`%w` throughout).

**Observation 5 (TTL window vs the manifest-inconsistency contract) — documented at the contract's home.** `docs/error-handling.md` `ErrParquetSetInconsistent` now carries the bounded exception explicitly: delete-during-TTL is *delay, not suppression* — the loss surfaces after the entry expires and the object is rescanned, up to one retention window late. Cross-referenced to §7.3.

**Observation 6 (SELECT * conservatism)** — kept explicit in §7.3's superset rationale; narrowing verification to projected columns would trade exclusion precision for the attribution-soundness argument, so any future change there must revisit that paragraph.

Verification for `65b9b58`: `go test -race -count=1 ./internal/federated` green (incl. the four new tests), full `make test` 31 packages, pinned `golangci-lint` exit 0, and the Docker e2e suites re-run locally: `TestManifestConsistency_*` 3/3, `TestParquetCorruption_*` 6/6, `TestParquetPermission_BadCredentials` — 10/10 PASS.

iceboundrock:
## Standards

No hard documented-standard violations found in `b2362276...65b9b58`.

Judgement call: `internal/federated/parquet_verify.go:44` reaches four nested control-flow levels (`for` -> failed drain -> failed re-drain -> cancellation check), exceeding the `coding-standard.md` preference to avoid nesting beyond three levels. Flattening the retry/cancellation flow would improve readability.

## Verification

`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -race ./internal/federated` passed.

iceboundrock:
## Review — Findings (round 3)

No blocking or actionable correctness findings at `65b9b58` after re-reviewing the full `main...PR` diff and the round-2 fixes.

The prior blockers are addressed:
- Two consecutive per-object drain failures are now required before exclusion, so a one-off object failure remains on the normal failure/breaker path.
- Half-open probe release is token-scoped, preventing a lapsed caller from clearing a newer callers reservation.
- Newly touched errors retain contextual wrapping.

Known, explicitly deferred product/API work remains outside this finding set: #348 (HTTP-visible partial-result signal) and #347 (silent parquet mis-decode integrity gap).

iceboundrock:
## Review — Spec (round 3)

**Pass for issue #251.**

- The `ignore_errors` question is evaluated with pinned DuckDB evidence and correctly rejected in `docs/federated-query/design.md`.
- Source-authored multi-object paths are verify-and-exclude candidates; confirmed exclusions are recorded in the internal execution plan.
- `ErrFederatedReadFailed` classification is retained, with missing-manifest-object classification preceding corruption verification. Confirmed corruption is breaker-neutral; inconclusive/transient verification errors remain breaker-worthy.
- The scenario-7 production contract exercises good-file plus hot-row recovery and records the excluded object.

The HTTP-visible marker is an acknowledged #348 API follow-up, not a regression in the issue’s internal-plan acceptance contract.

iceboundrock:
## Review — Standards (round 3)

**Pass.** No hard violation of `AGENTS.md` or `coding-standard.md` found in the final diff.

- Go files and functions remain below the 500/100-line caps.
- New/changed error paths wrap context and preserve sentinel matching through `%w` and `errors.Is`/`errors.As`.
- The round-2 bare-return finding is resolved.
- The token-scoped breaker and retry helpers are cohesively placed and covered by targeted tests.

Non-blocking judgement: `verifyParquetPaths` has nested retry/cancellation branches (`internal/federated/parquet_verify.go:35-49`). It remains short and understandable, but a future small extraction could further flatten the control flow.

iceboundrock:
## Review — Assessment (round 3)

**Approve from a code-review perspective; no merge action taken.**

The final change closes the substantive concerns from round 2 without weakening the bounded retry design:
- confirmation distinguishes a single transient drain failure from repeatable corruption;
- probe release cannot affect another callers half-open reservation;
- normal missing-object consistency behavior remains first in the classification order.

The remaining public partial-result contract belongs to #348. Per repository policy, this review does not auto-merge the PR.

iceboundrock:
## Review — Observations (round 3)

- Two back-to-back failed drains are a sound conservative policy without brittle DuckDB error-text parsing. A fault that persists through both remains intentionally treated as corruption; the bounded TTL and all-paths-unreadable guard limit that residual ambiguity.
- `ProbeToken` makes `ReleaseProbe` safe for the new post-verification path while retaining the existing conservative handling of stale `RecordFailure` calls.
- The latest tests cover both a one-shot transient drain and a lapsed reservation followed by a newer probe.
- The execution plan still exposes excluded paths only to internal consumers; #348 should remain the next product-facing follow-up.

iceboundrock:
## Review — Verification (round 3)

Verified against PR head `65b9b58d9c8be7d3e212bf0cac588da393c0685f` in `.worktrees/issue-251-partial-read-resilience`:

```text
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test -race -count=1 ./internal/federated  PASS
make test                                                            PASS
make lint                                                            PASS
go vet ./internal/federated                                         PASS
git diff --check main...65b9b58                                  PASS
```

Docker/testcontainers e2e suites and benchmarks were not re-run in this review pass. The PR head remained unchanged while verification and comments were prepared.

iceboundrock:
## Review — Summary (round 3)

**Verdict: approve from code review, no blocking findings.**

- Findings: none.
- Spec: issue #251 requirements and the internal scenario-7 acceptance contract are met.
- Standards: compliant; only a minor non-blocking nesting observation.
- Verification: focused race test, full unit suite, lint, vet, and diff check passed.
- Follow-ups: #348 for an HTTP-visible partial-result contract; #347 for the pre-existing silent mis-decode limitation.

No merge action was taken.

Issue #337, opened 2026-07-30
gofmt rewrites doubled-quote prose in internal/cdc/redact_test.go doc comments into typographic quotes


iceboundrock:
Surfaced during PR #335's final review; pre-existing on `main` (reproduces at b70867f, i.e. NOT introduced by #310/#335).

**The trap:** newer gofmt (observed with go1.26.5) rewrites bare `''` pairs in comment prose into typographic quotes. `internal/cdc/redact_test.go` has doc comments describing the doubled-quote forms (`password=''secret''`, `password=''''`) as flowing prose, so `gofmt -d` wants to corrupt exactly the sequences the comments exist to describe. This is the same corruption `internal/redact/connstring.go:39-41` explicitly warns about — and dodges by putting the quote-bearing lines in an indented block, which gofmt leaves alone.

**Why CI doesn't catch or suffer it:** `make lint` (golangci-lint pinned v1.64.8) does not run gofmt over these comments, so CI stays green either way. The risk is a human or agent running `gofmt -w ./internal/...` and silently corrupting the documentation of the #290 redaction contract — the doubled-quote forms in those comments are the spec of what `redactConnStr` must match.

**Fix:** apply the same indented-block defense used in `internal/redact/connstring.go` to the affected doc comments in `internal/cdc/redact_test.go` (move the literal quote sequences onto indented example lines). Verify with `gofmt -d internal/cdc/redact_test.go` → empty diff.

iceboundrock:
## Verification: already fixed on `main` by PR #380 — closing as completed

Picked this up to plan an implementation, but verification against `main` (`b31f026`) shows the fix described here **already landed** as part of the #276 cleanup (PR #380, merged as `b31f026`). PR #380's body does not reference #337, which is why this issue was never auto-closed or linked.

Verified locally with **go1.26.5** — the exact toolchain version named in the issue body.

### Requested fix — applied

`internal/cdc/redact_test.go` now carries all three literal doubled-quote sequences on tab-indented code-block lines, exactly the defense `internal/redact/connstring.go:39-41` uses:

| Line | Comment line (tab shown as `\t`) |
|---|---|
| `redact_test.go:81` | `//\tpassword=''` |
| `redact_test.go:94` | `//\tpassword=''''` |
| `redact_test.go:110` | `//\tpassword=''secret''` |

`redact_test.go:76-79` also records the reason inline, naming #276:

[...]

PR #350, opened 2026-08-03
feat(manifest): #256 schema stamping — column-set stamps in FileEntry, probe-fallback reads

### LINKED ISSUE #256: Manifest schema stamping: record each parquet file's column-set fingerprint in FileEntry
Spun out of #189 (PR #254).

The #189 pre-read validator probes each parquet footer via DESCRIBE once per process (cached; objects are write-once). That is cheap steady-state but re-probes the whole file set on every cold start (e.g. lambda), and the probe result is process-local.

Proposal: at write time (flush export, init export, compaction merge) record each file's column-set fingerprint (and optionally the full name→type map) in `manifest.FileEntry`. Reads then validate the invariant from the manifest and only fall back to footer probing on fingerprint mismatch or missing stamp (pre-existing files). This is a **cache layer on top of** the byte-truth probe, not a replacement — #187's corruption detection requires distrusting metadata.

Needs: manifest format versioning/backfill story for pre-existing entries, and a decision on whether the stamp also feeds the never-flushed-column projection (see the sibling follow-up).

iceboundrock:
Closes #256.

## What

Manifest schema stamping: each parquet writer (CDC flush, CDC init, compaction merge) records the file's footer schema (column name → DuckDB type) in `manifest.FileEntry.Columns` at write time, from a `DESCRIBE` of the bytes it wrote. Federated reads plumb the stamps from `manifest.QuerySource.Paths` through `federated.ParquetSource` into the #189 pre-read validator: a stamp that passes `parquetcheck.Check` short-circuits the footer probe and feeds the #255 column union; a stamp that is absent (every pre-existing entry) or fails the invariant falls back to the probe. Cold starts (e.g. Lambda) on fully-stamped manifests validate and compute #255 augmentation with **zero footer probes**.

## Design decisions

- **Full name→type map, no hash fingerprint.** `parquetcheck.Check` and the #255 union both need the actual columns; a hash could serve neither without a probe. The issue's "optionally the full map" resolves as: the map IS the stamp.
- **Stamps are write-time byte truth.** Writers DESCRIBE the object they wrote (flush/init describe the final key; compaction describes the staged tmp object — tmp→final is a byte-identical `CopyObject`, and the e2e suite pins the *published* object's footer against the stamp). Stamps are never derived from the export SQL projection.
- **Best-effort at write time.** A failed DESCRIBE warns and leaves `Columns` nil (`SizeBytes` precedent); stamping never fails a flush/init/compaction. When the manifest store is disabled, writers skip the probe entirely.
- **Stamps only short-circuit success — failures always require byte truth (#187).** A stamp failing the invariant triggers a probe; only the probe's verdict can fail the query. Corruption detection is untouched: unreadable footers stay inconclusive for the execution-path classifier, and #251 verify-and-exclude still drains real bytes.
- **Stamps feed the #255 union** (the issue's open decision, adjudicated yes). One documented exception to "never fail a healthy object": a Check-passing stamp that under-reports attribute columns causes a loud, classified binder failure that persists until the manifest entry is corrected (unlike the self-healing #255 glob skew). Accepted because stamps come from a write-time DESCRIBE, so this requires a corrupted/tampered manifest — and the failure is loud, never silent data loss. Documented in `Validate`'s contract comment and `docs/federated-query/design.md`.
- **Versioning/backfill = field presence.** `columns,omitempty`; legacy entries decode to nil and behave byte-for-byte as today (pinned by tests). No migration; entries get stamped naturally as compaction rewrites them.

## Correctness bar

`make test-e2e-production` green with new assertions: for every manifest entry written by flush (delta), init (base), and compaction (merged base, incl. the zero-row edge and promotion relabel), `entry.Columns` equals a live `DESCRIBE` of the S3 object through the harness duck session, asserted against the manifest re-loaded from S3 — "indistinguishable from probing", empirically.

## Out of scope (per plan)

- `reconcile --repair` promoted entries stay unstamped (probe fallback covers them) — follow-up candidate.
- Compaction's pre-merge **source** validation stays probe-based (it deletes sources afterwards — highest correctness bar).
- No backfill of existing manifests.

## Follow-up to file

- `.golangci.yml` has no `build-tags: [e2e]`, so the e2e harness (65 files) is invisible to `make lint`; three pre-existing gofmt violations live there undetected.
- Compaction is the only writer that swallows a stamp-probe failure without logging (DuckMerger has no logger) — minor observability gap.

## Verification

- Unit sweep (`make test`) green at branch tip; `make lint` clean (pre-existing `internal/cdc/redact_test.go` gofmt noise is #337, untouched).
- E2E: infra smoke + `make test-e2e-production` (112 PASS, 0 FAIL) with all six stamp==probe assertions logged.
- Two independent whole-branch reviews (per-task gates + final cross-task review); all findings fixed or triaged in-branch.


internal/federated/parquet_schema_validation.go
```diff
@@ -103,6 +130,20 @@ func (v *parquetSchemaValidator) validateConcrete(
 			mergeColumnUnion(union, cols)
 			continue
 		}
+		if stamp, ok := stamps[path]; ok && len(stamp) > 0 {
+			if parquetcheck.Check(path, stamp) == nil {
+				// Clone: the cache outlives the call and must not alias a
```
iceboundrock: P1: this stamp branch only runs `parquetcheck.Check`, which validates the three system columns but never verifies the full `Columns` map against the object footer. A stale or tampered stamp that contains those three columns therefore skips `DESCRIBE` and feeds an incomplete or incorrect #255 union. If the real object is missing a system column, `union_by_name` can still NULL-fill it and silently lose rows; if only an attribute is under-reported, the projection can fail later. This is not byte-truth validation as required by #256/#187. Please either compare the complete stamp with a footer probe (with an appropriate cache key) or make the stamp path unable to bypass byte-truth validation, and add a regression test for a valid-system-column but incomplete stamp.

internal/federated/parquet_schema_validation.go
```diff
@@ -103,6 +130,20 @@ func (v *parquetSchemaValidator) validateConcrete(
 			mergeColumnUnion(union, cols)
 			continue
 		}
+		if stamp, ok := stamps[path]; ok && len(stamp) > 0 {
+			if parquetcheck.Check(path, stamp) == nil {
```
iceboundrock: P1: the validator cache is still keyed only by URI and is checked before the current manifest stamp. That conflicts with the existing `cdc-init` contract: `BuildBasePath` uses deterministic `{minRowID}_{maxRowID}.parquet` keys, and rerunning init overwrites those objects. A server that validated path P, then observes an init rerun/schema change that overwrites P, will keep using the old cached columns forever and ignore the new stamp/footer. This can produce a stale #255 projection or suppress detection of a changed system schema. Please key invalidation by a manifest generation/stamp (or make init output immutable) and add a test that warms the validator, reruns init over the same key, and verifies the new schema is used.

iceboundrock:
## Standards

- **Hard:** `internal/cdc/flush_batch.go:39-144` is 106 lines, exceeding `coding-standard.md` and `AGENTS.md`'s 100-line function limit. The function was already over the limit, but this change adds more logic; the documented incremental-improvement rule says to refactor the immediate area when touching it.
- **Style:** `internal/cdc/mocks_test.go:201` names a private helper `lastEntry`, and `internal/federated/parquet_source_stamps_test.go:15` names one `stampedSystemCols`; both are noun/adjective-first rather than the documented verb-first private-helper convention.
- No changed source file exceeds 500 lines, and `git diff --check` is clean. I found no newly introduced production bare error returns.

iceboundrock:
## Findings

1. **P1: stamped metadata can bypass byte-truth validation** (`internal/federated/parquet_schema_validation.go:134-141`). The stamp path only runs `parquetcheck.Check`, which validates the three system columns but never compares the complete `Columns` map with the footer. A stale or tampered stamp containing those three columns skips `DESCRIBE` and feeds an incomplete or incorrect #255 union. If the real object is missing a system column, `union_by_name` can still NULL-fill it and silently lose rows; an under-reported attribute can cause a later binder/projection failure. This conflicts with the issue/#187 requirement that stamping remain a cache over byte truth.

2. **P1: the URI cache survives `cdc-init` overwrites** (`internal/federated/parquet_schema_validation.go:129-134`, `internal/cdc/init.go:328-330`). The validator cache is permanent and keyed only by URI, while `BuildBasePath` intentionally reuses deterministic `{minRowID}_{maxRowID}.parquet` keys. After a server validates path P, an init rerun can overwrite P with a different schema, but subsequent reads use the old cached columns and ignore the new stamp/footer. This can stale the #255 projection or suppress detection of a changed system schema.

3. **P2: compaction stamp failures are silent** (`internal/compaction/merge.go:85-90`). Flush and init warn when their best-effort DESCRIBE fails, but `DuckMerger` returns an unstamped entry without logging. Operators cannot distinguish a deliberately legacy entry from a compaction stamping degradation; please add equivalent observability or document the intentional absence.

iceboundrock:
## Verification

- `make test`: passed on PR head `e020dcd`.
- `make lint`: passed on PR head.
- Targeted package tests: `./internal/federated ./internal/manifest ./internal/cdc ./internal/compaction ./internal/parquetcheck`: passed.
- Targeted race tests for the same packages: passed.
- `make test-e2e-production`: passed locally with Docker (`ok .../internal/e2e_harness/production`, 130.870s).
- `git diff --check`: clean.
- GitHub CI: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all passed.

iceboundrock:
## Summary

Review result: **request changes**.

- Findings: 2 P1 correctness issues and 1 P2 observability issue.
- Spec: core stamping/plumbing behavior is present, but byte-truth validation and the legacy backfill contract need clarification/fix.
- Standards: one function-size violation and two private-helper naming violations.
- Verification: local unit, lint, race, production E2E, diff-check, and all reported CI checks passed.

The passing checks validate the covered fresh-object paths; they do not exercise the two identified invalidation/trust boundaries.

iceboundrock:
## Observations

- The writer-side coverage is strong: flush, init, merge, zero-row merge, and promotion are exercised, and the live stamp equals a DESCRIBE result in production E2E.
- The missing regression dimensions are stale/tampered stamps with valid system columns and a warm validator cache followed by deterministic init rerun/overwrite.
- The PR description already acknowledges the compaction logging gap; treating it as a follow-up is reasonable, but the runtime cache and byte-truth issues are not just observability concerns.

iceboundrock:
## Assessment

**Request changes.** The implementation is broad and the happy path is well covered, but the two P1 findings affect the correctness contract of the optimization: trusted metadata can bypass footer truth, and the permanent URI cache is invalidated by an existing init overwrite behavior. Please address those before merge; the compaction logging gap and function-size issue can be handled in the same follow-up or separately if policy permits.

iceboundrock:
## Spec

Issue #256 core requirements are substantially implemented: flush, init, and compaction stamp writer-produced parquet schemas; manifest entries preserve nil for legacy files; source plumbing reaches the pre-read validator; valid stamps feed the #255 column union; invalid/missing stamps fall back to probing.

The spec requirement that this remain a cache over byte-truth is only partially met, because a full stamp is not compared with the current footer before it is trusted (see P1 finding). The versioning/backfill story is also implicit rather than operational: field presence is the format signal and compaction naturally restamps rewritten files, but legacy entries that are never rewritten remain permanently probe-based and there is no explicit backfill procedure. That is acceptable only if the intended contract is documented as "lazy fallback, no backfill".

internal/federated/parquet_schema_validation.go
iceboundrock: Fixed in 8a4a4fd (remedy adjudicated with you: scan-level guard), plus an opportunistic cross-check in 52bb9bb.

Verified the finding first: the silent direction is real — with a stamp claiming valid system columns over an object that lacks `row_id`, `union_by_name` NULL-fills and rows drop out of the dirty anti-join. The red-first e2e (`manifest_stamp_rogue_e2e_test.go`) demonstrates it against the pre-fix build: *query returned 8 records, losing 5 row(s) silently*.

What changed:

1. **Byte-truth is now enforced at the scan itself**, not the metadata layer: `sqlgen.BuildParquetScanSource` wraps every federated parquet scan with `* REPLACE (COALESCE(row_id, error('…')) AS row_id)`. Any scanned row whose `row_id` is NULL — the only way a wrong-schema object manifests after union_by_name — fails the query loudly with the invariant named, classified as `ErrFederatedReadFailed` (degradable, same class as today's probe-detected violation). This closes the silent channel for every trust path (stamp, cache, or probe raced by an overwrite) at zero extra S3 reads, so the #256 cold-start win survives. Both template scan sites flow through the single chokepoint; the e2e now asserts loud failure + degraded-mode still answering.
   - Spike note: the guard is deliberately *untyped* `COALESCE(row_id, error(…))` — `error()` adopts the column's type (UUID in production, VARCHAR in the benchmark shape), verified by a `typeof()` cross-assertion against the bare scan; a `CAST(… AS UUID)` form provably fails to bind on benchmark parquet. Constant-folding and projection-pruning hazards were spiked against real DuckDB before wiring.
2. **Full-map comparison when a probe runs** (52bb9bb): whenever the footer is probed for a currently-stamped path, footer vs stamp are compared; on divergence the footer wins the union and the cache, and a warning names the path (`WithLogger` engine option, wired from factory).
3. **Regression test for the exact shape you named** — valid-system-column but incomplete stamp: covered at unit level (bad-stamp fall-through + cross-check tests in `parquet_schema_validation_stamps_test.go`) and at e2e level by the rogue-overwrite scenario above.

Trust-without-probe remains for the invariant check (that is the #256 optimization), but every divergence between stamp and bytes now surfaces loudly — under-report → #255 binder collision, over-report → binder failure, missing system column → scan guard.

internal/federated/parquet_schema_validation.go
iceboundrock: Fixed in 52bb9bb. Verified the premise first: `cdc.BuildBasePath` is indeed deterministic (`{minRowID}_{maxRowID}.parquet`, no UUID), so init reruns overwrite in place — the validator's "objects are write-once" comment premise was false for init, and the permanent URI-keyed cache would pin stale columns past a rerun.

What changed:

1. **Cache entries are now keyed by (URI, stamp)**: each entry records the manifest stamp it was validated under (nil for probe-validated-without-stamp). A lookup hits only when the current manifest stamp `maps.Equal`s the recorded one — an init rerun writes a new stamp via `ReplaceTierFiles`, so the next query misses the cache and re-validates against the new state. nil↔stamp transitions in both directions are unit-tested; the comparison sits inside the existing mutex, `-race` clean.
2. **The cold-start win is pinned, not sacrificed**: unchanged stamp → zero probes across repeated validations (unit + e2e assertions).
3. **The test you asked for** — warm the validator, rerun init over the same keys, verify the new schema is used — landed as `manifest_stamp_rerun_e2e_test.go`. One note from building it: the *added-column* rerun direction turns out to be inherently harmless (the stale union under-claims, #255 augments beside the real column, and DuckDB resolves the duplicate name to the physical column), so the discriminating red is the *dropped-column* direction — stale cache over-claims, projection binds a missing column (`Binder Error: Referenced column "score" not found`). The red transcript against the pre-fix build shows exactly that; post-fix the rerun is picked up and the query answers. The unchanged-rerun half pins that identical stamps still hit the cache.
4. The validator's doc comment now names the init exception instead of the false write-once claim, and `docs/federated-query/design.md` documents the invalidation rule. Chose invalidation over making init output immutable: unique init keys would ripple into `manifest-reconcile`'s init-shaped orphan classification (#292) and double transient storage on reruns.

iceboundrock:
## Review response — all findings addressed (e020dcd..c43516a, 5 commits)

**P1 findings** — both fixed; detailed responses in the two inline threads. Summary: the silent-loss channel is closed at the byte level by a scan guard on `row_id` (`8a4a4fd`, red-first e2e demonstrating silent 5-row loss → loud classified failure), and the validator cache is now keyed by (URI, manifest stamp) so init's deterministic-key overwrites invalidate it (`52bb9bb`, e2e: warm cache → init rerun over same keys → new schema used). A full stamp==footer comparison now also runs whenever a probe happens, footer wins + warning on divergence.

**P2 (compaction stamp failures silent)** — fixed in `4a5cb1f`: `DuckMerger` gains an optional nil-safe logger, warns with `tmp_uri` + cause on a failed self-describe; wired at both construction sites (compactor cmd, production harness). Unit tests cover warn-on-failure, silent-on-success, and nil-safety.

**Standards** — fixed in `4a5cb1f`/`c43516a`: `executeBatch` split to 91 lines (export+tmp-cleanup extracted as `exportBatchToTmp`, pure move, existing tests unmodified); `lastEntry`→`readLastEntry`, `stampedSystemCols`→`buildStampedSystemCols`. The wave's own re-review caught that it had pushed `engine.go` to 508 and `parquet_schema_validation_test.go` to 504 — resolved by pure moves into `engine_options.go` and `parquet_schema_validation_stamps_test.go` (`c43516a`); every wave-touched file now ≤500.

**Spec (backfill contract implicit)** — `docs/federated-query/design.md` now states it explicitly: **lazy fallback, no backfill** — legacy entries never rewritten stay probe-based indefinitely by design (one probe per process lifetime, the pre-#256 steady state). The section also documents the scan guard and the stamp-keyed invalidation, in lockstep with the code comments (§5 SQL blocks are machine-pinned to the builder output).

**Verification at c43516a**: focused packages + `make lint` + `make test` + e2e smoke + `make test-e2e-production` (114 PASS, 0 FAIL, incl. the two new rogue-overwrite / init-rerun scenarios) + `-race` on federated/sqlgen + `make benchmark-smoke` all green. One residual: benchmark smoke stops before query execution, so guard cost on the benchmark query path is bind-verified (real-DuckDB spike, both UUID and VARCHAR shapes) but not perf-measured — `benchmark-regression` pre/post is the check if you want a number before merge.

internal/sqlgen/duckdb_cold_scan.go
```diff
@@ -23,19 +23,51 @@ type NullScanColumn struct {
...
+//     100-102 render through this same source). An explicit CAST would bind
+//     one of the two and coerce the other (#147);
+//   - REPLACE puts the guard IN the row_id expression, so projection pushdown
+//     cannot prune it while row_id is still read — and both template scan
+//     sites read it (the anti-join and the semijoin's SELECT row_id).
+//
+// A scan set where NO object carries row_id fails to bind the REPLACE list
+// instead. Different message, same contract: loud, never silent.
+var parquetRowIDGuardItem = fmt.Sprintf(
+	"* REPLACE (COALESCE(row_id, error('%s')) AS row_id)", ParquetNullRowIDMessage)
```
iceboundrock: P1: this guard closes only the NULL `row_id` silent-loss case, not the full `parquetcheck` invariant. A stamp-valid object missing `changed_at` or `deleted_at` still passes the pre-read stamp path, and `union_by_name` supplies NULL for the missing system column; the new `REPLACE(COALESCE(row_id, error(...)))` does not inspect either column. I reproduced this with DuckDB: a healthy file plus a file missing `changed_at` returned the rogue row with `changed_at = NULL` and no error. Since `changed_at`/`deleted_at` feed version ordering and tombstone semantics, the read is no longer equivalent to the byte-truth validator. Please guard or otherwise validate all three required system columns (including their required types, while preserving the benchmark VARCHAR exception) and add regression coverage for missing `changed_at`/`deleted_at`.

iceboundrock:
## Summary (Re-review)

Result: **request changes**.

- Previous P1 init-cache issue: fixed and E2E-covered.
- Previous P1 silent missing-`row_id` issue: fixed by the scan guard and E2E-covered.
- Remaining P1: missing/wrong `changed_at` or `deleted_at` can still bypass the full parquet invariant.
- Remaining P2: concurrent cache misses can duplicate probes.
- Standards: trailing whitespace and a touched 107-line factory helper remain.
- Verification is otherwise green, with `git diff --check` as the only failed check.

iceboundrock:
## Standards (Re-review)

- **Hard:** `git diff --check` fails because the new SQL documentation lines in `docs/federated-query/design.md:231-238,248` contain trailing whitespace.
- **Hard:** `factory/factory.go:99-205` remains a 107-line helper after this PR touched it; this exceeds the 100-line function limit in `AGENTS.md`/`coding-standard.md`.
- **Style:** newly touched private helpers `validatedCols`, `stampCacheValidator`, `mergeStampFixture`, and `pathList` are not verb-first, contrary to `coding-standard.md:52-56`.
- The changed Go files are otherwise under 500 lines; the earlier `executeBatch` size and helper-name findings were fixed.

iceboundrock:
## Observations (Re-review)

- The new E2E tests are materially stronger: they keep one engine alive across an init overwrite and demonstrate the rogue missing-`row_id` failure is classified loudly.
- The scan guard is intentionally unconditional and works for both production UUID and benchmark VARCHAR `row_id`; however, that design currently leaves the other two invariant columns unguarded.
- The stamp/footer cross-check remains opportunistic: it runs only on probe fallback, so a stamp that passes the three-column `Check` is still trusted without a complete footer comparison.
- The concurrent probe race is safe but can duplicate S3/DuckDB work.

iceboundrock:
## Spec (Re-review)

- The writer stamping and manifest plumbing requirements are present.
- The init overwrite invalidation is now correctly keyed by `(path, stamp)`, and the lazy-fallback/no-backfill contract is explicitly documented.
- The byte-truth requirement is still partial: the scan-level defense protects missing `row_id`, but not missing/wrong `changed_at` or `deleted_at`; valid stamps can still bypass the complete three-column invariant.
- The new rogue-overwrite and init-rerun scenarios are appropriate and cover the two prior P1s, but add a regression for each required system column before calling the optimization indistinguishable from probing.

iceboundrock:
## Verification (Re-review at c43516a)

- `make test`: passed.
- `make lint`: passed.
- `go test -race ./internal/federated ./internal/sqlgen ./internal/compaction`: passed.
- `make test-e2e-production`: passed, including `TestManifestStamp_InPlaceRewriteInvalidatesValidatorCache` and `TestManifestStamp_RogueOverwriteFailsLoudly` (`ok`, 146.303s).
- Direct DuckDB check: the current `row_id` guard still allows a file missing `changed_at` to return with `NULL changed_at`.
- `git diff --check`: failed on trailing whitespace in `docs/federated-query/design.md`.
- GitHub CI checks currently report pass.

iceboundrock:
## Findings (Re-review at c43516a)

1. **P1: the new scan guard does not enforce the full parquet invariant** (`internal/sqlgen/duckdb_cold_scan.go:33-56`). It only errors for NULL `row_id`. A stamp-valid object missing `changed_at` or `deleted_at` still passes pre-read validation; `union_by_name` supplies NULL and the query continues with malformed version/tombstone data. I reproduced this directly with DuckDB: a healthy file plus a file missing `changed_at` returned the rogue row with `changed_at = NULL` and no error.

2. **P2: concurrent cache misses can still thrash the footer probe** (`internal/federated/parquet_schema_validation.go:167-196,252-267`). Lookup and publication are mutex-protected, but the expensive probe runs outside the lock. Concurrent requests for the same URI/stamp can issue duplicate DESCRIBEs, and an older validation can publish after a newer one, causing an avoidable probe on the next request. This is primarily a performance/observability issue, not a stale-result correctness issue; per-key single-flight or a generation recheck would close it.

The previous init-overwrite cache finding and compaction logging finding are fixed in this head.

iceboundrock:
## Assessment (Re-review)

**Request changes.** The two original P1s and the compaction observability issue were addressed well, with focused unit/E2E coverage. One P1 remains because the scan guard does not preserve the full three-system-column byte-truth invariant, and the current diff still has a standards failure from trailing whitespace plus a touched >100-line factory helper.

internal/sqlgen/duckdb_cold_scan.go
iceboundrock: Fixed for `changed_at`, types pinned for both timestamp columns, and the `deleted_at` presence case is documented as a bounded residual with a technical reason it cannot be closed at this layer (6b8aa73, comment refinement in bdfa925).

**What the guard now enforces** (`BuildParquetScanSource`, both template scan sites):
- `COALESCE(row_id, error(…))` — unchanged.
- `CAST(COALESCE(changed_at, error(…)) AS BIGINT) AS changed_at` — presence guard, unconditional for both shapes. Verified never-NULL at the source before wiring: flush exports `cl.changed_at` (`change_log.changed_at BIGINT NOT NULL`), init exports `m.ltbase_updated_at` (`BIGINT NOT NULL`), and the benchmark parquet carries `changed_at` directly.
- `CAST(deleted_at AS BIGINT) AS deleted_at` — type pin only (see below).

The CASTs also close the type channel you named: a rogue VARCHAR `changed_at`/`deleted_at` can no longer widen the union and turn LWW ordering lexicographic — numeric strings normalize value-preserving, garbage errors loudly (spike leg: unguarded `max(changed_at)` returned `"9"` over `"100"`; guarded returns 100). `row_id` deliberately keeps no cast — that is the benchmark VARCHAR exception, still pinned by the `typeof()` cross-assertion.

**Why `deleted_at` presence cannot be value-guarded here:** delta exports encode live rows as `deleted_at = NULL` (`cl.deleted_at`; only the init/base path writes `COALESCE(…, 0)`) — the #274 encoding asymmetry. A NULL-presence guard would therefore error on every live delta row in every existing delta object. A missing-column NULL is indistinguishable from a legit live-row NULL at row level, and any per-file presence check is a footer read — the probe this optimization exists to avoid. The residual is: documented at the guard site and in design.md §"#256", pinned by a characterization test (`TestParquetScanGuardTolerateNullDeletedAt`) that names #274, and mechanically unblocked once #274 normalizes the delta encoding to 0 — at which point the guard extends to `deleted_at` with the same one-line shape. Until then its type is pinned and its presence remains covered by the probe/stamp paths only.

**Regression coverage added:** unit (missing `changed_at` → loud error naming the column; non-numeric `changed_at` → loud CAST failure; benchmark shape unaffected) and e2e (`TestManifestStamp_RogueOverwriteFailsLoudly` extended with a missing-`changed_at` rogue). The e2e RED against the pre-fix build is worth noting: 13 records before and 13 after — no count delta at all; the NULL-version rogue *wins* the LWW merge and silently serves wrong contents. That is now a loud classified failure.

iceboundrock:
## Re-review response — wave 2 (c43516a..bdfa925, 5 commits)

**P1 (guard doesn't cover the full invariant)** — addressed in `6b8aa73`; details in the inline thread. Summary: `changed_at` gets the same unconditional presence guard as `row_id` (never-NULL verified at both export builders and the benchmark shape); both timestamp columns get `CAST AS BIGINT` type pins closing the lexicographic-LWW channel; `deleted_at` presence is a documented, test-pinned residual — it cannot be value-guarded while delta exports encode live rows as NULL (#274 asymmetry), and the guard extends mechanically once #274 normalizes the encoding. New unit + e2e regressions cover missing/mistyped `changed_at` (the e2e red is instructive: the rogue row *wins* LWW with no record-count change — now a loud classified failure).

**P2 (duplicate probes on concurrent misses)** — fixed in `8a5a320`: per-(path, stamp-fingerprint) single-flight, hand-rolled (no new dependency), probe outside the mutex, followers wake and re-run the normal lookup whose `maps.Equal` stamp check is the generation recheck. A leader that hits a schema violation publishes nothing, so every waiter still probes and fails loudly itself — the single-flight cannot weaken the loud-failure contract (test-pinned, all 6 concurrent callers get the error). Cancelled followers return the inconclusive path, no hang. `-race -count=5` clean.

**Standards** — `git diff dd67aff..HEAD --check` now exits clean (trailing whitespace stripped, `6b8aa73`); `factory/factory.go` helper split 107 → 75 lines by pure extraction (`requireCoreTables`, `newRepositoryAndEngine`), ordering vs the #314 comment preserved, factory tests unmodified (`68e0031`).

**Style** — `stampCacheValidator` deleted (it added nothing); `validatedCols`→`lookupValidatedCols`, `pathList`→`formatPathList`, `mergeStampFixture`→`buildMergeStampFixture`, all call sites (`8a5a320`).

**Verification at bdfa925**: focused packages, `gofmt`, `git diff --check`, `make lint`, `make test`, e2e smoke, `make test-e2e-production` (115 PASS incl. the new rogue scenarios), `go test -race ./internal/federated ./internal/sqlgen` — all green. Length caps re-swept over every grown file (the wave-1 lesson), all ≤500/≤100.

iceboundrock (commented):
## Summary

#256 manifest schema stamping. Each parquet writer (CDC flush, CDC init, compaction merge) records a `DESCRIBE`-derived footer column map (`FileEntry.Columns`) on the manifest entry it writes; federated reads consult these stamps in the pre-read validator so a valid stamp short-circuits the footer probe and feeds the #255 column union, with probe fallback for legacy/unstamped entries. Two hardening layers were added during the branch's own review process and are the right calls: a scan-level system-column guard (row_id/changed_at presence + type pins) that closes the silent-loss channel a stamp-trusted rogue object would otherwise open, and stamp-keyed cache invalidation so an init rerun (deterministic base key) can't serve stale columns for the life of the process.

I reviewed the full 4568-line diff, the branch source at HEAD (bdfa925), and verified the touched packages locally.

## Findings

P1: none found.

- **P2-1 — The scan guard fails loudly but nothing tells an operator which object failed.** When the guard fires it is precisely the case where the stamp short-circuited the probe, so no path-naming signal runs anywhere: the pre-read validator already trusted the stamp (no probe → no `warnStampDivergence`), the guard message omits the path (#306), and #251's verify-and-exclude never marks a schema-wrong object corrupt. An operator facing `ErrFederatedReadFailed: parquet scan produced NULL row_id` has no bucket/key signal. Not blocking — the failure is loud, classified, and degradable — but the observability gap is real for the exact failure mode this PR introduces. Follow-up: surface the scanned path-set in the classified error (or log it at the executor), or at minimum document the operational triage ("reconcile the manifest / identify by probe") in design.md.
- **P2-2 — Per-row guard cost on the healthy path.** The unconditional `REPLACE` (2× `COALESCE` + 2× `CAST`) now runs on every cold-scan row for every query, including legacy probe-validated sets that pre-#256 could never reach a rogue object. The schema-blind rationale for making it unconditional is sound; a quick guard-vs-no-guard benchmark sanity check would confirm the per-row overhead on the hottest path is within budget (benchmark-smoke passed, but no head-to-head is shown).
- **P3-1 — `DESCRIBE` scan loop is duplicated** between `parquetcheck.DescribeColumns` (over `*sql.DB`) and `federated.describeParquetColumns` (over `DuckDBQueryExecutor`) — compaction's private copy was consolidated, federated's was not. Same row shape, same scan; a drift risk. Consider sharing a row-scan helper over an interface.
- **P3-2 — Timing-window single-flight tests can flake on loaded CI.** `require.Eventually(2s)` / `require.Never(150ms, 5ms)` windows in `parquet_schema_validation_singleflight_test.go` assert wall-clock behavior. Consider widening the `Never` window or gating the count assertion on the release channel rather than elapsed time.

## Spec

The PR closes #256 and matches its contract:

- Write-time stamps from all three writers (flush delta, init base, compaction merged base incl. zero-row edge) ✓
- Probe fallback for unstamped/legacy entries, byte-identical behavior (pinned by `TestFileEntryColumnsLegacyCompat`, `TestValidateNilStampsUnchanged`) ✓
- Stamps short-circuit success only; a rejected stamp costs a probe, never a failure (`TestValidateBadStampFallsBackToProbe`) ✓
- Stamps feed the #255 union, with the documented under-reporting exception ✓
- Scan-level guard closes the silent-loss channel the stamp opened (`TestManifestStamp_Rogue*`, `TestParquetScanGuard*`) ✓
- Cache keyed by (path, stamp) so init reruns re-validate (`TestManifestStamp_InPlaceRewriteInvalidatesValidatorCache`) ✓
- Versioning = field presence (`columns,omitempty`), no migration, lazy backfill ✓
- `reconcile --repair` promoted entries, pre-merge source validation, and manifest backfill are explicitly out of scope ✓

## Standards

- Line caps respected, and enforced with mechanical moves: engine.go 508→451, `executeBatch` 106→91, factory 107→75. ✓
- Error wrapping with context (`%w`) and classified `ErrFederatedReadFailed` throughout. ✓
- Test seams (`describeColumns`) mirror the existing pattern; `t.Helper()`; deterministic `lastSavedPath` replaces map-iteration order; test twins for flush/init (manifestStore-nil short-circuit) keep the two writers symmetric. ✓
- Comments are thorough and load-bearing, consistent with the repo's style. One nit: the package-level doc on `parquet_schema_validation.go` and `Validate`'s doc restate near-identical caveat text — consider referencing one from the other to reduce the drift surface.

## Assessment

Approve. The design is carefully argued, every failure mode is enumerated and either closed or documented as accepted, and the e2e correctness bar (`stamp == live DESCRIBE` of the published object) is the right one. The branch's own review process produced several genuine hardening commits (single-flight, cache re-keying, scan guard, `requireCoreTables` split) and the final state reads like a well-reviewed PR. All findings are non-blocking (P2-1 observability, P2-2 perf-check, P3s below).

## Observations

- **Glob-expanded matches never carry stamps in production.** The fallback glob string maps to no manifest entry, so its concrete matches always probe. Correct and intended (the fallback exists only when the manifest is empty), but it means the cold-start win never applies to fallback-glob scans; `TestValidateGlobMatchesConsumeStamps` proves the validator could consume concrete-match stamps if the plumbing ever returned them.
- **Follower re-probe on the error path is a thundering herd.** When a leader's probe raises an invariant violation, every follower re-probes (N DESCRIBEs) to hear the same loud error. Correct by design ("the loud failure must reach every caller"), but N concurrent callers on a corrupt object each pay a full DESCRIBE; the success path is collapsed to one.
- **`WithLogger` narrowness is good.** Routing the validator's stamp/footer cross-check through `EngineOption.WithLogger` (default nop) is the minimal surface, and both composition roots pass their existing logger.
- **`error()` inside `COALESCE` not being bind-time folded** is the load-bearing engine property and is proven against the pinned DuckDB rather than assumed — good.
- The follow-ups listed in the PR body are reasonable; I especially agree the `.golangci.yml` missing `build-tags: [e2e]` gap (65 harness files invisible to `make lint`, pre-existing gofmt violations undetected) is worth a dedicated fix.

## Verification

Local (branch bdfa925, isolated worktree, no Docker):
- `go test ./internal/{manifest,parquetcheck,cdc,compaction,federated,sqlgen,factory}/...` — all pass.
- `go test -race -count=3` on federated, sqlgen, cdc, compaction, manifest, parquetcheck — all pass; the new single-flight concurrency is race-clean.

CI on the PR: Lint ✓, Test ✓, E2E Smoke ✓, k6 Smoke ✓, Benchmark Smoke ✓ (all SUCCESS). The Docker-based production e2e (112 PASS / 0 FAIL, six stamp==probe assertions) and the new rogue-overwrite / cache-invalidation scenarios are claimed green; I did not run those locally.

internal/sqlgen/duckdb_cold_scan.go
```diff
@@ -23,19 +23,87 @@ type NullScanColumn struct {
...
 }
 
+// ParquetNullRowIDMessage and ParquetNullChangedAtMessage are the texts the
+// scan-level system-column guard raises. Each names the offending column and
+// the invariant rather than the object: the scan source is schema-blind and
+// path-blind, and a path in a DuckDB error would be a storage-location leak
+// (#306). The manifest entry and the pre-read validator identify the offending
+// object.
+const (
+	ParquetNullRowIDMessage     = "parquet scan produced NULL row_id: a scanned object violates the export schema invariant (#189/#256)"
```
iceboundrock, line 37: P2-1 (observability): this is the one message an operator gets when the #256 guard fires, and it names neither the object nor the path. In the exact case this guard exists for — a stamp short-circuit, so the pre-read validator never probed — warnStampDivergence (the only path-naming log) never runs either, and #251's verify-and-exclude won't classify a schema-wrong object as corrupt. Net: an operator with ErrFederatedReadFailed: NULL row_id has no bucket/key to act on. Suggest a follow-up to correlate the failing scan back to a manifest entry or path-set (the classified error surfaces to operators, not clients), or at minimum a design.md note on the triage path.

internal/parquetcheck/describe.go
```diff
@@ -0,0 +1,37 @@
+package parquetcheck
+
+import (
```
iceboundrock, line 3: P3 (drift risk): this row-scan loop is a near-verbatim twin of federated.describeParquetColumns (internal/federated/parquet_schema_validation.go) — compaction's private copy was consolidated into this new helper, but the federated read path kept its own copy over the DuckDBQueryExecutor abstraction. A change to the DESCRIBE row shape (e.g. a DuckDB minor adding a describe column) would have to be made in both places. Consider sharing one row-scan helper (e.g. exposing a scan over *sql.Rows) rather than keeping two loops.

internal/federated/parquet_schema_validation_singleflight_test.go
```diff
@@ -0,0 +1,237 @@
...
+	var wg sync.WaitGroup
+	for _, stamp := range []map[string]string{stampA, stampB} {
+		wg.Add(1)
+		go func(stamp map[string]string) {
+			defer wg.Done()
+			validateConcurrently(t, v, duck, 4, stampCachePath, stamp)
+		}(stamp)
+	}
+
+	require.Eventually(t, func() bool { return duck.describes.Load() == 2 }, 2*time.Second, time.Millisecond,
```
iceboundrock, line 139: P3 (flake risk): the following `require.Never(..., 150*time.Millisecond, 5*time.Millisecond)` asserts wall-clock behavior — on a loaded CI runner a scheduled delay could push a follower past the window and flake. The count-based parts of this test are already gated deterministically on the `release` channel; consider asserting the probe-count while both leaders are parked rather than on elapsed time.

iceboundrock (commented):
Inline findings (anchored on added diff lines), complementing the review summary above.

internal/sqlgen/duckdb_cold_scan.go
iceboundrock, line 37: P2-1 (observability): this is the one message an operator gets when the #256 guard fires, and it names neither the object nor the path. In the exact case this guard exists for — a stamp short-circuit, so the pre-read validator never probed — warnStampDivergence (the only path-naming log) never runs either, and #251's verify-and-exclude won't classify a schema-wrong object as corrupt. Net: an operator with ErrFederatedReadFailed: NULL row_id has no bucket/key to act on. Suggest a follow-up to correlate the failing scan back to a manifest entry or path-set (the classified error surfaces to operators, not clients), or at minimum a design.md note on the triage path.

internal/parquetcheck/describe.go
iceboundrock, line 3: P3 (drift risk): this row-scan loop is a near-verbatim twin of federated.describeParquetColumns (internal/federated/parquet_schema_validation.go) — compaction's private copy was consolidated into this new helper, but the federated read path kept its own copy over the DuckDBQueryExecutor abstraction. A change to the DESCRIBE row shape (e.g. a DuckDB minor adding a describe column) would have to be made in both places. Consider sharing one row-scan helper (e.g. exposing a scan over *sql.Rows) rather than keeping two loops.

internal/federated/parquet_schema_validation_singleflight_test.go
iceboundrock, line 139: P3 (flake risk): the following `require.Never(..., 150*time.Millisecond, 5*time.Millisecond)` asserts wall-clock behavior — on a loaded CI runner a scheduled delay could push a follower past the window and flake. The count-based parts of this test are already gated deterministically on the `release` channel; consider asserting the probe-count while both leaders are parked rather than on elapsed time.

iceboundrock (commented):
Inline findings (anchored on added diff lines), complementing the review summary above.

internal/sqlgen/duckdb_cold_scan.go
iceboundrock: Addressed in efe1b8d + follow-up filed as #351.

design.md §5 now carries the triage note: the guard error names the invariant but not the object, and all three compounding causes are stated (single union scan → one error per set; the object exists so missing-key classification stays silent; the #251 drain is unguarded so a schema-wrong object reads clean). Until #351 lands, the documented bisect is: list the schema's manifest paths, run the guarded scan per file. The guard-site comment points at #351.

#351 is scoped to your first suggestion: on a guard-classified failure, extend the #251 verify pass with a *guarded* per-file drain to identify the offending key(s) in the classified error — identification only, no auto-exclusion (unlike byte corruption, excluding a schema-wrong object would silently drop rows it may legitimately own once repaired). The verify pass already visits one path at a time; the guard is the only missing piece, so the follow-up is small.

internal/parquetcheck/describe.go
iceboundrock: Fixed in dbf5fc5. The row-scan loop now lives once in parquetcheck: `ScanDescribeColumns(rows DescribeRows, uri string)` over a minimal structural interface (`Next/Scan/Err` — deliberately no `Close`, which stays with the callers). Both call sites delegate: parquetcheck's `*sql.DB` path and federated's `describeParquetColumns` — `duckDBRowsIterator` satisfies the interface structurally with an identical `Scan` signature, no adapter. The old federated loop is gone (`database/sql` dropped out of that file's imports), error texts unchanged, existing tests in both packages pass unmodified. A DESCRIBE row-shape change is now a one-place edit.

One residual noted for completeness: the production e2e harness has a third, non-twin DESCRIBE site (`artifacts.go` via a generic rows-to-maps helper) — different shape, harness-only, left alone deliberately.

internal/federated/parquet_schema_validation_singleflight_test.go
iceboundrock: Fixed in 0b5b84a — with one deviation from the obvious fix that's worth spelling out.

The natural replacement (assert the FINAL total probe count == 2 after releasing the gate and joining) turns out to be empirically false: the validator cache holds ONE entry per path (not per (path, stamp)), both generations in this test fail the invariant so both leaders probe and publish, the second publish clobbers the first, and each woken follower whose stamp no longer matches the entry legitimately re-probes (a follower must never inherit a downgrade). Measured over 200 runs the post-release total ranges 2..8 — all by design.

What landed instead: the count-based assertion moved to the window where it IS structurally invariant. While both leaders are parked on the `release` gate, a third DESCRIBE is impossible (`beginProbe` admits one leader per key under the mutex; followers can't reach the probe until `finishProbe` closes `done`, which requires the release). So: `Eventually(describes==2)`, then a single snapshot under `v.mu` asserting `describes==2 && len(inflight)==2` — no wall-clock window at all. Mutation-verified it still fails on the two real regressions: stamp dropped from `probeKey` (one leader → Eventually fails at ==1) and `beginProbe` admitting duplicate leaders (count races past 2 → snapshot fails). The eagerly-evaluated failure message in the old `Eventually` went with it. `-race -count=10` clean.

(For the record the old `Never(>2, 150ms)` couldn't false-fire — the count was structurally capped while the gate was shut — but you're right that a wall-clock assertion is the wrong shape; it burned 150ms to sample an invariant.)

iceboundrock:
## Round-3 response (bdfa925..0b5b84a, 3 commits)

All three findings addressed; details in the inline threads.

- **P2-1 (guard error names no object)**: design.md §5 triage note + guard-site pointer (`efe1b8d`); follow-up **#351** filed for guarded per-file identification on the failure path (identification only, no auto-exclusion).
- **P3 (describe-loop twins)**: single shared `parquetcheck.ScanDescribeColumns` over a structural `DescribeRows` interface; both call sites delegate, old loops gone, error texts unchanged (`dbf5fc5`).
- **P3 (wall-clock `Never`)**: replaced with a deterministic mutex snapshot while both leaders are parked (`0b5b84a`). One deviation documented in the thread: the final-total==2 form is empirically false (per-path single-entry cache + legitimate follower re-probes → totals 2..8 by design), so the assertion moved to the window where the cap is structural. Mutation-verified against both real single-flight regressions; `-race -count=10` clean.

Verification at 0b5b84a: focused packages + `make lint` + `make test` green; gofmt clean on touched files; no e2e delta (prose/comment, behavior-identical extraction, test-only — reasoning in the report). CI green on the push.

iceboundrock (commented):
## Summary (follow-up review @ HEAD 0b5b84a)

This is an updated review. Since my pass at bdfa925 the branch added three commits that close every finding I raised inline: `dbf5fc5` (shared DESCRIBE row-scan), `efe1b8d` (documented guard triage + #351), `0b5b84a` (deterministic single-flight assertion). I re-verified each against the new HEAD and re-ran the affected tests. The PR still closes #256: write-time `FileEntry.Columns` stamps from flush/init/compaction, probe fallback for legacy entries, stamp short-circuit-success-only, stamp-fed #255 union, the scan-level system-column guard, and (path, stamp)-keyed validator caching.

## Findings

All three inline findings from the prior review are now addressed:

- **P1 → resolved.** No P1 findings at either revision.
- **P2-1 (guard named no object) → addressed as triage.** `efe1b8d` adds a §5 "Triage: the guard names the invariant, not the object" section to `docs/federated-query/design.md` documenting manual bisection until per-file identification lands, and #351 is filed to track it. I audited the triage doc against the code: it correctly states why none of the existing signals help (single `union_by_name` error over the set; object exists so the missing-key classifier is silent; #251 drain is deliberately unguarded). Good scoping — documenting the operational path instead of expanding the guard's blast radius in this PR is the right call.
- **P3-1 (duplicated DESCRIBE loop) → fixed.** `dbf5fc5` extracts `parquetcheck.ScanDescribeColumns` over a `DescribeRows` interface (`Next/Scan/Err`), which both `parquetcheck.DescribeColumns` (*sql.DB) and `federated.describeParquetColumns` (executor seam) now call — the 6-column row shape lives in exactly one place. The interface omits `Close` so handle lifetime stays with the query issuer; good design. I verified the federated probe now `return parquetcheck.ScanDescribeColumns(rows, path)` and both call sites pass.
- **P3-2 (timing-window flake) → fixed and improved.** `0b5b48a` replaces `require.Never(150ms, 5ms)` with a single snapshot of the in-flight leadership map + probe count taken under the validator's mutex while both leaders are parked. This is structurally deterministic rather than wall-clock-based: `beginProbe` admits one leader per key under the mutex, and a follower cannot probe until `finishProbe` closes its `done` channel after release — so the counts are pinned for the whole parked window. Confirmed by `-race -count=5` on the single-flight tests. Also nice: the `Eventually` message no longer eagerly interpolates the pre-wait count.

**Remaining (open from prior review, P2-2):** the unconditional per-row guard (`COALESCE×2 + CAST×2` + <system-column REPLACE) on the healthy cold scan. The author has not addressed it — understandably, since it has no clean schema-blind fix — but I'd still like a guard-vs-no-guard benchmark comparison before merge, or an explicit accepted-overhead note in the PR body. This is the only open item from my prior review.

## Spec

Still conforms to the #256 contract, plus the review-added invariants:
- Write-time stamps from all three writers (flush delta, init base, compaction merged base incl. zero-row edge) ✓
- Probe fallback for unstamped/legacy entries, byte-identical behavior (pinned) ✓
- Stamps short-circuit success only; rejected stamp → one probe, never a failure ✓
- Stamps feed the #255 union, with the documented under-reporting exception ✓
- Scan-level guard closes the silent-loss channel; deleted_at presence residual (#274) pinned ✓
- Cache keyed by (path, stamp) so init reruns re-validate ✓
- Versioning by field presence, no migration, lazy backfill ✓
- `reconcile --repair` promoted entries, pre-merge source validation, manifest backfill — out of scope ✓

## Standards

- The three follow-up commits stay within the 500/100 line caps; the shared-scan extraction removes duplication rather than adding it. ✓
- `%w` wrapping and classified errors unchanged ✓
- The new `DescribeRows` interface's deliberate exclusion of `Close` is documented, not accidental. ✓

## Assessment

The findings are all addressed to my satisfaction; I still recommend confirming the single per-row overhead item. Net: strong, carefully-reasoned PR with an unusual honesty of its failure modes. I remain at Approve.

## Observations

- The nailed-down #717 tracks guard→object correlation; the manual bisection fallback in the docs is a reasonable interim.
- `ScanDescribeColumns` folding both probes is a genuine correctness/drift win, not just cleanup — the DESCRIBE column count (six) is now pinned in one place.
- The author replied to each thread with the addressing commit; process is clean.

## Verification

Local at 0b5b84a (isolated worktree, no Docker):
- `go test ./internal/{parquetcheck,federated,sqlgen,cdc,compaction,manifest,factory}/...` — all pass.
- `go test -race -count=5 -run TestValidatorSingleFlight ./internal/federated/` — passes (flake fix solid under repeat race runs).
- `go test -race -count=3` on the writer packages (cdc, compaction, manifest) — all pass.

CI at HEAD: Lint ✓, Test ✓, Benchmark Smoke ✓ (`E2E Smoke`/`k6 Smoke` still pending). Note: I did not run the Docker-based production e2e suite locally.

## Summary

All inline findings closed with corresponding commits. The only item still open is the per-row guard cost (P2-2), which is a performance question, not a correctness one — a quick benchmark head-to-head would settle it. Ready to merge once that's triaged (or explicitly accepted).

iceboundrock:
## Findings

This is an independent post-merge re-review of final head `0b5b84a`.

1. **P1: a passing manifest stamp can still bypass byte-truth validation for `deleted_at`.** [`validateConcrete`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/federated/parquet_schema_validation.go#L179-L190) trusts any stamp that satisfies `parquetcheck.Check` without probing the current object. If a listed object is subsequently replaced with bytes that omit `deleted_at`, the scan only casts the NULL-filled column ([`duckdb_cold_scan.go#L81-L95`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/sqlgen/duckdb_cold_scan.go#L81-L95)); it cannot distinguish an absent column from a legitimate live-delta NULL. The visibility predicate then treats the value as live ([`advanced_query_template_duckdb.go#L116-L121`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/sqlgen/advanced_query_template_duckdb.go#L116-L121)).

   Repro: retain a valid `Columns` stamp, replace one listed parquet object with `row_id` and `changed_at` but no `deleted_at`, and query it with a healthy sibling. Rows that were tombstones in the replaced object can surface as live. The behavior is explicitly characterized as successful at [`duckdb_cold_scan_guard_test.go#L315-L345`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/sqlgen/duckdb_cold_scan_guard_test.go#L315-L345). Deferring this to #274 leaves the #256 optimization with a correctness gap today.

2. **P2: a trusted stamp does not enforce the production UUID `row_id` type.** The guard deliberately leaves `row_id` uncast so benchmark VARCHAR IDs work ([`duckdb_cold_scan.go#L62-L69`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/sqlgen/duckdb_cold_scan.go#L62-L69)), while the final projection casts only after LWW partitioning ([`duckdb_schema_projection.go#L320-L327`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/sqlgen/duckdb_schema_projection.go#L320-L327)). A stale valid stamp can therefore hide a replacement object using VARCHAR UUID spellings. Two case variants of the same UUID partition as separate rows before both are cast to the same output UUID, allowing duplicate logical records and conflicting versions. Add a production-shape validation/guard that preserves benchmark compatibility, with an end-to-end regression.

## Spec

#256 asks for write-time stamps for flush/init/compaction, probe fallback for missing or rejected stamps, a versioning/backfill story, and explicitly says the metadata is a **cache layer on top of, not a replacement for, byte-truth probing** because corruption detection must distrust metadata. The writers, legacy lazy fallback/no-backfill contract, projection union plumbing, stamp-keyed cache, and probe single-flight are implemented. Finding 1 means the byte-truth requirement remains partial: a currently valid-looking stamp can still conceal an object missing one required invariant column.

## Standards

- **Hard:** `newEntityManagerWithConfigContext` returns dependency errors unchanged at [`factory.go#L108-L115`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/factory/factory.go#L108-L115). This is in the #256-touched/refactored area and conflicts with the contextual-wrapping requirement in `AGENTS.md:68-70` and `coding-standard.md:22-33`.
- **Hard:** the new [`DescribeRows`](https://github.com/Lychee-Technology/forma/blob/0b5b84ab895e06ce416aa3dc877ff04c38fbc30b/internal/parquetcheck/describe.go#L11-L19) interface does not use the documented `<Noun>er`/`<Noun>Service` naming convention (`coding-standard.md:52-56`); `DescribeRowScanner` would conform.
- **Judgement call:** the three writer-side footer-stamping implementations remain materially similar (`internal/cdc/flush_batch.go`, `internal/cdc/init_stamp.go`, `internal/compaction/merge.go`). The distinct logger/error policies may justify the duplication, but it exceeds the documented DRY review threshold.
- No changed Go source file exceeds 500 lines, no changed function exceeds 100 lines, and `git diff --check` is clean.

## Assessment

**Post-merge corrective action required.** The PR is already merged, so this is not a request to change its review state. Track and prioritize Finding 1 as a P1 corrective fix: it contradicts the core corruption/byte-truth safety claim of #256 and can resurrect rows. Finding 2 should be addressed in the same hardening work or explicitly bounded by a documented trust model.

## Observations

- The final branch materially improves prior iterations: the `row_id`/`changed_at` scan guards, `(path, stamp)` cache key, deterministic single-flight test, and shared DESCRIBE row scanner all close previously reported issues.
- The remaining `deleted_at` residual is intentionally documented, not accidental, but its characterization test demonstrates the unresolved safety boundary rather than mitigating it.
- Failed/inconclusive probes intentionally re-probe for each follower so every caller receives a loud invariant violation. This is correct but can amplify DESCRIBE traffic during an S3/DuckDB failure burst.

## Verification

- `GOCACHE=... GOFLAGS=-buildvcs=false go test ./internal/federated ./internal/sqlgen ./internal/manifest ./internal/cdc ./internal/compaction ./internal/parquetcheck ./factory` passed.
- `GOCACHE=... GOFLAGS=-buildvcs=false go test -race -count=3 ./internal/federated ./internal/sqlgen` passed.
- `git diff --check dd67aff...0b5b84a` passed.
- PR CI completed successfully: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

## Summary

Final re-review: 1 P1 byte-truth/correctness gap, 1 P2 row-ID type boundary, and 2 standards violations. Most previous findings are genuinely resolved, but #256 should receive a corrective follow-up before the manifest stamp is treated as a safe corruption-detection cache.

iceboundrock:
逐条回应 post-merge 复审（@0b5b84a）。纠正措施按你的指示落在 **PR #352 分支**（commits `d4dfb89` / `9cea774` / `dc7dd19`，已推送；`make test` + `make lint` 全绿）。

## 先说两个 finding 的共同信任类别

F1 与 F2 是同一类状态：**对象字节被带外替换、而 manifest stamp 保持形状有效**。系统内没有任何 writer 能产生它——init 重跑经 `ReplaceTierFiles` 换新 stamp（(path, stamp) 缓存键随之失效）、flush/compaction 写全新 key、#292 晋升按实际 footer 重算 stamp。所以读路径守卫之外的正确纠正面是**离线字节真值核查**，而 forma 已有这个工具的天然位置：`manifest-reconcile`（#292 刚给它接好 `StatsReader.FileColumns`）。

## F1（P1：stamp 可绕过 deleted_at 字节真值）— 纠正已落地：`manifest-reconcile --verify-stamps`

- **读路径闭合仍然不可行**，与 wave-2 结论一致：delta 导出把存活行编码为 `deleted_at = NULL`（#274 不对称），行级 NULL 无法区分"缺列填充"与"合法存活"；逐文件列存在性检查=footer 读=该优化要省掉的探测。`TestParquetScanGuardTolerateNullDeletedAt` 特征化的正是这个技术边界，#274 落地后守卫机械扩展。
- **新纠正**（`d4dfb89`+`dc7dd19`）：`manifest-reconcile --verify-stamps` 对每个在册带戳条目做 **stamp vs footer 全列映射比对**（比 scan guard 更强：含 deleted_at 缺列、类型漂移、属性列增删），分歧 → `stamp divergence: <key> (column "x": stamp "..." vs footer "...")` → 残余 → exit 2；探测失败 → exit 1（沿 confirmDangling 的存储故障规则）。skip 集 = legacy nil 戳（#256 lazy-fallback 契约）/ 不可解析或越前缀路径 / dangling 候选（皆已另行上报）。真-DuckDB 回归即你的 F1 复现形状：fixture 实际缺 `deleted_at`、stamp 声称有 → 工具点名捕获。
- design.md 新增「信任边界」段：stamp 免探测信任的边界=系统 writer + manifest 事务性；带外替换由 (a) scan guards（row_id/changed_at）、(b) `--verify-stamps` 离线全量核查、(c) #274 解锁的 deleted_at 读路径守卫三层围合。
- 运维语义：周期跑或疑似带外改写后跑；検出即 exit 2 并点名 key+列+两侧类型。

## F2（P2：row_id 类型边界）— 同一纠正覆盖；读路径 shape-guard 经权衡不做，理由如下

- `--verify-stamps` 直接捕获该通道：stamp 声称 `row_id: UUID`、footer 实为 `VARCHAR` → 类型分歧点名上报。
- 读路径小写-shape guard（`lower() 比对` 或分区键归一）被否决的技术理由：①逐行成本叠加在你此前已指出的 guard 每行开销（P2-2）上；②触碰两处已被测试钉死的语义——`typeof()` 交叉断言（guard 刻意 untyped、UUID 原样穿透）与 LWW 分区键形状；③其防御对象与 F1 同为带外替换，已由离线核查兜住。如你倾向仍要读路径守卫，我开 follow-up 单独评审成本。

## Standards

- **factory.go:108-115 裸返回**：已修（`9cea774`）——`collectTables`/`requireCoreTables` 两处均加上下文包装；factory 测试断言的子串不受影响。
- **`DescribeRows` 命名**：已改 `DescribeRowScanner`（`9cea774`）；federated 侧为结构性满足，零改动。
- **三处 writer 侧戳记实现相似（judgement）**：维持现状——三者的 logger/错误策略确实不同（flush 走 zap warn+继续、init 有 describeColumns seam+manifestStore nil 早退、merge 走可选 nil-safe logger），强行合并需要抽出一个跨包依赖注入面，收益低于耦合成本。DESCRIBE 行扫描的真正孪生已在 `dbf5fc5` 合一（`ScanDescribeColumns`）。需要时可开 follow-up 收敛剩余骨架。

## 备注

`--verify-stamps` 为串行逐条 DESCRIBE（单钉连接），大 manifest 上有耗时，文档已标注为低频运维操作；批量化留作可选 follow-up。该纠正在 #352 分支上，随该 PR 一并评审合入。

PR #353, opened 2026-08-04
fix(httpapi): finish UseNumber adoption — exact validation instances, pointer-shape sidecar, HTTP round-trip probe (#282)

### LINKED ISSUE #282: HTTP JSON decoding rides float64: integer payloads above 2^53 are rounded at the door (UseNumber decision)

iceboundrock:
Closes #282.

Finishes the `dec.UseNumber()` adoption decided on #282 (option 1). PR #284 already covered the entity create/update decode sites; this PR closes the remaining surface:

- **`internal/httpapi`** — `readJSONBody` and `readEntityJSONBody` are merged into a single `readJSONBody` that always decodes with `UseNumber`. For the two remaining plain callers this is provably a no-op (`handleDelete` decodes `[]string`; `handleAdvancedQuery` decodes the fully-typed `forma.QueryRequest`, whose custom `UnmarshalJSON` receives raw bytes and whose condition values are the `"op:value"` string DSL) — but it removes the float64-at-the-door trap for any future `any`-typed payload. Pinned by `TestReadJSONBodyDecodesNumbersExactly`.
- **`internal/schemavalidate`** — the real finding of the consumer audit: `Validate`'s marshal→unmarshal round-trip re-decoded with a plain decoder, so the *validation instance* rode float64 and numeric bounds silently rounded above 2^53 (a `maximum: 2^53` failed to catch `2^53+1` while storage kept the exact value). The round-trip now decodes with `UseNumber` and rewrites every `json.Number` to `int64`/`float64` via `exactNumberInstance` before validating. The rewrite is mandatory: jsonschema-go v0.4.2's `jsonType` classifies a raw `json.Number` as `"string"` (it switches on `reflect.Kind`), which would fail every `"type": "integer"` property, while `int64` classifies as `"integer"` and all numeric constraint checks (`minimum`/`maximum`/`multipleOf`, `enum`, `const`) compare via `big.Rat` — exact. The rewrite mutates only `Validate`'s private round-tripped copy; the caller's document is untouched (#312/#314 constraint).
- **`internal/transform`** — the #205-final-review fold-in: `toInt64ExactForEAV` now dereferences `*int`/`*int16`/`*int32`/`*float64`/`*string` (previously only `*int64`), so those pointer shapes reach the exact-int64 sidecar instead of riding the float64 fallback. `*float32` is deliberately absent (every float32-representable integer is losslessly representable in float64). Plus `toBoolForEAV` gains the `json.Number` case mirroring `toBool` (see audit table).
- **Acceptance (e2e)** — `TestHTTPBigintRoundTripAbove2p53`: the production harness's real `EntityManager` (validator wired) behind the real `httpapi` server via `httptest`; POST `9007199254740993` / `±MaxInt64` as raw JSON integer literals, GET back by `row_id`, assert the exact literal in the response. Full `make test-e2e-production` green.

## Consumer audit (mandated by the decision comment)

| Consumer of UseNumber-decoded maps | State |
|---|---|
| Transform chain — numeric (`numutil.Int64Exact`, `numutil.Float64`) and `toBool` | ✅ handled since #205/#284 |
| Transform chain — `toBoolForEAV` | ⚠️ had no `json.Number` case (gap dates to #284). Fixed here mirroring `toBool` (`f != 0`). Note: on the current production chain, bool values are normalized before reaching it (`populateTypedValue` → `FromEAVRecords` hands back Go `bool`), so this is defense-in-depth, not a reproducible failure |
| `schemavalidate.Validator.Validate` | ❌ was rounding the validation instance → **fixed in this PR** |
| Relation handling (`entity_relation_service.go` et al.) | ✅ no numeric type-switches; relation values are UUID strings |
| Query condition parsing | N/A — `KvCondition.Value` is a string DSL parsed by `conditionexpr` + `TryParseNumber` (int64-exact for integral literals); never sees decoder output. Comparison-side exactness is #281 |
| `model.ParseAttributesJSON` (`.(float64)` asserts) | ✅ decodes DB-side JSON_AGG output internally; never sees an HTTP body |
| `model.EntityAttribute.Numeric()/BigInt()` | ✅ no non-test callers |
| `schemameta/schema_parser.go` | ✅ parses schema documents loaded internally, not request payloads |

## Known boundaries (documented, deliberate)

- **Exactness ceiling:** integral literals ≥2^63 still fall back to float64 in the validation instance (and int64 storage cannot hold them anyway) — unchanged semantics.
- **Schema-side bounds round:** jsonschema-go's `Minimum`/`Maximum`/`ExclusiveMin`/`Max`/`MultipleOf` are `*float64`, so a schema-declared bound above 2^53 is itself rounded by the library. This PR makes the *instance* exact, not the *bound*.
- **Pre-existing since #284 (not introduced here):** a literal outside float64 range (e.g. `1e400`) now passes the boundary decode (json.Number is lazy) and fails in `Validate`'s round-trip as a plain error → 500, where pre-#284 the boundary decode 400'd. It is caller input and arguably deserves `forma.ErrInvalidInput` → 4xx; follow-up candidate.
- **e2e anchor note:** the MaxInt64 subtest is not a reliable pre-fix red anchor on arm64 (float→int saturation false-pass, the #174 hazard); the `2^53+1` and `-MaxInt64` subtests carry the platform-independent proof.

## Sequencing

This lands **before** #281 (per the decision comment): the comparison-side audit can now trace a final decoder that delivers `json.Number`, not a state about to change. No comparison/keyset code is touched here.

## Verification

- `make test` green; `make lint` (pinned golangci-lint v1.64.8) clean; `go test -race` on the three touched packages green.
- Targeted e2e 3/3 + full `make test-e2e-production` green (131s).
- Four-task subagent execution with per-task spec+quality reviews, one whole-branch final review (finding fixed in `b3bd76e`), scoped re-reviews after every fix round.

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -418,7 +418,7 @@ delegates to it, and `executeGet` (`internal/httpapi/server.go:175`) calls
 
 Handlers also call `writeError` directly — 33 sites in `server.go` — and those
 bodies are verbatim without passing the gate. They are safe because every one of
-them reports a request-parsing failure (`parsePath`, `readEntityJSONBody`,
+them reports a request-parsing failure (`parsePath`, `readJSONBody`,
 `parseUUID`, `parseCreateObjects`, `parseSortParams`); none touches the manager,
 the engine, S3, or `PG_CONN`. What holds that line is the source-level guard
 `TestWriteErrorAlwaysCarriesALiteral4xxStatus`
--- a/internal/e2e_harness/production/http_bigint_roundtrip_e2e_test.go
+++ b/internal/e2e_harness/production/http_bigint_roundtrip_e2e_test.go
@@ -1,0 +1,80 @@
+//go:build e2e
+
+package production
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/lychee-technology/forma/internal/httpapi"
+)
+
+// TestHTTPBigintRoundTripAbove2p53 is #282's acceptance probe: an integer
+// payload above 2^53 must survive the full HTTP surface — UseNumber decode,
+// JSON-schema validation, the exact-int64 transform sidecar, bound-column
+// storage — and read back over HTTP as the same literal. Pre-#284 the default
+// decoder rounded it at the door regardless of #205's exact write path.
+func TestHTTPBigintRoundTripAbove2p53(t *testing.T) {
+	cluster := SharedCluster(t)
+	env := NewEnv(t, cluster)
+	wide := DefaultSchemaFixtures()[1]
+
+	srv := httptest.NewServer(httpapi.NewServer(env.EntityManager(), httpapi.Options{}).Handler())
+	defer srv.Close()
+
+	literals := []string{
+		"9007199254740993",     // 2^53+1: first integer float64 cannot represent
+		"9223372036854775807",  // MaxInt64
+		"-9223372036854775807", // -MaxInt64 (pre-#205 rounded to MinInt64)
+	}
+	for _, lit := range literals {
+		t.Run(lit, func(t *testing.T) {
+			body := fmt.Sprintf(`{"title":"http-bigint-%s","amount":%s}`, lit, lit)
+			resp, err := http.Post(srv.URL+"/api/v1/"+wide.Name, "application/json", bytes.NewReader([]byte(body)))
+			if err != nil {
+				t.Fatalf("create request: %v", err)
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode != http.StatusCreated {
+				t.Fatalf("create status = %d, want %d", resp.StatusCode, http.StatusCreated)
+			}
+			var created struct {
+				RowID string `json:"row_id"`
+			}
+			if err := json.NewDecoder(resp.Body).Decode(&created); err != nil {
+				t.Fatalf("decode create response: %v", err)
+			}
+			if created.RowID == "" {
+				t.Fatal("create response carries no row_id")
+			}
+
+			getResp, err := http.Get(srv.URL + "/api/v1/" + wide.Name + "/" + created.RowID)
+			if err != nil {
+				t.Fatalf("get request: %v", err)
+			}
+			defer getResp.Body.Close()
+			if getResp.StatusCode != http.StatusOK {
+				t.Fatalf("get status = %d, want %d", getResp.StatusCode, http.StatusOK)
+			}
+			var record struct {
+				Attributes map[string]any `json:"attributes"`
+			}
+			dec := json.NewDecoder(getResp.Body)
+			dec.UseNumber()
+			if err := dec.Decode(&record); err != nil {
+				t.Fatalf("decode get response: %v", err)
+			}
+			got, ok := record.Attributes["amount"]
+			if !ok {
+				t.Fatalf("get response missing amount attribute: %v", record.Attributes)
+			}
+			if want := json.Number(lit); got != want {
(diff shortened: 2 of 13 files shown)


iceboundrock:
## Findings

No new blocking findings for #282.

- **Follow-up observation (pre-existing from #284):** `internal/schemavalidate/number_instance.go:29-31` returns a plain error for a valid JSON literal that fits neither `int64` nor `float64` (for example, `1e400`). `Validator.Validate` preserves that class, so a caller-supplied write becomes an operator-visible `500` rather than a `4xx` wrapping `forma.ErrInvalidInput`. This conflicts with the write-path error semantics in `docs/error-handling.md:7-20,119-130`. The PR description accurately documents this known boundary; it is not introduced by this PR, so I am recording it as follow-up work rather than blocking #353.

## Spec

Issue #282 and its 2026-07-25 decision are satisfied:

- `readJSONBody` now consistently calls `UseNumber`.
- The consumer audit addressed the live schema-validation gap and the boolean conversion case; relations and query condition parsing remain outside decoder-produced numeric maps.
- The exact-sidecar pointer gap is covered, including the required `*int` and `*int32` shapes.
- `TestHTTPBigintRoundTripAbove2p53` supplies the required HTTP `>2^53` POST/GET acceptance probe.
- #281 remains open, preserving the required sequencing.

The additional pointer shapes (`*int16`, `*float64`, `*string`) are adjacent, tested, and consistent with the audit goal; no harmful scope expansion found.

## Standards

No newly introduced file/function-size, naming, layering, or error-wrapping violation found in the final diff. The out-of-range-literal error classification above is an existing documented-semantics gap retained by the PR.

## Assessment

Approve after the required CI checks complete successfully. The implementation closes #282 without changing the documented deliberate ceiling for values outside `int64` storage.

## Observations

The validation-instance conversion is necessary because `jsonschema-go` classifies raw `json.Number` as a string. Converting integral in-range literals to `int64` preserves exact constraints above `2^53`, while fractional literals retain the existing `float64` behavior.

## Verification

Reviewed final PR diff, all seven commits, and issue #282 including the decision comment. Current GitHub checks: `Lint`, `Test`, `Benchmark Smoke`, and `k6 Smoke` passed; `E2E Smoke Tests` is still in progress. No local test run was performed because the local checkout remains on `main`, not the PR head.

## Summary

No blocking findings for #353. Merge only after the pending E2E check is green; track the `1e400` validation error-classification behavior separately.

Issue #355, opened 2026-08-04
GetPersistentRecordsByAttrValues binds numeric operands as []float64 (bigint exactness; plus EAV integer-typed attr edge)


iceboundrock:
Discovered while auditing the bigint comparison-value path for #281. #281 fixed the predicate binders in `internal/sqlgen` (both dialects); this batch lookup path was out of that issue's scope and still rides float64.

## The float64 hop

`internal/postgres_persistent_repository_by_attr_values.go:87-98` — the numeric-family arm of the batch `value_numeric = ANY($n)` builder:

```go
case forma.ValueTypeNumeric, forma.ValueTypeInteger, forma.ValueTypeBigInt, forma.ValueTypeSmallInt:
    numeric := make([]float64, 0, len(values))
    for _, v := range values {
        parsed, err := strconv.ParseFloat(v, 64)
        ...
```

The error message even hardcodes `"not parseable as float64"`. For `ValueTypeBigInt`, an operand above 2^53 (or any integral literal with ≥16 significant digits) is rounded before it reaches the query, so the `ANY` set is not the set the caller asked for.

## Proposed treatment

Same shape as the #281 fix: parse integral literals **int64-first**, fall back to float64 only for fractional values, and bind `[]int64` (or a `[]any` mixed slice) so pgx encodes into NUMERIC losslessly. Update the error message to name the numeric family rather than float64.

**This may end up document-only.** The EAV storage ceiling from #205 still applies: EAV-only bigint attributes round at *write* time at 2^53, so exact operands above that ceiling would match nothing anyway. If that is the accepted semantics, the resolution is a doc comment stating the ceiling — the same call that was made for the predicate path in #281. Decide explicitly rather than leaving the float64 parse unremarked.

## Related latent edge (fold in here)

#281 deliberately left **INTEGER/SMALLINT** DuckDB binds as float64, on the grounds that those physical column ranges cap at 2^31 / 2^15 — far below float64's exact-integer range. That reasoning holds for hot-field columns but **not** for an EAV-only *integer-typed* attribute: EAV storage has no physical range cap, so a value above 2^53 on such an attribute could produce a PG-vs-DuckDB filter disagreement of exactly the kind #281 set out to eliminate.

Whether this is reachable depends on whether writes to an integer-typed EAV attribute are range-validated. Worth confirming; if reachable, the integer family needs the same int64-first bind as bigint (in both this file and `parseDuckDBRawParam`).

[...]

Issue #354, opened 2026-08-04
fix(federated): hot-only routing silently drops the keyset cursor


iceboundrock:
Discovered while auditing the bigint comparison-value path for #281. This is an engine-correctness bug, **independent of numeric precision**.

## Symptom

A federated query that carries a keyset continuation cursor returns an **unfiltered first page** — the cursor is silently ignored — whenever the request routes hot-only. The caller sees rows it has already consumed, and pagination never advances.

## Mechanism

`internal/federated/engine.go:150-153`:

```go
if fq.PreferHot || (len(fq.PreferredTiers) == 1 && fq.PreferredTiers[0] == model.DataTierHot) {
    recordHotOnlyGatePlan(opts, tables)
    return e.queryPostgresOnlyWithPlan(ctx, tables, fq, opts)
}
```

The short-circuit fires **after** `validateKeysetTiebreak` has accepted the cursor (engine.go:140-144), and `queryPostgresOnlyWithPlan` has no keyset support at all — it neither applies the cursor predicate nor errors. So a non-nil `fq.KeysetCursor` is dropped on the floor rather than being honoured or rejected.

The keyset predicate itself is correct on the federated path (`generateKeysetWhereClause`); the defect is purely in the hot-only routing gate.

## Expected

One of:

1. Apply the keyset predicate on the Postgres-only path too (preferred — hot-only + keyset is a legitimate request shape), or
2. Fail closed: return an error when `KeysetCursor` is set and the request routes hot-only, so callers cannot receive a silently wrong page.

Whichever is chosen should be pinned by a test that sets `PreferHot` (and separately `PreferredTiers == [hot]`) together with a cursor and asserts the cursor is not ignored.

## Notes

- Not reachable via a continuation token today — no token codec exists (see the #281 audit) — so the exposure is limited to in-process callers that construct `KeysetCursor` directly plus any future HTTP cursor surface. That makes it cheap to fix now and expensive to discover later.

Refs #281.

iceboundrock:
## Design (approved) — fail closed by intent, override the cost heuristic

Posting the full design before implementation, per AGENTS.md. This supersedes the two-option framing in the issue body: the audit below **widens the scope** (four routing doors, not one) and **reclassifies the issue's stated preference** (PG-side keyset is a feature, not a fix).

---

## 1. Audit findings that changed the design

### 1.1 The cursor is dropped at one choke point reached by four doors

[...]

PR #356, opened 2026-08-04
fix(sqlgen): exact bigint comparison binds above 2^53 — predicates both dialects, keyset contract pinned (#281)

### LINKED ISSUE #281: bigint predicate and keyset-cursor bindings above 2^53: comparison-side float64 exactness audit

iceboundrock:
Closes #281. Refs #205, #282.

## What the audit found

The issue asked to *trace* the comparison-value path and fix or document what it revealed. The trace found one live bug, one ceiling worth documenting, and one hypothesised risk that turned out not to be reachable today.

**The path.** Condition values arrive as the `"op:value"` **string** DSL and stay strings end to end. So **#282 / `UseNumber` is a no-op for comparisons** — it fixes the write/validation path, not this one. Every float64 hop that mattered was inside `internal/sqlgen`.

**Fixed — predicates disagreed across tiers (the live bug).** Pre-fix, the DuckDB binder `parseDuckDBRawParam` (`internal/sqlgen/dualpath_sql_helpers.go:122-126`) parsed the literal with `ParseFloat`, and `ToDuckDBParam` rendered float64 with `%.15g` (`internal/sqlgen/duckdb_type_mapper.go:208-210`). So `equals:9007199254740993` bound the string `"9.00719925474099e+15"` (= 9007199254740990) — a value **held by no row**: the predicate could not address these values at all, and equality returned nothing. `%.15g` of MaxInt64-adjacent literals renders `"9.22337203685478e+18"`, outside INT64 range, so `CAST(? AS BIGINT)` throws. Meanwhile the PG main-table path (`ConvertPgMainValue`) already bound exact int64 — so **hot and cold tiers evaluated a different filter for the same query**, reachable from `POST /api/v1/advanced_query`. That tier disagreement is the real defect. Separately, the PG EAV path deliberately downcast int64→float64 (`internal/sqlgen/predicate_normalizer.go:255-257`). Both now bind exact int64.

**Documented — the EAV storage ceiling.** EAV-only bigint attributes round at *write* time (2^53, the #205 scope). With exact binds, values above that ceiling now miss on all tiers alike instead of false-matching a colliding rounded row — parity preserved, semantics honest.

**Keyset (M-2) — hypothetical, now fenced.** No continuation-token codec exists anywhere in the tree; `KeysetCursor` is in-process only and not HTTP-reachable, and `generateKeysetWhereClause` binds cursor values verbatim. No precision is lost today, so no code change — only a pinned contract.

## Changes

5 commits on `issue-281-bigint-comparison-exactness` (base `6a2de1c`):

- **2dc0167** `fix(sqlgen)`: bind integral bigint/numeric predicates as exact int64 on both dialects. `parseDuckDBRawParam` parses int64-first for `Numeric`/`BigInt` so `ToDuckDBParam` renders the exact decimal via `%d`; `normalizePgEavPayload` stops downcasting int64→float64. `ToDuckDBParam`'s float64 arm is untouched — the fix stops *routing* integral literals through it.
- **05c3455** `test(sqlgen)`: pin the keyset raw-bind int64 exactness contract (`TestKeysetArgsPreserveInt64Above2p53`), document the M-2 defense at the binder (`internal/sqlgen/keyset_where.go`) and as a decoder rule in `docs/federated-query/requirements.md` §6.3.
- **acaa056** `test(e2e)`: bigint filter probes above 2^53 on both dialects.
- **3b582ee** `test(e2e)`: correct the pre-#281 mechanism narrative in the filter-probe comments (the earlier "matched the wrong row" story was refuted — the pre-fix binder addressed a value no row held).
- **1ccb09f** `test(e2e)`: keyset continuation walk across the 2^53 boundary (`TestBigintKeysetBoundaryNoDupNoSkip`).

Also: characterization matrix gains a bigint fixture with >2^53 / MaxInt64 rows across all three emitters; integer-family EAV args flip to int64.

## Behavior changes reviewers should weigh

1. **Below 2^53 as well.** PG EAV args for the whole numeric family flip float64→int64 on integral literals — the comparison becomes `numeric = bigint`, which is param-promoting and index-friendly instead of casting the column to `float8`. And DuckDB integral numeric/bigint literals with **≥16 significant digits** now bind exactly; old `%.15g` corrupted those even below 2^53.
2. **User-visible semantics change for EAV-only bigint attrs.** Exact binds above 2^53 now *miss* on all tiers, where pre-fix equality could false-match via colliding rounding. This is intended per the #205 storage ceiling — tier parity is preserved and the miss is the honest answer.
3. **INTEGER/SMALLINT DuckDB binds deliberately stay float64** (column ranges cap at 2^31 / 2^15, far below float64's exact-integer range) — pinned in the characterization matrix. Latent edge noted: an EAV-only *integer-typed* attribute has no physical range cap, so a >2^53 value on such an attribute could still produce PG-vs-DuckDB disagreement. Folded into follow-up #355 rather than widened here.
4. **Keyset: no code change** — contract pin only. No token codec exists (M-2 is hypothetical), now fenced by the unit pin + binder doc comment + `requirements.md` §6.3 decoder rule, so a future decoder cannot silently reintroduce the float64 hop.
5. **Note for reviewers of the e2e tests:** the harness oracle is float64-blind by design, so the new tests use exact-set / exact-sequence assertions rather than oracle comparison. The cold leg uses the `RunInit` + `DELETE change_log` idiom — `RunInit` alone leaves rows dirty and hot-served, which would have made the DuckDB-binder claim vacuous.

## Plan deviations (both strengthened the plan)

- **Task 3:** the plan's cold-leg setup (`RunInit` alone) was insufficient — rows stayed dirty and hot-served. Fixed with the `RunInit` + `DELETE change_log` idiom, so the DuckDB-binder claim is genuine.
- **Task 4:** the plan's page-size-2 walk never landed a cursor on 2^53+1 (cursors fell on 2^53 and 2^53+2). Replaced with table-driven walks (page sizes 3/5/2; asc/asc/desc) so every walk pins a continuation cursor on a float64-unrepresentable value, with cursor positions asserted.

## Follow-ups filed

- #354 — hot-only routing silently drops the keyset cursor (`internal/federated/engine.go:150-153`): `PreferHot` / hot-only tiers short-circuit to `queryPostgresOnlyWithPlan`, which has no keyset support, so a cursor-bearing query returns an unfiltered page instead of paginating or erroring. Engine-correctness bug, independent of precision.
- #355 — `GetPersistentRecordsByAttrValues` binds `[]float64` (`internal/postgres_persistent_repository_by_attr_values.go:87-98`): batch `value_numeric = ANY($n)` builds args via `ParseFloat`. Same int64-first treatment as this PR if that path must be exact for bigint attrs (may be document-only under the #205 ceiling). Also carries the EAV integer-typed-attr edge from item 3 above.

## Verification

- `make lint` (pinned golangci-lint v1.64.8), red-anchor-verified.
- `make test` and the `-race` variant: all green.
- Full production e2e suite: **213/213 in 2m25s**, no flake re-runs needed.

### DIFF
--- a/docs/federated-query/requirements.md
+++ b/docs/federated-query/requirements.md
@@ -736,6 +736,15 @@ def decode_cursor(encoded: str) -> Cursor:
     )
 ```
 
+**Decoder numeric exactness (#281):** the decoder MUST NOT route numeric cursor
+fields through a default JSON number path — Go's `encoding/json` decodes into
+`float64`, which silently rounds integers above 2^53 and makes the resulting
+`BIGINT col > DOUBLE param` boundary skip or duplicate one row of the next page.
+Integer sort keys must either stay strings (as the pseudocode's `sk` already
+models) or decode via `json.Number` to an exact `int64` (`numutil.Int64Exact`)
+before reaching `model.KeysetCursor.Values`, which the SQL generator binds
+verbatim without coercion (pinned by `TestKeysetArgsPreserveInt64Above2p53`).
+
 ### 6.4 Backward Pagination SQL
 
 ```sql
--- a/internal/e2e_harness/production/bigint_filter_boundary_e2e_test.go
+++ b/internal/e2e_harness/production/bigint_filter_boundary_e2e_test.go
@@ -1,0 +1,114 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"math"
+	"testing"
+
+	"github.com/google/uuid"
+)
+
+// boundaryProbe is one filter probe plus the exact set of seeded rows it must
+// return. The expectation is a row-identity set, never an oracle comparison:
+// the oracle normalizes numerics to float64 (oracle.go) and is structurally
+// blind to a 2^53 miss.
+type boundaryProbe struct {
+	name   string
+	filter Filter
+	want   []*Event
+}
+
+// TestBigintFilterBoundaryBothDialects (#281) probes comparison-side exactness
+// for a bound bigint above 2^53 on both binder paths: PreferHot (Postgres,
+// ConvertPgMainValue int64) and federated (DuckDB, parseDuckDBRawParam → exact
+// decimal string). Pre-#281 the DuckDB binder rode float64 + %.15g, so
+// equals:9007199254740993 bound 9007199254740992 and matched the WRONG row —
+// hot and cold tiers disagreed on the same filter. #205 made those values legal
+// column state, so the filter has to be able to address them.
+func TestBigintFilterBoundaryBothDialects(t *testing.T) {
+	cluster := SharedCluster(t)
+	env := NewEnv(t, cluster)
+	ctx := context.Background()
+	wide := DefaultSchemaFixtures()[1]
+
+	atPow53 := CreateEvent(wide, map[string]any{"title": "b-2p53", "amount": int64(1) << 53})        // 9007199254740992
+	abovePow53 := CreateEvent(wide, map[string]any{"title": "b-2p53p1", "amount": int64(1)<<53 + 1}) // 9007199254740993
+	nearMax := CreateEvent(wide, map[string]any{"title": "b-maxm1", "amount": int64(math.MaxInt64) - 1})
+	atMax := CreateEvent(wide, map[string]any{"title": "b-max", "amount": int64(math.MaxInt64)})
+	seeded := []*Event{atPow53, abovePow53, nearMax, atMax}
+	if err := env.ApplyEvents(ctx, seeded...); err != nil {
+		t.Fatalf("apply boundary rows: %v", err)
+	}
+
+	probes := []boundaryProbe{
+		{"equals_2p53_plus_1", Filter{Attr: "amount", Op: "equals", Value: "9007199254740993"}, []*Event{abovePow53}},
+		{"equals_2p53_excludes_neighbor", Filter{Attr: "amount", Op: "equals", Value: "9007199254740992"}, []*Event{atPow53}},
+		{"gt_2p53_boundary", Filter{Attr: "amount", Op: "gt", Value: "9007199254740992"}, []*Event{abovePow53, nearMax, atMax}},
+		{"equals_maxint64_minus_1", Filter{Attr: "amount", Op: "equals", Value: "9223372036854775806"}, []*Event{nearMax}},
+		{"lte_maxint64_minus_1_excludes_max", Filter{Attr: "amount", Op: "lte", Value: "9223372036854775806"}, []*Event{atPow53, abovePow53, nearMax}},
+	}
+
+	// PG dialect: rows are unflushed, PreferHot short-circuits to postgres-only.
+	runBoundaryProbes(ctx, t, env, "hot-pg", Query{Schema: wide, PreferHot: true, Limit: 100}, false, probes)
+
+	// DuckDB dialect: export every row to base, then drop the change_log
+	// entries so the rows leave the dirty set and are served from cold parquet
+	// (the RunInit-plus-DELETE idiom of TestNullAndBoundaryRoundTripAcrossTiers;
+	// RunInit alone leaves the rows unflushed, i.e. still hot).
+	if _, err := env.RunInit(ctx, wide); err != nil {
+		t.Fatalf("run init: %v", err)
+	}
+	env.ExecSQL(ctx,
+		"DELETE FROM change_log WHERE schema_id = $1 AND row_id = ANY($2)",
+		wide.ID, rowIDs(seeded))
+	runBoundaryProbes(ctx, t, env, "cold-duck", Query{Schema: wide, Limit: 100}, true, probes)
+}
+
+// runBoundaryProbes executes every probe against base and asserts the routing
+// dialect plus the exact row set. wantDuck pins which binder actually ran:
+// without it a mis-routed leg would silently retest the other dialect.
+func runBoundaryProbes(ctx context.Context, t *testing.T, env *Env, label string,
+	base Query, wantDuck bool, probes []boundaryProbe) {
+	t.Helper()
+	for _, p := range probes {
+		q := base
+		q.Filters = []Filter{p.filter}
+		res := mustQuery(ctx, t, env, q)
+		if got := res.Plan.Routing.UseDuckDB; got != wantDuck {
+			t.Errorf("%s/%s: UseDuckDB = %t, want %t (routing %+v)",
+				label, p.name, got, wantDuck, res.Plan.Routing)
+		}
+		assertExactBigintRowSet(t, label+"/"+p.name, res, p.want)
+	}
+}
+
+// assertExactBigintRowSet fails unless res.Records is exactly the given events
+// (as an unordered set) and each carries the exact seeded amount. Both halves
+// matter: a float64 bind can drop a row that should match AND admit its
+// neighbour, and only an exact-set assertion sees both.
+func assertExactBigintRowSet(t *testing.T, label string, res *QueryResult, want []*Event) {
+	t.Helper()
+	got := map[uuid.UUID]int64{}
+	for _, r := range res.Records {
+		got[r.RowID] = r.Int64Items["bigint_01"]
+	}
+	if len(got) != len(want) {
+		t.Errorf("%s: %d rows, want %d (got %v)", label, len(got), len(want), got)
+	}
+	for _, ev := range want {
+		amt, ok := got[ev.RowID]
+		if !ok {
+			t.Errorf("%s: missing row %s (%v)", label, ev.RowID, ev.Attrs["title"])
+			continue
+		}
+		wantAmt, isInt := ev.Attrs["amount"].(int64)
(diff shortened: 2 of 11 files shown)


iceboundrock:
## Findings

1. **P1 - bigint precision remains lossy for accepted integral spellings.** `parseDuckDBRawParam` keeps precision only when `strconv.ParseInt` succeeds (`internal/sqlgen/dualpath_sql_helpers.go:129-133`). The condition DSL accepts arbitrary non-empty text after `op:` (`internal/conditionexpr/parser.go:40-50`), so `equals:9007199254740993.0` and `equals:9.007199254740993e15` fall through to `ParseFloat` and then the unchanged `%.15g` rendering. The PostgreSQL paths do likewise through `numutil.TryParseNumber` (`internal/numutil/convert.go:99-107`). Both spellings denote an integral bigint value but again lose the >2^53 boundary exactness that #281 set out to fix. The new E2E probes cover only bare decimal strings (`bigint_filter_boundary_e2e_test.go:48-53`).

   Either reject non-integer bigint spellings with `ErrInvalidInput`, or preserve exact decimal semantics for integral decimal/exponent forms; add hot and cold regression probes for the selected contract.

2. **P2 - the changed EAV binder has no execution-level boundary probe.** The PR changes `normalizePgEavPayload` to bind `int64` (`internal/sqlgen/predicate_normalizer.go:251-263`) and documents EAV-only bigint semantics, but the filter E2E exercises only bound `amount` (`bigint_filter_boundary_e2e_test.go:39-53`). The `total` EAV-only case is characterization-only. Add hot-Postgres and cold-DuckDB assertions that an EAV-only value beyond the documented write ceiling has the intended exact-bind/miss behavior, so both altered binder paths are demonstrated.

iceboundrock:
## Spec

Issue #281 requires tracing condition and cursor values to the DuckDB/PostgreSQL binders, adding >2^53 filter and keyset-continuation probes for both dialects, then fixing or documenting the result.

- The bare-integer filter path is traced and tested in hot PostgreSQL plus cold DuckDB.
- The DuckDB-only keyset path is documented as in-process and receives exact-int64 unit and E2E continuation coverage.
- Coverage is partial for the two findings above: accepted integral decimal/exponent spellings remain lossy, and the altered EAV PostgreSQL path lacks a runtime boundary probe.

No separate product spec/PRD was linked beyond #281.

iceboundrock:
## Standards

- `reversedInt64s` (`internal/e2e_harness/production/bigint_keyset_boundary_e2e_test.go:170`) is a private helper that is not verb-first. `coding-standard.md` §4 requires verb-first private-helper names; `reverseInt64s` conforms.
- `normalizePgEavPayload` remains 85 lines after this PR (`internal/sqlgen/predicate_normalizer.go:232-316`). This is below the 100-line hard limit but exceeds the >80-line refactoring trigger in `coding-standard.md` §7. The PR does not materially worsen its cohesion, so record this as a follow-up rather than a blocking defect.
- No new file exceeds 500 lines, and the reviewed change uses contextual wrapping for its new direct parse errors.

iceboundrock:
## Assessment

The change correctly repairs the primary bare-integer bigint bind path and adds meaningful tier-routing, exact-set, and keyset-boundary coverage. However, P1 leaves the same user-visible precision failure reachable through accepted equivalent numeric spellings. Had this review occurred before merge, that would warrant a request for changes. P2 is an important regression-coverage gap for a modified execution path.

iceboundrock:
## Observations

- The implementation intentionally broadens behavior beyond #281's bigint-focused audit: integral `numeric`, `integer`, and `smallint` paths receive changed PostgreSQL EAV binding behavior, and `numeric` shares the new DuckDB int64-first path. The PR describes this, but it is still scope expansion and should have explicit compatibility coverage.
- The cold-leg setup correctly removes change-log entries after `RunInit`; this prevents the E2E from silently exercising the hot path.
- The exact-row and exact-sequence assertions appropriately avoid the float64-normalizing oracle for this boundary.

iceboundrock:
## Verification

Reviewed `git diff 6a2de1c...408e695` (the PR merge base through merge commit) and the five PR commits.

Fresh local verification:

- `make test` passed.
- Focused `internal/sqlgen` tests passed: `TestToDualClauses_Characterization`, `TestToDuckDBParam_Numeric_FromInt64MaxValue`, and `TestKeysetArgsPreserveInt64Above2p53`.
- `git diff --check 6a2de1c...408e695` produced no whitespace errors.

GitHub CI is green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke. The Docker-backed full production E2E suite was not rerun locally in this review.

iceboundrock:
## Summary

Review of merged PR #356 against issue #281 and the repository standards found **two findings**: one P1 correctness defect in accepted bigint literal spellings, and one P2 EAV execution-coverage gap. There is one minor naming-standard deviation and one non-blocking refactoring trigger. The strongest part of the PR is its bare-integer cross-tier and keyset coverage; the highest-risk remaining issue is P1.

iceboundrock:
## Review response

Both findings verified against the code and resolved in **#357 / PR #358**; section-by-section:

**Findings P1 (integral spellings lossy)** — Confirmed: `parseDuckDBRawParam` kept exactness only on the `ParseInt` fast path, and both PG paths share the same fallthrough via `numutil.TryParseNumber`; `equals:9007199254740993.0` / `equals:9.007199254740993e15` reproduced the pre-#281 loss. Of the two offered resolutions, PR #358 takes **preserve-exact-semantics** rather than reject: rejecting would break currently-legal filters like `"42.0"` (same normalize-don't-reject reasoning as #219). Mechanism: `big.Rat` refinement gated on the existing `ParseFloat` acceptance, so the grammar neither widens nor narrows. Fixing this surfaced a second-order issue our re-review classified Important — ungated `big.Rat` on caller-supplied literals is a CPU amplifier (`"1e-1000000"`: 10 bytes → ~20ms × 3 emitters per leaf, unbounded leaf count). PR #358 bounds it (magnitude / zero-fold / 64-char guards, 22ms → 2.2µs, mutation-verified) and discloses the one residual: integral spellings >64 chars fall back to the pre-#357 float64 bind.

**Findings P2 (no EAV execution probe)** — Confirmed; `total` was characterization-only. PR #358 adds `TestEAVBigintFilterBoundaryBothDialects`: miss at the unstorable exact value / match at the stored rounded value, both legs, routing asserted, exact-set assertions, plus the P1 spellings probed at execution level on bound `amount`. One note from that work: the both-tiers-agree outcome is load-bearing on `eav_data.value_numeric` being `NUMERIC` — a `DOUBLE PRECISION` column would coerce the int64 operand back to float64 on the PG side and diverge from DuckDB's `TRY_CAST AS BIGINT` pivot.

**Standards** — Both items in PR #358: `reversedInt64s` → `reverseInt64s`; `normalizePgEavPayload` 85 → 48 lines via `parsePgEavValue` extraction (byte-identical behavior, characterization matrix as the guard), which clears the §7 trigger rather than deferring it.

**Observations (scope expansion)** — The broadened surface (integral `numeric`/`integer`/`smallint` PG EAV binds flipping float64 → int64; `numeric` sharing the duck int64-first path) is intentional: the four types share one normalizer arm, and splitting bigint out would have left the family inconsistent for the identical loss mechanism. Compatibility grounds: below 2^53 the int64 and float64 binds are value-identical in every consumer (PG `NUMERIC` comparison, duck `%d` vs `%.15g` rendering), pinned per-emitter by the characterization matrix, and the existing cross-tier e2e suites (filter-LWW, or-composite, query) continuously exercise below-2^53 numeric filters. PR #358's below-ceiling sanity row adds an execution-level EAV data point. Agreed it warranted explicit statement — the PR body now carries it.

**Spec / Assessment / Verification / Summary** — No dispute with the partial-coverage assessment; the two gaps named are exactly what PR #358 closes. The "had this occurred before merge, request-changes" framing is fair — P1 was a real reachable loss in the same class #281 targeted.

Issue #361, opened 2026-08-05
Disclosed 4xx with withheld operator detail carries no error_id for correlation


iceboundrock:
Follow-up candidate recorded by #313 (docs/error-handling.md, "Log levels are contract").

Since #313 a disclosed 4xx can withhold operator detail (`forma.WithOperatorDetail`): the body carries the published message, the detail goes to a WARN log line. Unlike redacted responses, such a body carries **no `error_id`**, so a caller reporting the failure cannot quote a correlation id and the operator must locate the WARN line by time/schema instead.

Six assertions currently pin "a disclosed 4xx emits no correlation fields", so adding `error_id` there is a deliberate public contract change, not a bug fix. File to decide whether operational experience justifies it.

iceboundrock:
**Artifact: implementation plan** (working copy: `docs/superpowers/plans/2026-08-18-issue-361-error-id-withheld-detail.md`, gitignored). Includes the design ruling taken 2026-08-18: **Option A — error_id only on withheld-detail disclosed 4xx (the Warnw branch)**; rejected: uniform error_id on all disclosed 4xx (dangling ids below the production Info threshold) and close-as-not-planned.

---

# #361: error_id on Withheld-Detail Disclosed 4xx — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** A disclosed 4xx that withholds operator detail (`forma.HasOperatorDetail`) carries an `error_id` in its body, matching the `error_id` field on its Warnw log line, so a caller can quote a correlation id and the operator no longer locates the WARN line by time/schema.

**Architecture:** One branch change in `respondErrorWithStatus` (`internal/httpapi/error_response.go`): the disclosed branch splits on `forma.HasOperatorDetail` already (for log level); the detail-withholding half additionally mints a UUID, puts it on the Warnw line and into `APIResponse.ErrorID`. The detail-less half (Debugw) is deliberately unchanged — its log line does not survive the production Info threshold, so an id there would correlate to nothing. `error_class` remains absent from every disclosed body and becomes the sole redacted-vs-disclosed discriminator.

**Tech Stack:** Go, zap (observer for log assertions), `github.com/google/uuid`. No new dependencies, no new files.

[...]

Issue #360, opened 2026-08-05
Request-parsing 4xx sites interpolate raw error text outside the disclosure gate


iceboundrock:
Deferred from #313 (user decision: follow-up, not in-PR).

`internal/httpapi/server.go` has 17 direct `writeError` sites that interpolate an error into the body (`fmt.Sprintf("invalid json body: %v", err)`, `uuid.Parse` echoes, one bare `err.Error()` at `parseCreateObjects`'s call site). They bypass `respondErrorWithStatus` deliberately: all are request-parsing failures that never touch the manager, the engine, S3, or `PG_CONN`, so the text is json/uuid parsing prose with no operator data. The AST guard `TestWriteErrorAlwaysCarriesALiteral4xxStatus` pins their statuses but explicitly does not inspect messages.

After #313 these are the only 4xx bodies not built from a deliberate publication. Aligning them (curated messages, or `forma.InvalidInputf` + the gate) would make the disclosure rule total; the cost is reworking the AST guard and renegotiating the parse-error body texts.

iceboundrock:
**Implementation plan (2026-08-17).** Working copy: `docs/superpowers/plans/2026-08-17-issue-360-parse-error-disclosure-gate.md` (gitignored, local). User ruling incorporated: **InvalidInputf + gate** over curated fixed messages. Note: the issue's `server.go` references are stale — all 17 sites now live in `internal/httpapi/handlers.go` after the file split.

---

# Issue #360 — Route request-parsing 4xx bodies through the disclosure gate

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Make the #313 disclosure rule total: the 17 `writeError` sites in `internal/httpapi` that interpolate an error into a 4xx body are converted to `forma.InvalidInputf` carriers routed through `respondError`, and a new source-level guard pins that every remaining direct `writeError` site passes a string-literal message.

[...]

Issue #359, opened 2026-08-05
Batch success bodies carry raw error chain text in OperationError.Error


iceboundrock:
Deferred from #313 (user decision: follow-up, not in-PR).

`entity_batch_service.go`'s best-effort path populates `forma.OperationError.Error` with raw `err.Error()` chain text, which serializes inside `BatchResult.Failed` on a **success** body (`writeSuccess`) — bypassing the `respondErrorWithStatus` gate and the credential scrubber entirely.

Currently latent over HTTP: every handler builds `Atomic: true` batches, which return the error to the gated path instead. The surface is live for Go embedders and for any future non-atomic handler.

Direction: populate `OperationError.Error` from the #313 publication (`forma.PublicError` resolution, falling back to a fixed redacted string when the error publishes nothing), so the batch item error follows the same deny-by-default disclosure rule as 4xx bodies.

iceboundrock:
**Verification note (2026-08-17): this issue was already implemented on `main` by #318's PR #392 — closing as completed.** PR #392 (merge commit `5fedafd`, merged 2026-08-12) carried the fix in commits `8c276c2` ("publish only the resolved message from a best-effort batch") and `01fb0f4` ("stop a best-effort batch publishing raw error text"), but its body had no `Closes #359` reference, so the issue stayed open.

What exists on `main`, checked against this issue's direction:

1. **Publication-based rendering.** `internal/entity_batch_error.go` defines `resolveBatchErrorMessage`, and `executeBestEffortBatch` (`internal/entity_batch_service.go:163`) populates `forma.OperationError.Error` exclusively through it. A publishing error answers `forma.ResolvePublicMessage` — the same #313 resolution the HTTP boundary uses — then passes through `redact.ConnStringPassword`, the same scrubber `respondErrorWithStatus` applies. That second step goes beyond this issue's ask: a published message can interpolate caller-supplied text (the registry publishes the caller's schema name on not-found), so the scrub is load-bearing even on the publishing branch.
[...]

PR #362, opened 2026-08-05
Typed public client messages: 4xx bodies carry only deliberately published text (#313)

### LINKED ISSUE #313: Mixed error chains can disclose an operator cause in a 4xx body (typed public client messages)
Follow-up from #307 (issue #301). This is #301's P2 residual, recorded rather than closed.

`isClientError` (`internal/httpapi/error_response.go`) uses `errors.Is`, which matches **any leaf** in the chain. So a multi-cause error such as:

```go
errors.Join(forma.ErrInvalidInput, credentialBearingDriverError)
```

takes the verbatim 4xx branch and discloses the second cause. Multi-cause wrapping (`fmt.Errorf("%w: %w", …)`) is used throughout `internal/federated`.

**Currently mitigated, not closed:**
- The credential scrubber runs on verbatim bodies too, so no chain shape can leak a `password=` value.
- A mixed-chain canary test pins that behaviour.
- Review confirmed every `%w: %w` / `errors.Join` in `internal/federated` today joins *typed read-path* sentinels (`ErrFederatedReadFailed`, `ErrPostgresReadFailed`) and never a client sentinel, so the gap is **theoretical in production right now**.

**What remains:** a mixed chain can still surface a *non-credential* operator cause — an S3 object key, a manifest path — in a 400 body, because the verbatim branch discloses raw chain text.

**Direction:** stop disclosing raw `err.Error()` for client errors. Give the client-facing sentinels a typed public message (an interface such as `PublicMessage() string`, or a dedicated carrier per class) so the HTTP boundary emits only text the error deliberately published, rather than whatever the chain happens to contain. That is deny-by-default on the disclosure axis, matching what #301 already did on the status axis.

This is a larger change than #307 could carry — it touches every write-path validator that currently relies on its raw message reaching the caller.

iceboundrock:
Closes #313.

## What

Client-facing errors now carry two messages with different audiences: `Error()` stays the operator's full-chain copy; `PublicMessage()` (new `forma.PublicError` carrier interface) is the only text the HTTP boundary emits on a 4xx. An error that wraps a client sentinel without a carrier keeps its status and answers the redacted shape (`error_class`/`error_id`) — deny-by-default on the disclosure axis, matching what #301 did on the status axis.

- **Mechanism** (`client_error.go`): `InvalidInputf`/`NotFoundf`/`Conflictf` reproduce the old `fmt.Errorf("<msg>: %w", sentinel)` text byte-for-byte (which is what spared ~50 non-httpapi message assertions); `WrapPublicf` prefixes both messages and degrades to a plain wrap over operator errors; `WithOperatorDetail` keeps operator facts in `Error()`/the chain and out of the publication.
- **Conversion sweep**: all 74 client-sentinel wrap sites (`ErrInvalidInput` 56, `ErrNotFound` 17, `ErrConflict` 1) across entity services, sqlgen/conditionexpr, transform/schemavalidate, federated, schemameta, and the repositories.
- **Boundary flip** (`internal/httpapi/error_response.go`): disclosure gate becomes `isClientError` + `errors.As`-resolved publication; `hasMultipleCauses`/`canDiscloseVerbatim` deleted; published messages still pass `redactCredentials`; log levels Debug (published) / **Warn** (published with withheld detail — the log is its only copy) / Error (redacted).
- **Guard** (`client_error_guard_test.go`): go/ast test — zero `fmt.Errorf`/`errors.Join` calls referencing a client sentinel, empty allowlist, mutation-verified.
- **Docs**: `docs/error-handling.md` rewritten where overturned, including the false "Blast radius: nil" claim (the `duckdb_query_build.go` mixed chain always existed and is HTTP-reachable) and the #309 sentinel-suffix clause; both "accepted disclosures" recorded as closed. `AGENTS.md` Error Semantics updated.

## Externally visible contract changes

1. 4xx bodies lose the sentinel suffix (`: invalid input` etc.) and any non-published chain text; the caller-actionable message itself is unchanged at every converted site.
2. The issue's motivating case flips usefully: an unrenderable caller-supplied `s3_parquet_path_template` now answers 400 naming the template and reason (was: opaque `internal error` 400); `text/template` internals stay log-only.
3. Unique-violation 409s publish a curated summary instead of `pgErr.Detail`'s physical column prose; ID-keyed registry 404s stop rendering internal `int16` ids (both were documented "accepted disclosures" — now closed; ids/detail remain in the operator log).
4. Four service-wiring guards in `entity_query_service.go` ("config is required" / "not initialized") were misclassified 400 → now plain errors → 500 (user decision; recorded in docs status-change log).
5. jsonschema-go violation prose is **deliberately published** (user decision; rationale + `anyOf`/`oneOf` revisit trigger recorded at the wrap site and in docs).

## Decisions taken with the user (2026-08-04)

- Single PR, full 74-site conversion (no partial-disclosure window)
- Publish third-party jsonschema violation text verbatim
- Side channels deferred: #359 (`OperationError.Error` raw chain in batch success bodies), #360 (17 request-parsing `writeError` interpolation sites); #361 records the no-`error_id`-on-disclosed-4xx candidate
- Wiring guards 400→500 in this PR

## Verification

- `make lint` + `make test` green at **every commit** (commits 2–6 change no existing test — the untouched suite is the no-behaviour-change proof for the conversion sweep; commits 1 and 7 are red-first)
- Mutation checks: injected bare sentinel wrap → AST guard reports exact position; reverting `duckdb_query_build.go` to the pre-#313 mixed chain → both the guard and the federated publication test fail; restore is byte-identical (empty diff)
- e2e: infra smoke green; **production harness suite green** (`make test-e2e-production`, 131s)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,10 +74,10 @@ Module: `github.com/lychee-technology/forma`.
 
 Two distinct error classes (`docs/error-handling.md`):
 
-- **Write-path validation** wraps `forma.ErrInvalidInput` → surfaces as user-facing 4xx.
+- **Write-path validation** builds `forma.InvalidInputf` (or `NotFoundf`/`Conflictf`) carriers → surfaces as user-facing 4xx whose body is the carrier's published message (#313). A bare `%w` wrap of a sentinel earns the status but answers a redacted body, and fails the `TestNoBareSentinelWraps` guard.
 - **Read-path consistency** errors (metadata drift, storage column mismatch) are plain errors → operator-visible failures, not 4xx.
 
-Error messages must name the logical value type, the offending column/attribute, and the expected state.
+Error messages must name the logical value type, the offending column/attribute, and the expected state. Detail an operator needs but a caller cannot act on goes behind `forma.WithOperatorDetail` — it stays in `Error()` and the log, never in the published message.
 
 ## Pull Request Rules
 
--- a/client_error.go
+++ b/client_error.go
@@ -1,0 +1,143 @@
+package forma
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Typed public client messages (#313).
+//
+// A client-facing error carries two messages with different audiences. Error()
+// is the operator's copy: the full chain, safe to log, never safe to publish.
+// PublicMessage() is the only text that may cross a public transport, and it
+// exists only because a wrap site deliberately authored it. The HTTP boundary
+// (internal/httpapi) resolves the outermost PublicError in the chain and emits
+// nothing else on 4xx — an error built by wrapping a bare sentinel keeps its
+// status but loses its body. Deny-by-default on the disclosure axis, matching
+// what sentinel classification already does on the status axis (#301).
+//
+// Constructors return error rather than a concrete type so no call site can
+// hold a typed nil or reach past the interface. The carriers stay unexported:
+// the contract is PublicError plus errors.Is on the sentinel, nothing more.
+
+// PublicError is implemented by an error that carries a message deliberately
+// published to API clients.
+type PublicError interface {
+	error
+	PublicMessage() string
+}
+
+// InvalidInputf builds a client error classified by ErrInvalidInput whose
+// formatted message is published verbatim. Error() renders
+// "<message>: invalid input".
+func InvalidInputf(format string, args ...any) error {
+	return &clientError{sentinel: ErrInvalidInput, public: fmt.Sprintf(format, args...)}
+}
+
+// NotFoundf builds a client error classified by ErrNotFound. Error() renders
+// "<message>: not found".
+func NotFoundf(format string, args ...any) error {
+	return &clientError{sentinel: ErrNotFound, public: fmt.Sprintf(format, args...)}
+}
+
+// Conflictf builds a client error classified by ErrConflict. Error() renders
+// "<message>: conflict".
+func Conflictf(format string, args ...any) error {
+	return &clientError{sentinel: ErrConflict, public: fmt.Sprintf(format, args...)}
+}
+
+// WrapPublicf adds caller-actionable context to err, prefixing BOTH its
+// operator message and its published message: Error() is identical to
+// fmt.Errorf("<prefix>: %w", err) either way. When err carries no PublicError
+// the result IS exactly that plain wrap — an operator error can never gain a
+// publication by being wrapped. Returns nil for a nil err.
+//
+// Use it only where the wrap adds identification the caller can act on (a
+// batch index, a caller-supplied name). A layer that adds operator context
+// should stay a plain fmt.Errorf so its prefix stays out of the body.
+func WrapPublicf(err error, format string, args ...any) error {
+	if err == nil {
+		return nil
+	}
+	prefix := fmt.Sprintf(format, args...)
+	var pub PublicError
+	if !errors.As(err, &pub) {
+		return fmt.Errorf("%s: %w", prefix, err)
+	}
+	return &publicWrap{prefix: prefix, err: err}
+}
+
+// WithOperatorDetail attaches operator-only detail to a client error. The
+// detail joins the chain (errors.Is/As still see it) and its text reaches
+// Error(), so logs keep it — but it never reaches PublicMessage(). Returns err
+// unchanged when detail is nil or err carries no PublicError, and nil for a
+// nil err.
+func WithOperatorDetail(err error, detail error) error {
+	if err == nil {
+		return nil
+	}
+	var pub PublicError
+	if detail == nil || !errors.As(err, &pub) {
+		return err
+	}
+	return &operatorDetail{err: err, detail: detail}
+}
+
+// HasOperatorDetail reports whether err withholds detail from its published
+// message. internal/httpapi keys a log level on it: a disclosed 4xx whose
+// chain still holds operator text makes the log line the only copy of that
+// text.
+func HasOperatorDetail(err error) bool {
+	var d *operatorDetail
+	return errors.As(err, &d)
+}
+
+// clientError is the leaf carrier: a sentinel for classification and a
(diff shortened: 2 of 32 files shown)


iceboundrock:
## Findings

1. **P1 - A bare client-sentinel branch can borrow an unrelated publication.** `isClientError` and `resolvePublicMessage` independently search the complete error tree. A multi-cause value such as `errors.Join(fmt.Errorf("bad input: %w", forma.ErrInvalidInput), foreignPublicError{message: manifestPath})` satisfies both gates: the first branch supplies status evidence, while the second supplies `PublicMessage()`. The HTTP response then discloses `manifestPath`, even though the sentinel-bearing branch is bare and should produce the redacted `error_class`/`error_id` shape. `forma.PublicError` is externally implementable at `client_error.go:25`; the independent lookups are at `internal/httpapi/error_response.go:242-271` and `:313`. Add a regression test for this split-branch construction and bind a published message to the same chain branch that carries the client sentinel.

2. **P2 - Batch operation wrappers publish internal phase names.** `internal/entity_batch_service.go:210,236,240,288,293,301,326,373` use `forma.WrapPublicf` for texts such as `failed to get schema` and `failed to transform persistent record to json`. When the nested error is public, these implementation phases enter the 4xx body. Preserve only `operation[%d]` as the caller-actionable wrapper; use plain `fmt.Errorf` for the phase context.

3. **P2 - Internal schema ID is published.** `internal/transform/transformer.go:342` returns `InvalidInputf("... schema %d", schemaID)`. `schemaID` is an internal identifier, not caller-actionable input. Publish the attribute-level explanation and retain the ID with `forma.WithOperatorDetail`.

4. **P3 - Source file exceeds the project size limit.** `internal/federated/parquet_source_test.go` grows from 486 to 518 lines. `coding-standard.md` caps non-generated source files at 500 lines; move the added test to a focused companion test file.

## Spec

Issue #313 requires that “only deliberately authored typed public messages may cross” the HTTP boundary, and that a bare client sentinel retain its status while receiving a redacted `error_class`/`error_id` response. Finding 1 violates both requirements because publication provenance is accepted from a separate multi-cause branch.

The four entity-query service guards reclassified from 400 to 500 are not reported as scope creep: the PR description explicitly records that as a user decision.

## Standards

Findings 2 and 3 violate the updated authorship rule in `docs/error-handling.md:495-503`: `WrapPublicf` is reserved for caller-actionable identification, while internal phase names and schema IDs must remain operator-only. Finding 4 violates `coding-standard.md:7-11` (source files must not exceed 500 lines).

## Assessment

The carrier model, conversion sweep, documentation, and boundary redaction direction are sound. However, the P1 mixed-branch disclosure defect reopens the issue's core security boundary, so this change does not meet the intended disclosure contract until the two gates are provenance-bound.

## Observations

- The test suite correctly covers carrier discovery behind an operator cause, but it does not cover a bare-sentinel branch joined with a separate `PublicError` branch.
- The PR body explicitly defers batch-success and request-parsing side channels to #359 and #360; those remain outside this review.
- The 400-to-500 initialization-guard change is documented as deliberate.

## Verification

- Reviewed `c7c0d0f...a8dd605` (the PR head range) and issue #313.
- `git diff --check c7c0d0f...a8dd605` passed.
- `make lint` passed.
- `go test ./internal/httpapi ./internal/transform ./internal/federated` passed.
- CI checks attached to the PR were successful; these checks do not exercise Finding 1's split-branch construction.

## Summary

**1 P1 spec/security finding, 2 P2 standards findings, and 1 P3 standards finding.** The P1 provenance gap is the blocking issue: a public message must be resolved from the same error branch as the client sentinel, rather than from any node in the aggregate error tree.

iceboundrock:
Response to the review — all four findings verified against the merged code and addressed in **PR #363**.

**Finding 1 (P1) — confirmed by reproduction, fixed.** Built the exact split-branch construction against the merged boundary first: the foreign branch's manifest path did cross on a 400. `resolvePublicMessage` is now a preorder left-first walk in which a `PublicError` node qualifies only when a client sentinel is reachable from that node's own subtree — the regression test uses your construction verbatim. Writing the companion test exposed a second face of the same defect the review didn't name: under the old `errors.As` first-hit resolution, `errors.Join(foreignPublicError, forma.InvalidInputf("bad filter"))` published the *foreign* text and dropped the legitimate carrier's message entirely. The walk therefore steps over non-qualifying nodes rather than treating them as terminal; both directions are pinned (`TestForeignPublicationCannotBorrowSentinelBranch`, `TestForeignNodeDoesNotBlockCarrierResolution`). One note on live exposure: no type in this module or its dependencies implements `PublicError` outside the forma constructors today, so the leak needed an embedder- or future-authored implementer — but the interface is public and the contract said branch-bound, so agreed this was blocking.

**Finding 2 (P2) — fixed.** The restructure keeps `Error()` byte-identical (`WrapPublicf(fmt.Errorf("<phase>: %w", err), "operation[%d]", i)`), so operator logs are unchanged while publications become index + leaf. The docs worked example, both hand-built httpapi tests, and a new production-path test (real registry + batch service) now assert the phase name present in `Error()` and absent from the body.

**Finding 3 (P2) — fixed** with `WithOperatorDetail`, message now `attribute '%s' is not defined for this schema`. Scope note: the `attrID=%d` texts elsewhere in transform (`missing required attribute 'email' (attrID=2)`, `invalid value for attribute … (attrID=…)`) are the same class of internal identifier but are deliberately unchanged — `docs/error-handling.md`'s message-style section explicitly prescribes naming the attrID, and its examples pin that style. If you want the internal-id rule applied uniformly, that's a docs-contract change plus ~4 sites and a handful of assertions; say the word and I'll fold it into #363.

**Finding 4 (P3) — fixed**, mechanical move to `parquet_source_publication_test.go`; `parquet_source_test.go` back to 486 lines.

Verification on #363: `make lint` + `make test` green at every commit; F1 red-first (both regression tests failed against the merged boundary before the resolver change); e2e infra smoke green.

PR #363, opened 2026-08-05
Provenance-bound publication resolution + review fixes from PR #362 (#313)

iceboundrock:
Addresses all four findings from the post-merge review on #362 (issue #313).

## F1 (P1) — publication must come from the sentinel-bearing branch

Confirmed by reproduction before fixing: `errors.Join(fmt.Errorf("bad input: %w", forma.ErrInvalidInput), foreignPublicError{manifestPath})` disclosed the foreign branch's text on a 400 under the merged boundary.

`resolvePublicMessage` is now a preorder left-first walk (errors.As order, so outermost-wins and prefix accumulation are unchanged) in which a `PublicError` node qualifies only when a client sentinel is reachable from **that node's own subtree**. All forma constructors bind message and sentinel into one branch, so every real carrier qualifies unchanged; a foreign implementer qualifies only by doing the same — the contract, not a loophole.

Writing the second regression test surfaced a worse half of the same defect: under the old resolver, `errors.Join(foreignPublicError, forma.InvalidInputf("bad filter"))` published the **foreign** text and lost the legitimate carrier's message (`errors.As` first-hit). Non-qualifying nodes are now stepped over, not terminal. Both tests were red against the merged boundary before the fix.

## F2 (P2) — batch phase context out of publications

The eight phase-text sites become `WrapPublicf(fmt.Errorf("<phase>: %w", err), "operation[%d]", i)`: `Error()` byte-identical, publication = index + leaf. Docs worked example updated (`operation[0]: schema not found: nosuchschema`); the unknown-schema and re-wrap tests assert the phase name absent from the body, and a new production-path test pins it through the real registry + batch service.

## F3 (P2) — internal schema id withheld at `transformer.go`

`attribute '%s' is not defined for this schema` published; `schema %d` moves to `WithOperatorDetail` so `Error()`/logs keep it. No test asserted the old text. Scope note: the `attrID=%d` occurrences elsewhere in transform are deliberately unchanged — docs' message-style section prescribes naming the attrID; happy to sweep those too if you want the internal-id rule applied uniformly.

## F4 (P3) — `parquet_source_test.go` back under 500

Mechanical move of the added test to `parquet_source_publication_test.go` (486 + 44 lines).

## Verification

`make lint` + `make test` green at every commit; F1 red-first with both regression tests failing against the merged boundary; e2e infra smoke green.

### DIFF
--- a/client_error.go
+++ b/client_error.go
@@ -27,6 +27,69 @@ type PublicError interface {
 	PublicMessage() string
 }
 
+// ResolvePublicMessage returns the message err deliberately published, if
+// any. It is the canonical resolution used by the HTTP boundary, by
+// WrapPublicf/WithOperatorDetail to qualify their input, and by the
+// decorators' own PublicMessage delegation — one traversal, so the
+// publication a decorator carries is exactly the one the boundary would emit.
+//
+// The walk is preorder, left-first (errors.As order, so wrap prefixes
+// accumulate outermost-wins), and each node is matched the way errors.As
+// matches: direct implementation or the node's As(any) bool protocol
+// (#363 review, P2). A node qualifies only when a client sentinel is
+// reachable from that node's OWN subtree (#362/#363 reviews, P1): two gates
+// searching the whole tree independently let a mixed tree borrow — sentinel
+// evidence from one branch, PublicMessage() from a foreign sibling — and
+// decorating such a tree must not manufacture the same borrow one level up.
+// Non-qualifying nodes are stepped over, not terminal, so a real carrier
+// behind a foreign publisher still resolves. An empty publication is treated
+// as no publication and the walk continues past it.
+func ResolvePublicMessage(err error) (string, bool) {
+	if err == nil {
+		return "", false
+	}
+	if pub, ok := publicErrorAt(err); ok && carriesClientSentinel(err) {
+		if msg := pub.PublicMessage(); msg != "" {
+			return msg, true
+		}
+	}
+	switch u := err.(type) {
+	case interface{ Unwrap() error }:
+		return ResolvePublicMessage(u.Unwrap())
+	case interface{ Unwrap() []error }:
+		for _, cause := range u.Unwrap() {
+			if msg, ok := ResolvePublicMessage(cause); ok {
+				return msg, true
+			}
+		}
+	}
+	return "", false
+}
+
+// publicErrorAt matches a single node the way errors.As matches nodes,
+// without descending: direct interface satisfaction first, then the node's
+// own As(any) bool protocol.
+func publicErrorAt(err error) (PublicError, bool) {
+	if pub, ok := err.(PublicError); ok {
+		return pub, true
+	}
+	if as, ok := err.(interface{ As(any) bool }); ok {
+		var pub PublicError
+		if as.As(&pub) && pub != nil {
+			return pub, true
+		}
+	}
+	return nil, false
+}
+
+// carriesClientSentinel reports whether err's subtree proves caller fault —
+// the same three sentinels internal/httpapi classifies status on.
+func carriesClientSentinel(err error) bool {
+	return errors.Is(err, ErrInvalidInput) ||
+		errors.Is(err, ErrNotFound) ||
+		errors.Is(err, ErrConflict)
+}
+
 // InvalidInputf builds a client error classified by ErrInvalidInput whose
 // formatted message is published verbatim. Error() renders
 // "<message>: invalid input".
@@ -60,8 +123,12 @@ func WrapPublicf(err error, format string, args ...any) error {
 		return nil
 	}
 	prefix := fmt.Sprintf(format, args...)
-	var pub PublicError
-	if !errors.As(err, &pub) {
+	// Qualification is the canonical resolution, not a bare errors.As: an
+	// input whose publication ResolvePublicMessage would reject — a foreign
+	// PublicError in a sibling branch of a bare sentinel — must degrade to
+	// the plain wrap, or the decorator would manufacture the very borrow the
+	// resolver exists to prevent (#363 review, P1).
+	if _, ok := ResolvePublicMessage(err); !ok {
 		return fmt.Errorf("%s: %w", prefix, err)
 	}
 	return &publicWrap{prefix: prefix, err: err}
@@ -76,8 +143,11 @@ func WithOperatorDetail(err error, detail error) error {
 	if err == nil {
 		return nil
 	}
-	var pub PublicError
-	if detail == nil || !errors.As(err, &pub) {
+	if detail == nil {
+		return err
+	}
+	// Same qualification rule as WrapPublicf (#363 review, P1).
+	if _, ok := ResolvePublicMessage(err); !ok {
 		return err
 	}
 	return &operatorDetail{err: err, detail: detail}
@@ -106,10 +176,14 @@ func (e *clientError) Error() string         { return e.public + ": " + e.sentin
 func (e *clientError) PublicMessage() string { return e.public }
 func (e *clientError) Unwrap() error         { return e.sentinel }
 
-// publicWrap prefixes both messages of an error already carrying a
-// PublicError; WrapPublicf guarantees the inner resolution succeeds. Being
-// outermost, it is the node errors.As finds first, so prefixes accumulate
-// outside-in exactly as they do in Error().
+// publicWrap prefixes both messages of an error whose publication
+// ResolvePublicMessage accepted; WrapPublicf guarantees that at construction.
+// Being outermost, it is the node the resolver finds first, so prefixes
+// accumulate outside-in exactly as they do in Error(). Delegation re-runs the
+// canonical resolution — never a whole-tree errors.As, which would adopt a
+// foreign sibling's text (#363 review, P1). The structure is immutable, so
+// the empty-string fallback is unreachable; if it were reached, the boundary
+// treats an empty publication as none and denies.
 type publicWrap struct {
 	prefix string
 	err    error
@@ -119,9 +193,11 @@ func (e *publicWrap) Error() string { return e.prefix + ": " + e.err.Error() }
 func (e *publicWrap) Unwrap() error { return e.err }
 
 func (e *publicWrap) PublicMessage() string {
-	var pub PublicError
-	errors.As(e.err, &pub)
-	return e.prefix + ": " + pub.PublicMessage()
+	msg, ok := ResolvePublicMessage(e.err)
+	if !ok {
+		return ""
+	}
+	return e.prefix + ": " + msg
 }
 
 // operatorDetail joins operator-only detail to a publishing error. The detail
@@ -137,7 +213,6 @@ func (e *operatorDetail) Error() string   { return e.err.Error() + ": " + e.deta
 func (e *operatorDetail) Unwrap() []error { return []error{e.err, e.detail} }
 
 func (e *operatorDetail) PublicMessage() string {
-	var pub PublicError
-	errors.As(e.err, &pub)
-	return pub.PublicMessage()
+	msg, _ := ResolvePublicMessage(e.err)
+	return msg
 }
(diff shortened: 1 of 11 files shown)


iceboundrock:
## Findings

1. **P1 - A `WrapPublicf` wrapper can still publish foreign text from a sibling branch.** `resolvePublicMessage` accepts every directly implemented `forma.PublicError` whose entire subtree contains a client sentinel (`internal/httpapi/error_response.go:276`). `WrapPublicf` creates such a `publicWrap` whenever `errors.As` finds any `PublicError` in its input (`client_error.go:63-67`), and `publicWrap.PublicMessage` then resolves the first publication from that whole input tree (`client_error.go:121-125`). Consequently, `forma.WrapPublicf(errors.Join(fmt.Errorf("bad input: %w", forma.ErrInvalidInput), foreignPublicError{manifestPath}), "operation[0]")` returns a 400 body containing `operation[0]: <manifestPath>`: the bare sentinel branch authorizes disclosure while the sibling foreign branch supplies its text. The new tests cover the unwrapped join only, so they miss this public-decorator variant. Preserve the actual publication-bearing branch when constructing/resolving wrappers, and add this regression case.

2. **P2 - The resolver drops supported `errors.As`-provided publications.** The old boundary and the public wrappers use `errors.As` to locate `forma.PublicError`, but the new walker only uses `err.(forma.PublicError)` (`internal/httpapi/error_response.go:276`). An error may expose a deliberately typed public carrier through Go `As(any) bool` without directly implementing the interface; it now receives a redacted 4xx unless another wrapper happens to materialize a direct `PublicError`. This contradicts the documented `errors.As` resolution semantics and makes the boundary inconsistent with `WrapPublicf` / `WithOperatorDetail`. Retain `errors.As` compatibility while enforcing the branch-provenance rule, and pin it with a test.

## Spec

Issue #313 requires that only deliberately authored typed public messages cross the 4xx boundary, while a client sentinel without a qualifying publication keeps its status and takes the redacted `error_class` / `error_id` shape. Finding 1 violates the core requirement: wrapping a mixed tree reconstructs the split-branch publication borrowing that this PR intends to close.

Finding 2 is a compatibility regression against the established public-carrier contract: the PR description and `docs/error-handling.md` describe publication resolution in `errors.As` terms, and the constructors continue to accept that form.

## Standards

Finding 1 violates the authorship and disclosure rules in `docs/error-handling.md:495-503` and `:715-732`: caller-actionable prefixes may be public, but operator branches never cross; publication must be deliberately authored for the caller. It also conflicts with `AGENTS.md:77-80` on carrier-based 4xx disclosure and withholding non-actionable detail.

Finding 2 conflicts with the repository rule to match error types through `errors.Is` / `errors.As` (`AGENTS.md:42`) and with the documented `errors.As` ordering contract. No additional file-size, function-size, or formatting violations found.

## Assessment

The PR correctly addresses the phase-name publication, internal schema-ID publication, and oversized test-file findings from #362. However, Finding 1 reopens the disclosure boundary through a one-level `WrapPublicf` composition, so the PR is not safe to merge until branch provenance is preserved across public wrappers. Finding 2 should be resolved in the same change because the corrected traversal needs a clear, tested `errors.As` policy.

## Observations

- `TestForeignPublicationCannotBorrowSentinelBranch` and `TestForeignNodeDoesNotBlockCarrierResolution` cover direct joins but not `publicWrap` or `operatorDetail` composition.
- `WithOperatorDetail` has the same structural risk when its publishing side is itself a mixed tree; its `PublicMessage` delegates through whole-tree `errors.As`.
- The moved federated test leaves `parquet_source_test.go` below the 500-line limit.

## Verification

- Reviewed PR #363 at `afc8e7946687a584c5b71c3d6ae213b920fbb9c0`, its four commits, issue #313, and the #362 review/follow-up record.
- `git diff --check 112dd5d...HEAD` passed.
- `make lint` passed.
- `make test` passed.
- Targeted packages `./internal/httpapi ./internal ./internal/transform ./internal/federated` passed.

## Summary

**1 P1 security/spec/standards finding and 1 P2 compatibility finding.** The P1 `WrapPublicf` composition path remains a blocker: resolve the publication from the same concrete branch that owns the client sentinel, not from a whole mixed subtree.

iceboundrock:
Round-2 response — both findings verified by reproduction and fixed in d504645.

**Finding 1 (P1) — confirmed on both decorators, fixed.** Reproduced your `WrapPublicf` construction against the branch head: `operation[0]: manifests/lead/22.json` crossed on a 400. Your Observations section also flagged `WithOperatorDetail` as structurally identical — reproduced that too; it leaked the same way. Root cause as you diagnosed: the #362 fix confined the provenance rule to the HTTP boundary while the decorators still qualified their input and delegated `PublicMessage()` via whole-tree `errors.As`, presenting the boundary with a directly-implemented carrier whose subtree legitimately held a sentinel.

The fix moves the resolution into the forma package as `ResolvePublicMessage` — one traversal used by the boundary, by both constructors for qualification (an input whose publication the resolver would reject now degrades to the plain wrap / pass-through instead of materializing a decorator), and by both decorators' `PublicMessage()` delegation. That closes the class rather than the instance: any future decorator composed from these primitives inherits the rule because there is no second resolution path to diverge. Regression coverage at both layers: forma-level (`TestDecoratorsDoNotAdoptForeignPublications`, including the positive case that a real carrier inside the same mixed tree still qualifies through the wrap) and boundary-level (`TestDecoratedForeignPublicationIsRedacted`, sub-tests for both decorators, asserting the deny shape). All red before the resolver change.

**Finding 2 (P2) — fixed in the same change.** Per-node matching now mirrors `errors.As` node semantics: direct interface satisfaction first, then the node's own `As(any) bool` protocol (`publicErrorAt`). Pinned in both directions by `TestAsProvidedPublicationResolves`: an As-provided carrier with a sentinel in its own subtree publishes; the same node without one is denied — so As-compatibility is restored without reopening the provenance rule. One side effect worth noting: the root-package test helper that previously "mirrored" the boundary with a raw `errors.As` now calls the canonical resolver, which is the structural point — there is exactly one resolution left to mirror.

`docs/error-handling.md`'s provenance paragraph now names the canonical resolver, the decorator-bypass lesson, and the As-protocol matching. Verification: `make lint` + `make test` green; both reproductions red-first against `afc8e79`.

iceboundrock:
## Round-3 review — both round-1 findings verified closed; no new findings

Reviewed PR head `d5046456` against the round-1 review point `afc8e79`, the full PR against `main` (merge base `112dd5d`), issue #313, and the #362 review record.

## Findings

**None open.** Both round-1 findings are closed by `d504645`, and I verified the claim red-first rather than taking the description at face value:

- **R1-F1 (P1, decorator borrow)** — closed. `WrapPublicf` and `WithOperatorDetail` now qualify their input via the canonical `ResolvePublicMessage` (not whole-tree `errors.As`), and their `PublicMessage()` delegation re-runs the same provenance-bound traversal. An input whose publication the resolver would reject degrades to a plain wrap / pass-through, so no decorator can manufacture the borrow one level up. I reproduced the exact round-1 construction (`forma.WrapPublicf(errors.Join(fmt.Errorf("bad input: %w", forma.ErrInvalidInput), foreignPublicError{manifestPath}), "operation[0]")`) against `afc8e79`: it emitted `operation[0]: manifests/lead/22.json` on a 400, and `WithOperatorDetail` leaked identically — matching the round-1 Observations prediction. The identical test passes at head (deny shape, both decorators). This is a genuinely red-first fix, not a paper change.
- **R1-F2 (P2, `errors.As`-provided publications dropped)** — closed. `publicErrorAt` restores the two-step node match (direct interface satisfaction, then the node's own `As(any) bool` protocol), mirroring `errors.As` node semantics. Pinned positively (`TestAsProvidedPublicationResolves` publishes an As-provided carrier with a sentinel in its own subtree) and negatively (same node without a sentinel is denied). No regression in the other direction: the resolver resolves strictly more than the round-1 walk, and qualification only tightened for text never authored for a caller.

## Spec

Issue #313 requires that only deliberately authored typed public messages cross the 4xx boundary, deny-by-default on the disclosure axis. Both round-1 findings were violations of that core requirement (a mixed tree disclosing foreign text; a supported public carrier receiving a redacted body). Both are now consistent with the issue: one canonical `forma.ResolvePublicMessage` used by the boundary, the decorators' qualification, and their delegation — so what a decorator carries is exactly what the boundary would emit. The new regression tests (`TestDecoratedForeignPublicationIsRedacted`, `TestDecoratorsDoNotAdoptForeignPublications`, `TestAsProvidedPublicationResolves`) each pin a spec-shaped property: foreign text never crosses, a legit carrier behind a foreign publisher still resolves, and As-provided publications survive. No scope creep observed: the exported `ResolvePublicMessage` is the mechanism, not an addition.

## Standards

- No hard violations. `client_error.go` is 218 lines (≤500), `error_response.go` 341, `parquet_source_test.go` back to 486 (F4), functions all far under 100 lines. Error wrapping is context-carrying throughout; the `return err` sites in `WithOperatorDetail` are documented idempotent pass-throughs of the *input*, not bare propagation of a failing call. Godoc present on the new exported resolver.
- **Observation (judgement call):** the client-sentinel list is now enumerated a **third** time — `carriesClientSentinel` (`client_error.go:87`), alongside `isClientError` (`error_response.go:241`) and `classifyManagerError` (`:193`). Three cross-package copies of the same three sentinels in security-adjacent classification is a drift hazard: a fourth client sentinel added in one place silently misses the disclosure gate. Not a violation (under the duplication threshold) but worth a single exported source of truth.
- **Observation (judgement call):** `ResolvePublicMessage` runs `carriesClientSentinel` (a full-subtree `errors.Is`) per visited node, so a chain of depth L walks in O(L²). Chains are short and this is not a documented breach; noting only.

## Assessment

The PR is safe to merge. The round-2 fix is architecturally the right shape: the provenance rule lives once in the forma package instead of being confined to the HTTP boundary, closing the decorator class rather than the instance. The `errors.As` compatibility is restored exactly as the documented contract requires. I could not construct a remaining borrow of the P1/P2 class against head.

## Observations

- Two residual disclosure surfaces remain, both **pre-existing** (present since the #362 merge) and both the **documented contract** ("a foreign `PublicError` qualifies only by doing the same — binding message and sentinel into the same branch"): (a) *ancestor shadow* — a foreign publisher that wraps a sentinel-bearing tree in its own subtree qualifies and can shadow a legit carrier beneath it; (b) *As-carrier shadow* — a node whose `As` yields foreign text qualifies if any sentinel sits anywhere in its subtree. Neither is a sibling-branch borrow; flagging for awareness, not for this PR.
- The `resolvePublicMessage` boundary wrapper is now a one-line delegate to `forma.ResolvePublicMessage`; it survives as a named seam that pins the boundary's contract in the httpapi layer. Fine.
- F3 scope note stands: `attrID=%d` publications elsewhere in transform are unchanged, consistent with the docs' message-style prescription to name the attrID (`docs/error-handling.md:1045-1046`); the withheld id here is the schema id, which the caller cannot act on. If uniform internal-id withholding is desired, those sites are the follow-up sweep.

## Verification

- Reproduced round-1 finding 1 against `afc8e79` (detached worktree): both decorators leaked `manifests/lead/22.json` on a 400; the same test passes at `d5046456` with the deny shape. Red-first claim confirmed empirically.
- `go test ./...` at head: **all packages pass**, including the new `TestDecoratorsDoNotAdoptForeignPublications`, `TestAsProvidedPublicationResolves`, `TestDecoratedForeignPublicationIsRedacted`, and the production-path `TestBatchCreateAtomicPublishesWithoutPhaseNames` (real registry + batch service).
- `golangci-lint` v1.64.8 (the pinned version, via the Makefile's install path) at head: **clean**.
- `git diff --check main...head`: clean.
- File-size limits verified: `parquet_source_test.go` 486, `parquet_source_publication_test.go` 44, `client_error.go` 218, `error_response.go` 341.

## Summary

**0 open findings.** Both round-1 findings (P1 decorator borrow, P2 `errors.As` compatibility) are closed by `d504645`, verified red-first and green against the full suite and pinned lint; two judgement-call observations (triplicated sentinel list, quadratic walk) and two pre-existing documented-contract residual shadows noted. **Approve.**

Issue #365, opened 2026-08-06
Extend the parquet scan presence guard to deleted_at once pre-#274 delta objects are retired


iceboundrock:
Follow-up to #274 (which normalizes the delta export to `COALESCE(cl.deleted_at, 0) AS deleted_at`) and to the #256 scan guard (PR #350).

## Context

#274 makes the delta exporter emit `deleted_at = 0` for live rows, so **newly written** delta objects never carry a NULL `deleted_at`. That unblocks the presence half of the scan-level system-column guard (`internal/sqlgen/duckdb_cold_scan.go`), which today type-pins `deleted_at` (`CAST AS BIGINT`) but cannot presence-guard it: pre-#274 the legit live-row encoding on the delta path *was* NULL, indistinguishable from a `union_by_name` NULL fill for an object that lost the column entirely.

Consequence of the residual: an object missing `deleted_at` still reaches the LWW merge NULL-filled and undetected at the scan — covered only by the pre-read footer probe and the #256 manifest stamp. Characterized by `sqlgen.TestParquetScanGuardTolerateNullDeletedAt`.

## Blocked on: retirement of pre-#274 delta objects

Objects written **before** #274 still encode live rows as NULL and remain readable until compaction rewrites them into base (`internal/compaction/merge_sql.go` normalizes survivors via `COALESCE(deleted_at, 0)`). Turning on the presence guard earlier hard-fails every query whose scan set touches a legacy delta object.

Gate: all pre-#274 delta objects retired for every active schema — verifiable via `--verify-stamps` or a `COUNT(*) FILTER (WHERE deleted_at IS NULL)` sweep over the delta inventory.

## Change

One line in `parquetSystemColumnGuardItem` (`internal/sqlgen/duckdb_cold_scan.go`), following the `changed_at` pattern (the only column carrying both guard shapes):

```go
// from
"CAST(deleted_at AS BIGINT) AS deleted_at)",
// to
"CAST(COALESCE(deleted_at, error('%s')) AS BIGINT) AS deleted_at)",
```

plus a new `ParquetNullDeletedAtMessage` constant alongside the row_id/changed_at messages.

## Lockstep sites

- `internal/sqlgen/duckdb_cold_scan_test.go` — `wantGuardREPLACE` (the REPLACE string is deliberately spelled out twice so a shape change is made twice, on purpose).
[...]

PR #366, opened 2026-08-06
fix(cdc): normalize delta live deleted_at to 0 and document the equal-ver_ts tie contract (#274)

### LINKED ISSUE #274: deleted_ts encoding asymmetry: base exports 0, delta exports NULL — equal-ver_ts ties resolved by a NULLS LAST accident, not an explicit rule

iceboundrock:
Closes #274.

## What

The delta export projected raw `cl.deleted_at`, so live delta rows carried `deleted_at = NULL` while base rows carried `0` (init's COALESCE). The federated dedup's equal-`ver_ts` tie was then resolved by DuckDB's NULLS-LAST default sorting `0` before `NULL` — base won by an encoding accident, not a designed rule. Per the adjudication on #274 (option A + documented contract):

1. **`internal/cdc/duckdb_exporter.go`** — the delta projection becomes `COALESCE(cl.deleted_at, 0) AS deleted_at`, matching base and what compaction's merge already writes. The `AS` alias is load-bearing (without it the parquet column is named `coalesce(cl.deleted_at, 0)`). New unit pin in `export_sql_shared_test.go` covers both export modes symmetrically.
2. **`internal/federated/merge.go`** — the Go-level LWW broke equal-`UpdatedAt` ties on `DeletedAt` *presence*, so a live cold copy (`&0`) beat a live hot copy (`nil`) before tier priority. Hardened to a value test (`!= nil && * != 0`), matching the SQL rank order. This was a pre-existing bug for base-served rows; #274 would have widened it to all delta-served rows.
3. **The documented contract** (design.md §5, new subsection): an equal-`ver_ts` live/live cold tie has an **unspecified winner** — post-#210 such ties are value-identical copies through production export paths, so the contract is row multiplicity + value, never scan order. Tombstone-wins (`deleted_ts DESC`, `T > 0` beats live `0`/`NULL`) remains the hard delete-wins rule, proven surviving by `testEqualVertsTombstoneWins`. The one window that can produce *divergent* values at equal `ver_ts` (same-millisecond write across a failed-init boundary) is fenced at promotion time — the #292 fences are re-derived as **load-bearing** for this contract, no longer defense-in-depth.

Not in scope (per adjudication): the hot leg keeps raw `cl.deleted_at AS deleted_ts` (tier priority 3 decides before `deleted_ts`; one hot row per row_id); no explicit base-vs-delta discriminator key (option B rejected).

## Behavior changes

- New delta parquet objects encode live rows as `deleted_at = 0`. Old objects keep NULL until compaction retires them; the reader filter `(deleted_ts IS NULL OR deleted_ts = 0)` is symmetric, so visibility is unaffected in the mixed-vintage window.
- Delta-served live rows now surface `record.DeletedAt = &0` instead of `nil` (parity with base-served rows, via `deleted_ts AS ltbase_deleted_at`); correspondingly `MainColumnDeletedAt` emits a 0-valued `ltbase_deleted_at` attribute for delta-served records where a schema binds it.

## Pre-existing doc bug fixed textually

`promote_fence.go` and `manifest-reconcile.md` claimed a "resurrection" mirror image: a surviving tombstone `R@T` losing the tie to a promoted live `R@T`. That was false **before** this change too — `deleted_ts DESC` always ranked a tombstone `T > 0` above a live `0`, so the delete holds. The sentence is deleted and replaced with the confined hazard (divergent live values only). No semantics touched.

## Follow-up

- #365 — extend the scan-level presence guard to `deleted_at`, gated on pre-#274 legacy delta objects being retired by compaction. `TestParquetScanGuardTolerateNullDeletedAt` stays green here and is the test that goes red when #365 lands (its comment now says so).

## Recorded risk checks

- **Read→write loop**: none found. Update paths copy `DeletedAt` from `repository.GetPersistentRecord` (OLTP Postgres read; `ltbase_deleted_at` is NULL for live rows → `nil`), never from federated read output. The federated `&0` only reaches API responses.
- **Go merge tombstone magnitudes**: after the hardening, two non-zero `DeletedAt`s fall through to tier priority instead of the SQL's `deleted_ts DESC` — a known, benign divergence from the SQL rank order, noted in #365.
- **Column type stability**: `COALESCE(BIGINT, 0)` stays BIGINT; pinned by the full-type e2e stamp assertions and the scan guard CAST.

## Verification

- `make lint` — clean (golangci-lint v1.64.8).
- `make test` — all 32 packages ok (includes the new `TestMergeLWW_ZeroDeletedAtIsLiveNotTombstone`, `TestDuckMerger_EqualVertsIdenticalZeroEncodedCopies`, and the dual-mode export SQL pin).
- `go test ./internal/federated -run TestDesignDocSQL` — the executable design.md §5 stays green after the doc edits.
- Red-first evidence: the export change was pinned red at the unit level, and `TestEntityLifecycle` failed at exactly the three delta-NULL helper sites before the e2e pins were flipped.
- `make test-e2e-production` — full tagged production harness (lifecycle, restart, concurrent-flush, tiebreak, full-type roundtrip, compaction) green.
- `go test ./internal/e2e_harness/federated/... -tags=e2e` — full federated suite green.


iceboundrock:
## Findings

- **P1 - stale values can win after concurrent same-millisecond updates.** `UpdatePersistentRecord` assigns `UpdatedAt` before it begins the transaction (`internal/postgres_persistent_repository.go:156`), so two serialized updates can retain the same timestamp while carrying different payloads. Flush A, then flush B, then run init: `RunInit` publishes base B but replaces only base manifest entries (`internal/cdc/init.go:406-422`), leaving the delta files listed. With this change all cold copies have `ver_ts = T`, tier `1`, and `deleted_ts = 0`; the final `row_id ASC` is constant inside the partition (`internal/sqlgen/advanced_query_template_duckdb.go:111`). `ROW_NUMBER()` can therefore serve stale delta A rather than current base B. Before this PR, base B's `0` ranked before delta A's `NULL`.

  The new contract limits divergent ties to failed-init boundaries (`docs/federated-query/design.md:348-360`), but this ordinary concurrent-write path is also reachable. `assertStrictlyNewer` explicitly avoids equal-timestamp divergent versions in the production harness (`internal/e2e_harness/production/filter_lww_helpers_test.go:77-80`), so current tests do not exercise it. Please retain a deterministic, semantically correct tie-break or guarantee strictly ordered write versions, and add a regression test for this sequence.

iceboundrock:
## Spec

The implementation otherwise follows the approved #274 plan: the delta alias is `COALESCE(cl.deleted_at, 0) AS deleted_at`; the raw change-log and main-column projections remain unchanged; Go LWW treats `&0` as live; production assertions pin the new encoding; legacy-NULL tolerance remains deferred to #365; and the hot leg is unchanged.

The P1 finding identifies an unstated but reachable divergent equal-timestamp path, so the documented value-identical-copy contract is not satisfied in all production paths. Resolution is required before this can be considered spec-complete.

iceboundrock:
## Standards

- **P2** `internal/cdc/duckdb_exporter.go:102` returns the planning error without context in a function modified by this PR. Wrap it with the failed operation. This violates the error-wrapping rule in `coding-standard.md` §3 and `AGENTS.md:68-71`.
- **P2** `internal/cdc/export_sql_shared_test.go:19-160` is a 142-line test function. This PR extends that immediate area without splitting it, exceeding the 100-line function limit in `coding-standard.md` §1 and `AGENTS.md:68-71`.

No other documented-standard violations found.

iceboundrock:
## Observations

- The `AS deleted_at` alias is correctly retained, preserving the Parquet schema contract.
- The Go merge change correctly aligns `DeletedAt = &0` with SQL liveness for hot-versus-cold ties.
- Legacy delta objects with NULL live encodings remain readable and are consistently documented as the #365 gate.
- The documented claim that post-#210 production export paths produce only identical equal-timestamp copies is too broad; concurrent updates need to be included in the tie analysis.

iceboundrock:
## Verification

- GitHub Actions passed: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- Locally at PR head: `GOCACHE=/var/folders/37/2bpjmt0129z96c4plfy52gkw0000gn/T/opencode/gocache-pr-366 GOFLAGS=-buildvcs=false go test ./internal/cdc ./internal/federated ./internal/compaction ./internal/reconcile ./internal/sqlgen` passed.
- `git diff --check $(git merge-base origin/main origin/pr-366) origin/pr-366` produced no whitespace errors.
- The above tests do not cover the concurrent equal-millisecond divergent-version sequence described in P1.

iceboundrock:
## Summary

Found 1 P1 correctness regression and 2 P2 standards issues. The approved #274 changes are otherwise implemented as intended. Do not merge until the cold-tier divergent equal-`ver_ts` case is made safe and covered by regression testing.

iceboundrock:
## Assessment

**Request changes.** The encoding normalization and Go-level `&0` fix are directionally correct, but P1 makes the new unspecified cold-tier tie unsafe for a normal concurrent-write sequence. That can return a stale persisted value, so it is a merge blocker.

iceboundrock:
## Review round 1 response

### P1 — divergent same-millisecond ties (fixed, root cause; commits 25ab2ce + a6e6c5f)

Confirmed, and verification showed it is **wider than reported**: the delta-vs-delta shape of the same tie predates this PR — flush A → same-ms update B → flush B yields two delta files with divergent values at equal `ver_ts`, which pre-PR NULL/NULL encoding also left unranked, and the compaction fold of that pair was already arbitrary. The pre-PR 0-vs-NULL accident only covered the post-init base-vs-delta shape.

Adjudicated resolution: **strictly ordered write versions** (the review's second arm), fixing all shapes at the source rather than adding a read-side discriminator:

- `buildUpdateMainStatement` now computes `ltbase_updated_at = GREATEST($now, ltbase_updated_at + 1)` and `RETURNING`s the effective version; `change_log` is stamped from that value in the same transaction, preserving the #210 same-source contract. Two serialized writes to one row — same millisecond, backwards clock jump, or across a failed-init snapshot boundary — can no longer share a `ver_ts`.
- Hard deletes (single + batch) capture the deleted row's version via `DELETE … RETURNING` and stamp the tombstone `GREATEST(now, prev+1)` (`tombstoneStamp`). Without this leg, a clock-ahead live version would outrank its own later tombstone — a lost delete.

Regression tests, each verified **red against the pre-fix code**:

- `TestSameMillisecondWritesStayStrictlyOrderedIntegration` (real PG, frozen clock via `withClock`): create/update/update/delete on one frozen millisecond stamp `T, T+1, T+2, T+3` in both stores — the review's serialized same-ms sequence directly.
- `TestMonotonicVersionsCloseSameMillisecondTies` (production e2e, two scenarios): forces the collision by restamping a row's stored version ahead of the wall clock (the tiebreak suite's ExecSQL restamp pattern — the harness cannot freeze the repository clock), then drives the review sequence end-to-end: API update must land at exactly `ahead+1` in both stores, and after flush → init → change_log truncation the cold read serves the fresh value with zero flapping; the delete scenario pins the tombstone landing at `ahead+1` and the row staying invisible. Pre-fix both scenarios fail with "stamped the clock read" evidence.
- `TestTombstoneStamp` + updated pgxmock suites pin the RETURNING plumbing (effective version reaches `change_log` and the caller; NotFound now derives from `ErrNoRows`).

One deliberate consequence, exercised by the e2e and documented in design.md: a clock-ahead version is not flush-eligible until the wall clock passes it (the export snapshot filters `changed_at <= snapshotTS`). Drift is bounded by writes-per-millisecond-per-row and decays as the clock catches up.

### Observations item — "the value-identical claim is too broad" (fixed in docs, commit a6e6c5f)

With write-side ordering the claim becomes a real invariant for post-fix writes. design.md's tie-contract subsection now derives it from the two mechanisms (#210 same-source stamps + #274 GREATEST ordering) and narrows the residual to **objects written before the ordering fix**; `promote_fence.go` and manifest-reconcile.md re-scope their same-millisecond scenarios to that legacy corpus (post-fix writes land at `T+1` and cannot tie), with the fences retained to keep legacy or unverifiable entries unpublished.

### P2 — bare error return (fixed, commit c04a5b1)

`buildExportSQL` now wraps: `failed to build delta export SQL plan for schema %d: %w`, matching the base twin in `init_exporter.go`.

### P2 — 142-line test function (fixed, commit c04a5b1)

`TestBuildExportSQL_CommonSemanticsAcrossModes` split into a named case type, fixture/case builders, and two assertion helpers — main function 22 lines, every helper under the limit, assertions unchanged.

### Verification at head

- `make lint`, `make test` — green (32 packages).
- `go test ./internal/federated -run TestDesignDocSQL` — executable design doc green after the contract edits.
- `TestSameMillisecondWritesStayStrictlyOrderedIntegration` — green against real PG (red pre-fix at the same-ms update assertion).
- Full `internal/e2e_harness/production` (tag e2e, includes the two new monotonic scenarios) and `internal/e2e_harness/federated/...` suites — green.

iceboundrock:
## Findings (Review Round 2)

- **P1 - future logical versions can indefinitely starve CDC flush.** The new `GREATEST(now, ltbase_updated_at + 1)` update (`internal/postgres_persistent_repository_main_table.go:133`) deliberately lets a row's `changed_at` run ahead of wall time. CDC still selects every dirty row without an eligibility condition (`internal/cdc/helpers.go:53`), but both export (`internal/cdc/duckdb_exporter.go:144`) and marking (`internal/cdc/helpers.go:109`) reject that row when `changed_at > snapshot`, where `snapshot` is the current wall clock (`internal/cdc/helpers.go:71`).

  A threshold-triggered run therefore writes and lists an empty delta object, marks zero IDs, and leaves the row dirty. For a lone row, `shouldFlush` uses `now - oldest >= MaxAgeMs` (`internal/cdc/flusher.go:343-354`); after the clock catches up it waits the default additional hour before trying again. If updates outpace one version per millisecond, the lead grows under sustained traffic and never drains. The new E2E explicitly waits for clock catch-up rather than exercising CDC behavior while ahead. Please make the flush snapshot/eligibility aware of the logical version watermark, or avoid using a future timestamp as CDC eligibility state, and add a regression test that verifies a future-stamped dirty row is exported and marked without an empty manifest entry.

iceboundrock:
## Standards (Review Round 2)

- **P2** `internal/cdc/export_sql_shared_test.go` is now 503 lines.
- **P2** `internal/postgres_persistent_repository_test.go` is now 507 lines.
- **P2** `internal/postgres_persistent_repository_repo_test.go` remains 758 lines after this PR extends it.

All three violate the 500-line source-file limit in `coding-standard.md` §1. The latter also conflicts with incremental-improvement requirements for touched noncompliant code (§95-98).

- **P3** `internal/postgres_persistent_repository.go:95`: private helper `tombstoneStamp` is noun-first; `coding-standard.md` §4 requires verb-first private helper names, e.g. `stampTombstone`.

No other current documented-standard violations found.

iceboundrock:
## Spec (Review Round 2)

The prior P1 is resolved as specified: updates return and propagate the effective `GREATEST` version, deletes generate a later tombstone, and the same-millisecond update/delete E2E coverage passes. The original #274 scope is otherwise implemented: delta canonicalizes `deleted_at` to `0`, `&0` is live in Go LWW, the hot leg remains unchanged, and legacy-NULL guarding remains deferred to #365.

However, the strict-ordering supplement is not behaviorally complete because its logical timestamps are reused as CDC snapshot eligibility. The documented claim that drift simply “decays” is false under sustained writes; P1 must be addressed before the new version contract can be accepted.

iceboundrock:
## Observations (Review Round 2)

- The `RETURNING ltbase_updated_at` plumbing correctly keeps entity-main and change-log versions aligned.
- Delete-side stamping correctly prevents a clock-ahead live copy from outranking its tombstone.
- The new parallel E2E subtests do not share mutable data: each `NewEnv` provisions its own database and S3 prefix, so they are isolated.
- The existing test intentionally waits for `time.Now()` to pass the future version. That validates the intended workaround but masks the CDC liveness defect during the waiting period.

iceboundrock:
## Assessment (Review Round 2)

**Request changes.** The previous stale-value merge blocker is fixed, but P1 transfers the failure to CDC liveness: hot data can fail to become durable and accumulate indefinitely while logical versions remain ahead of wall time. This is a merge blocker.

iceboundrock:
## Verification (Review Round 2)

- Local targeted unit tests passed: `go test ./internal ./internal/cdc ./internal/federated ./internal/compaction ./internal/reconcile ./internal/sqlgen`.
- Local production E2E passed: `go test -v ./internal/e2e_harness/production -tags=e2e -run TestMonotonicVersionsCloseSameMillisecondTies -timeout=10m`.
- `git diff --check $(git merge-base origin/main origin/pr-366) origin/pr-366` passed.
- GitHub Actions: Lint, Test, Benchmark Smoke, and k6 Smoke passed; E2E Smoke Tests were still pending at review time.
- No test covers CDC flush/mark behavior while a logical `changed_at` is ahead of the wall-clock snapshot.

iceboundrock:
## Summary (Review Round 2)

Found 1 P1 CDC liveness regression, 3 P2 file-size violations, and 1 P3 naming violation. The prior same-millisecond stale-read P1 is fixed. Do not merge until future-version rows can flush without waiting for wall-clock catch-up and the new behavior has regression coverage.

iceboundrock:
## Review round 2 response

### P1 — CDC starvation on clock-ahead versions (fixed, root cause; commit dcb5c41)

Confirmed on all four claims: the dirty listing had no eligibility condition, export and mark both rejected `changed_at > snapshot` with `snapshot` a bare wall-clock read, a threshold-triggered run shipped an empty listed delta and marked nothing, and the round-1 e2e masked the defect by waiting the drift out. The "drift decays" claim in my round-1 response was wrong under sustained faster-than-1-write/ms traffic — corrected in the docs (commit 84e2c22).

Resolution takes the review's second arm precisely: **the future timestamp is no longer CDC eligibility state.**

- `SelectBatchRowIDs` now records each listed row's `changed_at` and returns `snapshot = max(wall clock, listed versions)` — the export filter always covers what was listed, so a clock-ahead row exports in the same run.
- Marking becomes `MarkFlushedVersions`, exact against the listed `(row_id, version)` pairs (`changed_at <= v.version` per row). The wall clock is out of the marking predicate entirely. This is not just a fix but a tightening: the old global `<= snapshot` cutoff relied on "post-snapshot writes always stamp later than snapshot"; the per-row listed version relies only on per-row monotonicity, which the #274 ordering guarantees directly. A row whose version advanced past its listing was concurrently rewritten — its exported copy is superseded by a strictly greater version and the row stays dirty for the next run, the same accepted re-flush semantics as before (the "possible concurrent updates" arm, unchanged).
- One deliberate rejection: making the snapshot version-aware while **keeping** predicate marking is unsound — a concurrent write landing `<= watermark` between export and mark would be marked flushed with its payload in no delta (silent data loss). Exact pair marking is what makes the watermark snapshot safe.
- The predicate-only `MarkFlushed` helper had no callers and is deleted.
- Follow-up commit 6613ac2: `TestFlushFaultMarkFlushed` pins the failing step by error text and needed the wrap rename ("mark flushed at snapshot" → "mark flushed at listed versions") — caught by the full-suite sweep, fixed and re-run green.

Residual, documented at `shouldFlush` and in design.md: the age-based flush *trigger* still compares `now - oldest`, so a **lone** clock-ahead row age-triggers only after drift + MaxAgeMs. That delays only the trigger — any run started by the count threshold or by other rows exports and marks it normally. Production drift is bounded by writes-per-ms-per-row (ms-scale); flag if you want the trigger made version-aware too.

Regression coverage, red before the fix:

- The monotonic e2e now flushes **immediately** while the version is ahead of the wall clock and asserts `UnflushedAfter == 0` plus the row (`assertSoleLiveVersion`) / tombstone (`assertTombstoneParquet`) present in the just-written delta — no empty listed object. Pre-fix failure evidence: `flush left 1 unflushed rows`.
- `TestMarkFlushedVersions_MarksClockAheadVersions` (integration, real PG): a version one hour ahead of the wall clock marks normally. `_SkipsAdvancedRows` pins the stays-dirty arm and the re-list-then-mark convergence.
- The executor unit suite runs the new mark SQL against DuckDB, so both dialects of the `UPDATE … FROM (VALUES …)` shape are covered.

### P2 — file-size violations ×3 (fixed, commit 78a388a)

Pure moves along the seams this PR touched, assertions unchanged: `export_sql_shared_test.go` 503 → 418 + 92 (mode cases out); `postgres_persistent_repository_test.go` 507 → 328 + 192 (statement-builder + stamp tests out); `postgres_persistent_repository_repo_test.go` 758 → 372 + 406 (the update/delete/batch-delete pgxmock suites out — the pre-existing violation is now resolved, not merely not-worsened).

### P3 — noun-first helper (fixed, commit 78a388a)

`tombstoneStamp` → `computeTombstoneStamp` (definition, both call sites, test).

### Verification at head

- `make lint`, `make test` — green.
- `go test ./internal/cdc -tags=integration -run TestMarkFlushedVersions` — green against real PG (testcontainers).
- `TestMonotonicVersionsCloseSameMillisecondTies` — green with immediate flush (4.5s, no waiting).
- Full `internal/e2e_harness/production` (including the renamed fault pin) and `internal/e2e_harness/federated/...` suites (tag e2e) — green.

iceboundrock:
## Verification (Review Round 3)

- `make lint` passed at `6613ac2`.
- `make test` passed at `6613ac2`.
- `go test ./internal/cdc -tags=integration -run "TestMarkFlushedVersions_(SkipsAdvancedRows|MarksClockAheadVersions)" -v -count=1` passed.
- A temporary real-Postgres regression probe failed as predicted: after changing a listed row from `V` to `V-1`, `MarkFlushedVersions` returned that row as flushed. The probe was removed after execution.
- `git diff --check c855832...origin/pr-366` failed only on the trailing blank line noted above.
- GitHub Actions Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke are green.

iceboundrock:
## Standards (Review Round 3)

- **P2** `internal/cdc/flush_batch_test.go` is now 538 lines. It was already over the 500-line limit and this PR extends it, so it violates `coding-standard.md` §1 and the incremental-improvement rule in §95-98. Split the changed test area.
- **P2** The modified repository paths still directly propagate errors without operation context: `internal/postgres_persistent_repository.go:125-149,164-186,198-230`; `internal/postgres_persistent_repository_batch.go:19-20,61-62,115-116`; and `internal/postgres_persistent_repository_main_table.go:143-159,182-184,198-200`. `internal/cdc/flush_batch.go:60-62` likewise returns a child error bare. Wrap at these changed call boundaries, per `coding-standard.md` §3 and `AGENTS.md:68-71`.

No function exceeds the 100-line hard limit, and private-helper naming now conforms.

iceboundrock:
## Findings (Review Round 3)

- **P1 - delete/recreate can regress a row version and clear the dirty barrier for data that was never exported.** `InsertPersistentRecord` and `BatchInsertPersistentRecords` stamp a reused `row_id` from raw wall time (`internal/postgres_persistent_repository.go:129-149`, `internal/postgres_persistent_repository_batch.go:23-47`), while the new delete path can have published or selected a clock-ahead tombstone at `V+1` (`internal/postgres_persistent_repository.go:208-230`). A later recreate at `W < V+1` overwrites the same slot-0 change-log row with `W`. `MarkFlushedVersions` then accepts `changed_at <= listed_version` rather than the documented exact pair (`internal/cdc/helpers.go:88-120`).

  Concrete failure: select and export the tombstone at `V+1`; recreate the same ID after export but before marking; the recreated live row at `W` was not in that Parquet file, yet `<= V+1` marks it flushed. The dirty anti-join no longer serves hot data and the cold `V+1` tombstone wins, hiding the recreate. Even when later re-exported, a recreated `W` cannot outrank an already-flushed `V+1` tombstone. The prior test covers only an advanced replacement (`internal/cdc/flusher_snapshot_test.go:15-61`), not this legal regressed replacement.

  Please make create/recreate obtain a version strictly above retained row history, make the mark predicate equality against the listed version, and add an end-to-end regression covering recreate between export and mark.

iceboundrock:
## Observations (Review Round 3)

- The delta alias remains load-bearing and is correctly retained.
- The `&0` Go LWW change correctly makes cold live rows defer to hot-tier priority on equal versions.
- The watermark snapshot fix correctly makes listed clock-ahead versions eligible for export; the existing advanced-version integration tests cover that arm.
- The newly documented claim that marking is exact against listed pairs does not match the `<=` SQL predicate.
- `git diff --check main...PR` reports a new blank line at EOF in `internal/cdc/export_sql_mode_cases_test.go:92`.

iceboundrock:
## Spec (Review Round 3)

The approved #274 scope is otherwise met: delta exports `COALESCE(cl.deleted_at, 0) AS deleted_at`; Go LWW treats `&0` as live; the hot projection remains raw; tombstone-wins remains in place; legacy `deleted_at` presence validation is deferred to #365; and the contract/documentation and parity tests were updated.

The prior review supplements are not spec-complete. The PR documents strict per-row ordering and exact listed `(row_id, version)` marking (`docs/federated-query/design.md:348-377`), but recreated rows can regress below their tombstone and the actual predicate is `<=`, not equality. The P1 sequence can return a stale deletion despite a later successful recreate, so resolution is required before the documented post-#274 value/ordering contract is valid.

iceboundrock:
## Assessment (Review Round 3)

**Request changes.** The P1 delete/recreate sequence causes a later live version to lose to, or be hidden by, its earlier tombstone. This violates persisted-data correctness and is a merge blocker.

iceboundrock:
## Summary (Review Round 3)

Found 1 P1 persisted-data correctness defect and 2 P2 standards issues. The original normalization and the prior clock-ahead CDC-liveness fix are otherwise implemented as intended. Do not merge until recreate versioning and exact listed-version marking are corrected and covered.

iceboundrock:
## Review round 3 response

### P1 — delete/recreate version regression + inexact mark (fixed, both legs; commit d0e7269)

Confirmed on both legs, including your real-PG probe result. One note on reachability before the fixes: the EntityManager surface always mints a fresh UUIDv7 on create, so row_id reuse is not reachable through the HTTP API today — but it is legal at the repository contract (`InsertPersistentRecord` takes the caller's RowID), which is exactly the seam restore/import-style callers use, so the defect stands as filed.

**Create leg** — `nextRowVersion`: a (re)created row now stamps strictly above `MAX(changed_at)` over **all** change_log slots for the row (flushed tombstones included; change_log entries are never garbage-collected, so that MAX is the row's full retained history). `CreatedAt` stays the identity clock read; `change_log` receives the effective version, preserving the #210 same-source contract; a fresh row_id sees no history and stamps the plain clock read — pre-fix behavior unchanged for ordinary creates. All three writers (update `GREATEST`, delete `computeTombstoneStamp`, create `nextRowVersion`) now preserve strict per-row ordering; design.md's contract names all three legs.

**Mark leg** — exact equality: `MarkFlushedVersions` now matches `cl.changed_at = v.version`, as the documentation claimed. The `<=` was my error — I reasoned "versions only advance" while the delete/recreate slot overwrite was precisely the case where they don't. Equality also stays correct going forward: with the create leg in place a slot can never legally revisit a listed version, so "differs in either direction ⇒ concurrently rewritten ⇒ stays dirty" is the exact invariant.

Regression coverage, each red before the fix:

- `TestRecreateOutranksClockAheadTombstoneIntegration` (real PG, frozen clock): create/delete/recreate on one frozen millisecond stamp `T, T+1, T+2` in both stores, `CreatedAt` pinned at `T`.
- `TestMarkFlushedVersions_SkipsRegressedRows` codifies your probe: a slot regressed below its listing is not marked and stays dirty.
- `TestRecreateBetweenExportAndMark` (production e2e): the exact race, end-to-end — clock-ahead tombstone listed and exported at `ahead+1`, flush paused at its CopyObject (`runPausedFlush`), recreate lands at `ahead+2` through the repository contract while the flush hangs (mirrored into the oracle log per the tiebreak suite's restamp pattern), the mark leaves the slot dirty (`UnflushedAfter == 1`), the recreate stays hot-visible, a clean retry drains it into its own delta, and cold-only reads serve the recreate over its tombstone with zero flapping. Pre-fix failure evidence: `recreate stamped <wall clock>, want ahead+2`.

### P2 — flush_batch_test.go 538 lines (fixed, commit 1726373)

Split along the touched seam: the mark-flushed and before-export-hook suites move to `flush_batch_mark_test.go` (324 + 234) — the pre-existing violation is resolved.

### P2 — bare error propagation (fixed, commit 1726373)

Every flagged boundary now wraps with its operation: insert/update/delete and batch write paths (validate-tables, main-row, EAV, change-log steps), the update statement builder's per-column-family assignment collection, and `executeBatch`'s export step. Carrier-bearing returns (NotFound/Conflict) use plain `fmt.Errorf` wraps, which `ResolvePublicMessage` traverses (#296 precedent); `TestNoBareSentinelWraps` stays green.

### Observations

- "Marking exact vs `<=` mismatch" — resolved by the mark leg above; the docs' claim is now literally true.
- Blank line at EOF in `export_sql_mode_cases_test.go` — removed; `git diff --check` over the full range is clean.

### Verification at head

- `make lint`, `make test` — green.
- `go test ./internal/cdc -tags=integration -run TestMarkFlushedVersions` (3 tests incl. the new regressed-rows case) and the frozen-clock integration suite — green against real PG.
- `git diff --check c855832...HEAD` — clean.
- Full `internal/e2e_harness/production` (including `TestRecreateBetweenExportAndMark`) and `internal/e2e_harness/federated/...` suites (tag e2e) — green.

iceboundrock:
## Observations

- The delta projection retains the required `AS deleted_at` alias, so the Parquet system-column name remains stable.
- The hot leg remains raw as the approved plan requires.
- The new focused tests cover serialized same-millisecond update/delete, normalized export SQL, zero-as-live Go merge behavior, and the executable federated design SQL. They do not cover concurrent version allocation across delete and recreate, which is the remaining P1 path.

iceboundrock:
## Verification

Independently run at PR HEAD (`1726373`):

```text
GOCACHE=/var/folders/.../gocache-pr-366 GOFLAGS=-buildvcs=false go test ./internal -run 'Test(SameMillisecondWritesStayStrictlyOrderedIntegration|RecreateOutranksClockAheadTombstoneIntegration)$' -v
GOCACHE=/var/folders/.../gocache-pr-366 GOFLAGS=-buildvcs=false go test ./internal/cdc -run 'Test(BuildExportSQL_CommonSemanticsAcrossModes|MarkFlushedVersions|SelectBatchRowIDs)' -v
GOCACHE=/var/folders/.../gocache-pr-366 GOFLAGS=-buildvcs=false go test ./internal/federated -run 'Test(MergeLWW_ZeroDeletedAtIsLiveNotTombstone|DuckMerger_EqualVertsIdenticalZeroEncodedCopies|DesignDocSQL)' -v
git diff --check main...FETCH_HEAD
```

All executed tests passed and `git diff --check` was clean. The full production/federated E2E suites and `make lint`/`make test` were not independently rerun; the PR author reports them green.

iceboundrock:
## Standards

**Pass.** Reviewed the changed Go code against `AGENTS.md` and `coding-standard.md`: no documented hard-standard violation found. The PR keeps touched functions and files within the stated limits, uses contextual error wrapping at changed boundaries, and uses verb-first private helper naming.

iceboundrock:
## Assessment

**Request changes before merge.** The delta `deleted_at = 0` normalization and `DeletedAt == 0` live handling correctly align exported parquet semantics with the SQL rank. However, the P1 concurrent delete/recreate race breaks the PR's central guarantee that all writers allocate strictly ordered per-row versions, allowing a recreation to remain hidden after a normal CDC cycle.

The larger writer and CDC changes should also receive explicit scope approval as noted in the Spec comment.

iceboundrock:
## Summary

One P1 correctness finding: unprotected create-version allocation can tie a concurrent delete tombstone and permanently hide a recreated row in cold-tier reads. Core normalization work is implemented and standards review passes; merge should wait for serialized per-row version allocation plus a concurrent regression test. The PR also needs explicit approval for its broadened write/CDC scope.

iceboundrock:
## Spec

**Status: partial mismatch.** The approved #274 plan calls for Option A normalization, the equal-`ver_ts` contract, preserving the raw hot leg, and deferring the `deleted_at` presence guard. Those requested changes are present.

Later commits also introduce per-row version sequencing, clock-ahead CDC snapshot/mark semantics, and exact-version marking. These are material write-path and flush-contract changes beyond the approved one-line export change and its documented exception window; the issue plan instead describes the failed-init/promotion same-millisecond window as fenced at promotion. The additions may be justified by review discoveries, but they need an explicit issue/spec amendment or should be split into a follow-up PR so their operational and compatibility consequences receive independent approval.

iceboundrock:
## Findings

### P1: Serialize recreate version allocation with the delete path

`nextRowVersion` in `internal/postgres_persistent_repository.go:104` reads `MAX(changed_at)` without locking it. A recreate can read version `V` while a concurrent delete commits a tombstone at `V+1`; if the recreate then waits for that delete's main-row lock and proceeds, it inserts/upserts the live slot using its already-computed `V+1`. If CDC flushes the committed tombstone before that upsert, the later live delta has the same `ver_ts`, and DuckDB ranks the tombstone above live (`deleted_ts DESC`). The recreated row is then hidden in cold reads permanently. `BatchInsertPersistentRecords` has the same allocation path.

This violates the PR's stated strict per-row version invariant. Serialize version allocation across create, update, and delete for a `(schema_id,row_id)` within the transaction (for example, a transaction-scoped per-row advisory lock), then read history and allocate the stamp. Add a real-Postgres concurrent delete/recreate regression that flushes the tombstone between the recreate's history read and change-log upsert, and proves cold-only reads keep the recreation visible.

iceboundrock:
## Review round 4 response

### P1 — concurrent delete/recreate version allocation (fixed as prescribed; commit 4ec3520)

Confirmed. `nextRowVersion`'s history read had no lock coupling it to a concurrent delete — the two writers share no table row (delete locks the existing main row, create inserts a new one), so the recreate could read `V`, the delete commit a tombstone at `V+1`, and the recreate allocate its own `V+1`: an equal-`ver_ts` live/tombstone pair that `deleted_ts DESC` resolves tombstone-wins, permanently hiding the recreate once the tombstone flushed inside the window.

Fix follows the prescription: create and delete (single and batch) take a **transaction-scoped per-row advisory lock** before allocating. Details worth noting:

- Key: single-bigint `pg_advisory_xact_lock` over FNV-1a(schema_id, row_id). The single-bigint keyspace is disjoint from the `(int4, int4)` form the per-schema CDC lock uses, so cross-lock collision is impossible by construction; an FNV collision between two rows only serializes unrelated writes.
- Updates stay outside the advisory lock: their version is computed inside the row `UPDATE` under the row lock, which every competing delete also takes — update↔delete already serialize, and update↔create cannot interleave (the row either exists, failing the create's PK, or doesn't, failing the update). The advisory lock closes exactly the create↔delete pair that shares no row.
- Batch writers acquire the locks in input order — the same discipline as their existing row locks; an order inversion between two batches is detected and errored by PostgreSQL's deadlock checker like any row-lock inversion.

Regression: `TestConcurrentDeleteRecreateSerializesIntegration` (real PG, frozen clock), red before the fix ("recreate completed while the row version lock was held"), pins both halves:

- **Deterministic window** — a separate session holds the same lock (session- and xact-level advisory locks share one lock table), the recreate provably blocks, the tombstone is marked flushed **inside** the held window (the exact race you specified: flush between the recreate's history read and its change-log upsert), and the released recreate still lands strictly above the flushed tombstone, live slot verified in change_log.
- **True-concurrency sweep** — ten concurrent delete/recreate pairs under the frozen clock all come out `T, T+1, T+2`; never a tie, never a regression. (The unflushed-tombstone slot being overwritten by the recreate is correct behavior — it never reached parquet — so the pinned observable is the final live version.)

Cold-read visibility for the recreate remains pinned end-to-end by `TestRecreateBetweenExportAndMark` from round 3, which drives the flushed-tombstone-vs-recreate rank through DuckDB.

### Spec — scope amendment (adjudicated: amend in place, single PR)

Posted to #274 as the authoritative scope record: https://github.com/Lychee-Technology/forma/issues/274#issuecomment-5201437679. It consolidates the three review-driven supplements (strictly ordered per-row versions across all three writers incl. this round's allocation serialization; version-aware CDC eligibility + exact marking; Go-merge liveness parity), why each proved inseparable from the approved Option A contract, their operational consequences (clock-ahead drift bounds, advisory-lock round trip per create/delete, batch lock ordering), and the compatibility statement (no storage/wire contract change beyond the approved COALESCE; #365 gating unchanged; #292 fences keep guarding pre-fix legacy copies).

### Observations item — concurrent allocation not covered by prior tests

Resolved by the new integration test above; the prior frozen-clock tests covered serialized same-session sequences only, as you noted.

### Verification at head

- `make lint`, `make test` — green.
- Frozen-clock integration suite (now incl. the serialization test) and `internal/cdc` integration tags — green against real PG.
- Full `internal/e2e_harness/production` and `internal/e2e_harness/federated/...` suites (tag e2e) — green.

iceboundrock:
## Findings (Review Round 5)

Independent re-review of PR head `4ec3520` (full 15-commit diff against `c855832`). The four round-1–4 P1s are all resolved as claimed, with red-first regression coverage in place. No new correctness findings this round.

**P2 — `internal/postgres_persistent_repository_integration_test.go` exceeds the 500-line limit.** The file is **539 lines** at PR head (base was 287; this PR adds 252). This violates `coding-standard.md` §1 ("REJECT any changes that result in source code files >500 lines"). The PR correctly split the other oversized test files (`flush_batch_mark_test.go`, `export_sql_mode_cases_test.go`, `write_mock_test.go`, `statement_test.go`) to stay under the limit but let this one cross it. The three new integration tests added here are the natural material for a split (`same_millisecond_integration_test.go` + `recreate_integration_test.go`).

**P2 — `TestConcurrentDeleteRecreateSerializesIntegration` exceeds the 100-line function limit.** It spans `internal/postgres_persistent_repository_integration_test.go:429-539` = **111 lines** (`coding-standard.md` §1). Splitting the test file above resolves this at the same time.

**P3 — bare `return err` on the single-row lock paths.** `internal/postgres_persistent_repository.go:200` (`InsertPersistentRecord`) and `:281` (`DeletePersistentRecord`) return `lockRowVersion` errors bare, while the batch paths wrap the identical call (`postgres_persistent_repository_batch.go:38,150`: `fmt.Errorf("lock row version for record[%d]: %w", ...)`). The callee already adds the `row_id` via `%w` so no context is genuinely lost, but the single-row paths now match the documented DON'T pattern exactly and are internally inconsistent with the batch paths. Wrap for consistency (`coding-standard.md` §3).

**P3 — stale comment in `internal/e2e_harness/production/pausing.go:20`.** The rename to `MarkFlushedVersions` was applied in the text, but the clause "the window the `changed_at <= snapshot` guard protects" still describes the **old** `<=` predicate. The guard (`MarkFlushedVersions`) is now **exact equality** against the listed version; the comment is factually wrong and should say "the window the exact listed-version mark guard protects".

**P3 — stale doc in `internal/compaction/merge_sql.go:46`.** `buildMergeSQL`'s comment still says survivors are normalized "so the merged file keeps base-tier tie semantics". The adjacent `mergeLWWOrderBy` comment was correctly updated for #274 (unspecified winner, 0-canonical), but this one was missed — a reader looking at `buildMergeSQL` gets the retired #183 base-wins framing.

All three P2/P3 items are mechanical (file/function split + two comment edits) and none touch behavior.

iceboundrock:
## Spec (Review Round 5)

Checked against the authoritative scope record on #274 (approved Option A plan comment + the consolidated "Spec amendment: strictly ordered per-row versions and version-aware CDC" comment) and the PR body's explicit non-scope statements.

**(a) Missing / partial requirements — none substantive.**

- Option A is implemented with the load-bearing alias: `COALESCE(cl.deleted_at, 0) AS deleted_at` (`internal/cdc/duckdb_exporter.go:97`), pinned in both export modes by `export_sql_mode_cases_test.go` / `assertExportModeInvariants`.
- Tombstone-wins (`deleted_ts DESC`, `T > 0` beats live `0`/`NULL`) is intact and re-pinned (`testEqualVertsTombstoneWins`, `TestDuckMerger_EqualVertsIdenticalZeroEncodedCopies`).
- Amendment 1 (GREATEST+RETURNING update; DELETE…RETURNING + `computeTombstoneStamp`; `nextRowVersion` over all retained slots; per-row `pg_advisory_xact_lock` for create/delete incl. batch paths) matches the amended spec exactly. Lock keyspace verified disjoint from the `(int4,int4)` CDC lock (`schema_lock.go`).
- Amendment 2 (`SelectBatchRowIDs` snapshot = `max(wall clock, listed versions)`; `MarkFlushedVersions` exact-equality per listed `(row_id, version)`, row stays dirty if the slot moved in either direction) matches the spec, and both directions are pinned (`SkipsAdvancedRows`, `SkipsRegressedRows`, `MarksClockAheadVersions`).
- Amendment 3 (`chooseLWW` treats `DeletedAt == &0` as live) matches, and is verified to be a genuine pre-existing bug the COALESCE would have widened.
- Non-scope respected: hot leg keeps raw `cl.deleted_at AS deleted_ts`; no base-vs-delta discriminator key; `deleted_at` presence guard stays deferred to #365; the false "resurrection mirror image" claim is removed from `promote_fence.go` and `manifest-reconcile.md` (remaining `resurrect` references describe `verifyNoResurrection`, a different and real guard).

**(b) Scope creep — none functional.** The broad error-wrapping across the repository/CDC paths is AGENTS.md-mandated hygiene; the test-file splits are structural.

**(c) Implemented but slightly wrong / overstated — one cosmetic, one honest-but-nuanced.**

- **P3/observation — Go-vs-SQL ranking-term order.** `chooseLWW` (merge.go:120-127) evaluates the tombstone arm **before** tier priority, while the SQL rank is `ver_ts DESC, source_tier_priority DESC, deleted_ts DESC, row_id ASC` (`advanced_query_template_duckdb.go:111`) — tier first. The comment claims the change is "matching the SQL rank order"; the tombstone-vs-live *semantics* match, but the *evaluation order* does not. Post-#274 unreachable (tombstones stamp strictly above live versions, so a live-vs-tombstone equal-`UpdatedAt` pair cannot be minted), so this is benign — and the PR body's "Recorded risk checks" already discloses the two-nonzero-tombstone divergence for #365. Worth a one-line comment nuance, not a behavior change.
- **Observation — production-only normalization.** The "both tiers encode live rows as 0" claim is production-exporter-only: the benchmark delta shape (`duckdb_benchmark_projection.go:222`) still emits raw `cl.deleted_at AS deleted_ts`, so benchmark fixtures keep exercising the legacy NULL encoding. Benign (synthetic, and #365-tolerant), but the docs' blanket phrasing overstates slightly.

iceboundrock:
## Standards (Review Round 5)

Reviewed the changed Go code against `AGENTS.md` and `coding-standard.md`.

**Hard violations:**

1. **`internal/postgres_persistent_repository_integration_test.go` is 539 lines** — over the 500-line source limit (`coding-standard.md` §1; base was 287, PR adds 252). This is a REJECT-condition violation. The PR applied the split discipline to four other test files but missed this one.
2. **`TestConcurrentDeleteRecreateSerializesIntegration` is 111 lines** — over the 100-line function limit (`coding-standard.md` §1).

**Judgement calls:**

3. **Bare `return err`** at `internal/postgres_persistent_repository.go:200` and `:281` (single-row `lockRowVersion`), while the batch paths wrap the identical call. Callee already adds `row_id` context, so no information is lost — but the single-row paths violate the letter of the error-wrapping rule (§3) and are internally inconsistent with the batch paths.
4. **Stale comments** in `pausing.go:20` (still describes the retired `changed_at <= snapshot` predicate; the mark guard is now exact equality) and `merge_sql.go:46` (still claims merged survivors "keep base-tier tie semantics", the retired #183 framing). Both were touched by this PR and should have been updated with their neighbors.
5. **Naming (§4):** `rowVersionLockKey` and `exportSQLAttrCacheFixture` are not verb-first, but read naturally as getters/fixtures; defensible, not flagged as violations.

**Verified clean:**

- All changed `.go` files other than the one above are under 500 lines (checked every changed file).
- `validateWriteTables` and the other touched boundaries now wrap errors with operation context (the round-3/4 P2s are fixed).
- New helpers (`lockRowVersion`, `nextRowVersion`, `computeTombstoneStamp`, `restampMainVersionAhead`) are verb-first and documented.
- Doc/comment updates in `design.md`, `manifest-reconcile.md`, `duckdb_exporter.go`, `merge_sql.go`, `promote_fence.go`, `uncovered_rows.go`, and the scan-guard files are internally consistent with the #274/#365 encoding change.

Fix items 1–2 by splitting the integration test file; fix 3–4 inline. All are mechanical, non-behavioral.

iceboundrock:
## Observations (Review Round 5)

- The `AS deleted_at` alias is load-bearing and correctly retained; the parquet system-column name stays stable and the reader filter `(deleted_ts IS NULL OR deleted_ts = 0)` is symmetric across the mixed-vintage window.
- The `chooseLWW` value test (`!= nil && * != 0`) aligns Go liveness with the SQL `deleted_ts DESC` tombstone rule, and `TestMergeLWW_ZeroDeletedAtIsLiveNotTombstone` pins both argument orders.
- The clock-ahead machinery is coherent end-to-end: `SelectBatchRowIDs` watermarks the snapshot at `max(clock, listed versions)`, export uses that snapshot, and `MarkFlushedVersions` marks by exact listed `(row_id, version)` — a slot that advanced (update/delete) or regressed (recreate-overwrites-slot) stays dirty. The three `TestMarkFlushedVersions_*` arms plus `TestMonotonicVersionsCloseSameMillisecondTies` and `TestRecreateBetweenExportAndMark` cover both starvation and lost-recreate directions.
- The `pg_advisory_xact_lock` key design is sound: single-bigint keyspace is disjoint from the `(int4,int4)` per-schema CDC lock by construction, and an FNV collision between two rows only serializes unrelated writes. Batch writers acquiring in input order mirrors their existing row-lock deadlock discipline.
- `CreatedAt` staying the identity clock read while `UpdatedAt` runs ahead of it (recreate scenario) is documented and intentional; the READ/write loop risk was checked — `DeletedAt=&0` only surfaces on the federated read path, never copied back into a write.
- The spec amendment consolidation on #274 is a clean authoritative scope record; the round-4 response correctly re-derives the #292 fences as load-bearing rather than defense-in-depth.
- The remaining nits (Go-vs-SQL term-order comment nuance, production-only normalization phrasing in the docs) are documentation-scale, not design-scale.

iceboundrock:
## Verification (Review Round 5)

Independently run at PR head `4ec3520` (fresh worktree at the merge-base `c855832`):

- **Full unit suite:** `go test ./...` — all packages pass.
- **New #274 unit pins:** `TestMergeLWW_ZeroDeletedAtIsLiveNotTombstone`, `TestDuckMerger_EqualVertsIdenticalZeroEncodedCopies`, `TestDuckMerger_MergeToTmp`, `TestDesignDocSQL_*` (4 subtests) — pass.
- **CDC export/mark coverage:** `TestBuildExportSQL_CommonSemanticsAcrossModes` (both modes, incl. the new `expectedDeletedAtExpr` for COALESCE-with-alias) and the `TestExecuteBatch_*` / `TestMarkFlushedVersions_*` arms — 13 subtests pass.
- **Real-Postgres integration anchors:** `TestSameMillisecondWritesStayStrictlyOrderedIntegration`, `TestRecreateOutranksClockAheadTombstoneIntegration`, `TestConcurrentDeleteRecreateSerializesIntegration` (the round-4 serialization regression, incl. its true-concurrency sweep) — all pass against live PG.
- **Build:** `go build ./...` clean.
- **Whitespace:** `git diff --check c855832...pr-366` clean.
- Lint: `golangci-lint` not installed locally in this environment, so I could not re-run it; the round-4 response reports `make lint` green at head and the CI lint job passed on the latest commits.

The E2E production/federated suites are Docker-gated and were not re-run in this environment; the round-4 response reports them green, and the unit/integration layer independently corroborates the same invariants.

iceboundrock:
## Assessment (Review Round 5)

**Approve, conditional on the two mechanical standards fixes.**

The four round-1–4 P1s (same-millisecond stale copies, clock-ahead CDC starvation, delete/recreate version regression, concurrent delete/recreate allocation) are all resolved with red-first regression coverage, and my independent verification — full unit suite, the real-PG integration anchors, and the new #274 pins — confirms the write-side strictly-ordered-version invariant, version-aware CDC, and the unspecified-winner live/live contract are implemented and tested as specified.

No correctness findings remain. The two hard standards violations (`integration_test.go` at 539 lines; the 111-line `TestConcurrentDeleteRecreateSerializesIntegration`) are REJECT-condition per `coding-standard.md` §1, but both are fixed by a single mechanical test-file split, and the P3 items are comment/consistency edits. None touch production behavior.

Recommended sequence: split the integration test file (resolves both hard violations), wrap the two single-row lock errors, update the two stale comments (`pausing.go`, `merge_sql.go`), then merge. The #365 follow-up and the Go-vs-SQL term-order comment nuance can ride along with #365 rather than block this PR.

iceboundrock:
## Summary (Review Round 5)

No new correctness findings. All four prior P1s remain fixed and now carry regression coverage at every layer (unit, real-PG integration, production e2e).

Remaining: **2 P2 standards violations** (539-line `postgres_persistent_repository_integration_test.go`; 111-line `TestConcurrentDeleteRecreateSerializesIntegration`) and **3 P3 items** (two bare single-row `return err`, two stale comments in `pausing.go` / `merge_sql.go`). The delta `deleted_at = 0` normalization, the `&0`-is-live Go merge, version-aware CDC, and the documented unspecified-winner tie contract are implemented as adjudicated and independently verified. Merge after the mechanical test-file split.

iceboundrock:
## Review round 5 response

All five items fixed in commit 46cc32d; no behavior change.

### P2 — 539-line integration file + 111-line test function (fixed)

The three #274 version-contract anchors move to `postgres_persistent_repository_version_integration_test.go` (288 + 281 lines), and the true-concurrency sweep is extracted into `assertConcurrentDeleteRecreatePairsStayOrdered`, bringing `TestConcurrentDeleteRecreateSerializesIntegration` to 79 lines. Both hard violations resolved by the same split, as you suggested.

### P3 — bare single-row lock returns (fixed)

`InsertPersistentRecord` and `DeletePersistentRecord` now wrap `lockRowVersion` with their operation ("lock row version for insert/delete"), consistent with the batch paths.

### P3 — stale comments (fixed)

- `pausing.go` now describes the window as "the exact listed-version mark guard protects (#274)" — the `<=` predicate wording is gone.
- `buildMergeSQL`'s doc drops the retired base-wins framing: survivors normalize to 0 as "the canonical live encoding on every cold tier (#274)", consistent with its `mergeLWWOrderBy` neighbor.

### Ride-alongs from your Spec/Observations notes (also in 46cc32d)

Both were one-line comment/doc edits, so they ship now rather than riding with #365:

- `chooseLWW` notes the Go-vs-SQL nuance explicitly: tombstone-vs-live *semantics* match the SQL rank, the *evaluation order* differs (tombstone arm before tier priority), and the divergence is unreachable post-#274 since tombstones stamp strictly above live versions — noted for #365, where the tombstone-magnitude divergence already lives.
- design.md's live-encoding sentence now names the benchmark harness shape as the deliberate raw-encoding holdout, so "both tiers encode 0" no longer overstates.

### Verification at head

`make lint`, `make test` (32 packages), the full real-PG integration suite (9 tests incl. the split anchors), and `internal/federated` + `internal/compaction` — green. No production code path changed; the e2e suites were green at `4ec3520` and nothing behavioral moved since.

Per repo rules I won't merge; ready when you are.

Issue #372, opened 2026-08-06
Legacy scalar EAV rows under list metadata export/pivot as empty list


iceboundrock:
Found during #315 (task 3).

An EAV row with `array_indices = ''` for an attribute whose metadata now says `valueType: list` silently exports/pivots as `[]` — value loss with no error.

Mechanism: the list aggregate (`internal/cdc/export_sql_builder.go:143-159`, and the identical hot-tier pivot at `internal/sqlgen/duckdb_schema_projection.go:246-248`) filters elements with `array_indices <> ''` while `count(*)` still counts the row, so `coalesce(list(...) FILTER (...), [])` fires and the stored scalar is dropped.

Measured:

```
=== RUN   TestTW315_ListAggregateOverLegacyScalarRow
    legacy scalar row (array_indices='') through the list aggregate -> []interface {}{}
    legacy per-element rows through the list aggregate -> []interface {}{"alpha","beta"}
--- PASS
```

Such a row is only possible for data written before #314's array handling (the write path now rejects a scalar body where the schema declares an array). OLTP/EAV reads still reconstruct it correctly as a scalar; only the parquet export and the DuckDB pivot lose it.

Decide whether to detect and warn/reject (e.g. in `validate-schema-consistency` or at export time) rather than silently emitting `[]`.

Refs #315.

iceboundrock:
**Implementation plan** (working copy lives only in the session scratchpad; this comment is the record).

## Ruling

Both halves of the "detect or reject" question, resolved as follows:

1. **Detect in `validate-schema-consistency`, as a failure.** A row with `array_indices = ''` that carries a value under an attribute whose metadata now says `valueType: list` is the same class of finding as "numeric attribute stored in `value_text`": stored shape contradicts current metadata, the operator rewrites the rows. New check, `severityError`, category `scalar rows stored under list attributes in <eav_table>`, one line per `(schema_id, attr_id)` with the attribute name and row count. The empty-list marker row (`array_indices = ''`, both value columns NULL, #204) is excluded by the query itself, so it never trips the check.

[...]

Issue #371, opened 2026-08-06
cdc-init only replaces the base tier; delta files survive re-init


iceboundrock:
Found during #315's compatibility investigation (task 3).

`internal/cdc/init.go:421` calls `manifest.ReplaceTierFiles(..., "base", ...)` only — `cdc-init` replaces the **base** tier and leaves `/delta/` untouched.

Consequence: an operator re-initializing after a metadata change (e.g. #315's `text` -> `list` valueType flip) must purge `/delta/` manually, or stale-typed delta parquet keeps poisoning federated reads. In #315's case a single surviving pre-flip delta object with a non-NULL scalar value in an affected column is enough to make every federated read of that schema fail with a DuckDB `Conversion Error`, because `read_parquet(..., union_by_name=true)` and the `s3_source`/`pg_source` `UNION ALL` both force the old `VARCHAR` and the new `VARCHAR[]` together.

Decide whether `cdc-init` should also replace/purge the delta tier (or refuse to run while delta objects exist), and document the manual step until then.

Refs #315.

iceboundrock:
Operational cross-reference: the full migration runbook that depends on this issue's manual-delta-purge caveat lives in PR #373's body (section "Migration runbook (required before deploy)", steps 1–5). Anyone re-initializing tiers after the #315 text→list metadata flip must purge `/delta/` manually — `cdc-init` replaces `/base/` only (`internal/cdc/init.go:421`) — and verify legacy `_full` rows against the new required policies (runbook step 4) before deploy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

iceboundrock:
Scope update, re-verified at `dd58a69`. The surviving-delta hazard this issue describes (a stale-typed delta object failing every federated read after a valueType flip) has a quieter sibling worth folding into the same fix, because both come from the reader trusting `union_by_name` for attribute columns.

The reader re-pins **system** columns against exactly this class of drift: `parquetSystemColumnGuardItem` (`internal/sqlgen/duckdb_cold_scan.go`) emits a `REPLACE` that re-casts `changed_at`/`deleted_at` to BIGINT precisely so a rogue VARCHAR generation cannot make LWW ordering lexicographic. **Attribute** columns get no such re-pin: `read_parquet(..., union_by_name=true)` widens across generations and the `UNION ALL` with the typed hot pivot widens again.

[...]

Issue #370, opened 2026-08-06
Discuss: stable-by-name attributeID assignment for generate-attributes


iceboundrock:
Spun out of #315 (see its closing suggestion). Positional attributeID assignment is what turns any schema shape change into a physical EAV remap; #315 added $ref resolution plus byte-identical regeneration guards, which contain the hazard but do not remove it. A name-keyed assignment scheme would make the whole class of bug impossible. Needs its own design discussion: ID allocation, collision/migration story for the 12 shipped attribute files, interaction with the preserve-removed-attributes behavior.

Refs #315.

PR #373, opened 2026-08-06
fix(tools): #315 resolve $ref in generate-attributes; guard and repair schema artifact drift

### LINKED ISSUE #315: generate-attributes silently collapses object $ref and shifts every attributeID

iceboundrock:
Closes #315

**Spec:** issue #315 comment [issuecomment-5207776413](https://github.com/Lychee-Technology/forma/issues/315#issuecomment-5207776413) (approved design).
**Implementation plan:** [issuecomment-5207828511](https://github.com/Lychee-Technology/forma/issues/315#issuecomment-5207828511).

---

## What this fixes

`cmd/tools/generate_attributes.go` never resolved `$ref`. `traverseSchemaWithRequiredState` checks `schema["properties"]`; a `$ref`-only node has none, so it fell through to `getSchemaType` and was emitted as a **single scalar attribute**. A `$ref` to a scalar was accidentally correct; a `$ref` to an object silently collapsed the entire subtree — and because `attributeID` is assigned **positionally** over the surviving key set, and `attributeID` *is* the physical `attr_id` column of the EAV table, every subsequent attribute of every stored row was silently remapped.

The exposure is live, not theoretical: `visit.json`'s `contactSnapshot` is `{"$ref": "lead.json#/properties/contact", "x-relation": {...}}` — a 24-property object — as of #314's repointing. The committed `visit_attributes.json` was a hybrid: the collapsed scalar `contactSnapshot` (ID 25) plus 4 stale `contactSnapshot.*` keys (IDs 21–24) from an earlier inline era.

Fix: `generateAttributesJSON` loads the schema through `NewSchemaInliner(...).InlineFile(...)` before traversal, reusing the existing cross-file resolution, JSON-pointer handling, cycle detection and `$ref`-sibling merge instead of adding a second resolver.

---

## User rulings implemented (from the approved spec — do not re-litigate)

1. **list drift: repair by regeneration in this PR.** All `text` → `list`+`items_type` corrections are committed, attributeIDs unchanged. Implementation had to verify stored-row compatibility first — see the compatibility evidence below.
2. **`contactSnapshot` expansion: accepted.** Regenerating the visit family expands `contactSnapshot` into ~20 new `contactSnapshot.*` attributes with freshly appended IDs; the collapsed scalar (ID 25) and the 4 stale keys are preserved and demoted to optional (existing preserve semantics, consistent with #294 tolerate-and-preserve). `x-relation` is a read-time enrichment annotation; attributes are still generated from the schema shape.
3. **`*_full.json` drift: repair in this PR, plus a second guard** pinning every committed `*_full.json` to `inline-schema` output. Taken on `visit_full`; the same-day measurement found `lead_full` and `log_full` drifted the same way, so all three are repaired and the guard covers all six.
4. **ENOENT protection: explicit flag.** `loadExistingAttributes` returns an empty map when the output file does not exist, which renumbers every attribute from 1 in sorted-key order. The generator now refuses to run in that case unless `-init` is passed. `generateAttributesJSON` gained an `allowNew bool` parameter — fail-closed default, so a call site that goes to the zero value gets the *strict* behavior (no #314-style default-value blindness).

---

## Commits

| Hash | Subject |
|---|---|
| `b34285e` | `fix(tools): #315 resolve $ref via SchemaInliner before attribute traversal` |
| `5cb725b` | `fix(tools): #315 unify inline-schema error wrapping per review` |
| `d1a2149` | `feat(tools): #315 require -init to create a new attributes file` |
| `249c1bc` | `docs(tools): #315 document -init flag in cmd/tools README` |
| `835483c` | `fix(schemas): #315 repair drifted artifacts; guard byte-identical regeneration` |
| `d801e03` | `test(transform): #315 update lead_full fixtures for repaired contact required set` |
| `ff905dc` | `fix(transform): #315 exempt relation-root children from required-policy enforcement` |
| `e1cc862` | `review: #315 reattach Covers doc; pin _full-variant required enforcement` |

---

## ⚠️ Migration runbook (required before deploy)

**Plan-owner ruling:** the `text` → `list` flip ships despite the Parquet-tier conflict, because old tier data for these columns is **already silently wrong** — the pre-change `MAX(CASE …)` export collapsed `["alpha","beta"]` to the single string `'beta'`. The flip does not make existing parquet more wrong; it makes it unreadable, loudly.

Before deploying this change to any environment with existing tier data (**devo**):

1. **Stop CDC flush and compaction** for schemas `lead`, `lead_full`, `visit`, `visit_full`.
2. **Purge BOTH** `s3://<prefix>/<schema>/delta/` **and** `s3://<prefix>/<schema>/base/` for those schemas. Note: `cdc-init` replaces the **base** tier only — the delta purge is manual (`internal/cdc/init.go:421`; see follow-up issue #371).
3. **Run `cdc-init`** to rebuild base from Postgres. Hot data is unaffected and remains the source of truth — the EAV rows are physically identical before and after this change (proved below), so nothing is lost.
4. **Verify legacy `_full` rows against the new required policies** (final-review finding — the "hot data unaffected" statement above covers the list flips, not this): `lead_full`'s `contact.isAnonymous` is now `required_always`, so any pre-existing `lead_full` row **without** that EAV attribute fails on read (`FromEAVRecords` required check — plain operator error, incl. Get-before-Update); likewise `visit_full` rows carrying any `contactSnapshot.*` child but no `isAnonymous` (`required_if_parent_present`). Check devo for such rows and backfill `isAnonymous` before deploy, or accept loud read failures until backfilled. This failure class already existed for plain `lead` (which has carried `contact.isAnonymous: required_always` since before this branch); `_full` variants are catching up.
5. **Resume** flush and compaction.

If step 2 is skipped, the first federated query after the metadata flip fails with a DuckDB `Conversion Error`. **That is the designed failure mode, not data loss** — a classified read failure, not corrupted results. Postgres still holds every value.

---

## text→list compatibility evidence

The following is Task 3's evidence report, verbatim.

<details>
<summary>Task 3 — text→list stored-artifact compatibility evidence (click to expand)</summary>

**VERDICT: `INCOMPATIBLE`** — for any deployment whose S3 tier already holds parquet
objects carrying a non-NULL value in one of the 18 affected columns. The Postgres/EAV
side is provably compatible (physically identical rows); the **parquet tier is not**, and
the failure is a hard query error on every federated read, not a degradation.

Environment: worktree `/Users/ruoshi/code/Lychee/LTBase/forma-315`, branch
`fix/315-generate-attributes-ref`. All test output below is real output from throwaway
tests run against the repo's pinned DuckDB (`github.com/duckdb/duckdb-go/v2 v2.5.6`,
`go.mod:15`) with `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false`. The throwaway tests
were deleted after the run; nothing is committed by this task.

---

### 0. Scope: which attributes, which storage

The 18 entries Task 4 would flip are all **EAV-only** (no `columnBinding` in the shipped
`*_attributes.json`), so no `entity_main` hot column is involved. Their folded parquet
column names (`sqlgen.ParquetAttrColumn`, `internal/sqlgen/parquet_attr_column.go:27` —
dots fold to underscores) are:

| schema | attribute | attrID | parquet column |
|---|---|---|---|
| `lead` / `lead_full` | `tags` | 72 | `tags` |
| `lead` / `lead_full` | `communications` | 2 | `communications` |
| `lead` / `lead_full` | `visits` | 76 | `visits` |
| `lead` / `lead_full` | `contact.phones` | 19 | `contact_phones` |
| `lead` / `lead_full` | `requirement.mustHave` | 54 | `requirement_mustHave` |
| `lead` / `lead_full` | `requirement.niceToHave` | 55 | `requirement_niceToHave` |
| `lead` / `lead_full` | `requirement.unwantedConditions` | 64 | `requirement_unwantedConditions` |
| `visit` / `visit_full` | `attendees` | 3 | `attendees` |
| `visit_full` | `contactSnapshot.phones` | 20 | `contactSnapshot_phones` |
| `visit_full` | `logs` | 29 | `logs` |

(7 + 7 + 1 + 3 = 18, matching the plan.)

---

#### Static evidence

`flattenToAttributes`'s `[]any` branch (`internal/transform/transformer.go:305-337`) is
**metadata-independent for non-empty arrays**: it iterates elements and recurses with
`newIndices = append(indices, i)` (`:333-334`). It never consults `cache[attrName].ValueType`
on that path. The recursion bottoms out in the `default:` branch (`:338-364`), which sets
`ArrayIndices: joinIndices(indices)` (`:353`) and calls `populateTypedValue`.

`populateTypedValue` (`internal/transform/typed_value.go:12`) under `ValueTypeText`
(`:29-34`) writes `attr.ValueText`. Under `ValueTypeList` (`:62-78`) it asserts the row
already has indices (`:67`) and then **re-enters itself with `elemMeta.ValueType =
meta.EffectiveItemsType()`** (`:73-78`) — i.e. for `items_type: text` it takes the exact
same `ValueTypeText` branch and writes the exact same `ValueText`.

⇒ For a JSON array of strings the physical EAV rows are byte-identical under `text` and
under `list`+`items_type: text`.

The only metadata-sensitive spot in the `[]any` branch is the **empty** array
(`transformer.go:305-324`): `len(v) == 0` writes the `#204` empty-list marker row *only if*
`meta.ValueType == forma.ValueTypeList` (`:312`). Under `text` metadata it writes nothing.

#### Test evidence (`internal/transform`, throwaway)

```
=== RUN   TestTW315_WriteShapeArrayUnderTextVsList          # body: {"tags": ["alpha","beta"]}
    W1 text metadata (pre-change): attrID=18 indices="0" valueType=text value="alpha"
    W1 text metadata (pre-change): attrID=18 indices="1" valueType=text value="beta"
    W1 list metadata (post-change): attrID=18 indices="0" valueType=text value="alpha"
    W1 list metadata (post-change): attrID=18 indices="1" valueType=text value="beta"
--- PASS
=== RUN   TestTW315_WriteShapeEmptyArray                    # body: {"tags": []}
    W2 text metadata (pre-change): (no attributes)
    W2 list metadata (post-change): attrID=18 indices="" valueType=text value=<nil>
--- PASS
=== RUN   TestTW315_WriteShapeScalarValue                   # body: {"tags": "alpha"}
    W3 text metadata (pre-change): attrID=18 indices="" valueType=text value="alpha"
    W3 list metadata (post-change): ERROR convert value for attribute 'tags':
        invalid value for attribute 'tags' (attrID=18): list attribute requires an
        array value, got scalar alpha: invalid input
--- PASS
```

**Answer to the brief's Step-1 question: per-element rows with `array_indices`, not a
serialized scalar.** Three pre-change stored shapes are therefore possible:

- **(a)** per-element rows, `array_indices` = `"0"`, `"1"`, … — the normal case, identical
  to the post-change shape.
- **(b)** a scalar row with `array_indices = ""` — only if a client sent a bare string
  where the JSON Schema declares an array. Since #314 the write path validates the body
  against the registered JSON Schema (`internal/schemavalidate/`, wired at
  `internal/entity_crud_service.go:27`, `internal/entity_batch_service.go:29`), so such a
  body is rejected today; rows of shape (b) can only predate that guard.
- **(c)** no row at all for an explicit `[]` — pre-change, `{"tags": []}` persisted nothing.

#### #204 history check

`git log --all --oneline --grep='#204'` → `579e78a` is the only implementation commit. Its
diff on `internal/transform/transformer.go` touches only two places: the `FromAttributes`
empty-marker branch and the `len(v) == 0` empty-array branch. **The per-element decomposition
loop is untouched by #204** — it predates the list feature entirely.

The same commit message carries an explicit prior decision that is exactly this task's
question:

> *"Shipped `cmd/server/schemas/*_attributes.json` are intentionally NOT regenerated in this
> change: flipping existing scalar array columns to LIST would conflict with already-exported
> parquet under union_by_name. Existing deployments must re-export (or switch S3 prefix)
> before regenerating."*

The rest of this report verifies that claim empirically rather than inheriting it.

---

#### 2a. OLTP read / assembly — **COMPATIBLE**

`FromEAVRecords` (`internal/transform/attribute_converter.go:131`) substitutes the element
type for the container (`:168-173`), so a list attribute's rows are converted with
`ValueTypeText` — the same conversion the rows were written under.

`FromAttributes` (`internal/transform/transformer.go:99`) is metadata-sensitive in exactly
one place: the `attr.Value == nil` empty-marker branch (`:117-131`). Every other row goes
through `parseIndices(attr.ArrayIndices)` (`:134`, `internal/transform/array_paths.go:22`)
and `setValueAtPath` (`:140`), neither of which reads metadata.

Old rows of shape (a) reconstruct identically. Old rows of shape (b) (`array_indices=""`,
non-NULL value) never hit the marker branch (the branch requires `attr.Value == nil`), so
they take `parseIndices("") → nil` and reconstruct as a scalar — same as before the change,
no error. Shape (c) stays absent.

```
=== RUN   TestTW315_ReadOldPerElementRowsUnderListMetadata
    R1 text metadata (pre-change):  {"tags":["alpha","beta"]}
    R1 list metadata (post-change): {"tags":["alpha","beta"]}     # identical
--- PASS
=== RUN   TestTW315_ReadOldScalarRowUnderListMetadata
    R2 text metadata (pre-change):  {"tags":"alpha"}
    R2 list metadata (post-change): {"tags":"alpha"}              # identical, no error
--- PASS
```

#### 2b. CDC export — tolerates old rows, but **changes the parquet column type**

`buildSchemaDrivenProjection` (`internal/cdc/export_sql_builder.go:99`) branches on
`meta.ValueType == forma.ValueTypeList` at `:143`. Pre-change (`text`) it emits the scalar
collapse at `:158`:

```sql
MAX(CASE WHEN attr_id = 72 THEN value_text END) AS tags        -- → parquet VARCHAR
```

Post-change (`list`) it emits `:153-155`:

```sql
CASE WHEN count(*) FILTER (WHERE attr_id = 72) > 0
     THEN coalesce(list(value_text ORDER BY TRY_CAST(array_indices AS BIGINT))
                   FILTER (WHERE attr_id = 72 AND array_indices <> ''), [])
END AS tags                                                     -- → parquet VARCHAR[]
```

Two consequences:

1. **The parquet column type changes scalar `VARCHAR` → `VARCHAR[]`.** This is the whole
   problem; see 2c.
2. **Old rows are tolerated but shape (b) silently degrades to `[]`.** The `array_indices
   <> ''` filter excludes a legacy scalar row from the element aggregate while `count(*)`
   still sees it, so `coalesce(..., [])` fires. Verified:

```
=== RUN   TestTW315_ListAggregateOverLegacyScalarRow
    legacy scalar row (array_indices='') through the list aggregate -> []interface {}{}
    legacy per-element rows through the list aggregate -> []interface {}{"alpha","beta"}
--- PASS
```

The identical expression is the hot-tier pivot (`internal/sqlgen/duckdb_schema_projection.go:246-248`),
so a shape-(b) row also serves as `[]` from the hot tier. No error, but the value is lost —
worth noting even though it is not the gating failure.

Also worth recording as an improvement, not a regression: pre-change, the `MAX(CASE …)`
collapse already **destroyed** these arrays in parquet — a `["alpha","beta"]` row exported
as the single string `'beta'`. Every warm/cold read of these 18 attributes has been wrong
since the tiers were first populated. The flip fixes that going forward; it does not make
existing parquet more wrong, it makes it unreadable.

#### 2c. Federated / DuckDB read — **INCOMPATIBLE (hard error)**

Two independent binding sites force the old `VARCHAR` column and the new `VARCHAR[]` type
together:

- **Across parquet generations.** `s3_source` reads
  `{{.S3_SCAN_SOURCE}}` = `read_parquet(<paths>, union_by_name=true)`
  (`internal/sqlgen/duckdb_cold_scan.go:110`,
  `internal/sqlgen/advanced_query_template_duckdb.go` `s3_source` CTE). Its projection
  `buildS3Projection` (`internal/sqlgen/duckdb_schema_projection.go:138-151`) emits every
  attribute's folded column unconditionally (`:147-149`) — projection pushdown cannot spare it.
- **Across tiers.** `unified AS (SELECT * FROM s3_source UNION ALL SELECT * FROM pg_source)`
  (`advanced_query_template_duckdb.go:101-105`). `pg_source` carries the hot LIST pivot
  (`duckdb_schema_projection.go:203`, `:246-248`), i.e. `VARCHAR[]`.

Neither site has any tolerance for the type change. Note also that the `#255` cold-union
machinery does **not** help: `coldMissingColumns` (`internal/federated/cold_columns.go:22-39`)
only augments columns absent from *every* file, and `mergeColumnUnion`
(`internal/federated/parquet_schema_validation.go:366-372`) deliberately keeps only names,
first-seen type wins, delegating widening to `union_by_name` (`:365`). The pre-read validator
(`parquet_schema_validation.go:129`) inspects only the *system* columns, so it neither
rejects nor repairs an attribute-column type conflict.

**Real-DuckDB results (throwaway tests, `internal/sqlgen`):**

Case A — mixed parquet generations, old file `attr_tags` VARCHAR `'alpha'`, new file `VARCHAR[]`:

```
CASE A read_parquet(mixed VARCHAR/VARCHAR[]) ERROR:
  Conversion Error: Error while reading file ".../old.parquet":
  failed to cast column "attr_tags" from type VARCHAR to VARCHAR[]:
  Type VARCHAR with value 'alpha' can't be cast to the destination type VARCHAR[]
```

Case A' — `DESCRIBE` over that same mixed set resolves the union to `attr_tags -> VARCHAR[]`.
So the *binding* succeeds and only the *scan* fails: no pre-flight probe in the codebase can
see this coming, and the failure surfaces at query execution time.

Case E — through the production scan source, `sqlgen.BuildParquetScanSource(...)` (system-column
guard + `union_by_name`): **same Conversion Error**. The guard shape changes nothing.

Case B — the tier `UNION ALL` alone, no parquet involved:

```
SELECT 'alpha' AS attr_tags UNION ALL SELECT ['alpha','beta'] AS attr_tags
  → Conversion Error: Type VARCHAR with value 'alpha' can't be cast to the
    destination type VARCHAR[] when casting from source column attr_tags
```

Case G — **the worst case, and the one that decides the verdict**: a cold set containing
*only* pre-change parquet (no new-generation file anywhere) unioned with the post-change hot
leg — i.e. the very first federated query after the metadata flip, before any new export
exists:

```
CASE G ERROR: Conversion Error: Type VARCHAR with value 'alpha' can't be cast
              to the destination type VARCHAR[] when casting from source column attr_tags
```

So the breakage does **not** require a mixed corpus. It requires only (i) the metadata flip
and (ii) one pre-change parquet object with a non-NULL value in one of these columns.

Case D — compaction (`internal/compaction/merge.go:62-85`, `MergeToTmp` → `buildMergeSQL`
over `read_parquet(..., union_by_name=true)`) hits the identical error, so **compaction
cannot heal the corpus**; it fails on it too.

**Blast-radius bound (the one piece of good news).** The cast is per *value*, not per
*schema*, so an old column that is entirely NULL survives:

```
CASE F  old file attr_tags = CAST(NULL AS VARCHAR), new file VARCHAR[]
        → row a tags=<nil>;  row b tags=["alpha","beta"];  rows.Err=<nil>       OK
CASE G' old-only cold (all NULL) UNION ALL new hot LIST
        → row a tags=<nil>;  row b tags=["x","y"];         rows.Err=<nil>       OK
CASE H  CAST(NULL AS VARCHAR) AS VARCHAR[]  → NULL, no error
        CAST('' AS VARCHAR[])               → Conversion Error (empty string DOES fail)
        CAST('[a, b]' AS VARCHAR[])         → ['a','b']  (only bracket-literal text casts)
CASE I  count(*) not projecting the column  → 2, no error
        (irrelevant in production: buildS3Projection always projects it)
```

⇒ **The precise precondition for breakage:** at least one base/delta parquet object under the
deployment's S3 prefix carries a non-NULL, non-`[...]`-shaped string in one of the 10 distinct
folded column names above, for schema `lead`, `lead_full`, `visit`, or `visit_full`. If devo
has ever flushed a lead with `tags`, a contact with `phones`, or a visit with `attendees`, that
precondition is met.

---

### Step 3 — Verdict

**`INCOMPATIBLE`. STOP the plan; do not commit the regenerated `*_attributes.json` without a
paired storage remediation.**

Reasoning:

1. **Postgres/EAV is fine.** The flattener always decomposed JSON arrays into per-element
   rows regardless of declared `valueType` (`transformer.go:305-337`, unchanged by #204), and
   `populateTypedValue`'s list branch delegates to the items type
   (`typed_value.go:62-78`), so the stored rows are byte-identical before and after. OLTP
   reads are byte-identical too (test R1). If parquet were not in play, this would be a
   pure metadata relabel.
2. **Parquet is not fine.** The CDC export changes the physical column type from `VARCHAR`
   to `VARCHAR[]` (`export_sql_builder.go:143-159`), and every DuckDB read path that must
   reconcile the two — `read_parquet(union_by_name=true)`, the `s3_source`/`pg_source`
   `UNION ALL`, and the compaction merge — raises a hard `Conversion Error` on the first
   non-NULL legacy value (Cases A, B, D, E, G above).
3. **The failure fires immediately, not on the next flush.** Case G proves the hot tier's
   post-flip LIST pivot alone is enough to poison a union against a pre-flip cold set. There
   is no window in which the deployment is "flipped but not yet broken".
4. **It is loud, not silent** — a classified read failure, not corrupted results — which is
   the one mitigating fact, and it matches the failure mode #204's author predicted verbatim
   when he deliberately declined to regenerate these same files.
5. **Compaction cannot repair it**, so there is no self-healing path.

#### Affected surface

Schemas `lead`, `lead_full`, `visit`, `visit_full`; the 18 attribute entries / 10 distinct
folded parquet columns listed in §0. Every federated (warm/cold) query against those four
schemas fails outright once the metadata flips — including queries that never mention the
list attribute, because `buildS3Projection` projects all attribute columns.

#### Remediation options (for the user's decision)

- **A — New S3 prefix.** Flip the metadata and point the deployment at a fresh
  `S3Prefix`, then run `cmd/tools init-db` / `cdc-init` to build a base tier from Postgres
  under the new prefix. Postgres is the source of truth and its EAV rows are already
  correct, so no data is lost. Cleanest; the old prefix can be deleted after verification.
- **B — Full re-export in place.** Purge *both* the `/base/` and `/delta/` object sets and
  their manifests, then re-init. Note `cdc-init` calls
  `manifest.ReplaceTierFiles(..., "base", ...)` (`internal/cdc/init.go:421`) — it replaces
  the **base** tier only, so leftover pre-flip delta objects would still poison the scan.
  Delta must be cleared explicitly.
- **C — Ship the flip only to deployments with an empty/never-flushed S3 tier**, and gate
  devo behind A or B.
- **D — Split the change**: land the `generate_attributes` `$ref` fix (the actual #315
  defect) without regenerating the four affected `*_attributes.json` files, exactly as #204
  did, and track the regeneration + re-export as a separate operational issue.

Whichever is chosen, the metadata flip and the storage remediation must be **atomic from the
reader's point of view** — a server running new metadata against an old prefix is broken for
the duration.

---

#### Appendix — how the evidence was produced

Two throwaway test files (`internal/sqlgen/zz_throwaway_315_{compat,null,agg}_test.go`,
`internal/transform/zz_throwaway_315_test.go`) were written, run with
`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/sqlgen ./internal/transform -run TestTW315 -v`,
and deleted. Parquet fixtures were produced in-process with
`COPY (SELECT … ) TO '<tmp>.parquet' (FORMAT PARQUET)` against the same DuckDB build the
product links, so the Conversion Errors quoted above are the engine's own text on the pinned
version — not inference from the SQL. `git status` is clean; this task commits nothing.

</details>

**Resolution:** the plan owner chose **B (full re-export in place)**, scoped to devo, executed via the migration runbook above. The `$ref` fix and the regeneration ship together in this PR.

---

## Per-file delta

`git diff --stat cmd/server/schemas/`:

```
 cmd/server/schemas/lead_attributes.json       | 21 ++++---
 cmd/server/schemas/lead_full.json             | 26 +++++++-
 cmd/server/schemas/lead_full_attributes.json  | 49 ++++++++++++---
 cmd/server/schemas/log_full.json              | 14 ++++-
 cmd/server/schemas/log_full_attributes.json   | 15 ++++-
 cmd/server/schemas/visit_attributes.json      | 85 ++++++++++++++++++++++++++-
 cmd/server/schemas/visit_full.json            | 34 ++++++++---
 cmd/server/schemas/visit_full_attributes.json | 34 +++++++++--
 8 files changed, 243 insertions(+), 35 deletions(-)
```

The other 16 shipped schema files are byte-identical, as required.

### Attributes files

| File | text→list flips | new attrs (ID range) | required_policy changes |
|---|---|---|---|
| `activity_attributes.json` | — | — | — (no change) |
| `activity_full_attributes.json` | — | — | — (no change) |
| `communication_attributes.json` | — | — | — (no change) |
| `communication_full_attributes.json` | — | — | — (no change) |
| `contact_attributes.json` | — | — | — (no change) |
| `contact_full_attributes.json` | — | — | — (no change) |
| `lead_attributes.json` | **7** | — | — |
| `lead_full_attributes.json` | **7** | **6** (77..82) | 1 |
| `log_attributes.json` | — | — | — (no change) |
| `log_full_attributes.json` | — | **3** (11..13) | 1 |
| `visit_attributes.json` | **1** | **20** (29..48) | — |
| `visit_full_attributes.json` | **2** | **6** (42..47) | 1 |
| **total** | **17** | **35** | **3** |

**Precise counts.** **17 pre-existing `text` → `list` flips + 1 new-born list attribute (`contactSnapshot.phones`, ID 45 in `visit_attributes.json`) + 35 new attributes total.** Commit `835483c`'s message says "18 text->list corrections" (the spec's wording, used verbatim per instruction); the count above is the precise record — 18 is the number of attributes that end up typed `list`, of which 17 are corrections and 1 is brand new.

Detail:

- `lead_attributes.json` / `lead_full_attributes.json` — same 7 flips each, all `text -> list` with `items_type: text`: `communications`, `contact.phones`, `requirement.mustHave`, `requirement.niceToHave`, `requirement.unwantedConditions`, `tags`, `visits`.
- `lead_full_attributes.json` new (max old ID 76): `contact.contactCode` 77 (text), `contact.isAnonymous` 78 (bool), `contact.nickname` 79 (text), `createdBy` 80 (text), `ownerTeam` 81 (text), `updatedBy` 82 (text). Policy: `contact.name` `required_always -> optional`; the new `contact.isAnonymous` is `required_always`.
- `log_full_attributes.json` new (max old ID 10): `createdBy` 11, `ownerTeam` 12, `updatedBy` 13 (all text). Policy: `leadId` `required_always -> optional`.
- `visit_attributes.json` — 1 flip (`attendees`) + 20 new `contactSnapshot.*` at IDs 29..48: `annualIncome`(29,numeric), `annualIncomeCurrency`(30), `birthday`(31,date), `company`(32), `contactCode`(33), `currentResidence`(34), `email`(35), `employmentYear`(36,numeric), `familySize`(37,numeric), `isAnonymous`(38,bool), `isBlacklisted`(39,bool), `lineId`(40), `maritalStatus`(41), `memo`(42), `nickname`(43), `occupation`(44), `phones`(45,**list**), `preferredChannel`(46), `preferredLanguage`(47), `wechatId`(48).
- `visit_full_attributes.json` — 2 flips (`attendees`, `contactSnapshot.phones`) + 6 new at 42..47 (`contactSnapshot.contactCode` 42, `contactSnapshot.isAnonymous` 43, `contactSnapshot.nickname` 44, `createdBy` 45, `ownerTeam` 46, `updatedBy` 47). Policy: `contactSnapshot.name` `required_if_parent_present -> optional`.

### `_full.json` bodies

- `lead_full.json`: adds `contact.contactCode`, `contact.isAnonymous`, `contact.nickname`, and root `createdBy` / `ownerTeam` / `updatedBy`; `contact.required` changes `["name"] -> ["isAnonymous"]`.
- `visit_full.json`: same three contact-object additions and same three root audit fields; same `contact.required` flip; **removes `logs`** (an array of Log IDs) — `visit.json` no longer declares it.
- `log_full.json`: adds root `createdBy` / `ownerTeam` / `updatedBy`; **removes the description** on `data` (`log.json` has none); drops `leadId` from root `required`.

### Note on `visit_full_attributes.json` having 2 flips, not 3

The missing third is `logs`. The stale `visit_full.json` still carried `logs` as `array of string`, so a prediction made against the stale artifact expects it to flip to `list`. Because the repair regenerates `_full` bodies **first**, `logs` is gone from `visit_full.json` by the time attributes are generated, and the generator preserves it as an orphaned attribute `{"attributeID": 29, "valueType": "text"}`, unchanged. Correct under this ordering, not a defect.

---

## ⚠️ `_full` write-contract changes (client-visible)

Both are intended schema-truth alignment, not regressions, but both change accepted payloads.

**`visit_full`.** The relation-root carve-out (#314/#315) does **not** reach `_full` variants: inlining resolves `$ref` and strips `x-*`, so `visit_full.json` declares no `x-relation` and `RelationIndex` finds no relation roots for it (`grep -c 'x-relation' visit.json` → 1; `visit_full.json` → 0). `contactSnapshot` is therefore an ordinary nested object there. Combined with the #315 repair, `visit_full`'s write contract changes: `contactSnapshot.name` (ID 16) drops `required_if_parent_present` → optional, and the new `contactSnapshot.isAnonymous` (ID 43) carries `required_if_parent_present`. **A `visit_full` write carrying any `contactSnapshot` child but no `isAnonymous` now returns 400 where it previously passed.** Pinned by `TestFromEAVRecordsEnforcesRequiredPolicyOnFullVariant`. Plain `visit` is unaffected — the carve-out exempts it.

**`lead_full`.** `contact.isAnonymous` (ID 78) is now `required_always`, and `contact.name` drops to optional. This is `lead_full` **catching up to `lead`**, not a new demand: `lead_attributes.json` already carried `contact.isAnonymous: {attributeID: 78, required_always}` on `main@d941899`, before this branch. The two layers are now consistent, on the same attribute ID, matching `lead.json`'s own `contact.required: ["isAnonymous"]`.

---

## The relation-root carve-out (`ff905dc`, `e1cc862`)

The `contactSnapshot` expansion (ruling 2) made `contactSnapshot.*` attributes **exist** for `visit` for the first time. Their `required_policy` markers immediately collided with #314's relation-root write semantics: `TestCreateAcceptsDottedKeyUnderRelationRoot` went red —

```
failed to transform data to persistent record: failed to convert to attributes:
  convert to model.EntityAttribute: missing required attribute
  'contactSnapshot.isAnonymous' (attrID=38) in EAV records
```

A caller sending `contactSnapshot.name` alone now makes the parent "present", so the write path demanded `contactSnapshot.isAnonymous`. That is exactly the payload #314 decided must stay accepted; #314 carved out relation roots on the **validator** side, but the EAV `required_policy` side was never carved out because until this repair the attribute did not exist. The repair made a latent hole load-bearing.

**Plan-owner ruling: mirror the carve-out on the EAV side.** Implemented in `AttributeConverter.checkRequiredAttributes` (`internal/transform/attribute_converter.go`) — one `RelationRoots.Covers` guard at the top of the per-attribute loop. `RelationIndex` stays the single source of truth for what a relation root is; the set travels into `transform` via a new `RelationRootsLookup` func type and an optional `RelationRootsAware` interface, installed by `NewEntityManager` right after it loads the relation index.

Rejected alternatives: guarding `validateRequiredAttributesFromInput` (dead code for this case — after `StripComputedFields` the relation root is gone from the input map, so `required_if_parent_present` short-circuits); suppressing `required_policy` in the generator (no relation index there, and it would change bytes the new guards pin); re-deriving relation roots inside `transform` (two detectors that can disagree = silent write-acceptance bug).

**One real finding while wiring this:** the install initially had no effect in the integration harness — `writeSpy` wraps the transformer and did not implement the optional interface, so the lookup was dropped on the floor. The #313 decorator lesson again. `writeSpy` now forwards `SetRelationRoots`, and `NewEntityManager` logs a warning when the transformer is not `RelationRootsAware` while a relation index exists, so the next lossy decorator is visible instead of silent.

**Pinned in both directions** (`internal/transform/relation_root_required_policy_test.go`):

| Test | Pins |
|---|---|
| `TestFromEAVRecordsSkipsRequiredPolicyBeneathRelationRoot` | the carve-out itself |
| `TestFromEAVRecordsStillEnforcesRequiredPolicyOutsideRelationRoot` | the inverse — ordinary nested objects stay enforced; the carve-out must not widen to "nested policies are not enforced" |
| `TestFromEAVRecordsEnforcesRequiredPolicyOnRelationRootItself` | `Covers`' documented boundary — a name that *is* a root is not *beneath* one |
| `TestFromEAVRecordsWithoutRelationRootsKeepsEnforcement` | control: with no lookup installed, byte-for-byte the old behaviour |
| `TestFromEAVRecordsEnforcesRequiredPolicyOnFullVariant` | the `_full` write-contract change disclosed above, both directions |

`FromEAVRecords` was over the 100-line rule after the guard, so the required-policy block moved into `checkRequiredAttributes` (65 / 56 lines; `attribute_converter.go` 484 lines).

---

## attributeID invariant

The load-bearing safety property: **no stored row is remapped.** Checker compares every `*_attributes.json` on `main@d941899` against this branch:

```
$ python3 idcheck.py
checked 12 files
OK: all existing attributeIDs preserved across 12 files
```

Zero attributes disappeared; zero attributeIDs changed value. All new attributes take IDs strictly above each file's previous max (`visit_attributes.json`: max old ID 28, min new ID 29). The five pre-existing `contactSnapshot` entries (IDs 25, 21–24) keep both their IDs and their `text` type.

---

## Guards added

**`TestShippedAttributesRegenerateByteIdentical`** — every committed `*_attributes.json` must regenerate byte-identically (seed-first: copy the committed file onto the output path, regenerate in place, compare bytes).

**`TestShippedFullSchemasMatchInlineOutput`** — every committed `*_full.json` must match `inline-schema` output.

RED before the repair (exact match to the predicted drift set):

```
--- FAIL: TestShippedAttributesRegenerateByteIdentical
    --- PASS: activity  activity_full  communication  communication_full  contact  contact_full  log  log_full
    --- FAIL: lead  lead_full  visit  visit_full
--- FAIL: TestShippedFullSchemasMatchInlineOutput
    --- PASS: activity  communication  contact
    --- FAIL: lead  log  visit
```

with divergence windows like:

```
lead_attributes.json does not regenerate byte-identically (committed 7049 bytes, regenerated 7231 bytes)
  committed:   "communications": { "attributeID": 2, "valueType": "text" }
  regenerated: "communications": { "attributeID": 2, "valueType": "list", "items_type": "text" }
```

GREEN after:

```
--- PASS: TestShippedAttributesRegenerateByteIdentical (0.01s)   [12/12 subtests]
--- PASS: TestShippedFullSchemasMatchInlineOutput (0.00s)        [6/6 subtests]
```

---

## Mutation verification

**Mutant 1 — revert the `$ref` fix.** Task 1's `InlineFile` wiring in `generate_attributes.go` reverted to the old raw `os.ReadFile` + `json.Unmarshal`:

```
$ go test ./cmd/tools -run TestShippedAttributesRegenerateByteIdentical -v -count=1
    generate_attributes_regen_guard_test.go:82: visit_attributes.json does not regenerate
      byte-identically (committed 4104 bytes, regenerated 4051 bytes)
--- FAIL: TestShippedAttributesRegenerateByteIdentical/visit
    --- PASS: (all 11 others, incl. visit_full)
FAIL	github.com/lychee-technology/forma/cmd/tools	0.597s
```

The guard catches the exact #315 failure mode. `visit_full` correctly stays green — `visit_full.json` is the already-inlined artifact and contains zero `$ref` (`grep -c '$ref'` → 0, vs 3 in `visit.json`), so removing the inliner is a no-op for it. Note the mutant is caught by a **53-byte** difference, not by attributes vanishing: without inlining, the generator emits no `contactSnapshot.*` specs but *preserves* the orphaned existing ones, so all 47 survive and only their `required_policy` markers drop. Byte-exactness is what makes the guard sensitive enough to see it. Restored, re-run green.

**Mutant 2 — widen the carve-out** to `relationRoots.Covers(attrName) || strings.Contains(attrName, ".")`:

```
MUTATION: carve-out widened to every nested attribute
--- FAIL: TestAttributeConverterFromEAVRecords_NestedRequiredDependsOnParentPresence
--- FAIL: TestAttributeConverterFromEAVRecords_RequiredChildUsesArrayIndexContext
--- FAIL: TestAttributeConverterFromEAVRecords_NestedArrayRequiredUsesFullIndexPath
--- FAIL: TestAttributeConverterFromEAVRecords_RequiredAlwaysIgnoresParentPresence
--- FAIL: TestFromEAVRecordsStillEnforcesRequiredPolicyOutsideRelationRoot
--- FAIL: TestFromEAVRecordsEnforcesRequiredPolicyOnFullVariant
--- FAIL: TestFromEAVRecordsWithoutRelationRootsKeepsEnforcement
FAIL	github.com/lychee-technology/forma/internal/transform	0.194s
--- restoring ---
ok  	github.com/lychee-technology/forma/internal/transform	0.187s
```

The new `_full` pin is a real pin, not a tautology.

---

## Gates

All run on the branch head, working tree clean.

**`make lint`** — clean, exit 0. golangci-lint `v1.64.8` (the Makefile's pin), installed by the target itself:

```
$ make lint; echo "EXIT=$?"
Installing golangci-lint...
Running golangci-lint...
EXIT=0
$ golangci-lint --version
golangci-lint has version v1.64.8 built with go1.26.5 ...
```

**`make test`** — every package `ok`, zero failures:

```
Running unit tests...
ok  	github.com/lychee-technology/forma	0.315s
ok  	github.com/lychee-technology/forma/cdc	(cached)
ok  	github.com/lychee-technology/forma/cmd/benchmark	1.953s
ok  	github.com/lychee-technology/forma/cmd/lambda	0.595s
ok  	github.com/lychee-technology/forma/cmd/server	2.488s
ok  	github.com/lychee-technology/forma/cmd/tools	3.735s
ok  	github.com/lychee-technology/forma/factory	1.379s
ok  	github.com/lychee-technology/forma/internal	1.595s
ok  	github.com/lychee-technology/forma/internal/bootstrap	(cached)
ok  	github.com/lychee-technology/forma/internal/cdc	(cached)
ok  	github.com/lychee-technology/forma/internal/compaction	(cached)
ok  	github.com/lychee-technology/forma/internal/conditionexpr	(cached)
ok  	github.com/lychee-technology/forma/internal/duckdbinit	(cached)
ok  	github.com/lychee-technology/forma/internal/e2e_harness	(cached)
ok  	github.com/lychee-technology/forma/internal/e2e_harness/federated	(cached)
ok  	github.com/lychee-technology/forma/internal/e2e_harness/federated/benchmark	(cached)
ok  	github.com/lychee-technology/forma/internal/e2e_harness/production	(cached)
ok  	github.com/lychee-technology/forma/internal/federated	(cached)
ok  	github.com/lychee-technology/forma/internal/httpapi	2.597s
ok  	github.com/lychee-technology/forma/internal/manifest	(cached)
ok  	github.com/lychee-technology/forma/internal/model	(cached)
ok  	github.com/lychee-technology/forma/internal/numutil	(cached)
ok  	github.com/lychee-technology/forma/internal/parquetcheck	(cached)
ok  	github.com/lychee-technology/forma/internal/pgdsn	(cached)
ok  	github.com/lychee-technology/forma/internal/queryplan	(cached)
ok  	github.com/lychee-technology/forma/internal/reconcile	(cached)
ok  	github.com/lychee-technology/forma/internal/redact	(cached)
ok  	github.com/lychee-technology/forma/internal/schemameta	2.744s
ok  	github.com/lychee-technology/forma/internal/schemavalidate	2.960s
ok  	github.com/lychee-technology/forma/internal/sqlgen	(cached)
ok  	github.com/lychee-technology/forma/internal/sqlutil	(cached)
ok  	github.com/lychee-technology/forma/internal/transform	3.069s
```

**`make test-e2e-production`** — run because `ff905dc` changes runtime write-path behaviour and the production harness exercises the shipped schemas. **123 tests, 0 failures**:

```
--- PASS: TestProductionSmoke (1.65s)
... (123 --- PASS lines, 0 FAIL)
--- PASS: TestInitSchemaScopedIsolation (2.79s)
--- PASS: TestUpdateAfterAttributeRemoval (0.12s)
--- PASS: TestRecreateBetweenExportAndMark (3.75s)
PASS
ok  	github.com/lychee-technology/forma/internal/e2e_harness/production	132.678s
```

**`gofmt -l`** over the file list derived from git (`git diff --name-only d941899..HEAD -- '*.go'` plus `git ls-files --others --exclude-standard -- '*.go'`, #301 lesson — derived, not remembered) — **empty**. 15 changed `.go` files, 0 untracked `.go` files.

**`go vet ./...`** — clean, exit 0.

**Line limits** — every changed `.go` file is inside the 500-line rule; the largest is `internal/transform/transformer.go` at 485.

```
 58 internal/transform/relation_roots.go
 84 cmd/tools/main_test.go
124 cmd/tools/generate_attributes_regen_guard_test.go
129 cmd/tools/generate_attributes_ref_test.go
208 internal/transform/relation_root_required_policy_test.go
238 internal/entity_manager.go
280 cmd/tools/generate_attributes_test.go
335 internal/entity_write_validation_test.go
400 cmd/tools/generate_attributes_files_test.go
403 internal/transform/normalize_keys_arrays_test.go
407 cmd/tools/generate_attributes.go
423 cmd/tools/generate_attributes_merge_test.go
447 internal/transform/persistent_record.go
484 internal/transform/attribute_converter.go
485 internal/transform/transformer.go
```

**Not run:** the federated e2e suite (`-tags=e2e`, ~30m) — CI covers it, and this change's runtime surface is the write path, which the production harness exercises.

---

## Follow-up issues

- **#370** — `Discuss: stable-by-name attributeID assignment for generate-attributes`. Positional assignment is what turns any schema shape change into a physical EAV remap; this PR contains the hazard but does not remove it.
- **#371** — `cdc-init only replaces the base tier; delta files survive re-init` (`internal/cdc/init.go:421`). Directly load-bearing for the migration runbook above.
- **#372** — `Legacy scalar EAV rows under list metadata export/pivot as empty list`. Silent value loss for pre-#314 rows with `array_indices = ''`.
- **Cosmetic, no issue filed:** `log.json`'s audit-field descriptions say "visit record" (`"User ID of who created the visit record"`, etc. — `log.json:69,73,77`), a copy-paste from `visit.json`. It propagates into `log_full.json` through inlining. Fix in the source schema someday; it changes no behaviour.

---

### DIFF
--- a/cmd/server/schemas/lead_attributes.json
+++ b/cmd/server/schemas/lead_attributes.json
@@ -5,7 +5,8 @@
   },
   "communications": {
     "attributeID": 2,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contact.annualIncome": {
     "attributeID": 3,
@@ -86,7 +87,8 @@
   },
   "contact.phones": {
     "attributeID": 19,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contact.preferredChannel": {
     "attributeID": 20,
@@ -241,11 +243,13 @@
   },
   "requirement.mustHave": {
     "attributeID": 54,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "requirement.niceToHave": {
     "attributeID": 55,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "requirement.propertyType": {
     "attributeID": 56,
@@ -281,7 +285,8 @@
   },
   "requirement.unwantedConditions": {
     "attributeID": 64,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "score": {
     "attributeID": 65,
@@ -316,7 +321,8 @@
   },
   "tags": {
     "attributeID": 72,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "temperature": {
     "attributeID": 73,
@@ -338,6 +344,7 @@
   },
   "visits": {
     "attributeID": 76,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   }
 }
--- a/cmd/server/schemas/lead_full.json
+++ b/cmd/server/schemas/lead_full.json
@@ -46,6 +46,10 @@
           "description": "Employer or company name",
           "type": "string"
         },
+        "contactCode": {
+          "description": "Code for the contact when it is anonymous",
+          "type": "string"
+        },
         "currentResidence": {
           "description": "Current residential address or area",
           "type": "string"
@@ -65,6 +69,10 @@
           "description": "Number of family members excluding the lead",
           "minimum": 0,
           "type": "integer"
+        },
+        "isAnonymous": {
+          "description": "Whether the contact is anonymous",
+          "type": "boolean"
         },
         "isBlacklisted": {
           "description": "Whether the lead is blacklisted",
@@ -99,6 +107,10 @@
         },
         "nameNative": {
           "description": "Native language name",
+          "type": "string"
+        },
+        "nickname": {
+          "description": "Nickname for the contact. Used when the contact is anonymous.",
           "type": "string"
         },
         "occupation": {
@@ -147,13 +159,17 @@
         }
       },
       "required": [
-        "name"
+        "isAnonymous"
       ],
       "type": "object"
     },
     "createdAt": {
       "description": "Record creation timestamp",
       "format": "date-time",
+      "type": "string"
+    },
+    "createdBy": {
+      "description": "User ID of who created the visit record",
       "type": "string"
     },
     "firstTouchAt": {
@@ -175,6 +191,10 @@
     "nextFollowUpAt": {
       "description": "Scheduled next follow-up time",
       "format": "date-time",
+      "type": "string"
+    },
+    "ownerTeam": {
+      "description": "Team that owns the visit record",
       "type": "string"
     },
     "ownerUserId": {
@@ -493,6 +513,10 @@
     "updatedAt": {
       "description": "Last update timestamp",
       "format": "date-time",
+      "type": "string"
+    },
+    "updatedBy": {
+      "description": "User ID of who last updated the visit record",
       "type": "string"
     },
     "visits": {
--- a/cmd/server/schemas/lead_full_attributes.json
+++ b/cmd/server/schemas/lead_full_attributes.json
@@ -5,7 +5,8 @@
   },
   "communications": {
     "attributeID": 2,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contact.annualIncome": {
     "attributeID": 3,
@@ -23,6 +24,10 @@
     "attributeID": 6,
     "valueType": "text"
   },
+  "contact.contactCode": {
+    "attributeID": 77,
+    "valueType": "text"
+  },
   "contact.currentResidence": {
     "attributeID": 7,
     "valueType": "text"
@@ -38,6 +43,11 @@
   "contact.familySize": {
     "attributeID": 10,
     "valueType": "numeric"
+  },
+  "contact.isAnonymous": {
+    "attributeID": 78,
+    "valueType": "bool",
+    "required_policy": "required_always"
   },
   "contact.isBlacklisted": {
     "attributeID": 11,
@@ -57,8 +67,7 @@
   },
   "contact.name": {
     "attributeID": 15,
-    "valueType": "text",
-    "required_policy": "required_always"
+    "valueType": "text"
   },
   "contact.nameKana": {
     "attributeID": 16,
@@ -68,13 +77,18 @@
     "attributeID": 17,
     "valueType": "text"
   },
+  "contact.nickname": {
+    "attributeID": 79,
+    "valueType": "text"
+  },
   "contact.occupation": {
     "attributeID": 18,
     "valueType": "text"
   },
   "contact.phones": {
     "attributeID": 19,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contact.preferredChannel": {
     "attributeID": 20,
@@ -97,6 +111,10 @@
     "valueType": "date",
     "required_policy": "required_always"
   },
+  "createdBy": {
+    "attributeID": 80,
+    "valueType": "text"
+  },
   "firstTouchAt": {
     "attributeID": 25,
     "valueType": "date"
@@ -113,6 +131,10 @@
   "nextFollowUpAt": {
     "attributeID": 28,
     "valueType": "date"
+  },
+  "ownerTeam": {
+    "attributeID": 81,
+    "valueType": "text"
   },
   "ownerUserId": {
     "attributeID": 29,
@@ -221,11 +243,13 @@
   },
   "requirement.mustHave": {
     "attributeID": 54,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "requirement.niceToHave": {
     "attributeID": 55,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "requirement.propertyType": {
     "attributeID": 56,
@@ -261,7 +285,8 @@
   },
   "requirement.unwantedConditions": {
     "attributeID": 64,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "score": {
     "attributeID": 65,
@@ -296,7 +321,8 @@
   },
   "tags": {
     "attributeID": 72,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "temperature": {
     "attributeID": 73,
@@ -312,8 +338,13 @@
     "valueType": "date",
     "required_policy": "required_always"
   },
+  "updatedBy": {
+    "attributeID": 82,
+    "valueType": "text"
+  },
   "visits": {
     "attributeID": 76,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   }
 }
--- a/cmd/server/schemas/log_full.json
+++ b/cmd/server/schemas/log_full.json
@@ -7,8 +7,11 @@
       "format": "date-time",
       "type": "string"
     },
+    "createdBy": {
+      "description": "User ID of who created the visit record",
+      "type": "string"
+    },
     "data": {
-      "description": "Base64 encoded data content (audio, image, or text)",
       "type": "string"
     },
     "id": {
@@ -29,6 +32,10 @@
     },
     "ownerId": {
       "description": "Owner/creator of this log entry",
+      "type": "string"
+    },
+    "ownerTeam": {
+      "description": "Team that owns the visit record",
       "type": "string"
     },
     "summary": {
@@ -53,6 +60,10 @@
       "format": "date-time",
       "type": "string"
     },
+    "updatedBy": {
+      "description": "User ID of who last updated the visit record",
+      "type": "string"
+    },
     "visitId": {
       "description": "Visit identifier",
       "format": "uuid",
@@ -62,7 +73,6 @@
   },
   "required": [
     "id",
-    "leadId",
     "ownerId",
     "type",
     "createdAt"
--- a/cmd/server/schemas/log_full_attributes.json
+++ b/cmd/server/schemas/log_full_attributes.json
@@ -4,6 +4,10 @@
     "valueType": "date",
     "required_policy": "required_always"
   },
+  "createdBy": {
+    "attributeID": 11,
+    "valueType": "text"
+  },
   "data": {
     "attributeID": 2,
     "valueType": "text"
@@ -15,8 +19,7 @@
   },
   "leadId": {
     "attributeID": 4,
-    "valueType": "text",
-    "required_policy": "required_always"
+    "valueType": "text"
   },
   "noteId": {
     "attributeID": 5,
@@ -26,6 +29,10 @@
     "attributeID": 6,
     "valueType": "text",
     "required_policy": "required_always"
+  },
+  "ownerTeam": {
+    "attributeID": 12,
+    "valueType": "text"
   },
   "summary": {
     "attributeID": 7,
@@ -39,6 +46,10 @@
   "updatedAt": {
     "attributeID": 9,
     "valueType": "date"
+  },
+  "updatedBy": {
+    "attributeID": 13,
+    "valueType": "text"
   },
   "visitId": {
     "attributeID": 10,
--- a/cmd/server/schemas/visit_attributes.json
+++ b/cmd/server/schemas/visit_attributes.json
@@ -9,10 +9,68 @@
   },
   "attendees": {
     "attributeID": 3,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contactSnapshot": {
     "attributeID": 25,
+    "valueType": "text"
+  },
+  "contactSnapshot.annualIncome": {
+    "attributeID": 29,
+    "valueType": "numeric"
+  },
+  "contactSnapshot.annualIncomeCurrency": {
+    "attributeID": 30,
+    "valueType": "text"
+  },
+  "contactSnapshot.birthday": {
+    "attributeID": 31,
+    "valueType": "date"
+  },
+  "contactSnapshot.company": {
+    "attributeID": 32,
+    "valueType": "text"
+  },
+  "contactSnapshot.contactCode": {
+    "attributeID": 33,
+    "valueType": "text"
+  },
+  "contactSnapshot.currentResidence": {
+    "attributeID": 34,
+    "valueType": "text"
+  },
+  "contactSnapshot.email": {
+    "attributeID": 35,
+    "valueType": "text"
+  },
+  "contactSnapshot.employmentYear": {
+    "attributeID": 36,
+    "valueType": "numeric"
+  },
+  "contactSnapshot.familySize": {
+    "attributeID": 37,
+    "valueType": "numeric"
+  },
+  "contactSnapshot.isAnonymous": {
+    "attributeID": 38,
+    "valueType": "bool",
+    "required_policy": "required_if_parent_present"
+  },
+  "contactSnapshot.isBlacklisted": {
+    "attributeID": 39,
+    "valueType": "bool"
+  },
+  "contactSnapshot.lineId": {
+    "attributeID": 40,
+    "valueType": "text"
+  },
+  "contactSnapshot.maritalStatus": {
+    "attributeID": 41,
+    "valueType": "text"
+  },
+  "contactSnapshot.memo": {
+    "attributeID": 42,
     "valueType": "text"
   },
   "contactSnapshot.name": {
@@ -27,8 +85,33 @@
     "attributeID": 23,
     "valueType": "text"
   },
+  "contactSnapshot.nickname": {
+    "attributeID": 43,
+    "valueType": "text"
+  },
+  "contactSnapshot.occupation": {
+    "attributeID": 44,
+    "valueType": "text"
+  },
+  "contactSnapshot.phones": {
+    "attributeID": 45,
+    "valueType": "list",
+    "items_type": "text"
+  },
+  "contactSnapshot.preferredChannel": {
+    "attributeID": 46,
+    "valueType": "text"
+  },
+  "contactSnapshot.preferredLanguage": {
+    "attributeID": 47,
+    "valueType": "text"
+  },
   "contactSnapshot.primaryPhone": {
     "attributeID": 24,
+    "valueType": "text"
+  },
+  "contactSnapshot.wechatId": {
+    "attributeID": 48,
     "valueType": "text"
   },
   "createdAt": {
--- a/cmd/server/schemas/visit_full.json
+++ b/cmd/server/schemas/visit_full.json
@@ -50,6 +50,10 @@
           "description": "Employer or company name",
           "type": "string"
         },
+        "contactCode": {
+          "description": "Code for the contact when it is anonymous",
+          "type": "string"
+        },
         "currentResidence": {
           "description": "Current residential address or area",
           "type": "string"
@@ -69,6 +73,10 @@
           "description": "Number of family members excluding the lead",
           "minimum": 0,
           "type": "integer"
+        },
+        "isAnonymous": {
+          "description": "Whether the contact is anonymous",
+          "type": "boolean"
         },
         "isBlacklisted": {
           "description": "Whether the lead is blacklisted",
@@ -103,6 +111,10 @@
         },
         "nameNative": {
           "description": "Native language name",
+          "type": "string"
+        },
+        "nickname": {
+          "description": "Nickname for the contact. Used when the contact is anonymous.",
           "type": "string"
         },
         "occupation": {
@@ -151,13 +163,17 @@
         }
       },
       "required": [
-        "name"
+        "isAnonymous"
       ],
       "type": "object"
     },
     "createdAt": {
       "description": "Record creation timestamp",
       "format": "date-time",
+      "type": "string"
+    },
+    "createdBy": {
+      "description": "User ID of who created the visit record",
       "type": "string"
     },
     "feedback": {
@@ -176,17 +192,13 @@
       "pattern": "^[a-zA-Z0-9_-]+$",
       "type": "string"
     },
-    "logs": {
-      "description": "Visit log IDs tied to this visit",
-      "items": {
-        "description": "Log ID",
-        "type": "string"
-      },
-      "type": "array"
-    },
     "nextFollowUpAt": {
       "description": "Next follow-up time derived from the visit",
       "format": "date-time",
+      "type": "string"
+    },
+    "ownerTeam": {
+      "description": "Team that owns the visit record",
       "type": "string"
     },
     "propertyId": {
@@ -240,6 +252,10 @@
     "updatedAt": {
       "description": "Last update timestamp",
       "format": "date-time",
+      "type": "string"
+    },
+    "updatedBy": {
+      "description": "User ID of who last updated the visit record",
       "type": "string"
     },
     "userId": {
--- a/cmd/server/schemas/visit_full_attributes.json
+++ b/cmd/server/schemas/visit_full_attributes.json
@@ -9,7 +9,8 @@
   },
   "attendees": {
     "attributeID": 3,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contactSnapshot.annualIncome": {
     "attributeID": 4,
@@ -27,6 +28,10 @@
     "attributeID": 7,
     "valueType": "text"
   },
+  "contactSnapshot.contactCode": {
+    "attributeID": 42,
+    "valueType": "text"
+  },
   "contactSnapshot.currentResidence": {
     "attributeID": 8,
     "valueType": "text"
@@ -42,6 +47,11 @@
   "contactSnapshot.familySize": {
     "attributeID": 11,
     "valueType": "numeric"
+  },
+  "contactSnapshot.isAnonymous": {
+    "attributeID": 43,
+    "valueType": "bool",
+    "required_policy": "required_if_parent_present"
   },
   "contactSnapshot.isBlacklisted": {
     "attributeID": 12,
@@ -61,8 +71,7 @@
   },
   "contactSnapshot.name": {
     "attributeID": 16,
-    "valueType": "text",
-    "required_policy": "required_if_parent_present"
+    "valueType": "text"
   },
   "contactSnapshot.nameKana": {
     "attributeID": 17,
@@ -72,13 +81,18 @@
     "attributeID": 18,
     "valueType": "text"
   },
+  "contactSnapshot.nickname": {
+    "attributeID": 44,
+    "valueType": "text"
+  },
   "contactSnapshot.occupation": {
     "attributeID": 19,
     "valueType": "text"
   },
   "contactSnapshot.phones": {
     "attributeID": 20,
-    "valueType": "text"
+    "valueType": "list",
+    "items_type": "text"
   },
   "contactSnapshot.preferredChannel": {
     "attributeID": 21,
@@ -99,6 +113,10 @@
   "createdAt": {
     "attributeID": 25,
     "valueType": "date"
+  },
+  "createdBy": {
+    "attributeID": 45,
+    "valueType": "text"
   },
   "feedback": {
     "attributeID": 26,
@@ -121,6 +139,10 @@
   "nextFollowUpAt": {
     "attributeID": 30,
     "valueType": "date"
+  },
+  "ownerTeam": {
+    "attributeID": 46,
+    "valueType": "text"
   },
   "propertyId": {
     "attributeID": 31,
@@ -164,6 +186,10 @@
   "updatedAt": {
     "attributeID": 40,
     "valueType": "date"
+  },
+  "updatedBy": {
+    "attributeID": 47,
+    "valueType": "text"
   },
   "userId": {
     "attributeID": 41,
--- a/cmd/tools/README.md
+++ b/cmd/tools/README.md
@@ -18,10 +18,12 @@ generate-attributes
 - `-schema`：Schema 名（不含 `.json`）；与 `-schema-file` 互斥。
 - `-schema-file`：Schema 文件全路径；优先于 `-schema/-schema-dir`。
 - `-out`：输出文件路径；默认写到 schema 同目录的 `<name>_attributes.json`。
+- `-init`：允许创建全新的 attributes 文件；不指定此标志时，输出文件缺失会导致错误（因为生成会将所有 attributeID 从 1 重新编号）。
 
 示例：
-- `go run ./cmd/tools generate-attributes -schema lead`
-- `go run ./cmd/tools generate-attributes -schema-file /path/to/schema.json -out /tmp/schema_attributes.json`
+- `go run ./cmd/tools generate-attributes -schema lead`（更新既有 attributes 文件）
+- `go run ./cmd/tools generate-attributes -schema-file /path/to/schema.json`（再生成）
+- `go run ./cmd/tools generate-attributes -schema-file /path/to/schema.json -out /tmp/schema_attributes.json -init`（创建新文件）
 
 init-db
 -------
--- a/cmd/tools/generate_attributes.go
+++ b/cmd/tools/generate_attributes.go
@@ -70,14 +70,18 @@ func runGenerateAttributes(_ context.Context, args []string) error {
 }
 
 func generateAttributesJSON(schemaPath, outputPath string) error {
-	data, err := os.ReadFile(schemaPath)
+	inliner := NewSchemaInliner(filepath.Dir(schemaPath))
+	schema, err := inliner.InlineFile(schemaPath)
 	if err != nil {
-		return fmt.Errorf("read schema file: %w", err)
-	}
-
-	var schema map[string]any
-	if err := json.Unmarshal(data, &schema); err != nil {
-		return fmt.Errorf("parse schema JSON: %w", err)
+		// Adapt inliner error messages to match legacy test expectations
+		errMsg := err.Error()
+		if strings.Contains(errMsg, "read file") {
+			return fmt.Errorf("read schema file: %w", err)
+		}
+		if strings.Contains(errMsg, "parse JSON") {
+			return fmt.Errorf("parse schema JSON: %w", err)
+		}
+		return fmt.Errorf("inline schema: %w", err)
 	}
 
 	// Extract attributes from the schema
--- a/cmd/tools/generate_attributes.go
+++ b/cmd/tools/generate_attributes.go
@@ -73,15 +73,7 @@ func generateAttributesJSON(schemaPath, outputPath string) error {
 	inliner := NewSchemaInliner(filepath.Dir(schemaPath))
 	schema, err := inliner.InlineFile(schemaPath)
 	if err != nil {
-		// Adapt inliner error messages to match legacy test expectations
-		errMsg := err.Error()
-		if strings.Contains(errMsg, "read file") {
-			return fmt.Errorf("read schema file: %w", err)
-		}
-		if strings.Contains(errMsg, "parse JSON") {
-			return fmt.Errorf("parse schema JSON: %w", err)
-		}
-		return fmt.Errorf("inline schema: %w", err)
+		return fmt.Errorf("inline schema %s: %w", schemaPath, err)
 	}
 
 	// Extract attributes from the schema
--- a/cmd/tools/generate_attributes.go
+++ b/cmd/tools/generate_attributes.go
@@ -39,6 +39,7 @@ func runGenerateAttributes(_ context.Context, args []string) error {
 	schemaName := flags.String("schema", "", "Schema name without extension (mutually exclusive with -schema-file)")
 	schemaFile := flags.String("schema-file", "", "Path to the JSON schema file (overrides -schema and -schema-dir)")
 	outputFile := flags.String("out", "", "Path to write the generated attributes JSON (defaults next to schema file)")
+	initNew := flags.Bool("init", false, "Allow creating a brand-new attributes file (without it, a missing output file is an error because generation would renumber every attributeID from 1)")
 
 	if err := flags.Parse(args); err != nil {
 		if errors.Is(err, flag.ErrHelp) {
@@ -61,7 +62,7 @@ func runGenerateAttributes(_ context.Context, args []string) error {
 		resolvedOutputPath = filepath.Join(filepath.Dir(resolvedSchemaPath), base+"_attributes.json")
 	}
 
-	if err := generateAttributesJSON(resolvedSchemaPath, resolvedOutputPath); err != nil {
+	if err := generateAttributesJSON(resolvedSchemaPath, resolvedOutputPath, *initNew); err != nil {
 		return err
 	}
 
@@ -69,7 +70,16 @@ func runGenerateAttributes(_ context.Context, args []string) error {
 	return nil
 }
 
-func generateAttributesJSON(schemaPath, outputPath string) error {
+func generateAttributesJSON(schemaPath, outputPath string, allowNew bool) error {
+	if _, err := os.Stat(outputPath); err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("stat output file %s: %w", outputPath, err)
+		}
+		if !allowNew {
+			return fmt.Errorf("output file %s does not exist: generating without it would renumber every attributeID from 1 (attributeIDs are physical EAV keys); pass -init only for a genuinely new schema", outputPath)
+		}
+	}
+
 	inliner := NewSchemaInliner(filepath.Dir(schemaPath))
 	schema, err := inliner.InlineFile(schemaPath)
 	if err != nil {
--- a/cmd/tools/generate_attributes_files_test.go
+++ b/cmd/tools/generate_attributes_files_test.go
@@ -192,7 +192,7 @@ func TestGenerateAttributesJSONNewFile(t *testing.T) {
 		t.Fatalf("failed to write schema file: %v", err)
 	}
 
-	err := generateAttributesJSON(schemaPath, outputPath)
+	err := generateAttributesJSON(schemaPath, outputPath, true)
 	if err != nil {
 		t.Fatalf("expected no error, got %v", err)
 	}
@@ -240,7 +240,7 @@ func TestGenerateAttributesJSONPreserveExistingIDs(t *testing.T) {
 		t.Fatalf("failed to write schema file: %v", err)
 	}
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Read first generation result
 	data, _ := os.ReadFile(outputPath)
@@ -266,7 +266,7 @@ func TestGenerateAttributesJSONPreserveExistingIDs(t *testing.T) {
 		t.Fatalf("failed to write schema file: %v", err)
 	}
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Read second generation result
 	data, _ = os.ReadFile(outputPath)
@@ -309,7 +309,7 @@ func TestGenerateAttributesJSONRemoveAttributeFromSchema(t *testing.T) {
 	schemaData, _ := json.Marshal(schema1)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Now remove one attribute from schema
 	schema2 := map[string]any{
@@ -322,7 +322,7 @@ func TestGenerateAttributesJSONRemoveAttributeFromSchema(t *testing.T) {
 	schemaData, _ = json.Marshal(schema2)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Verify that email is still in the attributes file
 	data, _ := os.ReadFile(outputPath)
@@ -359,7 +359,7 @@ func TestGenerateAttributesJSONRemovedAttributeClearsRequiredFlag(t *testing.T) 
 	schemaData, _ := json.Marshal(initialSchema)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	if err := generateAttributesJSON(schemaPath, outputPath); err != nil {
+	if err := generateAttributesJSON(schemaPath, outputPath, true); err != nil {
 		t.Fatalf("expected no error on initial generation, got %v", err)
 	}
 
@@ -375,7 +375,7 @@ func TestGenerateAttributesJSONRemovedAttributeClearsRequiredFlag(t *testing.T) 
 	schemaData, _ = json.Marshal(updatedSchema)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	if err := generateAttributesJSON(schemaPath, outputPath); err != nil {
+	if err := generateAttributesJSON(schemaPath, outputPath, true); err != nil {
 		t.Fatalf("expected no error on regeneration, got %v", err)
 	}
 
--- a/cmd/tools/generate_attributes_merge_test.go
+++ b/cmd/tools/generate_attributes_merge_test.go
@@ -220,7 +220,7 @@ func TestGenerateAttributesJSONInvalidSchemaFile(t *testing.T) {
 	outputPath := filepath.Join(tempDir, "attributes.json")
 
 	err := generateAttributesJSON(schemaPath, outputPath)
-	if err == nil || !strings.Contains(err.Error(), "read schema file") {
+	if err == nil || !strings.Contains(err.Error(), "inline schema") {
 		t.Fatalf("expected read error, got %v", err)
 	}
 }
@@ -234,7 +234,7 @@ func TestGenerateAttributesJSONInvalidJSON(t *testing.T) {
 	_ = os.WriteFile(schemaPath, []byte("{invalid json"), 0o644)
 
 	err := generateAttributesJSON(schemaPath, outputPath)
-	if err == nil || !strings.Contains(err.Error(), "parse schema JSON") {
+	if err == nil || !strings.Contains(err.Error(), "inline schema") {
 		t.Fatalf("expected parse error, got %v", err)
 	}
 }
--- a/cmd/tools/generate_attributes_merge_test.go
+++ b/cmd/tools/generate_attributes_merge_test.go
@@ -25,7 +25,7 @@ func TestGenerateAttributesJSONUpdateValueType(t *testing.T) {
 	schemaData, _ := json.Marshal(schema1)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Update schema to add date format
 	schema2 := map[string]any{
@@ -40,7 +40,7 @@ func TestGenerateAttributesJSONUpdateValueType(t *testing.T) {
 	schemaData, _ = json.Marshal(schema2)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Verify value type was updated
 	data, _ := os.ReadFile(outputPath)
@@ -72,7 +72,7 @@ func TestGenerateAttributesJSONMarksRequiredAttributes(t *testing.T) {
 	schemaData, _ := json.Marshal(schema)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	if err := generateAttributesJSON(schemaPath, outputPath); err != nil {
+	if err := generateAttributesJSON(schemaPath, outputPath, true); err != nil {
 		t.Fatalf("expected no error, got %v", err)
 	}
 
@@ -112,7 +112,7 @@ func TestGenerateAttributesJSONUpdatesRequiredFlag(t *testing.T) {
 	schemaData, _ := json.Marshal(initialSchema)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	if err := generateAttributesJSON(schemaPath, outputPath); err != nil {
+	if err := generateAttributesJSON(schemaPath, outputPath, true); err != nil {
 		t.Fatalf("expected no error, got %v", err)
 	}
 
@@ -140,7 +140,7 @@ func TestGenerateAttributesJSONUpdatesRequiredFlag(t *testing.T) {
 	schemaData, _ = json.Marshal(updatedSchema)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	if err := generateAttributesJSON(schemaPath, outputPath); err != nil {
+	if err := generateAttributesJSON(schemaPath, outputPath, true); err != nil {
 		t.Fatalf("expected no error on regeneration, got %v", err)
 	}
 
@@ -194,7 +194,7 @@ func TestGenerateAttributesJSONMaxIDCalculation(t *testing.T) {
 	schemaData, _ := json.Marshal(schema)
 	_ = os.WriteFile(schemaPath, schemaData, 0o644)
 
-	_ = generateAttributesJSON(schemaPath, outputPath)
+	_ = generateAttributesJSON(schemaPath, outputPath, true)
 
 	// Verify new IDs start after maxID (10)
 	data, _ := os.ReadFile(outputPath)
@@ -219,7 +219,7 @@ func TestGenerateAttributesJSONInvalidSchemaFile(t *testing.T) {
 	schemaPath := filepath.Join(tempDir, "nonexistent.json")
 	outputPath := filepath.Join(tempDir, "attributes.json")
 
-	err := generateAttributesJSON(schemaPath, outputPath)
+	err := generateAttributesJSON(schemaPath, outputPath, true)
 	if err == nil || !strings.Contains(err.Error(), "inline schema") {
 		t.Fatalf("expected read error, got %v", err)
 	}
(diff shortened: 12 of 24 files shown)


iceboundrock:
## Final whole-branch review (pre-merge gate)

**Verdict: Ready to merge — Yes** (one Important disclosure gap found; resolved by runbook step 4 added to the PR body, no code change).

### Independently re-verified (not inherited from task reports)
- **attributeID invariant re-proven against branch base `d941899`**: zero shifted IDs, zero removed keys, zero duplicates across all 5 changed `*_attributes.json`; new IDs append above each file's prior max (`lead_full` 77–82, `log_full` 11–13, `visit` 29–48, `visit_full` 42–47).
- **Carve-out blast radius traced end to end**: CRUD and batch write paths both strip relation roots pre-transform (`entity_crud_service.go:76/217`, `entity_batch_service.go:216/306`), dotted descendants funnel through `FromEAVRecords` where the carve-out lives; CDC/compaction/federated paths bypass `FromEAVRecords` entirely; `factory → NewEntityManager` is the only production composition and installs the lookup. `RelationIndex.RelationRoots` is nil-receiver safe.
- **Guard soundness**: seed-first defeats the ENOENT renumber trap; an ID-shifting regression turns the guard red (mutation-verified on the visit family). Inherent limit: artifacts are their own ID oracle — a deliberate hand-edit passes trivially; no independent source exists.
- Fresh `-count=1` runs green on `cmd/tools`, `internal/transform`, `internal`; `gofmt`/`vet` clean.

### Important (resolved in PR body)
Legacy `_full` **read-side** exposure was not covered by the runbook: `lead_full` rows missing `contact.isAnonymous` (now `required_always`) fail on read post-merge; likewise `visit_full` rows with `contactSnapshot.*` but no `isAnonymous`. The runbook's "hot data is unaffected" holds for the list flips only. → Added runbook step 4 (verify/backfill before deploy). Mitigations: loud not silent; failure class already existed for plain `lead`; deploy scope is devo.

### Minor (all OK to defer)
- `entity_manager.go:117` warn branch is silent when a `RelationRootsAware` decorator swallows the forward (shipped-schema integration pin covers the real composition).
- `_full` guard re-implements the tool's `MarshalIndent` encoding — encoding drift fails red but misattributed to artifacts (safe direction).
- `lead_full.json`/`log_full.json` audit-field descriptions say "visit record" (faithful inlining of `log.json`/source copy-paste) — cosmetic follow-up candidate.
- `attribute_converter.go` at 484/500 lines; `generateAttributesJSON` at 96/100 — next in-body change forces extraction.

### Ledger triage
All deferred/parked items from per-task reviews ruled OK-TO-DEFER (details in review record); none block merge. Spec deviations are all disclosed with recorded rulings (17 flips + 1 new-born list attr vs spec's "18"; EAV carve-out entered scope findings-first per the spec's own clause).

iceboundrock:
## Findings

### F1 — Plan's Task-3 STOP gate was overridden (plan-owner sanctioned, documented) — *spec/process*
The implementation plan Task 3 is an evidence gate: *"If `INCOMPATIBLE`, STOP the plan and report to the user as a blocking finding — do not commit regenerated files."* Task 3's evidence came back **`INCOMPATIBLE`** (Parquet tier: `VARCHAR` → `VARCHAR[]` across tiers, Case G proves it fires on the first pre-flip query with no window). The PR then **did** commit the regenerated files. The override is a documented plan-owner ruling (resolution B: full re-export in place, scoped to devo, via migration runbook), and the runbook is thorough. This is defensible — but as a review observation the gate was not honoured as written; the diff carries the metadata flip while the incompatible storage is remediated operationally, not by code. Residual gaps are correctly tracked in #371 (cdc-init only replaces base tier) and #372 (legacy scalar rows export as `[]`).

### F2 — `internal/` runtime changes exceed the plan's declared scope, but are justified by ruling 2 — *spec/process*
Plan: *"No changes to `internal/` runtime code are planned. If the compatibility investigation forces any, they come back as findings first."* This PR adds the relation-root required-policy carve-out across `internal/transform/attribute_converter.go`, `persistent_record.go`, `relation_roots.go`, `transformer.go` and `internal/entity_manager.go`. Trigger is not the compatibility investigation but ruling 2: expanding `contactSnapshot` via `$ref` created `contactSnapshot.isAnonymous` with `required_if_parent_present` beneath a relation root, which would 400 payloads #314 ruled acceptable. The implementation mirrors #314's existing `RelationRoots.Covers` and correctly excludes `_full` variants (verified: `x-relation` present in `visit.json` only, `visit_full.json` 0). It is a faithful, well-pinned carve-out — but the "findings first" disclosure step is not visible in git history before these landed; it was folded into the PR body as a plan-owner ruling.

### F3 — Plan expected-delta table imprecise on visit/visit_full list counts — *doc, minor, reasoning sound*
Plan Step 6 expected `visit_attributes: 1 text→list`, `visit_full_attributes: 3 text→list`. Actual: `visit_attributes` has **2** list entries (`attendees` flip + **new** `contactSnapshot.phones` at ID 45), `visit_full_attributes` has **2** flips (`attendees`, `contactSnapshot.phones`) + the **orphaned `logs`** (ID 29, `text`) preserved because `logs` was dropped from `visit_full.json` before attribute regeneration. I independently verified: `logs` is present at ID 29/`text`, and total list entries = 7+7+2+2 = **18**, consistent with ruling 1 and the attributeID invariant. The PR body's "2 flips, not 3" explanation is correct; only the plan's per-file table was imprecise.

### F4 — Migration/deploy risk is real and operator-dependent — *observation*
The Parquet-tier flip (F1) requires the 5-step runbook executed exactly, in devo, before deploy; step 2's manual delta purge is the classic footgun (#371). `_full` write-contract changes (`lead_full` `contact.isAnonymous` now `required_always`; `visit_full` `contactSnapshot.isAnonymous` `required_if_parent_present`) can 400 pre-existing rows on read (incl. Get-before-Update). These are loud, classified failures — by design — but they are client-visible changes riding inside a tools-fix PR.

### F5 — `FromEAVRecords` now returns nondeterministic missing-attribute error — *pre-existing, carried into refactor*
The new `checkRequiredAttributes` retains the old loop-and-return over `missingRequired` map (`attribute_converter.go:256-273`): which missing attribute is named first in the error varies between runs. Carried over verbatim from pre-PR code; the split into `checkRequiredAttributes` was the opportunity to make it deterministic (e.g., sort keys) and it was not taken. Cosmetic, not correctness.

iceboundrock:
## Spec

**Spec source:** issue #315, comment "Design: generate-attributes `$ref` resolution + regeneration guards" (approved design, issuecomment-5207776413) + the implementation-plan comment (issuecomment-5207828511). Reviewed against both, on the merged state.

### Requirements — satisfied
- **Generator `$ref` fix (Design §1):** `generateAttributesJSON` now loads via `NewSchemaInliner(...).InlineFile(...)` instead of raw `os.ReadFile`+`Unmarshal` (`generate_attributes.go:83-87`). Reuses the existing resolver — no second implementation. ✔ Matches.
- **Scalar `$ref` semantics (§1):** `TestGenerateAttributesScalarRefUsesTargetKeywords` pins `format: date-time` → `date`. ✔
- **ENOENT `-init` flag (Design §2 / plan Task 2):** fail-closed default; missing output + no `-init` = error naming `-init` and "renumber"; refusal creates no file. Verified `TestRunGenerateAttributesRefusesMissingOutputWithoutInit` passes. ✔ Matches plan Task 2 Step 3 exactly (including the error string).
- **One-time repairs (Design §3):** all three `_full.json` regenerated via `inline-schema`; `lead` (7), `lead_full` (7), `visit` (2), `visit_full` (2) attributes files regenerated; the other 8 committed `*_attributes.json` stay byte-identical. ✔
- **attributeID invariant (absolute):** PR's own checker reports zero ID changes across all 12 files; my count check confirms new IDs strictly append (visit max 28 → new 29..48; visit_full max 41 → 42..47). ✔
- **Guard 1 — attributes regenerate byte-identically, seed-first, iterate discovered files, zero-file guard:** `TestShippedAttributesRegenerateByteIdentical` implements all four plan requirements. ✔ (red before repair, green after — verified on merged main).
- **Guard 2 — `_full` pinned to `inline-schema` output with identical `MarshalIndent(result, "", "  ")` encoding:** `TestShippedFullSchemasMatchInlineOutput` matches plan; I confirmed the encoding is the same as `runInlineSchema` uses. ✔
- **Mutation verification:** Mutants 1 & 2 documented with real red output proving both guards detect the exact failure modes. ✔
- **User rulings 1–4:** implemented as binding — list repair (17 corrections + 1 new list attr = 18 list-typed), contactSnapshot expansion with preserve-and-demote, `_full` repair + second guard, `-init` flag. No re-litigation. ✔

### Divergences / partial
1. **Task-3 STOP gate overridden** — see Findings F1. The plan said STOP on `INCOMPATIBLE`; plan owner ruled B (remediation + ship) instead. *Deviation from plan-as-written; owner-sanctioned.*
2. **`internal/` runtime changes** — plan said none planned; see Findings F2. *Scope expansion; justified by ruling 2; disclosed.*
3. **visit/visit_full list counts** — plan table said 1/3, actual 2/2 + orphaned `logs`; see Findings F3. *Table imprecision; aggregate unchanged.*

### Verification obligations (plan §5)
- text→list stored-row compatibility: **performed**, evidence in PR body (throwaway DuckDB tests, Cases A–I, byte-identical EAV proof). Verdict `INCOMPATIBLE` → resolved via runbook. Not reproducible from the diff (throwaway tests deleted), but that matches the plan's evidence-in-report requirement.
- Mutation verification of the attributes guard: **performed**, evidence in PR body.
- `make lint` / `make test`: claimed green; not independently re-run here (I ran the targeted guards and tests only).

iceboundrock:
## Standards

Standards sources: `coding-standard.md`, `AGENTS.md`, `docs/error-handling.md`.

### Pass
- **Line limits (`coding-standard.md` §1):** all changed `.go` files ≤500 lines. `attribute_converter.go` 484, `transformer.go` 485, `persistent_record.go` 447, `entity_manager.go` 238, `generate_attributes.go` 407; new test files 124–280. `checkRequiredAttributes` ~56 lines and `FromEAVRecords` ~64 — both under the 100-line function cap. Note: `attribute_converter.go` and `transformer.go` are now ~15 lines from the 500 cap — future edits there will need refactoring.
- **Error-class split (`docs/error-handling.md`, AGENTS.md):** the plain (non-sentinel) error from `checkRequiredAttributes` correctly stays plain — it fires on the read path (persisted drift) and must not surface as a verbatim 400. The long comment preserving the reasoning from the #301/#313 history is exactly right. ✔
- **Error wrapping (`coding-standard.md` §3):** `fmt.Errorf("inline schema %s: %w", ...)`, `stat output file %s: %w`, `resolve schema name for id %d: %w`, `resolve relation roots ...: %w` — all wrap with context. ✔
- **Doc comments:** exported `RelationRootsLookup`, `RelationRootsAware`, `SetRelationRoots`, and the wiring comment in `entity_manager.go` are clear and explain *why* (wiring-time install, nil semantics, the #314 gap). Private helpers `relationRootsFor`/`newConverter` commented. ✔
- **Flat structure / nesting (`coding-standard.md` §7):** `generateAttributesJSON`'s Stat-block and `firstDivergence` are flat; early returns throughout. ✔

### Judgement calls
- **`return nil, err` at `attribute_converter.go:212`** — bare return of `checkRequiredAttributes`' error. `coding-standard.md` §3 says never bare `return err`; the callee error is already fully contextual (names the attribute + attrID), so a further wrap is redundant. Matches the pre-existing pattern at `:161` (`GetSchemaMetadata`). *Letter of the rule says hard; spirit says acceptable — judgement call.*
- **`RelationRootsAware` naming** (`relation_roots.go:44`) — `coding-standard.md` §4 prescribes `<Noun>er`/`<Noun>Service`. `...Aware` is idiomatic Go but neither shape. *Judgement call.*

### Observations (not violations)
- **Nondeterministic error selection** in `checkRequiredAttributes` (map iteration) — see Findings F5. Refactor missed the chance to fix pre-existing behaviour.
- `entity_manager.go:115-116` relies on `RelationIndex.RelationRoots` being nil-receiver-safe (`relation_index.go:172-175`) when `SchemaDirectory` is empty or load fails — safe, but the nil-safety is implicit; a one-line comment there would help.
- The `-init` flag check happens before inlining but the README example drops the `-out` from the regeneration example; minor doc nit only.

iceboundrock:
## Assessment

| Axe | Result |
|---|---|
| **Standards** | Pass — no hard violations; 2 judgement calls (bare `return nil, err`, `RelationRootsAware` naming); error-class split respected. |
| **Spec** | Pass with 3 documented deviations (Task-3 gate override, `internal/` scope expansion, plan-table count imprecision) — all owner-sanctioned or soundly reasoned, none defects in the shipped code. |

**Does it work?** Yes. The core fix (inline before traversal) resolves the #315 defect class exactly as specified, and the two permanent guards are real, sensitive guards (the PR's own mutant evidence + my independent green run confirm they catch the exact failure mode and are not tautologies). The relation-root carve-out is the correct mirror of #314 and is pinned in both directions plus the `_full` negative.

**Is it safe to ship?** The code is safe; the *deployment* requires the runbook. The Parquet-tier type flip (F1/F4) is the load-bearing risk: without purging delta+base and re-initing, the first federated query fails loudly. That failure mode is by design and disclosed, and it is correctly treated as operational, not code. The `_full` write-contract changes (`lead_full` `contact.isAnonymous` required; `visit_full` `contactSnapshot.isAnonymous` required-if-parent) are client-visible and could 400 rows that previously passed — disclosed and pinned, but consumers must be told.

**Worth-flagging for future work:** `FromEAVRecords`/`checkRequiredAttributes` returning nondeterministic first-missing-attribute; `attribute_converter.go`/`transformer.go` sitting ~15 lines under the 500 cap; the positional attributeID hazard itself is now contained but not removed (tracked in #370).

**Overall: good engineering, exceptionally well-documented evidence trail, appropriate risk posture. Merge is justified.**

iceboundrock:
## Observations

1. **The `visit_full` "2 flips not 3" reasoning is verified correct.** `visit_full_attributes.json` has exactly 2 list entries (`attendees` ID 3, `contactSnapshot.phones` ID 20), plus the preserved orphaned `logs` at ID 29/`text` — because `logs` was removed from `visit_full.json` (visit.json no longer declares it) *before* attribute regeneration, the generator's preserve-removed-attributes semantics keep it. Ordering (`_full` bodies first) is what makes this correct; a reviewer could easily mistake the preserved `logs` for drift.

2. **The guard sensitivity is asymmetric and that is correct.** Mutant 1 (revert inliner) is caught on `visit` by a 53-byte difference in `required_policy` markers — *not* by attributes vanishing — because without inlining the generator preserves the orphaned `contactSnapshot.*` entries. Byte-exactness is what makes the guard strong; a semantic diff would have missed it. This is a subtle and impressive property of the test design.

3. **`NewEntityManager` relies on a nil-receiver-safe method.** `aware.SetRelationRoots(relationIdx.RelationRoots)` at `entity_manager.go:115-116` is only safe because `(*RelationIndex).RelationRoots` guards `idx == nil` (`relation_index.go:172`). If `SchemaDirectory` is empty or `LoadRelationIndex` errors, `relationIdx` is nil and the field access still works — but this is implicit. Worth a comment or an explicit nil-guard for future readers.

4. **The `else if relationIdx != nil` warning branch can never fire for a nil index** — it fires when the *transformer* is not `RelationRootsAware` while an index exists. That's the intended loud-failure for a lossy decorator (and `writeSpy` was fixed to forward `SetRelationRoots`), good. But note the inverse (index nil + aware transformer) silently installs an empty lookup — which is exactly the no-op the tests pin.

5. **Migration runbook step 2 (manual delta purge) is the highest-risk operational step.** #371 already tracks that `cdc-init` only replaces the base tier; the runbook correctly calls out the delta purge as manual. Anyone deploying this without following step 2 verbatim hits the designed loud failure — the PR is explicit about this, which is the right call.

6. **`_full` write-contract changes ride inside a tools-fix PR.** `lead_full`/`visit_full`/`log_full` payload requirements changed in ways a client can observe (400s). The PR body discloses them precisely (the "`_full` write-contract changes" section), but they are easy to miss under the `$ref`-fix headline. No change requested — flagging for the release notes.

7. **Cosmetic:** the plan's expected-delta table (1/3 flips) vs actual (2/2 + orphan) is reconciled in the PR body's "Note on `visit_full_attributes.json`" — good catch, but the plan table itself was written against a stale artifact. A future plan should state deltas as "regenerate and review" rather than exact counts.

8. **Testing strength is high overall:** 5 new test files, both mutation vectors demonstrated, `-init` refusal verified to create no file, `_full` negative control (`TestFromEAVRecordsEnforcesRequiredPolicyOnFullVariant`) proving the carve-out doesn't widen. This is above the repo's typical bar.

iceboundrock:
## Verification

Independent re-runs on the merged `main` (HEAD `c5bb715`), sandbox-safe env per AGENTS.md.

### Guard tests + new behaviour — all green
```
$ GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./cmd/tools \
    -run 'TestShippedAttributesRegenerateByteIdentical|TestShippedFullSchemasMatchInlineOutput|TestGenerateAttributesExpandsObjectRef|TestGenerateAttributesScalarRefUsesTargetKeywords|TestRunGenerateAttributesRefusesMissingOutputWithoutInit' -v
--- PASS: TestShippedAttributesRegenerateByteIdentical  (12/12 subtests: activity, activity_full, communication, communication_full, contact, contact_full, lead, lead_full, log, log_full, visit, visit_full)
--- PASS: TestShippedFullSchemasMatchInlineOutput        (6/6 subtests)
--- PASS: TestGenerateAttributesExpandsObjectRef
--- PASS: TestGenerateAttributesScalarRefUsesTargetKeywords
--- PASS: TestRunGenerateAttributesRefusesMissingOutputWithoutInit
```
Note: I ran the `$ref`-resolution tests and the two guards but not the `-init` success-path / merge tests in this invocation; the targeted set above is the load-bearing one.

### List-flip counts re-derived from committed files
```
lead_attributes       list count= 7   (communications 2, contact.phones 19, requirement.mustHave 54,
                                       requirement.niceToHave 55, requirement.unwantedConditions 64, tags 72, visits 76)
lead_full_attributes  list count= 7   (identical set/IDs)
visit_attributes      list count= 2   (attendees 3, contactSnapshot.phones 45 [NEW])
visit_full_attributes list count= 2   (attendees 3, contactSnapshot.phones 20)
```
Aggregate = 18 list-typed entries (17 corrections + 1 new-born `contactSnapshot.phones`), matching the PR body's "precise counts" and the plan's total. 7+7+2+2 = 18. ✔

### Orphaned `logs` + relation-root claims
- `visit_full_attributes.json["logs"]` = `{"attributeID": 29, "valueType": "text"}` — preserved orphan, confirming the "2 flips not 3" explanation. ✔
- `x-relation` occurrences: `visit.json` = 1, `visit_full.json` = 0 — confirms the carve-out reaches plain `visit` but not the `_full` variant (inlining strips `x-*`). ✔

### attributeID invariant (spot-check)
New IDs strictly append: `visit_attributes.json` max old ID 28, new range 29..48; `visit_full_attributes.json` max old 41, new 42..47. Existing IDs (contactSnapshot scalar 25, stale 21–24, `logs` 29) unchanged. ✔

### Not re-run here
`make lint`, full `make test`, `make test-e2e-production` (123 tests), `gofmt -l`, `go vet` — claimed green in the PR body and not disputed by this review. The production harness was appropriately run because `ff905dc` touches the runtime write path; the federated `-tags=e2e` suite was correctly left to CI (this change's runtime surface is the write path).

### Confidence in the PR body's evidence
Task-3's DuckDB evidence (Cases A–I, `Conversion Error` on first non-NULL legacy value, compaction cannot heal, NULL-column blast-radius bound) is the strongest part of this PR's evidence trail. The throwaway tests are deleted (repo clean, per PR), so it cannot be re-run from the diff — but the claims are internally consistent, cite exact file:line, and match DuckDB's documented cast semantics. I did not find any contradiction with the committed code.

iceboundrock:
## Summary

**Review of PR #373 (`fix/315-generate-attributes-ref`, merge `c5bb715`) — reviewed on the merged state against the approved design + implementation plan on issue #315.**

**Verdict: solid, approve.** The `$ref` fix, the `-init` fail-closed flag, both permanent byte-identical guards, the one-time artifact repairs, and the relation-root required-policy carve-out all match the approved spec, and I independently re-ran the key gates green on the merged `main`.

- **2 spec-process deviations, both plan-owner-sanctioned and documented in the PR body**, none blocking:
  1. The Task-3 STOP gate ("If `INCOMPATIBLE`, STOP") was overridden by plan-owner ruling B (full re-export in place) — shipped together with a migration runbook instead of blocking.
  2. `internal/` runtime changes were made despite the plan's "no changes to `internal/` runtime code" note — forced by ruling 2 (contactSnapshot expansion) colliding with #314's required-policy enforcement; functionally correct and disclosed as a plan-owner ruling.
- **1 spec-table inaccuracy** (visit/visit_full list-flip counts vs the plan's expected-delta table) with sound reasoning in the PR body; aggregate stays 18 and the attributeID invariant holds.
- **No hard coding-standard violations.** One judgement-call bare `return nil, err` carried into the refactor; one pre-existing nondeterministic map-order error return that the refactor had a chance to fix and did not.
- **Verification (mine, on merged main):** both guards green across all 18/6 subtests, `-init` refusal test passes, and I re-derived the list-flip counts and the orphaned `logs` entry independently.

Worst single issue: none blocking. The residual operational risks (Parquet-tier type conflict, `_full` write-contract changes) are disclosed, pinned, and tracked in #370/#371/#372.

iceboundrock:
Responses to the post-merge review, finding by finding.

**F1 (Task-3 STOP gate) — clarification, no change.** The gate fired exactly as written: Task 3 returned `INCOMPATIBLE`, execution stopped with nothing regenerated, the evidence went to the plan owner as a blocking finding, and ruling B was taken **before** Task 4 ran. What the review correctly observes is that this stop→ruling→resume sequence is not visible in git or on GitHub — it lived in the working session's ledger. That is a real observability gap, and it also covers F2's "not visible in git history before these landed". **Adopted going forward: mid-execution plan-owner rulings get posted to the issue at ruling time**, not reconstructed in the PR body afterwards.

**F2 (internal/ scope expansion) — clarification, no change.** Same shape as F1: the carve-out entered findings-first — Task 4 stopped on the red `TestCreateAcceptsDottedKeyUnderRelationRoot` pin, escalated it as a blocker, and `ff905dc` landed only after the ruling. The plan's "come back as findings first" clause was followed in process; the visibility gap is the same one as F1 and gets the same fix.

**F3 (plan expected-delta table) — adopted.** The table was written against measurements that could not see the `_full`-first regeneration ordering's knock-on effects (orphaned `logs`, `contactSnapshot.phones` born as list). Observation 7's prescription is right: future plans state artifact deltas as "regenerate and review against invariants" (ID invariant + guard) rather than exact per-file counts.

**F4 (migration risk) — pointer added.** No code change requested and none made; the runbook remains the deploy gate. A pointer to the PR-body runbook is now posted on #371, since anyone touching cdc-init's base-only behavior is the same operator who must run step 2's manual delta purge.

**F5 (nondeterministic missing-attribute error) — fixed in #374.** `checkRequiredAttributes` now names the alphabetically first missing attribute (`slices.Min`), pinned by a 25-run repeat test that reds under map-order selection (mutation-verified), and the pre-existing either/or assertion — itself evidence of the nondeterminism — is tightened to the deterministic value. Observation 3's nil-safety comment rides in the same PR.

**Standards judgement call: bare `return nil, err` at the `checkRequiredAttributes` call site — keeping as is.** The callee's error is fully self-contextual (names the attribute, attrID, and "EAV records"); an added wrap would restate the same facts, and the shape matches the adjacent `GetSchemaMetadata` return at `attribute_converter.go:161`. With the file now at 491/500 lines, the next change there forces an extraction — that sweep is the right moment to revisit both call sites together.

**Standards judgement call: `RelationRootsAware` naming — keeping as is.** It is an optional capability interface asserted via type assertion at wiring time, not a service: `-er`/`-Service` shapes claim behavior ("does X"), while this type claims receptivity ("can accept an install"). `-Aware` is the established Go idiom for that shape, and renaming would churn five files for no clarity gain. If optional-capability interfaces should be bound by §4's naming rule, that reads as a `coding-standard.md` amendment rather than a rename of this type.

PR #374, opened 2026-08-07
review: #373 deterministic missing-attribute error + nil-safety comment

iceboundrock:
Follow-up resolving the code-level items from PR #373's post-merge review (https://github.com/Lychee-Technology/forma/pull/373#issuecomment-5212927307).

## F5 — nondeterministic missing-attribute error (resolved)
`checkRequiredAttributes` returned whichever missing attribute Go's map iteration yielded first. It now names the **alphabetically first** via `slices.Min` — same message format, deterministic selection.

- Pinned by `TestAttributeConverterFromEAVRecords_MissingRequiredErrorIsDeterministic` (4 missing candidates, 25 repeat runs — mutation-verified: reinstating map-order selection fails run 0 with `got 'delta'`).
- The pre-existing either/or assertion in `TestAttributeConverterFromEAVRecords_NestedArrayRequiredUsesFullIndexPath` — itself evidence of the nondeterminism — is tightened to the deterministic value.

## Observation 3 — implicit nil-receiver safety (resolved)
`NewEntityManager`'s `aware.SetRelationRoots(relationIdx.RelationRoots)` now carries a two-line comment noting the call is safe when `relationIdx` is nil (nil-receiver-safe method, installs an empty lookup).

## Not changed (responses on PR #373)
Bare `return nil, err` at the `checkRequiredAttributes` call site and the `RelationRootsAware` naming — reasoning posted on the review thread.

Note: `attribute_converter.go` is now 491/500 lines — the next change there must extract (already flagged in the #373 review).

Gates: `make lint` clean, `make test` all packages ok, gofmt clean.

Refs #315.

### DIFF
--- a/internal/entity_manager.go
+++ b/internal/entity_manager.go
@@ -113,6 +113,8 @@ func NewEntityManager(
 	// EAV required-policy check enforces policies on relation-root children,
 	// which #314 ruled must never gate a write (#315).
 	if aware, ok := transformer.(transform.RelationRootsAware); ok {
+		// Safe when relationIdx is nil: RelationRoots is nil-receiver-safe
+		// (relation_index.go) and installs an empty lookup.
 		aware.SetRelationRoots(relationIdx.RelationRoots)
 	} else if relationIdx != nil {
 		// Loud, not silent: a transformer decorator that does not forward the
--- a/internal/transform/attribute_converter.go
+++ b/internal/transform/attribute_converter.go
@@ -253,25 +253,32 @@ func (c *AttributeConverter) checkRequiredAttributes(
 	}
 
 	zap.S().Infow("missing EAV records for attrIDs.", "idToName", missingRequired)
-	for missingAttrID, missingAttrName := range missingRequired {
-		// Plain error, deliberately. FromEAVRecords is not write-only: the read
-		// path rebuilds already-stored records through it
-		// (persistent_record.go's FromPersistentRecord), so a persisted row
-		// missing a required EAV row reaches here too. Wrapping
-		// forma.ErrInvalidInput here — as an earlier #301 sweep did — made the
-		// HTTP boundary answer that persisted-drift case with a verbatim 400,
-		// inverting the split AGENTS.md and this repo's error-handling doc
-		// draw: write validation carries the sentinel, read-path consistency
-		// failures stay plain and operator-visible.
-		//
-		// The write path's 400 does not depend on this wrap. It has its own
-		// write-only validator, validateRequiredAttributesFromInput
-		// (transformer.go), which ToAttributes runs against the caller's input
-		// before flattening and which does carry the sentinel.
-		return fmt.Errorf("missing required attribute '%s' (attrID=%d) in EAV records",
-			missingAttrName, missingAttrID)
+	names := make([]string, 0, len(missingRequired))
+	idsByName := make(map[string]int16, len(missingRequired))
+	for id, name := range missingRequired {
+		names = append(names, name)
+		idsByName[name] = id
 	}
-	return nil
+	// Name the alphabetically first missing attribute, not whichever the map
+	// yields first, so the same drift produces the same error on every run.
+	missingAttrName := slices.Min(names)
+
+	// Plain error, deliberately. FromEAVRecords is not write-only: the read
+	// path rebuilds already-stored records through it
+	// (persistent_record.go's FromPersistentRecord), so a persisted row
+	// missing a required EAV row reaches here too. Wrapping
+	// forma.ErrInvalidInput here — as an earlier #301 sweep did — made the
+	// HTTP boundary answer that persisted-drift case with a verbatim 400,
+	// inverting the split AGENTS.md and this repo's error-handling doc
+	// draw: write validation carries the sentinel, read-path consistency
+	// failures stay plain and operator-visible.
+	//
+	// The write path's 400 does not depend on this wrap. It has its own
+	// write-only validator, validateRequiredAttributesFromInput
+	// (transformer.go), which ToAttributes runs against the caller's input
+	// before flattening and which does carry the sentinel.
+	return fmt.Errorf("missing required attribute '%s' (attrID=%d) in EAV records",
+		missingAttrName, idsByName[missingAttrName])
 }
 
 // Helper functions for conversion
--- a/internal/transform/attribute_converter_test.go
+++ b/internal/transform/attribute_converter_test.go
@@ -277,8 +277,47 @@ func TestAttributeConverterFromEAVRecords_NestedArrayRequiredUsesFullIndexPath(t
 	if !strings.Contains(err.Error(), "missing required attribute '") {
 		t.Fatalf("unexpected error: %v", err)
 	}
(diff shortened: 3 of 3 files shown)


iceboundrock:
## Summary

Small, well-scoped follow-up resolving the two code-level items from PR #373's post-merge review — **F5** (nondeterministic missing-attribute error) and **Observation 3** (nil-receiver safety documentation). 3 files, +68/−20.

The core change is a strict improvement: `checkRequiredAttributes` now names the **alphabetically first** missing attribute (`slices.Min`) instead of whichever the `missingRequired` map happened to yield first, so the same drift produces the same error on every run. The behavior change is safe — selection was nondeterministic before, so no consumer could have depended on a specific name.

Verdict: **approve / ready to merge**. No blocking findings. See findings / spec / standards / assessment / observations / verification below.

iceboundrock:
## Findings

- **F1 — resolved, verified.** The #373 review's F5 is addressed: the loop-and-return over `missingRequired` is replaced by deterministic selection of the alphabetically first missing attribute via `slices.Min`. Correctness holds because names are unique (the cache is keyed by attribute name), so the reverse `idsByName` lookup is unambiguous, and the `len(missingRequired) == 0` early return guarantees `names` is never empty.
- **F2 — resolved, verified.** Observation 3 is addressed: the two-line comment at `internal/entity_manager.go:116-117` accurately documents the nil-receiver-safe wiring. Independently confirmed the claim:
  - `RelationIndex.RelationRoots` guards `idx == nil` and returns nil (`relation_index.go:172-175`);
  - `RelationRoots.Covers` short-circuits on an empty set (`relation_roots.go:27-30`);
  - `relationRootsFor` returns a nil set when no lookup is installed (`attribute_converter.go:42-50`).
  So `aware.SetRelationRoots(relationIdx.RelationRoots)` with a nil `relationIdx` installs a lookup that answers nil for every schema — i.e. "an empty lookup", exactly as the comment says.
- **No new findings.**

iceboundrock:
## Spec

- This PR is a code-level follow-up to the #373 post-merge review, not a spec-bearing feature: it resolves the two action items (F5, Observation 3) and explicitly responds on the two items intentionally not changed (the bare `return nil, err` at the `checkRequiredAttributes` call site and the `RelationRootsAware` naming). All four dispositions match the review thread.
- `Refs #315` is the correct trace-back — #315 is closed by #373 and the F5 refactor lived in that change; no open issue needs re-pointing.
- One spec-adjacent nuance worth recording: the deterministic error message is a **user-visible message-content change** (a client could previously observe *any* missing attribute named; now always the alphabetically first). Because the previous selection was nondeterministic, no consumer could have depended on a specific name — this is a message stabilization, not a contract break.

iceboundrock:
## Standards

- **`attribute_converter.go` is now 491/500 lines** — 9 lines of headroom. The PR body flags this and the #373 review already required extraction; agreed, and it stays inside the rule so it is compliant. But the margin is effectively zero: **the next change to this file must extract** (e.g. the `FromEAVRecords`→`checkRequiredAttributes` split has room to go further, or the required-policy helper block could move to `relation_roots.go`). No action required in this PR.
- `checkRequiredAttributes` is 63 lines (incl. signature/brace) — within the 100-line function rule.
- Error semantics unchanged and correct: still a **plain** read-path consistency error (no `forma.ErrInvalidInput` wrap), per `docs/error-handling.md` and AGENTS.md. The explanatory comment block was preserved rather than buried. No bare sentinel wraps introduced; `TestNoBareSentinelWraps` passes.
- The new entity_manager comment anchors to its source (`relation_index.go`) — consistent with the repo's location-anchoring convention.
- Test style: the repeat-loop red anchor with a `run %d` failure message is a good pattern. Minor: the deterministic test pins the literal name `'alpha'`, so its meaning depends on the stub's alphabetical ordering being deliberate — the in-test comment already explains this, which is sufficient.

iceboundrock:
## Assessment

- **Correctness.** `slices.Min` over the missing attribute names yields the alphabetically first; the reverse `idsByName` map is unambiguous since names are unique. The tightened either/or assertion in `TestAttributeConverterFromEAVRecords_NestedArrayRequiredUsesFullIndexPath` is valid: `"contact.phones.kind"` sorts before `"orders.items.price"` (`'c'` < `'o'`), so `attrID=6` is the deterministic answer.
- **Pin quality.** The new `TestAttributeConverterFromEAVRecords_MissingRequiredErrorIsDeterministic` is a genuine red anchor, not a tautology — verified by mutation (see verification).
- **Regression risk.** Low. The change is localized to error-message selection and one doc comment; no storage, SQL, or validation semantics change. The pre-existing either/or test assertion — itself the smoking gun of the nondeterminism — now asserts the stabilized value.
- **Minor inefficiency** (two passes + a temp map + a sort allocation) is confined to the error path — irrelevant.
- Overall: **mergeable as-is.**

iceboundrock:
## Observations

- **Message-content stabilization, not a regression.** The error text may name a *different* attribute than a given prior run did, because the old selection was map-order. This is the point of the change; worth noting for anyone who saw a specific missing-attribute name in an earlier incident report.
- **Operators keep the full picture.** The `zap.S().Infow("missing EAV records for attrIDs.", "idToName", missingRequired)` log still dumps the whole map, so the log names all missing attributes while the error names one — good split.
- **`idsByName` reverse-map.** Could be avoided by collecting `(name, id)` pairs together and sorting, but the current form is clear and the cost is confined to the error path. Cosmetic.
- **Carried-over debt:** the `attribute_converter.go` 500-line ceiling (see standards) is the standing extraction debt already tracked from #373's review; still no issue filed for it as far as this PR shows. Suggest filing a follow-up issue so it does not get lost.

iceboundrock:
## Verification

Ran against branch head `1b0cf5c` (worktree, mirroring the Makefile's `GOCACHE`/`GOFLAGS` env):

- `go test ./internal/transform ./internal -count=1` — **ok** (both packages).
- Targeted transform tests, incl. `TestAttributeConverterFromEAVRecords_MissingRequiredErrorIsDeterministic` (25/25 runs) and the tightened `..._NestedArrayRequiredUsesFullIndexPath` — **PASS**.
- `go vet ./internal/transform ./internal` — clean.
- `gofmt -l` over the 3 changed files — empty.

**Mutation check (red anchor):** reinstated the old map-order selection (`missingAttrName := names[0]`) →

```
--- FAIL: TestAttributeConverterFromEAVRecords_MissingRequiredErrorIsDeterministic
    run 0: expected deterministic 'alpha', got missing required attribute 'delta' (attrID=2)
```

matching the PR body's claim exactly. Reverted; tree clean and green.

Not re-run here: full `make lint` / `make test` (PR reports clean; CI will confirm). The touched packages are the two verified above.

iceboundrock:
Review dispositions:

- **Extraction-debt follow-up (observations)** — filed as #375: the 491/500 debt plus the deferred bare-`return` judgement calls now live there, so the sweep has a home and does not depend on anyone remembering the review threads.
- **`idsByName` reverse map / error-path allocations / literal `'alpha'` pin** — left as is per the review's own labeling (cosmetic; the in-test comment covers the pin's intent).
- Nothing else is actionable: F1/F2 verified resolved, no blocking findings.

Awaiting your merge per repo convention.

PR #376, opened 2026-08-07
fix(schemameta): #342 guard against attributeID reuse after attribute removal

### LINKED ISSUE #342: Guard against attributeID reuse after attribute removal (cross-generation)

iceboundrock:
Closes #342.

Formalizes `<schema>_attributes.json` as the **attributeID ledger** so an id freed by attribute removal can never be rebound to a different attribute — closing the trap #294's tolerate-and-preserve contract invited (same valueType → preserved values silently surface under the new name; different valueType → storage-type mismatch → row unreadable/500).

## ⚠️ Breaking change — read before merging

Two shipped ledgers carried pre-existing removal residue that the new generator behavior retires. **The two are not symmetric:**

| Attribute | Schema | Impact |
|---|---|---|
| `logs` (attrID 29) | `visit_full` | **Write-breaking.** `visit_full.json` declares no `logs` and no `additionalProperties`, so on `main` a client *can* write `logs` (array or scalar) and read it back. After this PR writes fail 400 `attribute 'logs' is not defined for this schema`, and existing values disappear from reads. Attribute 29 is effectively **unrestorable under its original array shape** — the docs now say so and give the workable alternative (new name, fresh id) instead of the remedy that does not work. |
| `contactSnapshot` (attrID 25) | `visit` | **Practically inert.** A pre-#315 `$ref`-collapse leftover; an object payload recurses per key and never reaches the bare leaf, so rows under id 25 could only exist if a client once sent it as a scalar (which #314 validation now rejects). |

EAV rows for both are preserved per #294 and both ids stay reserved. Retiring them is what the ledger contract requires; the alternative — leaving them unmarked — makes `TestShippedAttributesRegenerateByteIdentical` permanently red.

**Please confirm no client writes `visit_full.logs` before merging.** The remedy is cheap beforehand and awkward after (see the flush caveat below).

## Approach (per the rulings recorded on #342)

- **Mechanism: file ledger**, not a Postgres history table and not an EAV storage probe. The EAV table stores only `value_text`/`value_numeric` — no name, no type — so a storage probe could catch a cross-column type change but never a same-type silent rename. The ledger catches both.
- **Severity: hard error** at registration/startup, matching #314's fail-closed precedent.
- **Scope: guard only.** #341 (validator classification) and #343 (Warn demotion) stay separate; the retired marker is exactly the signal #341 needs.

## What changed

1. **`forma.AttributeMetadata.Retired bool`** + parser support for `"retired": true|false` (non-bool rejected).
2. **`generate-attributes`** marks schema-absent entries `"retired": true` and forces optional policy, instead of relying on humans never to hand-delete. Re-add of the same name with the same valueType+items_type clears the marker (values restore); a different type is a hard error naming attribute, old type/items, new type/items. `maxID` still scans retired entries, so freed ids are never reissued.
3. **`validateSchemaAttributeCache`** runs on the **full** cache at every registration path and emits reuse-specific errors when an attributeID or main-column collision involves exactly one retired entry. Retired entries also keep their folded parquet columns reserved (`sqlgen.ValidateParquetAttrColumns` still sees them) — those columns exist in already-flushed files.
4. **Retired entries are stripped from active caches** strictly *after* validation, at all four storage points. Downstream (transform, sqlgen, repositories, CDC) needs no changes: a retired attribute reads, writes, flushes and projects exactly as if its entry were absent, so #294 skip-on-read / preserve-on-write is unchanged.
5. **`MetadataCache.RegisterSchema` now validates** before snapshot+strip (ruled mid-flight): it was the only unvalidated registration door, and the strip would have silently swallowed a collision there. Consequence: #294's `TestReplaceEAVAttributes_DeleteScopedToCurrentSchemaAttrIDs` deliberately registered two names sharing attrID 12 to cover `knownAttrIDs`' `slices.Compact`; that state is now unconstructible through any path, so the test uses distinct ids (sort-determinism coverage preserved) and `slices.Compact` stays as belt-and-suspenders with an updated comment.
6. **Production e2e** `TestRetiredMarkerRemovalWorkflow` proves the retired-marker generation behaves identically to the hand-deleted one; the legacy `TestUpdateAfterAttributeRemoval` is untouched and still pins that old-style files are tolerated.
7. **Docs** — `docs/error-handling.md` and `docs/schema-consistency-migration.md` carry the blessed removal workflow, the guard's exact error shapes, re-add semantics, the rollback caveat, and the residual gaps.

## Operator caveat the docs now carry (inherited #294 semantics, previously over-promised)

Un-retiring restores values **only for rows not yet flushed to the lakehouse**. CDC derives its EAV pull from the active cache (`internal/cdc/export_sql_builder.go`), so a retired attribute's rows never reach parquet; a row flushed during the retired window reads NULL forever. The Postgres EAV row survives, so an unrelated later update makes the row hot again and the next flush *resurrects* the value — a value that read NULL yesterday reappears tomorrow. On CDC-enabled deployments treat retirement as effectively irreversible for flushed rows.

## Residual gaps (documented, not closed)

- **Live attribute type change is the same damage class, unguarded**: `generate-attributes` silently refreshes `valueType` for a still-present attribute, stranding existing EAV rows in the wrong physical column. Documented; follow-up candidate.
- **Hand-deleted ledger entries** (before or after #342) have no entry to guard; the rule stays "never hand-delete", which is convention, not enforcement.
- **Retired-in-ledger but still declared in `<schema>.json`** (old `tools` binary, or forgetting to regenerate) renders the attribute silently invisible at runtime; no cross-check today.
- **Rollback**: pre-#342 binaries ignore the `retired` key and load the entry as active — retired values become visible again, and the scoped DELETE resumes including those ids, so the next update of such a row destroys the preserved rows.

## Verification

`make test` green module-wide · `make lint` (golangci-lint v1.64.8) exit 0 · `gofmt` clean · e2e infra smoke green · both production removal e2e tests green under Docker · shipped-schema byte-identical regeneration guards green (12/12) · the `items_type` guard leg mutation-verified (deleting the leg reddens exactly the new test, nothing else).

Reviewed per task (spec + quality) plus a whole-branch review; findings, rulings and the deferred-minor triage are recorded on #342.

请不要自动合并。

### DIFF
--- a/cmd/server/schemas/visit_attributes.json
+++ b/cmd/server/schemas/visit_attributes.json
@@ -14,7 +14,8 @@
   },
   "contactSnapshot": {
     "attributeID": 25,
-    "valueType": "text"
+    "valueType": "text",
+    "retired": true
   },
   "contactSnapshot.annualIncome": {
     "attributeID": 29,
--- a/cmd/server/schemas/visit_full_attributes.json
+++ b/cmd/server/schemas/visit_full_attributes.json
@@ -134,7 +134,8 @@
   },
   "logs": {
     "attributeID": 29,
-    "valueType": "text"
+    "valueType": "text",
+    "retired": true
   },
   "nextFollowUpAt": {
     "attributeID": 30,
--- a/cmd/tools/generate_attributes.go
+++ b/cmd/tools/generate_attributes.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+
+	"github.com/lychee-technology/forma"
 )
 
 type attributeSpec struct {
@@ -127,7 +129,7 @@ func generateAttributesJSON(schemaPath, outputPath string, allowNew bool) error 
 	// Build the result. First, preserve ALL existing attributes (even ones the
 	// schema dropped): this file is the attributeID ledger (#342).
 	if err := mergeExistingAttributes(existingAttrs, newAttributes, result); err != nil {
-		return err
+		return fmt.Errorf("merge existing attributes from %s: %w", outputPath, err)
 	}
 
 	// Then, add new attributes from the schema
@@ -175,7 +177,7 @@ func mergeExistingAttributes(
 		}
 
 		if err := clearRetiredOnReAdd(name, existingData, spec); err != nil {
-			return err
+			return fmt.Errorf("re-add retired attribute: %w", err)
 		}
 		existingData["valueType"] = spec.ValueType
 		if spec.ItemsType != "" {
@@ -201,7 +203,7 @@ func clearRetiredOnReAdd(name string, existingData map[string]any, spec attribut
 
 	oldType, _ := existingData["valueType"].(string)
 	oldItems, _ := existingData["items_type"].(string)
-	if oldType != spec.ValueType || oldItems != spec.ItemsType {
+	if oldType != spec.ValueType || effectiveItemsType(oldItems) != effectiveItemsType(spec.ItemsType) {
 		return fmt.Errorf(
 			"attribute %s is retired with valueType %s/items_type %q but re-added as %s/%q: preserved EAV rows store the old type and would become unreadable; restore the original type or use a new attribute name",
 			name, oldType, oldItems, spec.ValueType, spec.ItemsType)
@@ -210,6 +212,18 @@ func clearRetiredOnReAdd(name string, existingData map[string]any, spec attribut
 	// Same name, same type: the re-add restores the preserved values.
 	delete(existingData, "retired")
 	return nil
+}
+
+// effectiveItemsType resolves an omitted items_type the way the runtime does:
+// forma.AttributeMetadata.EffectiveItemsType treats an empty items_type as
+// text. A legacy ledger entry stored as a bare `list` therefore describes the
+// same physical rows as one written `list`/`text`, and re-adding it must not be
+// refused for a difference that does not exist on disk (#342).
+func effectiveItemsType(itemsType string) string {
+	if itemsType == "" {
+		return string(forma.ValueTypeText)
+	}
+	return itemsType
 }
 
 // loadExistingAttributes reads an existing attributes file and returns its contents.
--- a/cmd/tools/generate_attributes.go
+++ b/cmd/tools/generate_attributes.go
@@ -124,24 +124,10 @@ func generateAttributesJSON(schemaPath, outputPath string, allowNew bool) error 
 		newIDMap[name] = maxID + i + 1
 	}
 
-	// Build the result
-	// First, preserve ALL existing attributes (even if they are removed from schema)
-	for name, existingData := range existingAttrs {
-		if spec, exists := newAttributes[name]; exists {
-			// Attribute still exists in schema: update valueType if changed
-			existingData["valueType"] = spec.ValueType
-			if spec.ItemsType != "" {
-				existingData["items_type"] = spec.ItemsType
-			} else {
-				delete(existingData, "items_type")
-			}
-			applyRequiredPolicy(existingData, spec.RequiredPolicy)
-		} else {
-			// Attribute no longer exists in schema path; it must not stay required.
-			applyRequiredPolicy(existingData, requiredPolicyOptional)
-		}
-		// Keep the attribute regardless of whether it exists in the new schema
-		result[name] = existingData
+	// Build the result. First, preserve ALL existing attributes (even ones the
+	// schema dropped): this file is the attributeID ledger (#342).
+	if err := mergeExistingAttributes(existingAttrs, newAttributes, result); err != nil {
+		return err
 	}
 
 	// Then, add new attributes from the schema
@@ -164,6 +150,65 @@ func generateAttributesJSON(schemaPath, outputPath string, allowNew bool) error 
 	}
 
 	fmt.Printf("Generated attributes total: %d, new: %d, maxID: %d\n", len(result), len(newAttrNames), maxID+len(newAttrNames))
+	return nil
+}
+
+// mergeExistingAttributes folds every existing ledger entry into result.
+// Entries still present in the schema get their type and required policy
+// refreshed (and a retired marker cleared when the re-add is type-compatible);
+// entries the schema dropped are branded retired so their attributeID stays
+// bound to the EAV rows it already owns instead of being handed to a later
+// attribute (#342), and forced optional so no reader requires a value that can
+// no longer be written.
+func mergeExistingAttributes(
+	existingAttrs map[string]map[string]any,
+	newAttributes map[string]attributeSpec,
+	result map[string]map[string]any,
+) error {
+	for name, existingData := range existingAttrs {
+		spec, stillInSchema := newAttributes[name]
+		if !stillInSchema {
+			existingData["retired"] = true
+			applyRequiredPolicy(existingData, requiredPolicyOptional)
+			result[name] = existingData
+			continue
+		}
+
+		if err := clearRetiredOnReAdd(name, existingData, spec); err != nil {
+			return err
+		}
+		existingData["valueType"] = spec.ValueType
+		if spec.ItemsType != "" {
+			existingData["items_type"] = spec.ItemsType
+		} else {
+			delete(existingData, "items_type")
+		}
+		applyRequiredPolicy(existingData, spec.RequiredPolicy)
+		result[name] = existingData
+	}
+
+	return nil
+}
+
+// clearRetiredOnReAdd un-retires an attribute the schema has re-added under
+// its original physical type, which restores the preserved EAV rows. A
+// re-add under a different type is refused: the stored rows carry the old
+// type and would become unreadable (#342).
+func clearRetiredOnReAdd(name string, existingData map[string]any, spec attributeSpec) error {
+	if retired, _ := existingData["retired"].(bool); !retired {
+		return nil
+	}
+
+	oldType, _ := existingData["valueType"].(string)
+	oldItems, _ := existingData["items_type"].(string)
+	if oldType != spec.ValueType || oldItems != spec.ItemsType {
+		return fmt.Errorf(
+			"attribute %s is retired with valueType %s/items_type %q but re-added as %s/%q: preserved EAV rows store the old type and would become unreadable; restore the original type or use a new attribute name",
+			name, oldType, oldItems, spec.ValueType, spec.ItemsType)
+	}
+
+	// Same name, same type: the re-add restores the preserved values.
+	delete(existingData, "retired")
 	return nil
 }
 
--- a/cmd/tools/generate_attributes_retired_test.go
+++ b/cmd/tools/generate_attributes_retired_test.go
@@ -83,6 +83,33 @@ func TestGenerateAttributes_ReAddDifferentTypeRejected(t *testing.T) {
 	}
 }
 
+// #342: a list attribute keeps its element type in items_type, so a re-add
+// that preserves valueType "list" but swaps the element type is just as
+// destructive as a scalar type change and must be refused too. This pins the
+// items_type leg of the guard, which the valueType cases cannot reach: both
+// sides here are "list", so only the items_type comparison can fail.
+func TestGenerateAttributes_ReAddDifferentItemsTypeRejected(t *testing.T) {
+	dir := t.TempDir()
+	schema := writeSchemaFixture(t, dir, "user.json",
+		`{"type":"object","properties":{"name":{"type":"string"},"old_list":{"type":"array","items":{"type":"number"}}}}`)
+	out := writeSchemaFixture(t, dir, "user_attributes.json", `{
+	  "name":     { "attributeID": 1, "valueType": "text" },
+	  "old_list": { "attributeID": 3, "valueType": "list", "items_type": "text", "retired": true }
+	}`)
+
+	err := generateAttributesJSON(schema, out, false)
+	if err == nil {
+		t.Fatalf("expected items_type-mismatch error on re-add")
+	}
+	// "text" is the retired element type, "numeric" the one the schema now asks
+	// for; naming both is what tells the operator which rows are at stake.
+	for _, want := range []string{"old_list", "text", "numeric"} {
+		if !strings.Contains(err.Error(), want) {
+			t.Fatalf("error %q does not name %q", err.Error(), want)
+		}
+	}
+}
+
 // #342: a retired entry still occupies its attributeID, so a genuinely new
 // attribute must be numbered above it rather than reusing the freed id.
 func TestGenerateAttributes_NewAttributeDoesNotReuseRetiredID(t *testing.T) {
--- a/cmd/tools/generate_attributes_retired_test.go
+++ b/cmd/tools/generate_attributes_retired_test.go
@@ -110,6 +110,36 @@ func TestGenerateAttributes_ReAddDifferentItemsTypeRejected(t *testing.T) {
 	}
 }
 
+// #342: a legacy ledger entry may record a list without any items_type, which
+// the runtime reads as text (forma.AttributeMetadata.EffectiveItemsType). The
+// re-add guard must compare effective types, not literal strings: refusing this
+// re-add would hand the operator an instruction they cannot satisfy, since the
+// generator never writes items_type "text" back as an empty field.
+func TestGenerateAttributes_ReAddListWithoutItemsTypeClearsRetired(t *testing.T) {
+	dir := t.TempDir()
+	schema := writeSchemaFixture(t, dir, "user.json",
+		`{"type":"object","properties":{"name":{"type":"string"},"old_list":{"type":"array","items":{"type":"string"}}}}`)
+	out := writeSchemaFixture(t, dir, "user_attributes.json", `{
+	  "name":     { "attributeID": 1, "valueType": "text" },
+	  "old_list": { "attributeID": 3, "valueType": "list", "retired": true }
+	}`)
+
+	if err := generateAttributesJSON(schema, out, false); err != nil {
+		t.Fatalf("re-add of a legacy items_type-less list must be accepted: %v", err)
+	}
+	attrs := readGeneratedAttributes(t, out)
+	oldList := attrs["old_list"]
+	if _, exists := oldList["retired"]; exists {
+		t.Fatalf("re-added attribute must lose the retired marker: %v", oldList)
+	}
+	if id, _ := oldList["attributeID"].(float64); id != 3 {
+		t.Fatalf("re-add must keep attributeID 3, got %v", oldList["attributeID"])
+	}
+	if items, _ := oldList["items_type"].(string); items != "text" {
+		t.Fatalf("re-add must record the now-explicit element type text, got %v", oldList["items_type"])
+	}
+}
+
 // #342: a retired entry still occupies its attributeID, so a genuinely new
 // attribute must be numbered above it rather than reusing the freed id.
 func TestGenerateAttributes_NewAttributeDoesNotReuseRetiredID(t *testing.T) {
--- a/cmd/tools/generate_attributes_retired_test.go
+++ b/cmd/tools/generate_attributes_retired_test.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// #342: an attribute dropped from the schema keeps its ledger entry, but the
+// generator must brand it retired so the freed attributeID stays bound to the
+// EAV rows it already owns instead of being handed to a new attribute.
+func TestGenerateAttributes_RemovalMarksRetired(t *testing.T) {
+	dir := t.TempDir()
+	schema := writeSchemaFixture(t, dir, "user.json",
+		`{"type":"object","properties":{"name":{"type":"string"}}}`)
+	out := writeSchemaFixture(t, dir, "user_attributes.json", `{
+	  "name":    { "attributeID": 1, "valueType": "text" },
+	  "old_col": { "attributeID": 3, "valueType": "text", "required_policy": "required_always" }
+	}`)
+
+	if err := generateAttributesJSON(schema, out, false); err != nil {
+		t.Fatalf("generate: %v", err)
+	}
+	attrs := readGeneratedAttributes(t, out)
+	oldCol, exists := attrs["old_col"]
+	if !exists {
+		t.Fatalf("removed attribute must stay in the ledger")
+	}
+	if retired, _ := oldCol["retired"].(bool); !retired {
+		t.Fatalf("removed attribute must be marked retired, got %v", oldCol)
+	}
+	if id, _ := oldCol["attributeID"].(float64); id != 3 {
+		t.Fatalf("retired attributeID changed: %v", oldCol["attributeID"])
+	}
+	if _, ok := oldCol["required_policy"]; ok {
+		t.Fatalf("retired attribute must be forced optional, got %v", oldCol["required_policy"])
+	}
+}
+
+// #342: re-adding a retired name with the identical physical type restores the
+// preserved rows, so the retired marker must be cleared and the original
+// attributeID kept.
+func TestGenerateAttributes_ReAddSameTypeClearsRetired(t *testing.T) {
+	dir := t.TempDir()
+	schema := writeSchemaFixture(t, dir, "user.json",
+		`{"type":"object","properties":{"name":{"type":"string"},"old_col":{"type":"string"}}}`)
+	out := writeSchemaFixture(t, dir, "user_attributes.json", `{
+	  "name":    { "attributeID": 1, "valueType": "text" },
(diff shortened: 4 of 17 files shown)


iceboundrock:
## Review — Standards

Verified against `coding-standard.md`, `AGENTS.md`, `CLAUDE.md`; diff `45dc962...0ff559a` (11 commits).

**Passing:** all touched source files ≤500 lines; all functions ≤100 lines; flat nesting with early returns throughout (`mergeExistingAttributes`, `clearRetiredOnReAdd`, `activeAttributeCache`, new test bodies). No new bare `return err`. Error messages follow the naming rule (logical value type, offending column/attribute, expected state, remedy) — `attrIDCollisionError` / `bindingCollisionError` (`internal/schemameta/validation.go`) name schema, id/column, retired name + valueType, both attribute names, and the fix.

**Findings:**

1. *(Judgement call)* **Pre-existing bare `return err` in directly-touched hunks.** `loader.go` (validate/parse error returns around the new strip) and `file_registry.go:176,181` still return unwrapped errors. Pre-existing, not introduced — but AGENTS.md's "when touching code that doesn't meet standards, fix the immediate area" suggests wrapping (`fmt.Errorf("validate schema %s: %w", schemaName, err)`). The new `RegisterSchema` path *does* wrap, so the inconsistency is visible.

2. *(Judgement call)* **`loader.go:325-326` — both cache fields now alias one map.** `attributeMetadata[schemaID]` and `schemaCaches[schemaID]` share the same `active` map (`map[string]forma.AttributeMetadata(active)` is a type conversion, not a copy). Pre-change this path kept two independent maps. Low practical risk (nothing mutates either post-load; `RegisterSchema` already aliases), but the file path used to be copy-safe.

3. *(Nit)* **Duplicate validate-then-strip comment blocks** in `file_registry.go:178-179`, `file_registry.go:376-377`, `loader.go:317-318` with near-identical prose — could live once on `activeAttributeCache` (already commented at `validation.go:52-56`).

*Nothing machine-enforced (gofmt/golangci-lint) is violated.*

iceboundrock:
## Review — Spec

Diff `45dc962...0ff559a` vs issue #342 + approved Implementation Plan (posted on #342).

**(a) Missing / partial:**

1. **`loadSchemasFromDB` strip site is untested.** The plan's Task 3 names four storage points; three are covered (`loadSchemasFromDirectory`, `loadAttributeMetadataFromFiles`, `RegisterSchema`), but the DB-path strip in `file_registry.go:178-181` has no unit test. Acknowledged as follow-up #4 in the plan — still a gap against Task 3's "four storage points" wording.
2. **Generator re-add error omits the schema name.** Global constraint: "Error messages must name the schema, the offending attributeID/attribute, the retired name+valueType, and the expected state." `clearRetiredOnReAdd` (`cmd/tools/generate_attributes.go`) names attribute + old/new types + expected state but not the schema (the outer `merge existing attributes from %s` wrap names only the file path). Minor, and the plan's own Task 4 snippet had the same shape.

**(b) Scope creep (defensible):**

1. **`effectiveItemsType` normalization + `TestGenerateAttributes_ReAddListWithoutItemsTypeClearsRetired`** — beyond the plan's literal `items_type` string comparison. Justified: fixes a latent plan bug for legacy `list`-without-`items_type` ledgers and matches runtime `EffectiveItemsType` (`schema_registry.go:166-171`).
2. **Docs expansion (~177 lines)** — far beyond Task 6's two short edits, but the lakehouse irreversibility / resurrect-asymmetry caveats are mandated by the recorded 2026-08-07 rulings, so this is correct scope.
3. **`knownAttrIDs` comment rewrite** (`internal/postgres_persistent_repository_eav.go`) — not in the plan's file list, tied to the RegisterSchema ruling. Fine.

**(c) Implemented but looks wrong:**

1. **Docs call them "two production registration paths — the file registry and the DB-backed loader"** (`docs/error-handling.md`). There are three load paths that matter (MetadataLoader file path, file-registry DB path, file-registry directory path) plus `RegisterSchema`; the phrasing understates the guard's reach, and `cmd/sample` reaches the directory path. Suggest: "every registration path".
2. **Retired↔retired collision falls back to the generic duplicate message** — `splitRetiredPair` requires exactly one retired side, so two retired entries on one id get the old generic text. Pre-existing behaviour, low value (retired↔retired is not a reuse trap), but follow-up #5's determinism note is accurate.

**Confirmed correct (spot-checked):** all four strip sites validate the FULL cache before stripping; `RegisterSchema` validates pre-snapshot (`loader.go:62-68`), closing the silent-swallow hole; retired ids stay in the `maxID` scan so never reissued; `logs` (29) / `contactSnapshot` (25) both retired with write-breaking + unrestorable caveats documented; e2e parity test `TestRetiredMarkerRemovalWorkflow` mirrors the hand-deleted flow including un-retire restore.

iceboundrock:
## Review — Assessment / Observations / Verification / Summary

### Assessment
Approve with minor nits. This is a carefully executed guard: the file-ledger mechanism matches the #342 rulings, the strip-after-validate ordering is enforced at all four storage points, and — critically — the docs were corrected (post whole-branch review) on two points that could have cost data: un-retire only restores hot-tier rows, and `logs` (29) is unrestorable under its original array shape. The `clearRetiredOnReAdd` items_type leg is mutation-verified and the `effectiveItemsType` normalization is a sound correction of the plan's own latent bug.

The breaking-change callout for `visit_full.logs` is appropriately prominent; the write-breaking / read-breaking asymmetry between `logs` (29) and `contactSnapshot` (25) is well explained.

### Observations
- **`effectiveItemsType`** — good catch; the literal-comparison plan would have refused a legacy `list` ledger with no `items_type` on re-add, handing the operator an unsatisfiable instruction. The test pins it.
- **Shared-map aliasing in `loader.go`** — harmless today, but if either cache field is ever mutated in place the other silently changes; a defensive `copySchemaAttributeCache` per field would restore the old copy-safety.
- **Error-message determinism** — collision message selection depends on map iteration order when a collision involves two entries; matches the pre-existing pattern, follow-up #5 is the right home for it.
- **`loadSchemasFromDB` untested strip** — the one production path without direct coverage; the e2e infra tests exercise it indirectly through the harness. Follow-up #4.

### Verification (run in a clean detached-HEAD worktree of `0ff559a`)
```
GOCACHE=/tmp/pr376-wt/.gocache GOFLAGS=-buildvcs=false go test ./internal/schemameta ./cmd/tools -run 'Retired|GenerateAttributes|LoadMetadata_StripsRetired' -count=1
ok  internal/schemameta   0.495s
ok  cmd/tools             0.610s
```
New parser/guard/strip tests, all four generator re-add tests, and the loader strip test pass. I did not re-run the full `make test` / e2e suite; that verification is per the PR body (green) and CI.

### Summary
- **Standards:** 3 minor findings (pre-existing bare `return err` in touched hunks; shared-map aliasing; duplicated comments) — no hard violations.
- **Spec:** 2 partials (untested `loadSchemasFromDB` strip; schema name missing from generator error), 3 defensible scope additions, 2 doc/behaviour nits. Core contract implemented as ruled.
- Worst single issue: none blocking. The untested DB-path strip is the most substantive gap and is already tracked as follow-up #4.

Non-blocking; no merge requested. 请不要自动合并。

iceboundrock:
## Response to review — 4 fixed (`dbb4484`), 3 answered

Thanks — the two I want to call out as genuinely mine: the docs "two production registration paths" line was an over-correction I introduced in the whole-branch fix wave, and the shared-map aliasing was a property this PR silently removed. Both fixed.

### Fixed

**Standards 1 — bare `return err` in touched hunks.** Wrapped, but under a rule rather than mechanically: *each wrap must add context the inner error does not already carry*, because a redundant wrap makes the operator message worse. The facts that drove it — `parseAttributeMetadata` errors already end with the attributes-file path; `validateSchemaAttributeCache` errors already begin with `schema <name>`; `loadSchemaArtifacts` errors carry file paths but not the schema name.

- loader validate site gains the attributes-file path: `attributes file cmd/server/schemas/visit_attributes.json: schema visit has duplicate attribute id 3 ...`
- both `loadSchemaArtifacts` sites gain the schema name and the load mode that demanded the file.
- **`RegisterSchema`'s double-naming (your deferred T3 nit) is fixed by making the wrap additive rather than by deleting it.** The validation error names the schema by *name*; the wrap now carries the *numeric id*, which it cannot know: `register schema id 100: schema user reuses attribute id 3: retired attribute old_col (valueType text) still owns preserved EAV rows and cannot be rebound to nickname; ...` The same reasoning wrapped the two `registerSchema` call sites — in directory mode that id is auto-assigned at runtime, appears in no readable file, and is the key every EAV row is written under.
- **Left bare, deliberately:** the two `parseAttributeMetadata` loops. Pre-existing context lines (not PR-authored), the message already carries the attributes file, `LoadMetadata` adds the phase, and the schema id is not actionable for a malformed `valueType`. Say the word if you want uniformity — it is a two-line change.

**Standards 2 — shared-map aliasing.** You are right that this PR removed the property: before it, that path built `attrMap` and `schemaCache` independently. `attributeMetadata` now takes a `copySchemaAttributeCache` deep copy (which also copies `*MainColumnBinding`), pinned by a new `TestLoadMetadataCacheFieldsAreIndependent` — red-checked against the pre-fix loader, and it fails on **both** the map injection and the shared binding pointer. `RegisterSchema`'s own aliasing is pre-existing and left alone, so the two paths now differ in copy-safety; flagging that rather than expanding scope.

**Standards 3 — triplicated comments.** Full rationale stays on `activeAttributeCache`; each call site keeps one line that preserves the ordering warning, so a future reorder still has to delete a comment telling it not to.

**Spec (c)1 — docs understate the reach.** Corrected to name all three load paths (`MetadataLoader` file path; file registry in registry-table and directory modes, with `cmd/sample` on the latter), keeping the separate `RegisterSchema`-has-no-production-callers note.

### Answered, not changed

**Spec 1 — `loadSchemasFromDB` strip untested.** Staying as follow-up #4, and I want to name the reason it deserves its own change rather than a tail-end patch here: making it testable means widening `fileSchemaRegistry.pool` from `*pgxpool.Pool` to an interface, and `NewFileSchemaRegistryFromDirectory` sets `pool: nil` — assigning a nil `*pgxpool.Pool` into an interface field produces the non-nil-interface-holding-nil trap this repo has already been bitten by (#302). That is a change that wants its own red test and its own review, not a patch appended after approval. The strip line itself is character-identical to the tested directory path and shares the `registerSchema`/`buildAttrIDToName` seam.

**Spec 2 — generator re-add error omits the schema name.** Keeping as is. The outer wrap carries `outputPath` — `merge existing attributes from cmd/server/schemas/visit_full_attributes.json: re-add retired attribute: attribute logs is retired with valueType text/items_type "" but re-added as list/"text": ...`. The file path identifies the schema more precisely than a bare name would, and `generate-attributes` processes one schema per invocation, so the ambiguity the global constraint guards against (many schemas registering in one process) does not exist here. Threading a name parameter through two signatures would add a fact the message already carries.

**Spec (c)2 — retired↔retired collision falls back to the generic message.** Agreed, and deliberately so: `splitRetiredPair` requires exactly one retired side because two retired entries sharing an id is not a reuse trap — neither is active, so nothing can bind to those preserved rows. It is still rejected, just with the generic duplicate text. Follow-up #5 is the right home if the message selection is made deterministic.

### Verification

`make test` 32 packages ok / zero FAIL · `make lint` exit 0 · gofmt clean · no existing assertion needed changing (all are substring checks on surviving text). CI was green on `0ff559a` (5/5) and is re-running on `dbb4484`.

请不要自动合并。

iceboundrock:
## Re-review — Standards

Re-review of `45dc962...dbb4484` (12 commits), fix-wave `0ff559a...dbb4484` verified against `coding-standard.md`, `AGENTS.md`, `CLAUDE.md`.

**Prior findings — all resolved:**

1. **Bare `return err` → wrapped, additively, with the rule recorded once per error source.** The wrap-rule is now explicit and self-consistent: each wrap adds only context the inner error lacks. Verified per site:
   - loader validate site (`loader.go:320`): `attributes file %s` — inner error begins with `schema <name>`, so the file path is the added fact.
   - both `loadSchemaArtifacts` sites (`file_registry.go:181,378`): add schema name + the load mode that demanded the file (registry-table vs directory) — inner errors already carry file paths.
   - both `registerSchema` call sites + `RegisterSchema` (`loader.go:66`, `file_registry.go:188,385`): add the numeric schema id, which `buildAttrIDToName`/`validateSchemaAttributeCache` cannot know — inner errors name the schema by name only. The old `register schema %s` double-name is gone.
   - Left bare deliberately, now documented: the two `parseAttributeMetadata` loops. `parser.go:12-13` states the contract (errors end with `in <source>`; `LoadMetadata` adds the phase). This matches the author's stated rule and the pre-existing context lines — accepted judgement call, clearly flagged in-code.

2. **Shared-map aliasing → fixed and pinned.** `loader.go:328-329` now stores `attributeMetadata` via `copySchemaAttributeCache(active)` (deep copy incl. `*MainColumnBinding`) while `schemaCaches` holds `active` itself. `TestLoadMetadataCacheFieldsAreIndependent` is mutation-sensitive on **both** legs — I reintroduced the old aliasing (`map[string]forma.AttributeMetadata(active)`) and the test reddened on the binding pointer; the map-injection leg would catch the map sharing too.

3. **Triplicated validate-then-strip comments → deduped.** Each call site keeps one line that preserves the ordering warning; the full rationale lives once on `activeAttributeCache` (`validation.go:59-64`).

**New observations (judgement calls, no violations):**

- **`RegisterSchema` still aliases the two fields** (`loader.go:74-75`): `schemaCaches[schemaID] = snapshot` and `attributeMetadata[schemaID] = map[string]forma.AttributeMetadata(snapshot)` are the *same* underlying map — the conversion is not a copy. The loader path is now copy-safe while `RegisterSchema` is not, an asymmetry the author explicitly flagged as pre-existing and left alone. Not a regression (main had it), but the doc comment `The two fields never share one map` on the loader path reads like a cache-wide invariant; a reader could miss that `RegisterSchema` violates it. Consider a one-line pointer on `RegisterSchema` acknowledging the alias, or `copySchemaAttributeCache` there too — the cost is one line and the two paths would then share the property.
- **Wrap granularity asymmetry (nit):** `file_registry.go:185,382` say `attribute metadata in %s` (schemaDir) where the loader path says `attributes file %s` (full path). Both are additive and locatable; only cosmetic.
- **Generator error text** verified byte-for-byte against the PR body's example: `merge existing attributes from <path>: re-add retired attribute: attribute logs is retired with valueType text/items_type "" but re-added as list/"text": ...` — the outputPath carries the schema.

*Nothing machine-enforced (gofmt/golangci-lint/go vet) is violated; gofmt-clean on all PR-touched files (the three gofmt-flagged e2e files are pre-existing and outside this diff).*

iceboundrock:
## Re-review — Spec

Diff `45dc962...dbb4484` vs issue #342 + approved Implementation Plan (posted on #342), re-checking the fix-wave's disposition of the prior Spec findings.

**Prior findings — dispositions verified:**

1. **Spec (c)1 — docs understate the guard's reach → FIXED.** `docs/error-handling.md` now reads "Every production registration path — the DB-backed `MetadataLoader` file path, and the file registry in both its registry-table and directory modes (`cmd/sample` uses the latter)". I confirmed against the code that this is accurate and complete: `MetadataLoader`'s file path (`loader.go:274`), the file registry's registry-table mode (`loadSchemasFromDB`), its directory mode (`loadSchemasFromDirectory`, used by `cmd/sample` via `NewFileSchemaRegistryFromDirectory`), plus the separate `RegisterSchema` note that it "has no production callers" — verified: every `RegisterSchema` caller in the tree is a test. This matches the prior suggestion ("every registration path") and now names the real three.

2. **Spec 1 — `loadSchemasFromDB` strip site untested → acknowledged, stays follow-up #4.** The author's refusal-to-tail-patch is sound and I want to validate the technical reason: making it unit-testable requires widening `fileSchemaRegistry.pool` from `*pgxpool.Pool` to an interface, and `NewFileSchemaRegistryFromDirectory` sets `pool: nil` — the nil-interface-holding-nil-pool trap this repo already hit in #302. That is a real hazard, not a dodge. Note the strip line itself is character-identical to the tested directory path (`file_registry.go:184-189` vs `376-385`) and shares the `registerSchema`/`buildAttrIDToName` seam, so the untested surface is one line, not a path.

3. **Spec 2 — generator re-add error omits schema name → answered, kept.** Verified the reasoning is defensible: the outer wrap carries the full `outputPath` (`merge existing attributes from %s`), which identifies the schema more precisely than a bare name, and `generate-attributes` processes one schema per invocation. The global constraint's ambiguity concern (many schemas registering in one process) does not apply to the generator. Agree.

4. **Spec (c)2 — retired↔retired collision falls back to generic message → deliberate, kept.** Confirmed `splitRetiredPair` requires exactly one retired side (`validation.go:102-111`); two retired entries sharing an id is not a reuse trap (neither is active, nothing can bind to those preserved rows). It is still rejected, just without the reuse-specific text. Follow-up #5 remains the right home for deterministic message selection.

**New observations:**

- **`TestLoadMetadataCacheFieldsAreIndependent`** (`loader_test.go:382`) is the one test that proves the fix-wave actually restored copy-safety; I mutated the loader back to the aliased assignment and it failed on the binding-pointer leg. The test is genuinely load-bearing, not decorative.
- **`effectiveItemsType` normalization** (`generate_attributes.go:222-227`) now demonstrably matches runtime `EffectiveItemsType` (`schema_registry.go:166-171`): both treat empty `items_type` as text. The prior plan's literal string comparison would have refused a legacy `list`-without-`items_type` re-add — this is a sound correction of a latent plan bug.
- **`TestGenerateAttributes_ReAddListWithoutItemsTypeClearsRetired`** pins that the re-add *records* the now-explicit `items_type: text` — the generator does not write an empty `items_type` back. This closes the loop the plan left open.

**Spot-checked correct:** all four strip sites validate the FULL cache before stripping; retired ids stay in the `maxID` scan so never reissued; `logs` (29) / `contactSnapshot` (25) retirement rationale matches the actual shipped schema state (verified `visit_full.json` has no `logs`; `visit.json:29-34` `contactSnapshot` is a `$ref` to an object); the production e2e `TestRetiredMarkerRemovalWorkflow` (with `-tags=e2e`) passes under Docker and mirrors the hand-deleted flow including un-retire restore.

iceboundrock:
## Re-review — Assessment / Observations / Verification / Summary

### Assessment
Approve. This re-review confirms the fix-wave (`dbb4484`) resolves all four actionable findings from the previous review round — three standards items (additive error wraps, restored loader copy-safety with a mutation-sensitive test, deduped comments) and the spec docs-reach correction. The two deferred items (untested `loadSchemasFromDB` strip, generator error schema name) were answered with technical reasoning I verified as sound, and both remain appropriately tracked as follow-ups on #342 rather than tail-end patches. The breaking-change handling for `visit_full.logs` (write-breaking, unrestorable under its original shape) is correctly documented with two honest remediation paths.

### Observations
- **The wrap-rule is now a documented convention.** Each error source carries one comment stating what callers may wrap it with (parser.go, validation.go, file_registry.go `registerSchema`). This is the right pattern — it turns "always wrap" from a mechanical rule into a rule that produces additive messages, and it is recorded where future edits will see it.
- **The `RegisterSchema` aliasing asymmetry** (loader path copy-safe, `RegisterSchema` still aliasing the two cache fields) is the one lingering smell. It is pre-existing and non-blocking, but since this PR now asserts copy-safety as a property worth pinning, making `RegisterSchema` symmetric would be a one-line follow-up.
- **The e2e parity test is the load-bearing proof.** `TestRetiredMarkerRemovalWorkflow` passing under Docker (with `-tags=e2e`) demonstrates the retired-marker generation behaves identically to the hand-deleted one end-to-end — seed, retire, preserve-on-update, invisible-on-read, un-retire restore, round-trip. The invisibility assertion would fail if a retired entry leaked into any consumer cache.

### Verification (run in a clean detached-HEAD worktree of `dbb4484`)
```
go build ./...                                      # OK
go test . ./cdc ./cmd/... ./factory ./internal/...  # 32 packages ok, 0 FAIL
go test -race ./internal/schemameta ./cmd/tools     # ok
go test ./cmd/tools -run TestShippedAttributesRegenerateByteIdentical   # 12/12 PASS
go test -tags=e2e ./internal/e2e_harness/production \
  -run 'TestRetiredMarkerRemovalWorkflow|TestUpdateAfterAttributeRemoval'  # both PASS (Docker)
```
Mutation checks performed: (a) deleted the `items_type` comparison leg → reddens exactly `TestGenerateAttributes_ReAddDifferentItemsTypeRejected`, nothing else; (b) reintroduced the loader map aliasing → reddens `TestLoadMetadataCacheFieldsAreIndependent` on the binding-pointer leg. gofmt clean on all PR-touched files; go vet clean. I did not re-run the full e2e/federated suite or `make lint` (golangci-lint not installed locally) — those are green per CI on `dbb4484` (Lint / Test / E2E Smoke / k6 / Benchmark all pass).

### Summary
- **Standards:** 3 prior findings all resolved; 2 new judgement-call observations (pre-existing `RegisterSchema` alias asymmetry; wrap-granularity nit) — no hard violations.
- **Spec:** prior (c)1 docs finding fixed; the two deferred items answered with verified-sound rationale; 2 new positive observations. Core contract implemented as ruled, and this round the copy-safety fix is proven by a mutation-sensitive test rather than asserted.
- Worst single issue: none blocking. The `RegisterSchema` alias asymmetry is the most substantive nit and is pre-existing; a one-line `copySchemaAttributeCache` there would close it.

Non-blocking; no merge requested. 请不要自动合并。

iceboundrock:
## Response to re-review — both observations closed (`5674fff`)

Thanks for the mutation checks — reproducing the aliasing and the `items_type` leg independently is the kind of verification that makes the approval mean something.

Both new observations are now fixed rather than deferred. The `RegisterSchema` one I agree was more than cosmetic: the problem was not the alias itself (pre-existing, harmless today) but that **this PR wrote a comment asserting an invariant that one live writer violated**. Leaving that in place would have been worse than the alias.

### Fixed

**`RegisterSchema` alias asymmetry.** It now takes a second `copySchemaAttributeCache`, so both writers of `MetadataCache` hold the property. The two copies carry different guarantees and are commented as such, because otherwise a future reader deletes one as redundant:

- first copy — **caller ownership** (#142): severs the caller's map from what is stored;
- second copy — **field independence**: severs `schemaCaches` and `attributeMetadata` from each other.

The invariant moved to `MetadataCache`'s field declarations, where it binds every writer, and it explicitly warns that a `map[string]forma.AttributeMetadata(...)` conversion is *not* a copy — that conversion is exactly how both of these bugs got written. The per-site comments now point at it instead of restating it.

`TestRegisterSchemaCacheFieldsAreIndependent` pins the same two legs as the loader-path test. Mutation-verified by hand-restoring the aliasing conversion: **both** legs reddened (map injection `should not contain injected`; binding pointer `text_01` vs `text_09`), then hand-restored. No `git checkout`/`restore` — this repo has lost uncommitted work that way before.

**Wrap-granularity nit.** Added `(*fileSchemaRegistry).attributesFilePath` and routed `loadSchemaArtifacts` plus both validate sites through it, so the wraps now name the exact file (`attributes file <dir>/<name>_attributes.json: ...`) like the loader path does — and the inline path construction is no longer duplicated. Wraps stay additive: the inner validation error names the schema, the wrap contributes only the path.

While there, `validateSchemaAttributeCache`'s doc comment still promised a "schema directory" wrap variant that no longer exists — corrected.

### Noted, not done

`MetadataLoader` builds the same `<name>_attributes.json` string from its own `schemaDirectory` field. Unifying it with the new registry method needs a package-level helper and both types re-pointed — wider than this nit, so it stays a follow-up rather than getting folded in after approval.

### Verification

`make test` all packages ok / zero FAIL · `make lint` exit 0 · gofmt clean on touched files · no existing assertion needed changing. CI re-running on `5674fff`.

请不要自动合并。

PR #121, opened 2026-06-04
fix: harden schema metadata consistency

iceboundrock:
## Summary
- harden schema/metadata parsing and loading with shared validation for IDs, duplicate schema mappings, duplicate attribute IDs, and duplicate hot-column bindings
- reject unknown or invalid write attributes and preserve update consistency by returning validation errors instead of silently skipping user input
- make read-time metadata drift explicit by failing on unknown attr IDs, storage/type mismatches, and per-array required-field gaps

## Testing
- go test ./internal ./factory -count=1
- go test ./... -count=1

## Related Issues
- Closes #112
- Related to #108
- Related to #109
- Related to #110

PR #377, opened 2026-08-08
fix(tools): #341 classify #294-preserved EAV rows as informational

### LINKED ISSUE #341: validate-schema-consistency flags #294-preserved EAV rows as failures; migration doc coherence residue

iceboundrock:
Closes #341.

Spec: https://github.com/Lychee-Technology/forma/issues/341#issuecomment-5226173089

## The problem

`validate-schema-consistency` reported **every** EAV row whose `attr_id` is absent from current metadata as a failure and exited non-zero. Since #294 those rows are a supported steady state — skipped on read, preserved on update, restored on re-add — so any deployment that ever retired an attribute could never get a green pre-flight. The ledgers shipped by #342/#376 made this universal: `visit` `attr_id` 25 and `visit_full` `attr_id` 29 are retired, so every deployment holding rows under them was flagged on every run.

The signal to fix it already existed (#342's `"retired": true` marker) but the validator could not see it: `MetadataCache` strips retired entries before any consumer reaches them (`activeAttributeCache`).

## What changed

**`internal/schemameta`** — `MetadataCache` now records the retired ledger it previously discarded, at both registration points, reading the un-stripped cache strictly after `validateSchemaAttributeCache` (which is what makes the attributeID key unambiguous). Exposed through:

```go
func (mc *MetadataCache) RetiredAttributeIDs(schemaID int16) map[int16]string
```

Ids and names only, copy-on-read. It deliberately returns neither `forma.AttributeMetadata` nor the live map, so it cannot become a read, write, flush, or projection path for a retired attribute — the narrow return type, not convention, is what preserves `activeAttributeCache`'s invariant. Existing strip guards stay green.

**`cmd/tools/validate_schema_consistency.go`** — undecodable EAV rows are now classified three ways:

| Condition | Result |
| --- | --- |
| `schema_id` absent from the registry | failure (unchanged) |
| `attr_id` in that schema's retired ledger | informational |
| otherwise | failure (unchanged) |

Classification is per attributeID, not a per-schema amnesty, and the ledger lookup is scoped per schema. Only failures reach the exit code and the `%d issue(s)` count.

**`docs/schema-consistency-migration.md`** — the three residues #341 names (front-matter bullet, case (i) carrying an option that belongs to a third cause, rollback step 3 readable as licence to delete preserved rows), plus the Go Validator checklist, the Deployment Checklist, a ledger-reconstruction recipe for attributes retired before #342, and a rewrite of `### Unknown attribute IDs in EAV` so the lead-in, the cases, and the checklist all state the same three-way taxonomy and lead with what the tool now decides for you. `docs/error-handling.md` gains one clause naming the accessor as the sanctioned exception to "retired entries never reach a consumer".

## Verification

`make lint` (golangci-lint v1.64.8) and `make test` both clean. Nine unit tests added across the two packages, including per-id-not-per-schema and unknown-`schema_id` pins.

End-to-end against real Postgres and the **shipped** ledgers, with probe rows under `visit`/25, `visit_full`/29 and an unledgered id 199:

```text
- unknown attribute IDs in eav_data_dev: schema_id=110 attr_id=199 rows=1
informational (not a failure):
- preserved EAV rows for retired attributes in eav_data_dev: schema=visit schema_id=110 attr_id=25 attribute=contactSnapshot rows=1
- preserved EAV rows for retired attributes in eav_data_dev: schema=visit_full schema_id=111 attr_id=29 attribute=logs rows=1
validate-schema-consistency: schema consistency validation failed with 1 issue(s)
```

With the orphan removed, exit code 0:

```text
schema consistency checks passed for 12 schema(s), 2 informational finding(s)
informational (not a failure):
- preserved EAV rows for retired attributes in eav_data_dev: schema=visit schema_id=110 attr_id=25 attribute=contactSnapshot rows=1
- preserved EAV rows for retired attributes in eav_data_dev: schema=visit_full schema_id=111 attr_id=29 attribute=logs rows=1
```

## Exit-code change

This changes an operator-facing pre-flight tool's exit semantics — deliberately, and that is the issue. Nothing in CI gates on it: the in-repo callers are the `validate-schema-consistency` Makefile target, a hint echoed by `scripts/validate_schema_consistency.sql`, and the migration guide. `.github/workflows/ci.yml` never invokes it.

## Not in scope

Per the spec's decision record: retired attributes are **not** folded into the storage-column mismatch check (those rows are unreadable until re-add, and including them would create new failures for existing deployments); no `--fail-on-retired` mode; no `--allow-attr-ids` escape hatch. Deployments that retired an attribute before #342 have no ledger entry and still fail — the remedy is the documented hand-rebuild recipe, not code.

## Known residue

One-word polish left open: `docs/schema-consistency-migration.md` "startup will fail naming both" now follows a three-item antecedent. The claim is accurate (every guard message names two attributes), so it reads as imprecise rather than wrong.

### DIFF
--- a/cmd/tools/validate_schema_consistency.go
+++ b/cmd/tools/validate_schema_consistency.go
@@ -26,9 +26,20 @@ type schemaConsistencyValidator struct {
 	out         io.Writer
 }
 
+// issueSeverity separates findings that must block a pre-flight from findings
+// the operator only needs to know about. severityError is the zero value, so
+// every existing construction site stays a failure by default.
+type issueSeverity int
+
+const (
+	severityError issueSeverity = iota
+	severityInfo
+)
+
 type validationIssue struct {
 	category string
 	details  string
+	severity issueSeverity
 }
 
 func runValidateSchemaConsistency(ctx context.Context, args []string) error {
@@ -113,15 +124,49 @@ func (v schemaConsistencyValidator) run(ctx context.Context) error {
 		return err
 	}
 
-	if len(issues) == 0 {
-		fmt.Fprintf(v.out, "schema consistency checks passed for %d schema(s)\n", len(cache.ListSchemas()))
+	failures, notices := partitionIssues(issues)
+	v.report(failures, notices, len(cache.ListSchemas()))
+	if len(failures) == 0 {
 		return nil
 	}
+	return fmt.Errorf("schema consistency validation failed with %d issue(s)", len(failures))
+}
 
+// partitionIssues splits collectIssues' sorted output into the two report
+// blocks while preserving order inside each.
+func partitionIssues(issues []validationIssue) (failures, notices []validationIssue) {
 	for _, issue := range issues {
+		if issue.severity == severityInfo {
+			notices = append(notices, issue)
+			continue
+		}
+		failures = append(failures, issue)
+	}
+	return failures, notices
+}
+
+// report prints failures first, then the informational block. Informational
+// findings are EAV rows #294 preserves under a retired attribute: a supported
+// steady state, so they are surfaced for awareness and never gate the exit
+// code (#341).
+func (v schemaConsistencyValidator) report(failures, notices []validationIssue, schemaCount int) {
+	if len(failures) == 0 {
+		fmt.Fprintf(v.out, "schema consistency checks passed for %d schema(s)", schemaCount)
+		if len(notices) > 0 {
+			fmt.Fprintf(v.out, ", %d informational finding(s)", len(notices))
+		}
+		fmt.Fprintln(v.out)
+	}
+	for _, issue := range failures {
 		fmt.Fprintf(v.out, "- %s: %s\n", issue.category, issue.details)
 	}
-	return fmt.Errorf("schema consistency validation failed with %d issue(s)", len(issues))
+	if len(notices) == 0 {
+		return
+	}
+	fmt.Fprintln(v.out, "informational (not a failure):")
+	for _, issue := range notices {
+		fmt.Fprintf(v.out, "- %s: %s\n", issue.category, issue.details)
+	}
 }
 
 func (v schemaConsistencyValidator) collectIssues(ctx context.Context, cache *schemameta.MetadataCache) ([]validationIssue, error) {
@@ -148,8 +193,17 @@ func (v schemaConsistencyValidator) collectIssues(ctx context.Context, cache *sc
 	return issues, nil
 }
 
-func (v schemaConsistencyValidator) checkUnknownAttributeIDs(ctx context.Context, cache *schemameta.MetadataCache) ([]validationIssue, error) {
-	knownAttrIDs := make(map[int16]map[int16]struct{})
+// schemaLedger is the per-schema view checkUnknownAttributeIDs classifies
+// against: the active attributeIDs, the retired ledger #342 records, and the
+// schema name for the report line.
+type schemaLedger struct {
+	name    string
+	active  map[int16]struct{}
+	retired map[int16]string
+}
+
+func buildSchemaLedgers(cache *schemameta.MetadataCache) map[int16]schemaLedger {
+	ledgers := make(map[int16]schemaLedger)
 	for _, schemaName := range cache.ListSchemas() {
 		schemaID, ok := cache.GetSchemaID(schemaName)
 		if !ok {
@@ -159,12 +213,21 @@ func (v schemaConsistencyValidator) checkUnknownAttributeIDs(ctx context.Context
 		if !ok {
 			continue
 		}
-		ids := make(map[int16]struct{}, len(schemaCache))
+		active := make(map[int16]struct{}, len(schemaCache))
 		for _, meta := range schemaCache {
-			ids[meta.AttributeID] = struct{}{}
+			active[meta.AttributeID] = struct{}{}
 		}
-		knownAttrIDs[schemaID] = ids
+		ledgers[schemaID] = schemaLedger{
+			name:    schemaName,
+			active:  active,
+			retired: cache.RetiredAttributeIDs(schemaID),
+		}
 	}
+	return ledgers
+}
+
+func (v schemaConsistencyValidator) checkUnknownAttributeIDs(ctx context.Context, cache *schemameta.MetadataCache) ([]validationIssue, error) {
+	ledgers := buildSchemaLedgers(cache)
 
 	query := fmt.Sprintf(`
 SELECT e.schema_id, e.attr_id, COUNT(*) AS record_count
@@ -185,20 +248,38 @@ ORDER BY e.schema_id, e.attr_id`, quoteIdentifier(v.eavTable))
 		if err := rows.Scan(&schemaID, &attrID, &count); err != nil {
 			return nil, fmt.Errorf("scan unknown attribute ids: %w", err)
 		}
-		if ids, ok := knownAttrIDs[schemaID]; ok {
-			if _, exists := ids[attrID]; exists {
-				continue
-			}
+		// A schema_id absent from the registry yields the zero ledger: both maps
+		// are nil, so the row falls through to the failure branch as before.
+		ledger := ledgers[schemaID]
+		if _, exists := ledger.active[attrID]; exists {
+			continue
 		}
-		issues = append(issues, validationIssue{
-			category: "unknown attribute IDs in " + v.eavTable,
-			details:  fmt.Sprintf("schema_id=%d attr_id=%d rows=%d", schemaID, attrID, count),
-		})
+		issues = append(issues, classifyUndecodableRows(v.eavTable, ledger, schemaID, attrID, count))
 	}
 	if err := rows.Err(); err != nil {
 		return nil, fmt.Errorf("iterate unknown attribute ids: %w", err)
 	}
 	return issues, nil
+}
+
+// classifyUndecodableRows splits EAV rows current metadata cannot decode into
+// the two cases the migration guide names. Rows under a retired ledger entry
+// were preserved by an attribute removal (#294) — an expected, supported steady
+// state, reported informationally. Everything else is genuinely orphaned and
+// still fails the pre-flight (#341).
+func classifyUndecodableRows(eavTable string, ledger schemaLedger, schemaID, attrID int16, count int64) validationIssue {
+	if attrName, ok := ledger.retired[attrID]; ok {
+		return validationIssue{
+			category: "preserved EAV rows for retired attributes in " + eavTable,
+			details: fmt.Sprintf("schema=%s schema_id=%d attr_id=%d attribute=%s rows=%d",
+				ledger.name, schemaID, attrID, attrName, count),
+			severity: severityInfo,
+		}
+	}
+	return validationIssue{
+		category: "unknown attribute IDs in " + eavTable,
+		details:  fmt.Sprintf("schema_id=%d attr_id=%d rows=%d", schemaID, attrID, count),
+	}
 }
 
 func (v schemaConsistencyValidator) checkStorageMismatches(ctx context.Context, cache *schemameta.MetadataCache) ([]validationIssue, error) {
--- a/cmd/tools/validate_schema_consistency_test.go
+++ b/cmd/tools/validate_schema_consistency_test.go
@@ -203,3 +203,176 @@ func writeSchemaConsistencyArtifacts(t *testing.T, dir, schemaName, schemaJSON, 
 		t.Fatalf("write attrs: %v", err)
 	}
 }
+
+// newSchemaConsistencyMock wires the four queries validator.run issues, in
+// order: the schema registry, the attr_id census, then the two storage-column
+// censuses. Only the attr_id census varies across the #341 classification
+// tests, so the rest is fixed here.
+func newSchemaConsistencyMock(t *testing.T, attrCensus *pgxmock.Rows) pgxmock.PgxPoolIface {
+	t.Helper()
+	mock, err := pgxmock.NewPool()
+	if err != nil {
+		t.Fatalf("create mock pool: %v", err)
+	}
+	t.Cleanup(mock.Close)
+
+	mock.ExpectQuery(`SELECT schema_name, schema_id FROM "schema_registry_dev"`).
+		WillReturnRows(pgxmock.NewRows([]string{"schema_name", "schema_id"}).AddRow("contact", int16(100)))
+	mock.ExpectQuery(`SELECT e\.schema_id, e\.attr_id, COUNT\(\*\) AS record_count FROM "eav_data_dev" AS e`).
+		WillReturnRows(attrCensus)
+	mock.ExpectQuery(`SELECT e\.schema_id, e\.attr_id, COUNT\(\*\) AS record_count FROM "eav_data_dev" AS e WHERE e\.value_text IS NOT NULL`).
+		WillReturnRows(pgxmock.NewRows([]string{"schema_id", "attr_id", "record_count"}))
+	mock.ExpectQuery(`SELECT e\.schema_id, e\.attr_id, COUNT\(\*\) AS record_count FROM "eav_data_dev" AS e WHERE e\.value_numeric IS NOT NULL`).
+		WillReturnRows(pgxmock.NewRows([]string{"schema_id", "attr_id", "record_count"}))
+	return mock
+}
+
+// retiredLedgerSchemaDir writes a contact schema whose ledger has one active
+// attribute (id 1) and one retired entry, legacy (id 9) — the #342 shape.
+func retiredLedgerSchemaDir(t *testing.T) string {
+	t.Helper()
+	dir := t.TempDir()
+	writeSchemaConsistencyArtifacts(t, dir, "contact",
+		`{"type":"object","properties":{"id":{"type":"string"}}}`,
+		`{"id":{"attributeID":1,"valueType":"text"},"legacy":{"attributeID":9,"valueType":"text","retired":true}}`)
+	return dir
+}
+
+func attrCensusRows() *pgxmock.Rows {
+	return pgxmock.NewRows([]string{"schema_id", "attr_id", "record_count"})
+}
+
+// TestValidateSchemaConsistencyClassifiesRetiredAttrRows pins the #341
+// contract: EAV rows whose attr_id belongs to a retired ledger entry are the
+// #294 tolerate-and-preserve steady state, so they report as informational and
+// leave the pre-flight green.
+func TestValidateSchemaConsistencyClassifiesRetiredAttrRows(t *testing.T) {
+	mock := newSchemaConsistencyMock(t, attrCensusRows().AddRow(int16(100), int16(9), int64(12)))
+
+	var out strings.Builder
+	validator := schemaConsistencyValidator{
+		pool:        mock,
+		schemaDir:   retiredLedgerSchemaDir(t),
+		schemaTable: "schema_registry_dev",
+		eavTable:    "eav_data_dev",
+		out:         &out,
+	}
+
+	if err := validator.run(context.Background()); err != nil {
+		t.Fatalf("preserved rows must not fail the run: %v", err)
+	}
+	got := out.String()
+	if strings.Contains(got, "unknown attribute IDs") {
+		t.Fatalf("preserved rows must not be reported as unknown attribute IDs, got %q", got)
+	}
+	for _, want := range []string{
+		"informational (not a failure):",
+		"preserved EAV rows for retired attributes in eav_data_dev",
+		"schema=contact", "attr_id=9", "attribute=legacy", "rows=12",
+		"schema consistency checks passed for 1 schema(s), 1 informational finding(s)",
+	} {
+		if !strings.Contains(got, want) {
+			t.Fatalf("output missing %q, got %q", want, got)
+		}
+	}
+	if err := mock.ExpectationsWereMet(); err != nil {
+		t.Fatalf("unmet db expectations: %v", err)
+	}
+}
+
+// TestValidateSchemaConsistencyFailsUnledgeredAttrID pins that classification
+// is per attributeID, not a per-schema amnesty: a schema that owns a retired
+// entry still fails for a different, unledgered id.
+func TestValidateSchemaConsistencyFailsUnledgeredAttrID(t *testing.T) {
+	mock := newSchemaConsistencyMock(t, attrCensusRows().AddRow(int16(100), int16(99), int64(2)))
+
+	var out strings.Builder
+	validator := schemaConsistencyValidator{
+		pool:        mock,
+		schemaDir:   retiredLedgerSchemaDir(t),
+		schemaTable: "schema_registry_dev",
+		eavTable:    "eav_data_dev",
+		out:         &out,
+	}
+
+	err := validator.run(context.Background())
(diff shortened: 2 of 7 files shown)


iceboundrock:
## Review record

Built task-by-task with a fresh implementer per task and an independent reviewer after each, then one whole-branch review. Recording the outcome here so the decision trail lives with the PR.

### Per-task reviews

| Task | Scope | Verdict |
| --- | --- | --- |
| 1 (`41fad62`) | `schemameta` retired-ledger accessor | spec ✅, quality approved, 0 Critical/Important |
| 2 (`4d481fa`) | validator three-way classification + two-tier output | spec ✅, quality approved, 0 Critical/Important |
| 3 (`94f10ad`) | migration-guide residue | spec ✅, quality approved, 0 Critical/Important |

Checks worth naming, because they were verified rather than assumed:

- **The accessor cannot leak a retired attribute.** It returns `map[int16]string` — no `ValueType`, no `ItemsType`, no `ColumnBinding` — so a `forma.AttributeMetadata` cannot be reconstituted from it and re-enter the read/write/flush/projection path. Copy-on-read under `RLock` closes the mutation vector.
- **Both capture sites read the un-stripped cache** (`RegisterSchema` from the argument, not `snapshot`; `loadAttributeMetadataFromFiles` from `schemaCache`, not `active`) and both run strictly after `validateSchemaAttributeCache` — which is what makes the attributeID key unambiguous, since that guard rejects duplicate ids across the full ledger, retired entries included.
- **No alternate construction path can leave the new map nil**: the only `MetadataCache{` literal in the repo is inside `NewMetadataCache`.
- **The zero-value severity is load-bearing and safe**: all four `validationIssue{` sites use named fields, so `checkStorageMismatches` findings stay failures with no edit. No positional literal exists that adding a field could have broken.
- **The mock expectations genuinely bind**: `pgxmock.NewPool()` is ordered, and a SQL mismatch hard-fails rather than being swallowed — the new tests cannot pass vacuously.
- **An unregistered `schema_id` still fails**: the zero-value `schemaLedger` has nil maps, so the row falls through to the failure branch exactly as before the refactor.

### Whole-branch review

Zero Critical. Zero findings in the Go code — tasks 1 and 2 were signed off as-is. Two Important findings, both in the documentation deliverable, both fixed in `a7829a0`:

1. **The guide stated three mutually inconsistent counts of the same taxonomy.** The pre-existing lead-in said "two very different reasons", the new paragraph said "three things", and the Deployment Checklist said "one of two things" — and the checklist's two were *incomplete*, omitting accidental metadata loss for a still-legitimate attribute, which is exactly the case where the operator must not touch the rows. Compounding it, "the determination is yours in exactly one situation: removed before #342" stated a *conclusion* of the determination as its precondition, nudging the operator toward hand-adding a retired ledger entry — which, done for genuinely junk rows, permanently masks them as informational. `### Unknown attribute IDs in EAV` was rewritten: three labelled cases, the tool's automatic classification stated up front, the manual determination scoped to what the validator still reports as a *failure*, and the checklist extended to all three outcomes. Case (i) stays deletion-only, case (ii) keeps its "do not delete" force, and the "if you cannot establish which case applies, treat the rows as case (ii)" safety default survives.
2. **The ledger-rebuild recipe restored only part of the guard it promised.** It required `attributeID`/`valueType`/`items_type` and then claimed "the reuse guard protects the id" — but #342 also reserves `column_binding.col_name` for retired entries. An operator rebuilding an entry for an attribute that had a hot column would get id protection but not column protection: a later attribute could bind that main-table column with the guard silent and read the retired attribute's stale values. The recipe now requires `column_binding.col_name`, explains that nothing derives it so it must come from version history, and notes that the folded parquet column needs no separate field because it is name-derived.

Also folded into that commit: `docs/error-handling.md` now names the accessor as the one sanctioned exception to "retired entries never reach a consumer", and Recommended Migration Flow step 5 gets the same do-not-delete guard the branch had given the structurally identical Rollback step 3.

Scoped re-review of the fix: all four findings ADDRESSED; the quoted tool output verified byte-identical against the emitting code rather than accepted on the implementer's word; fences balanced; all in-document anchors resolve; no `.go` file touched.

### Verification

`make lint` (pinned golangci-lint v1.64.8) and `make test` clean at head. The end-to-end run against real Postgres and the shipped ledgers is in the PR description — it exercises all three classification arms and both exit codes.

### Deferred / parked

Not blocking merge, recorded so they are not lost:

- **Parked (one-word polish).** `docs/schema-consistency-migration.md`: "startup will fail naming both" now follows a three-item antecedent. Ruling: the claim is accurate — every guard message names two attributes — so this reads imprecisely rather than wrongly.
- **Follow-up candidate.** Pin the validator's informational output shape as a contract. The guide now quotes it verbatim, which makes it one, but the tests assert the detail line as separate substrings, so field order and the `schema_id=` component are unpinned; nor is failures-before-informational ordering pinned. Worth one issue covering all three.
- **Minor, deferred.** `buildSchemaLedgers` and `checkStorageMismatches` now run the same `ListSchemas → GetSchemaID → GetSchemaCache` triple; they build different structures, so extraction is not obviously worth it today, but a third consumer would change that.
- **Minor, deferred.** `RetiredAttributeIDs` is exported, so "validation tooling only" is convention; the narrow return type is the real guard. If a second caller ever appears, that is the moment to revisit the seam rather than widen the type.
- **Minor, deferred.** No test pins the accessor on a *known* schema with zero retired entries — the nil case is covered only via the unknown-schema-id arm.

iceboundrock (commented):
## Review — #377 fix(tools): classify #294-preserved EAV rows as informational

Two-axis review (Standards vs Spec) run against `origin/main...origin/fix/341-classify-preserved-eav-rows`, with each axis an independent sub-agent, then re-verified first-hand (tests re-run on the branch, doc claims checked against the emitted text).

---

### Findings

**Zero Critical. Zero Important.** No hard violation of a documented standard, and no spec requirement missing or misimplemented.

The only items are judgement calls, both in `docs/schema-consistency-migration.md`:

1. **"What Changed" front-matter bullet (line 14)** — the bullet "EAV rows whose `attr_id` is not present in metadata" still sits under the heading "The new checks fail on:" while its own text now says the runtime path no longer errors on these since `#294`. It is explicitly qualified as historical ("as originally shipped in `#121`") and cross-references the rewritten taxonomy section, so it is defensible — but at face value the heading/bullet pair reads as a live failure, which is the exact confusion the issue set out to remove. Suggested polish: move the qualifier into the lead-in or reword the heading to "The new checks introduced:".

2. **Case-(ii) label reused at two scopes (lines 171–201)** — the intro defines case (ii) as the whole "preserved by attribute removal (`#294`)" class; the "Determine which case you are in" paragraph narrows case (ii) to the pre-`#342` *unledgered* subset ("three possibilities remain at that point"), while the Case (ii) heading spans both. This is *not* a contradiction: the paragraph scopes itself with "The determination is yours whenever the validator reports a **failure**" and the Case (ii) body explains "reaches you as a *failure* only when… removed before `#342`". Readable, but the same label doing double duty is a mild cognitive load for the operator this doc is written for. Nit only.

### Spec

**Verdict: faithful.** Read the full approved design record (`spec341.md`) and diffed against it.

- `RetiredAttributeIDs(schemaID int16) map[int16]string` — implemented exactly: copy-on-read under `RLock`, nil for unknown/empty, ids+names only, no `forma.AttributeMetadata` and not the live map. Invariant guard intact (verified: retired entries stay out of `GetSchemaCache`/`GetSchemaCacheByID`/`attributeMetadata`).
- Both capture sites populated after `validateSchemaAttributeCache`: `RegisterSchema` reads the un-stripped `cache` argument (`loader.go:90`); `loadAttributeMetadataFromFiles` reads `schemaCache` (`loader.go:369`). `file_registry.go` stripping points untouched, as spec'd.
- Validator three-way classification, per-id not per-schema amnesty, unknown-`schema_id` falls through to failure via the zero-value ledger — all present and correct.
- Two-tier output: `severity` field on `validationIssue` (zero-value `severityError` keeps all 4 existing construction sites failing), `report` helper prints failures block then informational block, exit code and `%d issue(s)` count failures only; success line keeps the pinned "schema consistency checks passed for N schema(s)" prefix and appends the informational count.
- All six spec'd tests present and passing (accessor visibility + absence from consumer caches, copy-on-read, retired-only→exit 0 + no "unknown attribute IDs", unrelated third id still fails, mixed blocks with 1 issue(s), unknown schema_id fails).
- Non-goals respected: no `--fail-on-retired`, no `--allow-attr-ids`, `checkStorageMismatches` untouched.
- Docs: all six spec'd items (front-matter, case (i) hoist, Rollback step 3 never-delete clause, Go Validator checklist bullet, Deployment Checklist rewrite, ledger-reconstruction recipe requiring `column_binding.col_name`) are present and match.

Minor, non-blocking scope notes:
- `docs/error-handling.md` was edited though the spec's doc list names only `schema-consistency-migration.md` — defensible (it is the doc that states the invariant the accessor carves out), and it matches the PR description's stated scope.
- More restructuring than the one named `report` helper (also `partitionIssues`, `schemaLedger`, `buildSchemaLedgers`, `classifyUndecodableRows`) — justified by the 100-line function cap; each is small and single-purpose.
- Line-count estimate in the spec said ~380; the file is 404. Within the 500 cap.

### Standards

**Verdict: compliant.** (coding-standard.md + AGENTS.md + docs/error-handling.md)

- Caps hold: `validate_schema_consistency.go` 404 ≤ 500; largest function 70 lines (`checkStorageMismatches`) ≤ 100; all new functions 12–32 lines.
- Error wrapping: no bare `return err` introduced. `run` returns a fresh `fmt.Errorf("schema consistency validation failed with %d issue(s)", …)` (nothing to wrap); scan/iterate errors keep `%w` context. Pre-existing bare returns in `collectIssues`/`checkStorageMismatches` are untouched, not introduced.
- Flat structure: max nesting 2 in new code, early returns throughout.
- No unrelated rewrites: the case-(i) remediation edit is required by the spec's doc items, not scope creep.
- Judgement calls (see Findings): the two doc items above; `retiredAttributeIDs` is noun-first while the standard prefers verb-first helpers — but it deliberately mirrors sibling `activeAttributeCache`, matching local convention.

### Assessment

Both axes pass. The implementation is a faithful, minimal realization of the spec; the invariant guard (narrow return type) is real and verified, not conventional. The worst issue is a one-line front-matter framing judgement call in the docs — not a defect. The three findings the branch's own review record fixed in `a7829a0` are all addressed as claimed; the quoted tool output I verified against the emitting code rather than taking on trust.

### Observations

- `RetiredAttributeIDs` is exported, so "validation tooling only" is convention; the narrow return type is the real guard. Worth revisiting only if a second caller appears — agreed with the branch's own deferral.
- No test pins the accessor on a known schema with zero retired entries (nil case is covered only via the unknown-schema-id arm) — agreed, minor.
- The "informational finding(s)" suffix pluralizes oddly for 1 ("1 informational finding(s)") but matches the existing "schema(s)" convention; acceptable.
- `buildSchemaLedgers` and `checkStorageMismatches` both run the `ListSchemas → GetSchemaID → GetSchemaCache` triple; not worth extracting today.

### Verification

- Re-ran the branch's tests first-hand: `go test ./cmd/tools ./internal/schemameta -run 'ValidateSchemaConsistency|Retired|Unknown'` — all pass (9 schemameta + 4 tools relevant tests).
- Function/file caps and error wrapping verified by direct inspection.
- CI green on the PR: Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke all SUCCESS.

### Summary

Standards: 0 hard violations, 2 doc judgement calls. Spec: 0 missing, 0 wrong, minor scope additions. No blocking issues — safe to merge. Follow-ups worth tracking separately (all non-blocking): pin the informational output shape + block ordering as a test contract; optional one-line front-matter rewording (Finding 1).

iceboundrock:
## Response to the review

Both findings addressed in `42ed519`; one follow-up filed. Per-item below, including the ones I did not act on and why.

### Finding 1 — front-matter bullet under a "fail on" heading — **fixed**

Agreed on the defect: at face value the heading/bullet pair reads as a live failure. I did not take either suggested wording, for a reason worth stating.

Rewording the heading to "The new checks introduced:" would have weakened all six bullets to make room for one exception — the other five genuinely do still fail, and blurring them costs more than the bullet gains. Moving the qualifier into the lead-in has the mirror problem: the lead-in covers the whole list, so a caveat about one bullet reads as a caveat about all of them.

What landed instead scopes the heading to its release and pulls the exception out of the list entirely:

- heading → "The checks `#121` introduced fail on:"
- bullet 14 → "…for that schema — **since superseded, see below**"
- a standalone paragraph after the list:

> **Superseded since `#294`.** The runtime path no longer errors on EAV rows whose `attr_id` is absent from metadata: it skips them on read and preserves them on write. Since `#341` the validator likewise reports the ones a `retired` ledger entry accounts for as informational rather than as failures. Only that one bullet changed; the rest of the list still fails as written. See [Unknown attribute IDs in EAV](#unknown-attribute-ids-in-eav).

The last sentence is doing real work: it is what stops a reader from generalising the exception across the list.

### Finding 2 — case (ii) at two scopes — **fixed**

Recorded as a nit, and I agree it was not a contradiction — both sites scope themselves. But the fix is one clause, so the cognitive load is not worth keeping. The determination paragraph now reads "(case (ii), *unledgered variant* — the ledgered majority of case (ii) never reaches you as a failure at all)". That names the subset and simultaneously tells the operator why the broad class is mostly absent from the failure list they are staring at.

### Observations — no action, agreed

- **Exported accessor, "validation tooling only" is convention.** Agreed, and deliberately so: the narrow return type is the enforcement. Widening it is the thing to refuse; a second caller is the signal to revisit the seam. Recorded in #378 as an explicit non-goal so a future reader does not "fix" it by adding a richer accessor.
- **`buildSchemaLedgers` / `checkStorageMismatches` share the `ListSchemas → GetSchemaID → GetSchemaCache` triple.** Agreed, not worth extracting at two consumers building different structures. Also recorded in #378 as a deliberate non-goal.
- **`1 informational finding(s)`.** Agreed acceptable — it matches the existing `schema(s)` convention in the same line, and diverging would make the two halves of one sentence inconsistent.

### Scope notes — no action, with reasoning

- **`docs/error-handling.md` edited though the spec named only the migration guide.** Deliberate and, I would argue, required rather than merely defensible: that file states the invariant "retired entries… never reach a consumer", and this PR creates the one sanctioned exception to it. Landing the accessor while leaving that sentence unqualified would have shipped a false statement in the document that defines the invariant.
- **More helpers than the one named `report`.** `partitionIssues`, `schemaLedger`, `buildSchemaLedgers`, `classifyUndecodableRows` — driven by the 100-line cap and by wanting the three-way classification to be one readable function rather than three nested conditionals inside a row loop.
- **404 lines vs the spec's ~380 estimate.** Noted; the estimate was mine and it was low. Within the cap.
- **`retiredAttributeIDs` is noun-first against the verb-first preference.** Intentional: it is the exact counterpart of `activeAttributeCache` and sits directly beneath it. Local symmetry beat the general rule here — the pair reads as "one strips, one keeps the books".

### Follow-up filed

**#378** — pin the validator's informational output shape and block ordering as a test contract. Covers the whole-line detail assertion (field order and the `schema_id=` component are currently unpinned by four independent substring checks), failures-before-informational ordering, the suffix's absence at zero notices, and the accessor's zero-retired-entries nil case. Test-only, no behaviour change. It also records the two non-goals above so they are not re-litigated.

Nothing else outstanding from the review. The branch is at `42ed519`.

iceboundrock:
## Review — re-review at `42ed519`

Whole-branch re-review of the current head (`42ed519`) against the spec in [#341#issuecomment-5226173089](https://github.com/Lychee-Technology/forma/issues/341#issuecomment-5226173089) and the repo's coding standards. Reviewed independently from the existing review record: I read the full diff and each changed file at head, ran the test suites, and verified the claims rather than accepting them.

## Findings

Zero Critical. Zero Important. Two nits, both non-blocking:

1. **Nit — `report`'s doc comment overstates its input contract.** `validate_schema_consistency.go:148-151` says "Informational findings are EAV rows #294 preserves under a retired attribute", but `report` is generic — it prints whatever `notices` slice it is given and knows nothing about #294. The comment reads as a contract on the function's input rather than a description of the current sole producer. One-line rewording ("the current producer classifies preserved retired-attribute rows as informational") would make it accurate without coupling the function to the producer.
2. **Nit — pre-existing bare `return err` at `run()`:124** (`return err` after `collectIssues`). Not introduced by this diff, but per AGENTS.md's "fix the immediate area" a `fmt.Errorf("collect issues: %w", err)` would have been in scope since this diff rewrote `run`. Low priority; either fix inline or leave and it is fine.

Both are below the bar for blocking. Everything else checked out.

## Spec

Faithful implementation. All eight verification points from the spec pass:

- **Accessor** — `RetiredAttributeIDs` returns `map[int16]string`, copy-on-read under `RLock`, nil for unknown/empty schema (`loader.go:200-218`); populated at both registration points from the **un-stripped** cache strictly after `validateSchemaAttributeCache` (`loader.go:81→97`, `:360→369`); map initialised in `NewMetadataCache` (`loader.go:64`).
- **`file_registry.go` untouched** — the diff touches exactly the 7 listed files.
- **Three-way classification, per-id** — `classifyUndecodableRows` (`validate_schema_consistency.go:269-287`); unknown `schema_id` yields the zero `schemaLedger` (nil maps) → falls to the failure branch, pinned by `TestValidateSchemaConsistencyFailsUnknownSchemaID`.
- **`severity` field with `severityError` as zero value** — load-bearing and safe: all four `validationIssue` construction sites use named fields, so `checkStorageMismatches` findings stay failures with no edit (`validate_schema_consistency.go:29-43`).
- **Two-tier `report` + exit semantics** — failures first, then informational; error fires only when failures exist and counts `len(failures)` only; success line keeps the `schema consistency checks passed for N schema(s)` prefix and appends the informational count (`validate_schema_consistency.go:152-170`).
- **Attribute name in informational lines** — `attribute=%s` present (`:276`).
- **Tests** — both schemameta pins (visible via accessor yet absent from the consumer caches; copy-on-read mutation test) and all four cmd/tools pins (retired-only → exit 0, no "unknown attribute IDs"; unrelated id → still fails; mixed → fails with `1 issue(s)` + both blocks; unknown `schema_id` → still fails) are present and green.
- **Docs** — all six items present (front matter qualifier + superseded callout, case (i) deletion-only + new case (iii), step-3 never-delete clauses in both places, Go Validator checklist bullet, Deployment Checklist rewrite, ledger-reconstruction recipe with the `column_binding.col_name` requirement and side-effect note).

Two judgment-call deviations, both coherent rather than creep:

- `docs/error-handling.md` is edited though the spec named only the migration guide. **Correct, not creep**: the accessor makes the existing "never reach a consumer" sentence false, and the 5-line edit (`error-handling.md:329-334`) is the minimal fix that keeps that doc truthful. Landing the accessor without it would have shipped a false statement in the document that defines the invariant.
- Spec says populate "before `activeAttributeCache`" in `loadAttributeMetadataFromFiles`; the implementation stores `retiredAttrIDs` at `loader.go:369` *after* the stripped caches (`:367-368`). Unobservable — same critical section — but it reverses the spec's wording. Note only, no change needed.

## Standards

Pass, per AGENTS.md / `coding-standard.md`:

- **Line counts** — `validate_schema_consistency.go` is 404/500 (the spec's ~380 estimate was low but within the cap); all new functions ≤100 lines (`buildSchemaLedgers` 22, `classifyUndecodableRows` 13, `report` 19, `partitionIssues` 10). Loader 390, validation.go 132, tests within cap.
- **Error wrapping** — all new error paths contextual; `run():132` returns a fresh `fmt.Errorf` with no `%w` and no cause, which is correct here. The one bare `return err` (`:124`) is pre-existing.
- **Invariant upheld** — the "no consumer may see retired entries" contract is preserved and now enforced by type shape (no `forma.AttributeMetadata`, not the live map), verified by grep: exactly two writers of `retiredAttrIDs` (`loader.go:97`, `:369`) and one reader (`:209`). The strip-guard tests still green.

## Assessment

This is a well-scoped fix that does exactly what the spec asked and no more. The narrow accessor is the right seam: it gives validation tooling the one thing it needs (id → name) while making it structurally impossible to reconstitute a retired `forma.AttributeMetadata` back into the read/write/flush/projection path. The exit-code change is deliberate, documented, and (verified) nothing in CI gates on it — the only callers are the Makefile target and the migration guide. The three-way classification is genuinely per-attributeID, not a per-schema amnesty, and the unknown-`schema_id` behavior is byte-identical to pre-refactor semantics. Ready to merge.

## Observations

- **"Nine unit tests added" in the PR description is off by three.** The diff adds six new `Test*` functions: four in `cmd/tools`, two in `internal/schemameta` (`git diff | grep -E '^\\+func Test'`). Not a defect — the count in the description is simply wrong.
- The informational line carries the schema name (`schema=...`) while the failure line does not. Matches prior failure output, so no regression — noting for symmetry only.
- Agreed with the parked items in the earlier review record: the "startup will fail naming both" one-word polish (accurate, reads imprecise after the three-item antecedent at `schema-consistency-migration.md:410-413`) and the three follow-up candidates (informational-output contract pin → #378, the shared `ListSchemas→GetSchemaID→GetSchemaCache` triple, the exported-accessor seam).

## Verification

Independently run at head `42ed519`:

- `make lint` (golangci-lint v1.64.8) — **exit 0, clean**.
- `go test . ./cdc ./cmd/... ./factory ./internal/...` (the `make test` set) — **exit 0, all 32 packages ok**.
- `go test -race ./cmd/tools ./internal/schemameta` — **clean**.
- All six new test functions pass (`-run` for both `Retired` and `ValidateSchemaConsistency`), including the ordered pgxmock expectations (a SQL mismatch hard-fails rather than passing vacuously).
- Doc anchors referenced in the new text (`#unknown-attribute-ids-in-eav`, `#removing-an-attribute-342`) resolve against the headings in `docs/schema-consistency-migration.md`.

## Summary

Spec ✅, standards ✅, 0 Critical/Important, 2 non-blocking nits (a doc comment on `report` that overstates its input contract, and a pre-existing bare `return err`). Merge as-is.

Issue #379, opened 2026-08-08
cdc: wrap the remaining bare error returns in internal/cdc


iceboundrock:
Follow-up from #276, which wrapped only the two export entry points by explicit scope ruling (`ExportSnapshotToTmp` in `internal/cdc/duckdb_exporter.go` and its twin `ExportBaseFileToTmp` in `internal/cdc/init_exporter.go`).

Seven bare `return err` sites remain in `internal/cdc`, each dropping the operation context `coding-standard.md` requires:

- `internal/cdc/flusher.go:65,71,91,117`
- `internal/cdc/flush_batch.go:212`
- `internal/cdc/init.go:318`
- `internal/cdc/runner.go:166`

Each needs `fmt.Errorf("<operation> ...: %w", err)` naming what the caller was doing. Check whether the callee already names the schema before repeating it — the #276 wraps deliberately named the *operation* (`export snapshot to tmp for schema %d`) because the plan builders already name themselves.

Scope note: the repo-wide count of bare returns outside tests is 92. This issue covers `internal/cdc` only; a repo-wide sweep would need its own decision.

Gate: `make fmt-check && make lint && make test`.

Refs #276.

iceboundrock:
Adding one cosmetic item surfaced by the #380 review, so a sweep here can normalize the whole chain at once rather than one layer at a time.

The export paths now nest operation + schema twice:

```
export snapshot to tmp for schema 7: failed to build delta export SQL plan for schema 7: export snapshot: no row ids provided
```

Three layers each add context and the schema is named twice. Nothing is wrong — every layer satisfies the wrap rule and the meaning is right — but a sweep of this package is the natural moment to decide the layering convention: the entry point names the caller operation, the builder names the schema, the innermost names the failure. Under that convention `buildExportSQL`'s middle wrap (`failed to build delta export SQL plan for schema %d`, from #274's review commit `c04a5b1`) is the redundant one, since `ExportSnapshotToTmp` is its only caller.

Deliberately not changed in #380: it would churn #274's wording for cosmetics, and the entry-point wraps were the scoped deliverable there.

iceboundrock:
# Follow-up status — parent PR #380 merged

PR #380 (`fix/276-clock-ordering-cleanup`) merged `2026-08-08` via `b31f026`, closing #276. This issue remains the tracking item for the **seven bare `return err` sites in `internal/cdc`** still outstanding (scope re-confirmed on merge):

[...]

PR #380, opened 2026-08-08
fix: #276 deterministic clock ordering for the remaining LWW guard, export error wraps, and a real gofmt gate

### LINKED ISSUE #276: test/cdc cleanup: deterministic clock ordering sweep for LWW probes + remaining bare-error/gofmt debt

iceboundrock:
Closes #276. Refs #204, #210, #274, #275, #280, #290; follow-up #379.

Design, plan, and the mid-execution gofmt finding are on the issue: [design](https://github.com/Lychee-Technology/forma/issues/276#issuecomment-5227915070), [plan](https://github.com/Lychee-Technology/forma/issues/276#issuecomment-5227947663), [finding](https://github.com/Lychee-Technology/forma/issues/276#issuecomment-5228378227).

## Why this is smaller than the filed checklist

The issue was filed 2026-07-17. `#274` has landed since and made row versions **strictly monotonic per row in the write path** — `ltbase_updated_at = GREATEST($now, ltbase_updated_at + 1)` with `RETURNING` (`postgres_persistent_repository_main_table.go:133`), `computeTombstoneStamp` for deletes, `nextRowVersion` for recreates, and `change_log.changed_at` stamped from that same effective value. A successor version of the **same** `row_id` therefore cannot tie its predecessor regardless of clock resolution.

Auditing the eight listed `assertStrictlyNewer` call sites against that (line numbers had drifted):

| Site | Guard compares | Disposition |
| --- | --- | --- |
| `filter_lww_e2e_test.go:59,107` | same row, v1 → v2 | no change — `#274` guarantees it |
| `filter_lww_matrix_e2e_test.go:93` | same row, v1 → v2 | no change |
| `filter_lww_tombstone_e2e_test.go:32` | same rows, live → tombstone | no change (`computeTombstoneStamp`) |
| `filter_lww_tombstone_e2e_test.go:86` | same row, v1 → v2 | no change |
| `keyset_lww_e2e_test.go:75,222` | same row, v1 → v2 | no change |
| `keyset_lww_e2e_test.go:132` | `[a,b,c]` vs `[b,c,upd]` — **cross-row**, and `b`/`c` shared one `ApplyEvents` batch | **fixed** |
| `sort_stability_e2e_test.go:110` (issue omitted it) | same rows, reverse-order updates | no change; kept in the regression set as evidence |

Rather than edit seven sites the write path already covers, the invariant is now written into `assertStrictlyNewer`'s doc comment: which pairs are ordered by construction, and which two shapes (different `row_id`s; comparison against an independent anchor) still need an explicit wait.

Item 3 of the checklist was already fixed by `#274`'s review commit `c04a5b1`.

## What changed

**1. `keyset_lww_e2e_test.go` — the one real race.** `testKeysetCreatedAtCursorResurrection` needs `t1 < t2 < t3 < t4` across four writes on three rows, and its cursor boundary is literally `c.ChangedAt` with `B` expected to qualify strictly below it. `b` and `c` were created in one `ApplyEvents` call, where each `Create` takes its own clock read. They are now applied one per call with `waitClockPast` between, so a tie cannot silently change what the probe asserts.

**2. `cdc_concurrent_flush_e2e_test.go` — the observed CI flake.** `testUpdateBeforeExport`'s hook mutates a row and then asserts `update.ChangedAt > hookSnapshot`, where the snapshot is the flush's own independent clock read — nothing in `#274` orders a write against it. The hook now waits past the snapshot first. New `waitClockPastMillis(ms int64)` takes no `*testing.T` because a CDC hook reports failure through the flush error.

The `runPausedFlush` siblings (`:257`, `:310`, `:360`) are deliberately untouched: their pause point is the first `CopyObject`, i.e. after a full DuckDB export plus S3 tmp `PUT`, and their guard legs are same-row.

**3. Both `Export*ToTmp` entry points** now wrap their plan errors (`export snapshot to tmp for schema %d` / `export base file to tmp for schema %d`). The issue named only the delta side; wrapping one caller while leaving its byte-identical twin bare would rebuild the asymmetry this issue exists to clear. The seven remaining bare returns in `internal/cdc` (`flusher.go:65,71,91,117`, `flush_batch.go:212`, `init.go:318`, `runner.go:166`) are #379; repo-wide the count is 92, which needs its own decision.

**4. gofmt gate — and the item was not cosmetic.** The issue called the drift "comment-alignment"; that holds for only 2 of the 7 files. For the other 5, gofmt's doc-comment printer rewrites `''` into a typographic quote, corrupting documentation about SQL literals: the doubled-quote DSN forms `redact_test.go` exists to pin (#290) and the empty-string `array_indices` marker (#204). Those files could never have been gofmt-clean without editing prose — which is why the drift survived. The nine sites now keep the literal in a doc-comment code block (verified preserved) or name it in prose, so the tree is a genuine gofmt fixed point.

The gate is `make fmt-check` (`gofmt -l`), **not** golangci's `gofmt` linter, because that linter is build-tag scoped: it never loads `//go:build e2e` files, where 3 of the 7 drifted files lived. Pulling them in with `--build-tags=e2e` surfaces 6 unrelated pre-existing findings (`federated/cdc_flush_test.go` errcheck ×3, `suite_test.go` unused ×2, `consistency_test.go` staticcheck SA4010 ×1). `gofmt -l` is purely syntactic and therefore tag-agnostic. CI runs the same make target; the v1.64.8 pin and local/CI parity are untouched.

**5. `AGENTS.md`** — the Specs and Design Docs rule now covers every non-code artifact of an issue (drafts, investigation and verification notes, mid-execution rulings), posted when produced rather than after approval.

## Verification

- `make fmt-check && make lint && make test` — all pass.
- `go test -tags=e2e -run 'TestFilterLWW|TestKeysetLWW|TestConcurrentFlushSnapshot|TestSortStability|TestWaitClockPastMillis' ./internal/e2e_harness/production` — 19/19 pass, including the untouched suites that the audit claims need no wait.
- `make test-e2e-production` — 125 pass, 0 fail.

**Mutation verification (item 2).** A green run proves nothing about a race, so the guard was driven red on purpose: with the hook's anchor moved to `snapshot + 50` and the wait removed, `TestConcurrentFlushSnapshot/UpdateBeforeExport` fails at `cdc_concurrent_flush_e2e_test.go:207` — `concurrent update changed_at 1786226477115 must be > snapshot 1786226477164`, 49ms behind. Restoring the wait alone, same anchor, turns it green. Both probe lines were then reverted.

**Measurement (item 1).** Instrumenting the chain on the *unfixed* code across 5 runs gave `b→c` gaps of 2–3ms and `c→upd` gaps of 3–4ms — the ordering held entirely on per-event transaction latency, not by construction. No zero gap reproduced locally, and the post-fix gaps are unchanged (2–4ms) because the waits are inert whenever the clock has already advanced. The value of the change is removing the dependence on that margin, not widening it; under a loaded CI runner the margin is what closes.

**Helper mechanism** is pinned directly by `TestWaitClockPastMillis` (`clock_guard_test.go`): returns only once the clock is strictly past the anchor, and returns immediately for a past one. It starts no container (the shared cluster is lazy) and runs in ~10ms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,9 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Check gofmt
+        run: make fmt-check
+
       - name: Run golangci-lint
         run: |
           go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,6 +10,7 @@ Forma is a general-purpose data management system on PostgreSQL. Entities are de
 
 ```bash
 make test          # unit tests (wraps: go test . ./cdc ./cmd/... ./factory ./internal/...)
+make fmt-check     # gofmt gate over git-listed Go files — CI runs it ahead of the linter
 make lint          # golangci-lint, pinned to v1.64.8 — same as CI; do not upgrade the pin
 make coverage      # unit tests + coverage report in build/
 make build-all     # build server/tools/sample into build/ with platform symlinks
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,9 @@ Error messages must name the logical value type, the offending column/attribute,
 
 ## Specs and Design Docs
 
-Planning artifacts — specs, design docs, implementation plans — are not committed to the repo (`docs/superpowers/` is gitignored). Post them to the GitHub issue they serve instead: once a spec or design is approved, add its full content as a comment on the relevant issue so the issue carries the complete decision record. This covers per-issue planning artifacts only; reference designs that document the system itself (e.g. `docs/federated-query/design.md`) live in `docs/` as before.
+Every non-code artifact produced while working an issue goes to that issue as a comment — never into the repo (`docs/superpowers/` is gitignored). This covers design drafts, specs, implementation plans, investigation and verification notes, and rulings made mid-execution. Post the full content, not a summary, and post it when it is produced: a design draft goes up as a draft (say so), a plan goes up when written, a mid-execution decision goes up at the moment it is taken. The issue is the complete decision record; nothing load-bearing may live only in a chat transcript or a local file.
+
+Scope: per-issue artifacts only. Reference documentation of the system itself (e.g. `docs/federated-query/design.md`) lives in `docs/` and is committed as before. Review artifacts belong on the PR — see Pull Request Rules.
 
 ## Pull Request Rules
 
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,22 @@ test-e2e-production:
 # //go:build e2e harness files, where 3 of this issue's 7 drifted files lived.
 # gofmt -l is purely syntactic and therefore tag-agnostic, so it is the only
 # check that covers the whole tree. CI runs this exact target.
+#
+# The file list comes from git, not a directory walk: tracked files plus
+# untracked ones git would not ignore. Walking the checkout would also inspect
+# gitignored directories (docs/superpowers/, build/, .gocache/), so a local run
+# could fail on scratch files that do not exist in CI's fresh checkout —
+# breaking the local/CI parity this gate exists to protect. Untracked-but-not-
+# ignored files stay in scope so a new file is caught before it is committed;
+# CI has none, so the two runs see the same set.
 fmt-check:
 	@echo "Checking gofmt..."
-	@unformatted=$$(gofmt -l .); \
+	@files=$$(git ls-files --cached --others --exclude-standard '*.go'); \
+	if [ -z "$$files" ]; then \
+		echo "fmt-check found no Go files — not a git checkout?"; \
+		exit 1; \
+	fi; \
+	unformatted=$$(gofmt -l $$files); \
 	if [ -n "$$unformatted" ]; then \
 		echo "Not gofmt-clean:"; \
 		echo "$$unformatted"; \
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test test-unit test-e2e-production lint coverage build build-tools build-benchmark benchmark-smoke benchmark-regression benchmark-heavy benchmark-heavy-live build-all build-lambda clean all create-build-dir link validate-schema-consistency
+.PHONY: test test-unit test-e2e-production fmt-check lint coverage build build-tools build-benchmark benchmark-smoke benchmark-regression benchmark-heavy benchmark-heavy-live build-all build-lambda clean all create-build-dir link validate-schema-consistency
 
 # Binary names
 BINARY_SERVER=server
@@ -48,6 +48,21 @@ test-unit:
 test-e2e-production:
 	@echo "Running production E2E harness tests..."
 	@$(GOENV) go test -v ./internal/e2e_harness/production/ -tags=e2e -timeout=10m
+
+# Formatter gate (#276). golangci-lint's default linter set carries no
+# formatter, and its gofmt linter is build-tag scoped: it never loads the
+# //go:build e2e harness files, where 3 of this issue's 7 drifted files lived.
+# gofmt -l is purely syntactic and therefore tag-agnostic, so it is the only
+# check that covers the whole tree. CI runs this exact target.
+fmt-check:
+	@echo "Checking gofmt..."
+	@unformatted=$$(gofmt -l .); \
+	if [ -n "$$unformatted" ]; then \
+		echo "Not gofmt-clean:"; \
+		echo "$$unformatted"; \
+		echo "Fix with: gofmt -w ."; \
+		exit 1; \
+	fi
 
 # Run linter (same as CI lint job)
 lint:
--- a/internal/cdc/duckdb_exporter.go
+++ b/internal/cdc/duckdb_exporter.go
@@ -80,7 +80,7 @@ func NewDuckExporter(ctx context.Context, cfg CDCConfig, s3AccessKey, s3Secret, 
 func (e *DuckExporter) ExportSnapshotToTmp(ctx context.Context, cfg CDCConfig, pgConnStr string, s3TmpPath string, schemaID int16, snapshotTS int64, rowIDs []uuid.UUID, attrCache forma.SchemaAttributeCache) error {
 	sql, clQuery, mQuery, eQuery, err := buildExportSQL(pgConnStr, s3TmpPath, cfg, schemaID, snapshotTS, rowIDs, attrCache)
 	if err != nil {
-		return err
+		return fmt.Errorf("export snapshot to tmp for schema %d: %w", schemaID, err)
 	}
 
 	return e.executeExportSQL(ctx, cfg, "duckdb export sql", "duckdb copy exec", sql, clQuery, mQuery, eQuery)
--- a/internal/cdc/duckdb_exporter_sql_test.go
+++ b/internal/cdc/duckdb_exporter_sql_test.go
@@ -1,6 +1,7 @@
 package cdc
 
 import (
+	"context"
 	"strings"
 	"testing"
 
@@ -208,4 +209,30 @@ func TestBuildExportSQL_UsesCustomTableNames(t *testing.T) {
 		t.Fatalf("eav query not using custom table name: %s", eQuery)
 	}
 	_ = sql
+}
+
+// TestExportToTmpWrapsPlanErrors pins the caller-side context on both export
+// entry points: a plan failure must name the operation the caller was
+// performing, not arrive as a bare pass-through (#276, coding-standard.md).
+// Empty rowIDs is the cheapest plan failure — it returns before any DB or S3
+// use, so a zero-value DuckExporter is enough.
+func TestExportToTmpWrapsPlanErrors(t *testing.T) {
+	exporter := &DuckExporter{}
+	cfg := CDCConfig{}
+
+	t.Run("snapshot export", func(t *testing.T) {
+		err := exporter.ExportSnapshotToTmp(context.Background(), cfg,
+			"postgres://user@host/db", "s3://bucket/7/_tmp/f.parquet", 7, 0, nil, testAttrCache())
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "export snapshot to tmp for schema 7")
+		require.Contains(t, err.Error(), "no row ids provided")
+	})
+
+	t.Run("base export", func(t *testing.T) {
+		err := exporter.ExportBaseFileToTmp(context.Background(), cfg,
+			"postgres://user@host/db", "s3://bucket/base/7/_tmp/f.parquet", 7, nil, testAttrCache())
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "export base file to tmp for schema 7")
+		require.Contains(t, err.Error(), "no row ids provided")
+	})
 }
--- a/internal/cdc/init_exporter.go
+++ b/internal/cdc/init_exporter.go
@@ -14,7 +14,7 @@ import (
 func (e *DuckExporter) ExportBaseFileToTmp(ctx context.Context, cfg CDCConfig, pgConnStr string, s3TmpPath string, schemaID int16, rowIDs []uuid.UUID, attrCache forma.SchemaAttributeCache) error {
 	sql, mQuery, eQuery, err := buildBaseExportSQL(pgConnStr, s3TmpPath, cfg, schemaID, rowIDs, attrCache)
 	if err != nil {
-		return err
+		return fmt.Errorf("export base file to tmp for schema %d: %w", schemaID, err)
 	}
 
 	return e.executeExportSQL(ctx, cfg, "duckdb base export sql", "duckdb base copy exec", sql, "", mQuery, eQuery)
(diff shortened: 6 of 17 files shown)


iceboundrock:
## Review — PR #380 (`fix/276-clock-ordering-cleanup`)

> Note on numbering: **#379 is the follow-up *issue*** (the 7 remaining bare returns in `internal/cdc`), not this PR. This review targets PR #380, which closes #276. The two items you flagged — the gofmt/SQL-literal comment corruption and the keyset race held up by a transaction-latency margin — are both here.

Two-axis review against the full #276 decision record (issue body + design draft + implementation plan + mid-execution gofmt ruling + verification note).

## Findings

No hard violations on either axis. Both axes are independently clean; the findings below are judgement calls and residual risks, not blockers.

1. **`make fmt-check` scans gitignored directories.** The target runs `gofmt -l .` over the whole checkout, including `docs/superpowers/`, `build/`, `.gocache`. A local run with stray `.go` files in those dirs can diverge from a clean-checkout CI run — quietly undermining the "local `make lint` byte-identical to CI" property the gate is meant to protect. Consider `gofmt -l $(git ls-files '*.go')` or scoping the walk to tracked files.
2. **AGENTS.md adds a load-bearing gate but not the command.** The same PR widens the Specs-and-Design-Docs rule but doesn't add `fmt-check` to the Commands list in AGENTS.md, so a future agent reading AGENTS.md won't know the gate exists. Minor doc drift for a now-load-bearing target.
3. **Cosmetic message nesting.** The new wraps nest operation+schema twice, e.g. `export snapshot to tmp for schema 7: failed to build delta export SQL plan for schema 7: export snapshot: no row ids provided`. Meaning is right; the redundancy is cosmetic only.

## Spec

The diff faithfully implements the **re-scoped** spec. All four filed checklist items are disposed exactly as the design draft + mid-execution ruling prescribed, and every load-bearing claim in the PR body checked out against source:

- **Item 1 (waitClockPast sweep):** re-scoped to one real fix + documentation, as ruled. `keyset_lww_e2e_test.go:120-145` interleaves `apply a; flush; wait; b; wait; c; wait; upd` with no trailing wait needed (nothing compares against `upd` afterward), and `assertStrictlyNewer` is kept as the guard. The `#274` invariant (same-`row_id` successor strictly newer by construction) is written into the guard's doc comment; the seven dissolved sites get no mechanical edit.
- **Item 2 (UpdateBeforeExport snapshot race):** the hook waits on the **real** `snapshot` param (`waitClockPastMillis(snapshot)`, `cdc_concurrent_flush_e2e_test.go:179`); the `snapshot + 50` probe anchors were reverted as the verification note claims.
- **Item 3 (bare export return):** both `ExportSnapshotToTmp` and `ExportBaseFileToTmp` are wrapped with operation + schema context in the ruled wording; the remaining seven sites are correctly deferred to #379.
- **Item 4 (gofmt):** the mid-execution finding is real — gofmt rewrites `''` → `”` in doc-comment prose (verified on gofmt 1.26), so 5 of the 7 files could never be gofmt-clean without prose edits. All 9 sites are rewritten per the ruling; the SQL literals survive verbatim in tab-indented code blocks (verified below). The gate is `make fmt-check` (`gofmt -l .`), not golangci's build-tag-scoped `gofmt` linter, and CI runs the same target ahead of golangci-lint.

**Sanctioned scope deviations** (documented in plan/design, not unruled creep): wrapping `ExportBaseFileToTmp` alongside `ExportSnapshotToTmp` (one line, flagged in plan self-review); the item-2 sibling `assertStrictlyNewer` sites at 288/341 left untouched per the design's `runPausedFlush` justification (pause point is after a full DuckDB export + S3 PUT).

**Residual risks (noted, not blockers):**
1. No zero-gap reproduction — the keyset fix rests on reasoning plus a 2–3ms measured margin, honestly reported as such; a loaded runner is where the margin closes.
2. `waitClockPastMillis` is wall-clock + 1ms sleep; a scheduler/NTP stall between wait-return and the write's clock read can still theoretically tie a cross-row leg. Narrower than the original race, not eliminated.
3. The `#274` invariant claim's regression evidence is the green-but-untouched adjacent suites (`TestFilterLWW`, `TestSortStability`), which is deliberate but is assertion-by-prose.

## Standards

Compliant across the board; the diff is itself a compliance fix.

- **Error wraps** (`duckdb_exporter.go:83`, `init_exporter.go:17`): `fmt.Errorf("…for schema %d: %w", …)` is exactly the "never bare `return err`" rule (coding-standard.md, AGENTS.md). Error class is right per docs/error-handling.md: CDC flush-path, operator-visible — a plain wrap is owed, no `InvalidInputf` carrier, and no sentinel is wrapped at the new layer, so the redacted-4xx trap doesn't apply.
- **`TestExportToTmpWrapsPlanErrors`** uses `require.Contains(err.Error(), …)` — the substring match the repo mandates for published text. ~29 lines, within limits.
- **Makefile/CI**: `fmt-check` joins `.PHONY`, sits after setup-go and before golangci-lint; the v1.64.8 pin is untouched (AGENTS.md "do not upgrade the pin"). `docs/ci-cd-design.md` imposes no workflow-layout constraint.
- **Helpers** (`clock_guard_test.go`, `filter_lww_helpers_test.go:79`): extraction is DRY not duplication; `waitClockPastMillis` takes no `*testing.T` with the CDC-hook rationale documented. All touched files stay under the 500-line limit (largest touched: `full_type_parquet_e2e_test.go`, 452).
- **AGENTS.md change** is consistent with the Pull Request Rules section it cross-references.

## Assessment

Ship it. The two real fixes (T1 deterministic `created_at` chain, T2 snapshot-anchored hook wait) are sound and the second is mutation-verified red→green. The gofmt item is the hunk most worth your eyes — it is genuinely *not* cosmetic, the ruling is well-executed, and the gate closes the class of drift.

The one thing I'd weigh before merge: finding #1 above (`gofmt -l .` over untracked dirs) is the only point where the gate's local/CI parity claim has a hole. Low severity, easy to tighten.

## Observations

- The `''` → `”` rewriting is real: gofmt 1.26 rewrites doubled single quotes in doc-comment prose, and only tab-indented code blocks / single-quote-free prose survive — which is exactly the shape all 9 rewrites took.
- The keyset measurement reporting is exemplary: "no zero gap reproduced locally; ordering held on a 2ms transaction-latency margin" — no overclaiming, and the post-fix gaps being unchanged (2–4ms) is the correct signature of an inert wait.

## Verification

PR's own (verification note on #276): `make fmt-check && make lint && make test` PASS; targeted `-tags=e2e` run 19/19; `make test-e2e-production` 125/0.

Independently spot-checked on the branch (`/Users/ruoshi/code/Lychee/LTBase/forma-276` @ `7b2bd6d`):
- `make fmt-check` → exit 0 (genuine gofmt fixed point).
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/cdc -run TestExportToTmpWrapsPlanErrors -v` → both subtests PASS.
- `gofmt -d internal/cdc/redact_test.go` → clean; the literals `password=''`, `password=''''`, `password=''secret''` sit verbatim in tab-indented code blocks.

Not re-run here: the full e2e suites (Docker-backed; run by the author).

## Summary

Two axes, both clean: 0 hard standards violations, 0 material spec gaps (2 sanctioned scope deviations), 3 judgement-call findings, 3 residual risks — none blocking. Worst single issue: the `fmt-check` scan scope (judgement call, not a violation). Recommend merge after weighing finding #1.

iceboundrock:
# Review response — round 1

Pushed as `018b39b`. Findings 1 and 2 fixed; finding 3 kept as-is with reasoning and recorded on #379. Residual risks acknowledged below, no action taken on any.

## 1. `fmt-check` scanned gitignored directories — fixed

Confirmed: `.gocache/`, `build/` and `docs/superpowers/` are all gitignored (`.gitignore:49,51,79`), and `gofmt -l .` walked into them. That is exactly the local/CI divergence the gate exists to prevent, so it was worth closing rather than deferring.

The list now comes from git:

```make
files=$(git ls-files --cached --others --exclude-standard '*.go')
```

`--cached --others --exclude-standard` keeps untracked-but-not-ignored files in scope, which the suggested `git ls-files '*.go'` would have dropped — a file you just created should still be caught before you commit it, and CI has no untracked files, so both runs see the same set. The target now also fails loudly if the list comes back empty, so a non-git checkout cannot turn the gate into a silent no-op.

Verified four ways rather than by re-running the happy path:

| Scenario | Expected | Result |
| --- | --- | --- |
| clean tree | pass | pass |
| mis-formatted `docs/superpowers/scratch.go` (gitignored) | pass | pass — old form flagged it, new form does not |
| mis-formatted untracked `internal/scratch_probe.go` | fail | fail |
| mis-formatted tracked file | fail | fail |

The second row is the finding, shown directly:

```
old form (gofmt -l .) would have flagged:
docs/superpowers/scratch.go
new form flags:
(none)
```

## 2. `fmt-check` missing from the AGENTS.md command list — fixed

Added next to `make lint`, noting that CI runs it ahead of the linter. The omission was worse than ordinary doc drift given the same PR widens the artifact rule in that file.

## 3. Cosmetic message nesting — keeping it, recorded on #379

Not changed here, deliberately. Trimming the duplicate schema means editing `buildExportSQL`'s middle wrap, which is #274's review wording (`c04a5b1`) — churning it for cosmetics inside a PR whose scoped deliverable was the entry-point wraps is the wrong trade. It is now a note on #379 (issuecomment-5228535569), which owns the `internal/cdc` wrap sweep and is the right place to settle the layering convention across all three layers at once: entry point names the operation, builder names the schema, innermost names the failure. Under that convention the middle wrap is the redundant one, since `ExportSnapshotToTmp` is its only caller.

Say the word if you would rather have it trimmed in this PR instead.

## Residual risks — acknowledged, no action

1. **No zero-gap reproduction.** Agreed, and stated that way in the PR body and verification note rather than dressed up. A loaded runner is the only honest lever.
2. **`waitClockPastMillis` is wall-clock.** Correct and worth naming precisely: `time.Now().UnixMilli()` carries no monotonic guarantee across an NTP step, so a backwards jump between the wait returning and the write's own clock read could still tie a cross-row leg. There is no better anchor available — the repository stamps `changed_at` from the same wall clock, so a monotonic-source wait would be measuring a different clock than the one under test. Narrower than the original race, not eliminated; not worth engineering around at this severity.
3. **Invariant evidence is the adjacent green suites.** Accurate. The `#274` claim is proved at the source (`GREATEST`+`RETURNING`, `computeTombstoneStamp`, `nextRowVersion`) and the green untouched suites are corroboration, not proof. A test that *pins* the invariant — asserting two same-row writes inside one millisecond still order — would be the real evidence, and it belongs in the write-path package, not the e2e harness. Happy to file it if you want that pinned.

## Gate after the change

`make fmt-check` pass · `make lint` pass · `go test ./internal/cdc` pass. The Makefile/AGENTS.md change touches no test path; CI re-runs the full set on the push.

Issue #382, opened 2026-08-09
feat(federated): apply keyset cursors on the Postgres-only path


iceboundrock:
The deferred half of #354. That issue's body proposed applying the keyset predicate on the Postgres-only path as the preferred fix; the audit reclassified it as a feature rather than a bug fix, and #354 shipped the fail-closed/reroute split instead. This issue carries the feature.

**Sequence #381 before this one.** That issue settles what a cursor column may be and stops `col.Attribute` from reaching SQL unquoted — both of which get sharper the moment a caller-facing cursor surface exists, which is what this work introduces.

## Why it is a feature, not a port

Reaching parity with the DuckDB path requires all of:

1. **A `$n`-dialect keyset predicate generator.** The DuckDB generator emits positional `?` placeholders because DuckDB treats `$n` as an absolute parameter index and shifts every `?` after it (#161/#212). An older `$n`-emitting copy used to live in `internal/federated/keyset.go` with no production callers and was deleted in #217 precisely so the repo carries exactly one keyset codegen. This work brings a second one back.
2. **A cursor field on `model.PersistentRecordQuery`**, which today has none — that absence is the mechanical cause of #354.
3. **Wiring into the OLTP CTE + JSON_AGG template** (`internal/postgres_persistent_repository_query.go`), including making ORDER BY agree with the cursor column order, which is a new validation surface.
4. **EAV-attribute cursor support.** Without it this is not parity: the `Query` seam already accepts attribute cursors and the existing e2e cursors on `count`, an EAV attribute. EAV values live in a joined table with typed columns, so the predicate is materially harder than the main-table case.

## The duplication risk to plan for up front

Two keyset predicate generators must stay semantically locked to each other, or hot and cold tiers will filter differently — exactly the drift class #310 (three `EscapeLiteral` copies) and #260 (dual rendering) were spent removing, and exactly the bug #281 found when DuckDB predicate formatting diverged from Postgres. Whatever shape this takes, the two dialects should render from one IR with a differential test across them, not from two hand-maintained code paths.

## What #354 shipped that this changes

#354 made cursor-bearing requests fail closed on three routes (explicit hot-only, DuckDB disabled, degraded fallback) and reroute onto DuckDB on the fourth (the cost heuristic). Once Postgres can apply a cursor, the first and second of those refusals become servable and the reroute becomes optional rather than the only correct answer. Revisit all four decisions here rather than layering on top of them; the four-door table in #354's design comment is the map.

[...]

PR #383, opened 2026-08-09
fix(federated): #354 never serve a keyset cursor from the postgres-only path

### LINKED ISSUE #354: fix(federated): hot-only routing silently drops the keyset cursor

iceboundrock:
Closes #354.

A federated query carrying a keyset cursor was silently served by a Postgres-only path with no keyset support, so the caller got an **unfiltered first page** and pagination never advanced.

## The audit widened the issue

#354's body named one route. There are four, and they converge on one function — `queryPostgresOnly`, whose `model.PersistentRecordQuery` has no cursor field at all:

| Door | Trigger | In the issue? |
|---|---|---|
| 1 | hot-only gate: `PreferHot` or `PreferredTiers == [hot]` | yes — the only one |
| 2a | `!decision.UseDuckDB` — hybrid's `Limit < 1000` rule | no |
| 2b | `!decision.UseDuckDB` — DuckDB globally disabled | no |
| 3 | degraded fallback after a DuckDB failure | no |

**Door 2a is the widest and it is the default production configuration.** `config_duckdb.go` defaults to `RoutingStrategyHybrid`, and keyset pagination is by definition a small-`Limit` shape. The whole existing keyset e2e suite is green only because the harness leaves `routing` at `""`, which falls through `EvaluateRoutingPolicy`'s `default:` arm with `UseDuckDB` still true — every one of those tests sidesteps all four doors. `tier_hints_e2e_test.go` already warned that hybrid collapses small limits to Postgres-only; nobody had connected it to keyset.

## The fix: split by intent

An explicit hot-only request is the caller's **semantic** declaration and must not be silently reinterpreted. Hybrid's small-limit rule is only a **cost heuristic**, and correctness outranks it.

- **Choke-point guard** (`queryPostgresOnly`) — a plain read-path error wrapping the new `ErrKeysetUnsupportedOnPostgres`. Placed at the confluence rather than at each gate, so a fifth route added later cannot bypass it.
- **Routing override** (`EvaluateRoutingPolicy`) — a cursor reroutes a cost-driven Postgres-only decision onto the DuckDB federated path, the only path with keyset SQL generation. Never fires for hot-only; the reroute is recorded in `dec.Reason` so it is visible in the execution plan.
- **Degrade refusal** — the fallback *is* the Postgres-only path, so absorbing a DuckDB failure for a cursor request would return the very silent-loss answer the degraded-mode contract exists to prevent. Same reasoning as the five error-class exemptions already in `degradableFederatedError`. The real DuckDB cause now surfaces instead of the guard's rejection.
- **E2E gate** under an explicitly hybrid Env — the only test that proves a cursor survives the real default configuration.

Two decisions worth recording:

- **Plain error, not `forma.InvalidInputf`.** The submitted cursor is well-formed; it is the *route* that cannot serve it, which makes this an engine-capability error rather than caller-input validation. Consistent with the two sibling validators in the same file.
- **No tier repair needed.** `decision.Tiers` is execution-plan bookkeeping — `engine.go` passes `fq`, not `decision`, into the DuckDB path, and tier pruning reads `fq.PreferredTiers`. Verified independently during review.

## Accepted behavior changes

- Cursor queries no longer take the Postgres fast path under hybrid routing. That is correctness over a cost heuristic, and it affects no production traffic today: nothing caller-facing populates `KeysetCursor` (no HTTP parsing; `entity_query_service.go` never sets it).
- Cursor queries no longer degrade. Four previously-silent-wrong outcomes become hard failures: DuckDB disabled, `duck == nil`, circuit breaker open, and manifest misconfiguration. **Keyset pagination now hard-depends on DuckDB availability with no degrade path** — intended under this design, lifted by #382.
- Because the fail-closed error is a plain error, `httpapi` answers **500**, not 4xx. Correct per the settled ruling, but once cursors reach an HTTP surface, `PreferHot` + cursor becomes a caller-constructible 500. Noted so the eventual HTTP work inherits it.

## Verification

Eight commits, each red-first (seven at review time, plus one comment-only commit answering review finding F1). Beyond the suites, the review loop verified the load-bearing claims by mutation (`go test -overlay`, worktree untouched): the guard, the override's `!isHotOnlyRequest` conjunct, both degrade sites, the annotation gate, and both error-wrap chains each have a mutant that the new tests kill. Two mutants deliberately survive and are documented rather than hidden — the override's `cfg.Enabled` conjunct is unreachable-redundant, and widening `hotOnly` in the strategy heuristics is inert on every reachable path (tracked in #381).

`make test`, `make fmt-check`, `make lint` (v1.64.8, pin untouched), `-race` on `./internal/federated`, and `make test-e2e-production` all green.

Two corrections to the plan posted on the issue, for anyone re-running it:

- Its Task 4 command names `TestPaginationMutation`, which matches nothing. `go test -run` treats an unmatched pattern as **success**, so run literally it reports green while covering only `TestKeysetLWW`. The real names are `TestPaginationUnderInsert`, `TestDeepPagination`, `TestBigintKeysetBoundaryNoDupNoSkip` — all were run and are green.
- Its Task 3 red-state prediction says `pg.queryCalls == 1` pre-fix. It is 0, because Task 1's guard is already the first statement of `queryPostgresOnly` by then. The defect that task fixes is *which error surfaces*, not a wasted Postgres query.

## Follow-ups filed

- **#381** — the keyset cursor contract is not uniform across the engine's entry points. Includes a live correctness finding the doc correction surfaced: the `Query` seam admits attribute columns, but the keyset codegen emits `col.Attribute` unfolded, so a cursor over a dotted attribute references a name the `visible` CTE does not expose.
- **#382** — apply keyset cursors on the Postgres-only path (the deferred half of this issue). Sequence #381 first.

`docs/federated-query/design.md` is updated where this branch made it stale.


iceboundrock:
## Summary

**Verdict: Approve.** This is a high-quality, well-verified fix. The diagnosis (four doors converging on `queryPostgresOnly`) is correct, the fail-closed error class (plain read-path error, not `forma.InvalidInputf`) matches `docs/error-handling.md`, the red-first discipline is genuine, and the mutation-testing claims are load-bearing and mostly verified independently below. No blocking findings.

I independently re-verified the claims that matter rather than trusting the body:

- Production `KeysetCursor` claim: `internal/entity_query_service.go:225` never sets `KeysetCursor`; the only non-test setters are the e2e harness (`production/query.go:147`), the benchmark (`benchmark/execute.go:237`), and `queryplan/shape_test.go`. Correct.
- The "no tier repair" decision: `engine.go:173` passes `fq` (not `decision`) into `ExecuteDuckDBFederatedQuery`; the DuckDB path derives tiers from `fq.PreferredTiers` (`duckdb_query_helpers.go:364`, `duckdb_template_renderer.go:223`). `dec.Tiers` is plan bookkeeping. Correct.
- Door 2b / degrade-refusal mechanics: `duckdb_query.go:169-183` classifies nil-duck and breaker-open as the degradable `ErrDuckDBUnavailable`; the new `mayDegradeToPostgres` now returns false for a cursor and `explainDeclinedDegradation` surfaces the real cause. Correct.
- The hybrid-default claim: `config_duckdb.go:285` defaults `Strategy` to `RoutingStrategyHybrid`; the e2e harness default `""` falls to `applyStrategyHeuristics` default arm and keeps `UseDuckDB` true — so the existing keyset e2e suite genuinely sidesteps all four doors. Correct.
- The two plan corrections (Task 4 `-run` name matching nothing; Task 3 red-state `queryCalls == 1` being wrong) — both independently confirmed.

**Gaps in my verification:** I could not run `make lint` (golangci-lint not invoked here) or `make test-e2e-production` (needs Docker). Everything else passed: full unit suite, `-race ./internal/federated`, `go vet` on the changed packages, gofmt clean, and the 13 new/refactored federated tests green.

Minor doc nits only: the body says "Six commits" but the branch carries 7 (`0e60d22` through `93c0aab`), and the error-message prose in the body/plan differs slightly from the shipped `rejectKeysetOnPostgresOnly` message ("reach the federated path instead" vs "enable DuckDB"). Neither matters.

Follow-ups #381/#382 are appropriately filed and the `docs/federated-query/design.md` update correctly documents the per-seam cursor contract (I confirmed `sqlgen/keyset_where.go:54,57,95` emits `col.Attribute` unfolded, so the `contact.annualIncome` finding in #381 is real).

iceboundrock:
## Spec

**Origin:** #354 names only door 1 (the hot-only gate). The audit widened it to four doors converging on `queryPostgresOnly`; the four-door table in the body is accurate against the current tree.

**Requirements coverage — the issue offered two options; this implements option 2 (fail closed) fully and defers option 1 (Postgres-side keyset) to #382, sequenced behind #381. Consistent with the plan posted on the issue.**

- Door 1 (hot-only gate): guard rejects with `ErrKeysetUnsupportedOnPostgres`, `pg.queryCalls == 0`. Covered by `TestQueryPreferHotRejectsKeysetCursor` + `TestQueryHotOnlyTiersRejectsKeysetCursor`.
- Door 2a (hybrid small-limit, default production config): override reroutes onto DuckDB. Covered at policy level and engine seam.
- Door 2b (DuckDB disabled): guard fails the request. Covered.
- Door 3 (degraded fallback): refuses to absorb a cursor failure at both the page-fetch and recount sites. Covered.

**Deviations from the posted plan — both are improvements, not drift:**

1. Plan Task 1 Step 5 showed the guard returning `return nil, err` bare; the implementation wraps with `fmt.Errorf("validate keyset cursor routing: %w", err)`. Correct per the error-wrapping standard.
2. Plan Task 3\`s `explainDeclinedDegradation` guarded only on `opts.AllowPartialDegradedMode && hasKeysetCursor`; the shipped version adds `!degradableFederatedError(err)`, making it the exact complement of `mayDegradeToPostgres`. This is strictly more correct — it stops blaming the cursor for an error class that refuses degradation with or without a cursor — and is what makes `TestQueryWithKeysetCursorLeavesNonDegradableFailureUnannotated` meaningful.

**No scope creep:** the widened door coverage is a direct consequence of the audit (2a *is* the default production shape), and each widening is tested and documented.

**One spec gap to acknowledge (already tracked):** the `ExecuteFederatedPaginatedQuery` seam (`pagination.go`) is excluded from the guard. With a cursor and `KeysetEnabled` unset it falls into the merge path where the Postgres side (`RunOptimizedQuery`) applies no cursor while the DuckDB side does — a partial application, not fail-closed. Only the benchmark reaches this seam today (it always sets `KeysetEnabled`), so it is latent; #381/#382 scope it.

iceboundrock:
## Standards

Checked against `AGENTS.md`, `coding-standard.md`, and `docs/error-handling.md`. All load-bearing rules hold; no hard violations found.

- **File sizes** ≤ 500: `engine.go` 456 (was 451), `routing.go` 180, `keyset.go` 124, `errors.go` 66, tests 383/108. All within limit; the new unit tests correctly went into a fresh file rather than `engine_test.go` (near the cap).
- **Function sizes** ≤ 100: `EvaluateRoutingPolicy` ~90, `applyStrategyHeuristics` ~78, all new helpers short.
- **Error wrapping:** every new return wraps with context and `%w` (`rejectKeysetOnPostgresOnly`, the guard, `explainDeclinedDegradation`, both degrade-site wraps). No bare `return err`. `TestNoBareSentinelWraps` (root) passes.
- **Error-class discipline** (`docs/error-handling.md`): the fail-closed error is a plain read-path error wrapping a package sentinel, never `forma.InvalidInputf` — consistent with the two sibling validators in `keyset.go` and with the read-path-consistency guidance. The tests pin `NotErrorIs(err, forma.ErrInvalidInput)` so a later refactor cannot silently reclassify it into a 4xx.
- **Message quality:** `rejectKeysetOnPostgresOnly` names the value type (keyset cursor), the offending column (the trailing attribute), and the expected state (route to the federated path). Meets the "name the logical value type, the offending column, the expected state" bar.
- **Flat/early-return structure** throughout; no nesting >3.
- **Naming:** verb-first private helpers (`hasKeysetCursor`, `rejectKeysetOnPostgresOnly`, `mayDegradeToPostgres`, `explainDeclinedDegradation`, `applyStrategyHeuristics`).
- gofmt clean on all six changed files; `go vet` clean.

**Judgement calls (not violations):** the override clause comment block in `routing.go` (~30 lines) is long, but it documents four conjuncts of which two (`cfg.Enabled`, and `!isHotOnlyRequest` through the engine) are admitted to be unreachable/redundant — this is deliberate belt-and-braces with each conjunct's fate explained, which is the right way to carry dead-guard hygiene in a repo that mutation-tests. The `explainDeclinedDegradation`/`rejectKeysetOnPostgresOnly` messages embed caller remediation prose, which is consistent with the sibling validators' style here.

iceboundrock:
## Findings

Non-blocking. Ranked by how much a future reader should care.

**F1 — Plan bookkeeping is mutated before the guard rejects (doors 1 and 2b).** Both `recordHotOnlyGatePlan` (door 1) and `recordRoutedPostgresSource` (door 2b) write a `postgres` source + routing decision into `opts.ExecutionPlan` *before* `queryPostgresOnly`'s new guard fires. On the failure path the page is nil so the plan never reaches a response — the accepted consequence is documented in the plan comment for door 1 only. But a caller that inspects `opts.ExecutionPlan` after the error sees "postgres served this query" for a query that never ran. Since the branch's own "the plan reflects actual access" contract (#243/#185) is what these recorders exist for, door 2b should get the same one-line acknowledgment as door 1, or the recorders should move after the guard. Cosmetic today; a plan-truthfulness footgun later.

**F2 — The `ExecuteFederatedPaginatedQuery` seam stays outside the guard, and it is the one place a cursor can still be partially applied.** With a cursor and `opts.KeysetEnabled` unset, `pagination.go` falls into the merge path: the Postgres side (`RunOptimizedQuery`) applies no cursor while the DuckDB side (`ExecuteDuckDBFederatedQuery`, which always renders the keyset clause) does — so the merged page is cursor-filtered on cold but unfiltered on hot, exactly the silent-wrong-answer class #354 closes. Only the benchmark reaches this seam today (it sets `KeysetEnabled`), so it is latent, and the `rejectKeysetOnPostgresOnly` doc comment is honest about the exclusion. #381 should record that this is a *partial-application* gap, not merely a validation-gating gap.

**F3 — Exported `EvaluateRoutingPolicy` disagrees with `isHotOnlyRequest` for the `[hot]`-tiers form under freshness-first.** `applyStrategyHeuristics`'s local `hotOnly` is `PreferHot` only (routing.go:104), so freshness-first + `PreferredTiers: [hot]` + cursor returns `UseDuckDB: true` (no branch fires, and the override is correctly blocked by `isHotOnlyRequest`). The engine gate makes this shape unreachable in practice (it intercepts `[hot]` before the policy runs), so there is no live behavior bug — but the exported function and the engine now render opposite verdicts for one input shape, and a standalone caller of the exported function gets `UseDuckDB=true` for an explicitly hot-only request. Pre-existing (the local `hotOnly` was always PreferHot-only); #381's "contract not uniform" framing should list it.

**F4 — `hasKeysetCursor` keys on columns, not on value alignment.** A cursor with non-empty `Columns` but nil/short `Values` is "active", so it reroutes to DuckDB where the renderer binds nil `?`-args and fails with a raw driver error instead of a clean validation. Neither `validateKeysetTiebreak` nor `validateKeysetColumns` checks count alignment either, so this is a pre-existing shape-contract gap that the new routing logic now makes more likely to surface. Observation, not a defect this PR should carry.

**F5 — Cosmetic.** "Six commits" in the body vs 7 in the branch (`0e60d22`…`93c0aab`); the error message quoted in the body/plan ("drop PreferHot … enable DuckDB") differs slightly from the shipped `rejectKeysetOnPostgresOnly` prose ("reach the federated path instead"). Neither is worth a change.

**Where the mutation discipline pays off (positive):** the tests are engineered around wrap-chain semantics (`%w → %v` mutations would flatten `errors.Is` while prose stays green), the `!dec.UseDuckDB` truthfulness conjunct is pinned by `TestEvaluateRoutingPolicyKeysetCursorLeavesAlreadyFederatedReasonIntact`, and the `NotErrorIs(err, ErrKeysetUnsupportedOnPostgres)` assertions on both degrade sites prove the fallback was never entered rather than merely un-annotated. This is the right way to use mutation reasoning in tests.

iceboundrock:
## Observations

Things I checked that shape how much to trust the change, but are not findings.

- **The override fires only when a strategy already chose Postgres-only** (`!dec.UseDuckDB`). Every strategy that lands on DuckDB on its own merits (cost-first small scan → `UseDuckDB` stays true; hybrid deep-pagination; cold-only) skips the override and applies the cursor in the DuckDB template directly — correct, and the reason string stays un-suffixed so the plan never claims a reroute that did not happen. Verified across all three strategies.
- **Error precedence is consistent:** `validateKeysetTiebreak` (engine.go:140) runs before the hot-only gate and before the guard, so a hot-only request carrying a cursor *without* the trailing `row_id` tiebreak gets the tiebreak error, not the routing error. Well-formedness is judged before routability, which is the right order; both fail loud.
- **`hasKeysetCursor`'s empty-columns contract is uniform across every consumer** — the engine gate, the guard, `mayDegradeToPostgres`, the DuckDB renderer (`duckdb_template_renderer.go:310`), and `duckdb_compiled_query.go:134` all key on `len(Columns) > 0`. The negative control `TestQueryWithoutCursorStillServesPostgresOnly` (nil cursor + empty-columns cursor) pins the no-op contract on the Postgres path. One less class of drift than #310/#260 fought.
- **The `isHotOnlyRequest` refactor is behavior-preserving for cursor-free queries.** It is an exact copy of the inline predicate plus a `nil` guard (unreachable in `engine.Query` which rejects nil `fq` first); `applyStrategyHeuristics` keeps its own `PreferHot`-only `hotOnly` local untouched, so no strategy branch changes for existing traffic. The concern recorded on the issue — that widening `hotOnly` would flip freshness-first `[hot]` cursor-free queries — is correctly avoided.
- **Recount site:** a keyset query with `Offset > 0` reaching an empty page triggers the recount DuckDB query; the recount's degrade branch is wired through the same `mayDegradeToPostgres`/`explainDeclinedDegradation` pair, and `TestQueryWithKeysetCursorDoesNotDegradeOnRecountFailure` pins it — so door 3 cannot be reverted at one site while the other stays fixed.
- **Harness wrap compatibility:** the e2e `prefer_hot_fails_closed` subtest asserts `errors.Is(err, fedengine.ErrKeysetUnsupportedOnPostgres)` through `env.Query`, which wraps the engine error with `%w` (`production/query.go:91`) — so the sentinel survives to the test. Good.
- **The docs update** (`design.md` §1 and the keyset-columns paragraph) is accurate against the tree; the per-seam cursor contract it now describes is genuinely per-seam (`validateKeysetColumns` allowlist in `pagination.go` vs `validateKeysetTiebreak`-only in `engine.go`), and the `contact.annualIncome` example matches `generateKeysetWhereClause` emitting `col.Attribute` unfolded.

## Assessment

Ship it. The fix is architecturally right (choke-point guard at the confluence, not per-gate), the error class is the correct one for an engine-capability failure, and the two hard behavior changes (cursor queries never take the PG fast path under hybrid; cursor queries never degrade) are exactly the right trade against a cost heuristic. The three genuinely unresolved loose ends are all either latent (F2), engine-unreachable (F3), or cosmetic (F1), and F2/F3 are already scoped by #381.

One design point worth carrying into the eventual HTTP surface: because the fail-closed error is a *plain* read-path error, `httpapi` will answer 500 — correct per the settled ruling, but as the body notes, `PreferHot` + cursor becomes caller-constructible once cursors reach HTTP. That is a latent 500-for-caller-input; the HTTP work should decide whether a cursor+PreferHot request deserves the routing error before the boundary classifies it. Not this PR's problem.

iceboundrock:
## Verification

What I ran against the PR head (`93c0aab`), in a detached worktree:

| Check | Command | Result |
|---|---|---|
| Build | `go build ./...` | PASS |
| New federated tests | `go test ./internal/federated -run 'TestQueryPreferHotRejectsKeysetCursor\|TestQueryHotOnlyTiersRejectsKeysetCursor\|TestQueryRejectsKeysetCursorWhenDuckDBDisabled\|TestQueryWithoutCursorStillServesPostgresOnly\|TestEvaluateRoutingPolicyKeysetCursor\|TestQueryRoutesKeysetCursorToDuckDBUnderHybridSmallLimit\|TestQueryWithKeysetCursorDoesNotDegrade' -v` | PASS (all, incl. subtests) |
| Federated package | `go test ./internal/federated` | PASS |
| Race | `go test -race ./internal/federated` | PASS |
| Full unit suite | `go test . ./cdc ./factory ./internal/...` | PASS (mirrors `make test`; all `ok`, zero failures) |
| Sentinel guard | root package `TestNoBareSentinelWraps` | PASS |
| fmt | `gofmt -l` on all six changed files | clean |
| vet | `go vet ./internal/federated ./internal/sqlgen` | clean |
| E2E | `make test-e2e-production` / `-tags=e2e` | **not run** — requires Docker; not available here |

The two `-run`-correction claims in the body were independently confirmed: `TestPaginationMutation` matches nothing (the real names `TestPaginationUnderInsert`, `TestDeepPagination`, `TestBigintKeysetBoundaryNoDupNoSkip` all exist), and the Task 3 `queryCalls == 1` red-state prediction is wrong as described because the Task 1 guard already precedes the fallback.

Gaps I could not close without the e2e cluster: the actual red-state of the new e2e gate against pre-fix code, and `make lint` (golangci-lint not invoked in this environment). Everything else above is first-hand.

iceboundrock:
Thanks — the independent re-verification of the four load-bearing claims (production `KeysetCursor` writers, the no-tier-repair decision, door-2b/degrade mechanics, and the hybrid-default premise) is the part that matters most, since those are exactly the claims a reader would otherwise have to take on trust.

Disposition of the five findings: **F1 fixed, F3 and F4 folded into #381, F2 was already recorded there, F5 half-fixed and half-disputed.**

## F1 — plan bookkeeping mutated before the guard rejects · **fixed in `4412fe2`**

You're right, and it was worse than "door 2b lacks door 1's acknowledgment": **neither door carried it in code.** The acknowledgment existed only in a commit message, so a reader of `queryPostgresOnly` had no way to tell a deliberate trade from an oversight.

Recorded on the guard's doc comment rather than at the two gates, because that is where the question arises and because the trade being made is precisely the one the guard placement embodies — a single confluence beats a set of gate checks that must be kept in sync, and these two stale plan entries are what that choice costs. Comment-only; `make fmt-check`, `make lint`, `go build ./...` and the `internal/federated` package all green.

I did consider your alternative of moving the recorders after the guard. It doesn't work cleanly: the recorders are called by `Query` while the guard lives inside `queryPostgresOnly`, so honouring the ordering would mean either threading a recorder closure through the call or duplicating the guard at each gate — trading a real invariant for a cosmetic one.

## F2 — `ExecuteFederatedPaginatedQuery` partial application · **already recorded as such**

#381 §3 states it as partial application rather than a gating gap, in those terms: *"the Postgres side (`RunOptimizedQuery`) applies no cursor while the DuckDB side does — `duckdb_template_renderer.go:310` keys off `q.KeysetCursor` and never off `opts.KeysetEnabled`. The merged page is then hot-rows-unfiltered ∪ cold-rows-filtered: the same silent-wrong-answer family as #354."* No change needed.

## F3 — exported `EvaluateRoutingPolicy` disagrees with the engine for `[hot]` · **added to #381 as item 8**

Agreed, and it sharpens item 5 rather than duplicating it: the existing item argued only that widening `hotOnly` is unsafe, which is a statement about a fix. Your framing is a statement about the current contract — the exported function returns `UseDuckDB=true` for an explicitly hot-only request, the opposite of what `engine.Query` does with the same input. That is what the characterization test should pin, so I rewrote the item to say so.

## F4 — `hasKeysetCursor` keys on columns, not value alignment · **added to #381 as item 7, with a correction**

Real, and quieter than your description. You expected a raw driver error; `generateKeysetWhereClause`'s `valueAt` returns **`nil`** for an index past the end of `Values` rather than failing:

```go
valueAt := func(i int) interface{} {
    if i < len(cursor.Values) {
        return cursor.Values[i]
    }
    return nil
}
```

So a short or nil `Values` binds SQL NULL, `col > NULL` evaluates to NULL, `WHERE` treats that as not-true, and every disjunct carrying the unfilled arm drops out. The caller gets a **silently empty page**, not an error — the same silent-wrong-answer family as #354 itself, which makes it a better fit for #381 than "shape-contract gap" suggested. Recorded with that mechanism spelled out, to be fixed alongside the column-validation item (whatever validates columns should also require `len(Values) == len(Columns)`).

Agreed it isn't this PR's to carry: it is pre-existing, and #354 only raises the odds of reaching it by rerouting cursors onto the DuckDB template from shapes that previously never rendered it.

## F5 — cosmetic · **first half fixed, second half disputed**

The commit count was wrong: the body said six, the branch carried seven, and it now carries eight with `4412fe2`. Body corrected — worth one edit even though you flagged it as not worth changing, since a body that miscounts its own commits is the kind of thing that makes a reader discount the rest of it.

The error-message half I don't think holds. Both phrases you contrasted are in the same shipped string — `rejectKeysetOnPostgresOnly` emits *"…reach the federated path instead (drop PreferHot and any hot-only PreferredTiers, and enable DuckDB) or paginate with limit/offset"*, and the plan's Task 1 Step 4 code block carries it verbatim. The `"drop PreferHot … enable DuckDB"` form you saw is my own elided quotation of that same message in the #354 comments, not a second wording. Nothing to reconcile.

## On the two gaps you flagged in your own verification

`make lint` (v1.64.8) and `make test-e2e-production` both ran green locally before the PR opened, and CI has since gone 5/5 green on `93c0aab` — Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke. `4412fe2` is comment-only; I re-ran `make lint`, `make fmt-check` and the federated package on it, and CI will re-run the rest.

Your closing point about the HTTP surface is the right one to carry forward: `PreferHot` + cursor becoming a caller-constructible 500 is a boundary-classification decision, not an engine one, and it should be settled where the cursor is parsed rather than by weakening the engine's error class. Noted on #382, which is where that surface gets built.

iceboundrock:
## Re-review — Summary

**Verdict: Approve, unchanged.** This is a re-review at head `4412fe2` — one comment-only commit past the prior review's `93c0aab`. The delta answers finding F1, and every one of the author's dispositions of F2–F5 verifies cleanly against the tree, including a correction to my own prior finding. No new findings; nothing to block.

The one thing I got wrong in the prior pass is withdrawn below (F5 second half): the shipped `rejectKeysetOnPostgresOnly` message is a single string containing *both* quoted phrases, so there was never a second wording to reconcile.

Ready to merge.

iceboundrock:
## Re-review — Spec

**The delta is comment-only, so the spec surface is unchanged; the prior Spec verdict stands.** All four doors are still covered at head: the `queryPostgresOnly` guard (1/2b), the `EvaluateRoutingPolicy` override (2a), and `mayDegradeToPostgres` + `explainDeclinedDegradation` at both the page-fetch and recount sites (3). The fail-closed error remains a plain read-path error wrapping `ErrKeysetUnsupportedOnPostgres`, pinned by `NotErrorIs(err, forma.ErrInvalidInput)`.

**Disposition verification — all three claims check out against #381:**

- **F2 → #381 §3.** The issue already frames it as *partial application*, not merely a gating gap, with the exact mechanism (`duckdb_template_renderer.go:310` keys off `q.KeysetCursor`, never `opts.KeysetEnabled`). The prior review's ask was already satisfied.
- **F3 → #381 item 8.** Present and correctly sharpened: not "widening `hotOnly` is unsafe" (a statement about a fix) but "the exported contract currently disagrees with the engine for `PreferredTiers: [hot]`" (a statement about the current contract). That is exactly the characterization the finding asked for.
- **F4 → #381 item 7, with a mechanism correction that is *correct*.** `generateKeysetWhereClause`'s `valueAt` (sqlgen/keyset_where.go:38-43) returns `nil` for an index past `len(Values)`, so a short `Values` binds SQL NULL, `col > NULL` is NULL, `WHERE` drops the disjunct → **silently empty page**, not a raw driver error. The prior review mischaracterised the failure mode; the author's quieter-but-worse description is right, and it is the same silent-wrong-answer family as #354. Correctly filed ahead of the Postgres-side keyset feature.

**No new deviations.** The override's `cfg.Enabled` conjunct is still documented as unreachable-redundant (routing.go) rather than load-bearing, which is a doc-honesty improvement over the plan's original claim. Nothing else changed.

iceboundrock:
## Re-review — Standards

**No new standards surface: the entire delta is an 11-line doc comment.** Verified against `AGENTS.md`, `coding-standard.md`, and `docs/error-handling.md`:

- **`4412fe2` is comment-only** — 11 insertions in `internal/federated/keyset.go`, nothing else. File sizes unchanged (engine.go 456, routing.go 180, keyset.go 135, all ≤500; `EvaluateRoutingPolicy` ~72, `applyStrategyHeuristics` ~78, all ≤100).
- **The new comment is accurate**, not aspirational. Checked each claim against the tree:
  - Both `recordHotOnlyGatePlan` (engine.go:228-243) and `recordRoutedPostgresSource` (engine.go:282-292) do write an `Engine: "postgres"` source into `opts.ExecutionPlan` *before* the guard in `queryPostgresOnly` fires — for door 1 and door 2b respectively. ✓
  - "Inert on the response path — the page is nil on error" holds: `queryPostgresOnlyWithPlan` returns `nil` on error and `attachExecutionPlan` no-ops on a nil page. ✓
  - The degraded-fallback recorder is correctly *absent* from the comment: door 3 (`degradeToPostgresOnly`) calls `recordDegradedFallbackPlan`, but a cursor request never reaches it because `mayDegradeToPostgres` returns false — so there is no third stale entry to acknowledge. The comment's two-door enumeration is complete, not incomplete. ✓
  - Referencing the plan-truthfulness contract (#243/#185) is the right citation — those are precisely the recorders' stated purpose.
- **Error-class discipline unchanged.** The guard still wraps a plain read-path error; `TestNoBareSentinelWraps` and the `NotErrorIs(err, forma.ErrInvalidInput)` pin both pass.

No violations; the only judgement call is the same one already recorded — a long but honest comment in a repo that treats comments as the decision record.

iceboundrock:
## Re-review — Findings

Nothing new. Disposition of the prior five, verified against the tree at head `4412fe2`:

**F1 — Resolved by `4412fe2`.** The stale-plan consequence is now recorded on the guard's doc comment, covering **both** doors in one place — strictly better than the "door 2b one-liner" I asked for, because neither door carried it in code before (it lived only in a commit message). The author's rationale for *not* moving the recorders after the guard is sound: the recorders are called by `Query`, the guard lives inside `queryPostgresOnly`, so honoring the ordering would mean threading a recorder closure through the call or duplicating the guard at each gate — trading a real invariant for a cosmetic one. Accepted as-is.

**F2 — Already recorded as partial application in #381 §3.** No further action needed; the issue's wording is exactly the framing the finding asked for.

**F3 — Folded into #381 item 8** as a statement about the current contract (exported `EvaluateRoutingPolicy` returns `UseDuckDB=true` for `PreferredTiers:[hot]`), which is the sharper framing. Good disposition.

**F4 — Folded into #381 item 7 with a mechanism correction that is *correct*.** My prior finding expected a raw driver error from nil `?`-args; the author showed `valueAt` (sqlgen/keyset_where.go:38-43) returns `nil` instead, so a short/nil `Values` binds SQL NULL and yields a **silently empty page** — quieter and worse than I said, same family as #354. I agree with the correction and with it being tracked ahead of the Postgres-side keyset feature.

**F5 — First half fixed, second half withdrawn by me.** The commit count is now correct in the body (8). The error-message half: the author is right and I was wrong — the shipped `rejectKeysetOnPostgresOnly` message is a *single* string containing both "reach the federated path instead" and "(drop PreferHot and any hot-only PreferredTiers, and enable DuckDB)". There was never a second wording. Withdrawn.

No new findings at head. The mutation-discipline observation from the prior review still holds and is the reason the delta being comment-only is low-risk.

iceboundrock:
## Re-review — Assessment

Overall: a disciplined, well-verified bug fix whose re-review delta is small, correct, and honest.

**What I most trust:**
- The F1 documentation fix is truthful — I verified each sentence of the new guard comment against the recorder implementations, the nil-page path, and the degraded-fallback flow rather than taking the prose on faith.
- The F4 mechanism correction improves the record: the issue now carries the *actual* failure mode (silent empty page from NULL-bound keyset arms), not my mistaken "raw driver error" description. That matters because #382's Postgres-side keyset work will build on the same `valueAt` shape.
- The decision record is complete: every non-blocking finding has a disposition in #381 or in the PR reply, and the "why" of each disposition is stated.

**Residual risk (unchanged, not this PR's to carry):**
- The `ExecuteFederatedPaginatedQuery` seam (#381 §3) remains outside the guard — latent, benchmark-only, correctly tracked.
- `PreferHot` + cursor answering a caller-constructible 500 is a boundary-classification decision; it should be settled where the cursor is parsed, not by weakening the engine's error class. Noted on #382, which is the right home.

**Net:** the branch delivers exactly what #354 asked — fail closed where a cursor cannot be honored, override where correctness outranks a cost heuristic — with zero silent-wrong-answer paths left reachable through the `Query` seam. Merge-ready.

iceboundrock:
## Re-review — Observations

Things I checked that shape how much to trust the change, but are not findings.

- **The `4412fe2` doc comment does not over-claim.** It enumerates exactly two recorders and I confirmed those are the only two that can run before the guard for a cursor request: door 1 (`recordHotOnlyGatePlan`) and door 2b (`recordRoutedPostgresSource`). Door 3's `recordDegradedFallbackPlan` cannot run for a cursor because `mayDegradeToPostgres` already returned false, so omitting it is precise rather than lazy.
- **`hasKeysetCursor` is now spelled three ways.** The engine's `validateKeysetTiebreak` gate (engine.go) still inlines `fq.KeysetCursor != nil && len(...) > 0` rather than calling the new `hasKeysetCursor` helper. The PR went to effort to share `isHotOnlyRequest` across the gate and the override specifically so the two "cannot drift apart"; the same discipline was not applied to the active-cursor predicate. Pre-existing inline, zero behavior risk (the guard dominates), and a one-line follow-up if the drift-avoidance principle is to be consistent.
- **The `cfg.Enabled` conjunct remains un-testable** — `TestEvaluateRoutingPolicyKeysetCursorDoesNotOverrideDisabledDuckDB` documents that its `UseDuckDB` assertion cannot catch a mutation of that conjunct (the `!cfg.Enabled` early return already fixed `UseDuckDB`), and it pins `Reason` instead. The comment in routing.go admits it is unreachable-redundant and kept as a cheap invariant. Consistent with the repo's mutation-first test philosophy — worth a standing note, not a defect.
- **CI is green at the exact re-review head.** The status rollup for `4412fe2` (run 31293188885) shows all five gates SUCCESS: Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke. So the "CI will re-run the rest" from the author's reply has since landed.
- **Mutation-survivor transparency holds:** the two deliberately surviving mutants (the `cfg.Enabled` conjunct and widened `hotOnly`) are documented in the code comment and tracked in #381, not hidden.

iceboundrock:
## Re-review — Verification

What I ran against head `4412fe2` (worktree `/Users/ruoshi/code/Lychee/LTBase/forma-354`, GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false):

| Check | Command | Result |
|---|---|---|
| All 12 new/refactored routing+degrade tests | `go test ./internal/federated -run 'PreferHotRejects\|HotOnlyTiersRejects\|DuckDBDisabled\|WithoutCursorStillServes\|OverridesCostHeuristic\|DoesNotOverride\|LeavesAlreadyFederated\|RoutesKeysetCursorToDuckDB\|DoesNotDegrade\|RecountFailure\|LeavesNonDegradable' -v` | PASS (all) |
| Full federated package | `go test ./internal/federated/...` | PASS |
| Race | `go test -race ./internal/federated/...` | PASS |
| Error-class guard | `go test . -run TestNoBareSentinelWraps -v` | PASS |
| Full unit suite | `go test . ./cdc ./factory ./internal/...` | PASS (all packages) |
| Build | `go build ./...` | PASS |
| E2E compiles | `go vet -tags e2e ./internal/e2e_harness/production/` | PASS |
| gofmt | `gofmt -l internal/federated/ internal/e2e_harness/production/` | clean |

**Claims re-verified against the tree (not taken on trust):**
- F1 doc comment accuracy — both recorders write `Engine: "postgres"` before the guard; nil page means `attachExecutionPlan` no-ops. Verified at engine.go:228-243, 282-292, 294-310.
- F4 mechanism correction — `valueAt` returns `nil` past the end of `Values` (sqlgen/keyset_where.go:38-43), confirming silent-empty-page rather than raw-driver-error.
- F5 message — the shipped `rejectKeysetOnPostgresOnly` string contains both phrases in one `fmt.Errorf` (keyset.go:103).
- #381 items 7 and 8 exist with the claimed content.
- CI rollup at `4412fe2` is 5/5 SUCCESS (run 31293188885).

**Not run by me:** `make lint` (golangci-lint not installed here) and `make test-e2e-production` (needs Docker). CI's Lint and E2E Smoke gates cover both at head and are green. `4412fe2` is comment-only, so the unit-suite + gofmt + vet coverage above is the meaningful local check.

iceboundrock:
Re-review acknowledged — and thank you for withdrawing F5's second half explicitly rather than letting it sit. A retracted finding that stays on the record unretracted is worse than one never raised, because the next reader can't tell which half survived.

The re-review carries **no findings**, but one Observation is actionable, so I acted on it. Everything else needs no change.

## `hasKeysetCursor` spelled three ways · **fixed in `5c392d7`**

You're right, and the sharpest part of the observation is that it is *the branch's own argument turned back on it*: PR #383 extracted `isHotOnlyRequest` specifically so the gate and the override "cannot drift apart", then left the active-cursor predicate inlined at two sites while introducing a helper for three others. Applying the drift-avoidance principle to one predicate and not the other is the inconsistency — not the inlining.

Now routed through the helper:

- `engine.go` — the `validateKeysetTiebreak` gate
- `pagination.go:36` — `ExecuteFederatedPaginatedQuery`'s keyset branch

Behaviour-identical at both: `hasKeysetCursor` adds an `fq != nil` arm that both callers already guarantee (`Query` rejects a nil query at its top; `ExecuteFederatedPaginatedQuery` at the top of its own body), and the column test is character-for-character what was there. All five in-package call sites now agree by construction. Full unit suite, `-race ./internal/federated/...`, `make fmt-check` and `make lint` green locally; CI running.

**One site deliberately left inline:** `pagination.go:154`'s `if fq.KeysetCursor != nil`. It asks a different question — does a cursor *object* exist, guarding the validators — and it is reached only after line 36 has already established the column list is non-empty. Folding it in would swap a nil-guard for an activity test and blur two distinct predicates into one.

**Two sites I could not fold, now recorded as #381 item 9:** `sqlgen/duckdb_template_renderer.go:310` and `sqlgen/duckdb_compiled_query.go:134` are in a different package and cannot reach an unexported `federated` helper. They are also the two sites that decide whether the keyset clause is *rendered*, which makes them exactly the ones that must not disagree with the sites deciding whether a cursor is *honoured or refused* — and the renderer keying off `q.KeysetCursor` rather than `opts.KeysetEnabled` is what produces the split-brain in #381 §3. So it is the same seam seen from the codegen side. The cross-package unification is a predicate on the type (`func (c *KeysetCursor) IsActive() bool` on `model.KeysetCursor`), which touches three packages and belongs with whatever settles items 1, 3 and 7 — a single predicate buys nothing while the *contract* it encodes is still per-seam.

## The rest

**F1–F4 dispositions confirmed** — nothing further from me. I'd note one thing you verified that I hadn't stated explicitly and should have: that door 3's `recordDegradedFallbackPlan` is correctly *absent* from the new guard comment because `mayDegradeToPostgres` returns false before it can run, so the two-door enumeration is complete rather than partial. That is the right reading and it is why the comment stops at two.

**The `cfg.Enabled` conjunct's untestability** stands as documented. Worth restating why it isn't hidden: the test's own comment records that its `UseDuckDB` assertion cannot catch a mutation of that conjunct — the `!cfg.Enabled` early return has already fixed `UseDuckDB` — so it pins `Reason` instead, and `routing.go` admits the conjunct is unreachable-redundant rather than claiming it as a guard. A dead conjunct that says it is dead costs a reader nothing; one that claims to be load-bearing costs them a debugging session.

No further changes planned. PR remains unmerged pending your call.

Issue #386, opened 2026-08-09
conditionexpr.ParseNumeric is a dead fifth numeric parse rule that diverges from numutil.TryParseNumber


iceboundrock:
Found by the final whole-branch review of the #355 branch, verified against `main` @ `c819301`.

#355 unified the numeric-family parse rule across all four live binders — `ConvertPgMainValue`, `parsePgEavValue`, `parseDuckDBRawParam`/`ToDuckDBParam`, and `buildAttrValuesAnchor` — onto `numutil.TryParseNumber`. A fifth parse rule survived that unification by being dead:

```go
// internal/conditionexpr/parser.go:104-113
// ParseNumeric parses a numeric literal into int64 or float64.
func ParseNumeric(raw string) (any, error) {
	if i, err := strconv.ParseInt(raw, 10, 64); err == nil {
		return i, nil
	}
	if f, err := strconv.ParseFloat(raw, 64); err == nil {
		return f, nil
	}
	return nil, fmt.Errorf("value '%s' is not numeric", raw)
}
```

`grep -rn 'ParseNumeric' --include='*.go' .` returns only its own definition and `internal/conditionexpr/parser_test.go:83-99`. There are no production callers.

## Why it matters

It is exported, its name and signature are binder-shaped, and it **diverges from the unified rule**. `TryParseNumber` (`internal/numutil/convert.go:115`) refines any ParseFloat-accepted literal that denotes an integer into an exact `int64` — bare, decimal, or exponent spelling (#357). `ParseNumeric` tries `ParseInt` first and falls straight through to `ParseFloat`, so:

| literal | `TryParseNumber` | `ParseNumeric` |
| --- | --- | --- |
| `"42"` | `int64(42)` | `int64(42)` |
| `"42.0"` | `int64(42)` | `float64(42)` |
| `"9.007199254740993e15"` | `int64(9007199254740993)` | `float64(9007199254740992)` — rounded |

Anyone who wires this up as a binder — which its name invites — reopens exactly the divergence #281, #357 and #355 closed, and does so silently, because the wrong value is only observable above 2^53.

## Options

1. **Delete it** along with its test. Nothing calls it; the git history keeps it if it is ever wanted.
2. **Keep it but retarget it**: make it delegate to `numutil.TryParseNumber` so a future caller inherits the unified rule.
3. **Keep it and mark it**: rename to something that does not read as a binder and document that it must not be used to build query parameters.

[...]

Issue #385, opened 2026-08-09
CI provisions a Postgres service the integration suite never connects to — hardcoded DSN ignores DB_* env, suite silently skips


iceboundrock:
Found while choosing a verification venue for #355, against `main` @ `c819301`.

## The hole

CI's Test job provisions a Postgres service and injects connection env vars for it (`.github/workflows/ci.yml:35-71`):

```yaml
services:
  postgres:
    image: postgres:16
    env:
      POSTGRES_USER: test
      POSTGRES_PASSWORD: test
      POSTGRES_DB: forma_test
...
      - name: Run tests
        env:
          DB_HOST: localhost
          DB_USER: test
          DB_PASSWORD: test
          DB_NAME: forma_test
        run: go test -v -race -coverprofile=coverage.out . ./cdc ./cmd/... ./factory ./internal/...
```

The only consumer of a real Postgres in those packages ignores every one of those variables:

```go
// internal/postgres_persistent_repository_integration_test.go:189
dsn := "postgres://postgres:postgres@localhost:5432/forma?sslmode=disable"
```

Neither the role `postgres` nor the database `forma` exists in the CI service, so `pool.Ping` fails and the test calls `t.Skipf` (`:199`, `:204`) — a skip, not a failure. Every test built on `setupIntegrationEnv` (`internal/integration_suite_test.go`) is therefore dark in CI, and has been silently reported as passing.

`grep -rn 'DB_HOST\|DB_NAME\|DB_USER' --include='*.go' .` confirms no test in the `go test` target reads those variables; the only consumers are `cmd/tools` and `internal/bootstrap` production config paths.

## Why it matters beyond the wasted service

`internal/integration_suite_test.go` is the only venue in the repo that wires a real Postgres to a real `EntityManager` with `Entity.SchemaDirectory` set — i.e. the only place relation enrichment runs end to end against a real database (`internal/entity_manager.go:102-106` loads the relation index from that directory). Its coverage is not duplicated by the production e2e harness, which builds its own manager over fixture schemas.

## Fix

Build the DSN from the same `DB_*` variables the rest of the repo already reads (`internal/bootstrap.Env` with the current hardcoded values as defaults), so a local developer's `scripts/local_server.sh` Postgres and CI's service both work.

[...]

Issue #384, opened 2026-08-09
EAV-only integer/smallint values above 2^31 are NULL on every DuckDB tier but matched in Postgres


iceboundrock:
Found while auditing #355's "latent edge" against `main` @ `c819301`. #355 asks whether an EAV-only *integer*-typed attribute can produce a PG-vs-DuckDB filter disagreement above 2^53. It can — but the disagreement starts at **2^31**, and it is not a parameter-binding problem, so #355 explicitly leaves it here.

## The divergence

An `integer`-typed attribute with no column binding is stored in `eav_data.value_numeric` (unconstrained PG `NUMERIC`) and read back by the two engines under different types:

- **Postgres** compares the stored `NUMERIC` normally — a value of, say, `4294967296` is matched by an equality filter.
- **DuckDB, all three tiers**, projects it through a 32-bit cast and gets NULL:
  - hot pivot: `TRY_CAST(MAX(CASE WHEN attr_id = N THEN value_numeric END) AS INTEGER)` (`internal/sqlgen/duckdb_schema_projection.go:259`; `eavElementCastExpr` applies the same cast to list elements at `:280`);
  - parquet export: `TRY_CAST(value_numeric AS INTEGER)` via `duckTypeForValue` (`internal/cdc/duckdb_exporter.go:344`, `:255-278`).

`smallint` has the same shape at 2^15 (`duckdb_schema_projection.go:261`).

So the same filter answers "1 row" on the hot Postgres leg and "0 rows" on any DuckDB leg — the class of tier disagreement #281 set out to eliminate, and one that no change to parameter binding can affect, since the column value is already NULL before the predicate runs.

## Why the value gets in

The write path does not range-validate. `populateTypedValue` handles the whole numeric family — `numeric`, `bigint`, `integer`, `smallint` — with a single `numutil.Float64` call and no bound check against the declared type (`internal/transform/typed_value.go:35-45`). JSON Schema `"type": "integer"` carries no implicit magnitude constraint, so the #315 write-path validation does not catch it either.

## Reachability

`generate-attributes` maps every JSON Schema `integer`/`number` to value type `numeric` (`cmd/tools/generate_attributes.go:459`), so `integer`/`smallint` value types only arise from hand-authored attributes files. Those exist: `internal/e2e_harness/production/schemas/e2e_wide_attributes.json:76-77` declares `level` (smallint) and `qty` (integer), both EAV-only.

## The decision this needs

Two independent halves, and the answer to the first constrains the second:

[...]

Issue #357, opened 2026-08-04
bigint predicate exactness: integral decimal/exponent spellings still lossy; EAV binder lacks execution-level boundary probe (PR #356 post-merge review)


iceboundrock:
Post-merge review of PR #356 (issue #281) surfaced two findings plus two standards items — this issue tracks their resolution.

**P1 — integral spellings beyond bare integers remain lossy.** `parseDuckDBRawParam` keeps exactness only when `strconv.ParseInt` succeeds (`internal/sqlgen/dualpath_sql_helpers.go:129-133`); the condition DSL accepts arbitrary text after `op:`, so `equals:9007199254740993.0` and `equals:9.007199254740993e15` fall through to `ParseFloat` + `%.15g`. The PG paths do likewise via `numutil.TryParseNumber`. Resolution (normalize-not-reject, per the #219 precedent): any accepted spelling denoting an integral value within int64 range binds exact int64 on every path; genuinely fractional literals keep float64. Accepted grammar itself must not widen or narrow (gate on the existing `ParseFloat` acceptance).

**P2 — the changed PG EAV binder has no execution-level boundary probe.** PR #356 changed `normalizePgEavPayload` to bind int64 but the filter e2e only exercises the bound `amount`; the EAV-only `total` case is characterization-only. Resolution: hot-Postgres + cold-DuckDB e2e assertions that an EAV-only bigint beyond the write ceiling shows the documented exact-bind/miss behavior (miss at the unstorable exact value, match at the stored rounded value, both tiers alike).

**Standards:** rename `reversedInt64s` → verb-first (`coding-standard.md` §4); `normalizePgEavPayload` at 85 lines exceeds the §7 >80-line refactoring trigger — extract the numeric-family parse arm (which P1 touches anyway).

Refs #281, PR #356 review comments (issuecomment-5185811765 ff).

iceboundrock:
Resolved by merged PR #358 (merge commit 65e535043975c0485883f991e155e6cc7a332ef3). All required CI checks passed before merge.

PR #387, opened 2026-08-09
fix: #355 bind batch attr-value operands int64-first and unify the numeric-family parse rule

### LINKED ISSUE #355: GetPersistentRecordsByAttrValues binds numeric operands as []float64 (bigint exactness; plus EAV integer-typed attr edge)

iceboundrock:
Closes #355.

Approved design: https://github.com/Lychee-Technology/forma/issues/355#issuecomment-5229859768 · Implementation plan: https://github.com/Lychee-Technology/forma/issues/355#issuecomment-5229990260

## What this fixes

`buildAttrValuesAnchor` — the builder behind the batch `value_numeric = ANY($n)` lookup that relation enrichment uses — parsed every numeric operand with `strconv.ParseFloat` and bound `[]float64`. Since #281/#357 the predicate binders (`ConvertPgMainValue`, `parsePgEavValue`, `parseDuckDBRawParam`) go int64-first, and that contract is pinned by e2e (`TestEAVBigintFilterBoundaryBothDialects`): an operand the write path rounded away must **not** find a row by colliding with the same rounding error, while the value actually stored must stay addressable.

This anchor was the only numeric binder that disagreed. Same attribute, same operand: a filter said "no match", relation enrichment said "match".

## Audit — the issue's scope changed

Three of the filed issue's premises did not survive checking against `main`:

1. **The numeric arm is reachable only for EAV-only attributes, through one production caller.** `QueryPersistentRecordsByAttrValues` is called only by `entityRelationService.fetchParents`, and the anchor's first arm rejects any non-text column binding, so a column-bound bigint never reaches the numeric arm.
2. **"This may end up document-only" no longer held.** That hedge predated #357. Once the predicate path pinned the opposite contract, leaving this binder alone meant shipping a documented fork.
3. **The filed "latent edge" is real but the bind is not the load-bearing part.** For an EAV-only integer attribute the divergence starts at **2^31**, not 2^53, and no parameter type affects it — filed separately as #384.

## The three changes

**1. `buildAttrValuesAnchor` binds per element** (`dfa2243`). `numutil.TryParseNumber` per operand: integral literals in any accepted spelling → exact `int64`, genuine fractions → `float64`, bound as `[]any`. Per-element typing is the point — parity with `ConvertPgMainValue` is guaranteed *by construction* only when the same input string yields the same Go type on both paths. Two tempting alternatives were rejected in the design: collapsing to one array type loses exactness for the integral members of a mixed set, and rendering exact decimals into a `::numeric[]` cast is a net regression, because a fractional operand would become an exact `NUMERIC` and stop matching the float64 value the write path stored. The error message no longer claims "not parseable as float64" and names the numeric family instead; it stays a plain read-path error.

**2. A real-Postgres probe** (`d2ba313`). `pgxmock` cannot see pgx's wire encoding, so the unit tests cannot prove a per-element-typed `[]any` reaches Postgres as `numeric[]`. The probe calls the repository directly against the harness cluster and asserts the same contract the predicate path pins. **Empirical result: pgx encodes the mixed `int64`/`float64` slice into `numeric[]` with no error**, so the design's documented fallback was not needed.

**3. Integer/smallint join the numeric family's parse rule** (`63faf4c`). `parseDuckDBRawParam`'s integer/smallint arm merges into its numeric/bigint sibling, **and** `ToDuckDBParam` gains a matching `int64` passthrough — without the second edit the first is a no-op, because `toOptionalFloat64Param` converts the exact `int64` straight back to `float64` one call later at the hop that actually binds. The bigint/numeric arm never hit this because it exits through `toDuckDBDecimalParam`.

**Change 3 alters no query result.** It removes a divergence, nothing more. Below the arm's own cast bound (2^31 for `CAST(? AS INTEGER)`, 2^15 for `CAST(? AS SMALLINT)`) `int64` and `float64` denote the same number; above it both parameter types raise the same class of conversion error, so the query fails identically either way. *(Note: the design originally justified this with the column-side `TRY_CAST(… AS INTEGER)` → NULL projection. Review showed that reasoning is wrong on the parameter side — the parameter cast is a strict `CAST`, so the query errors before the column's NULL-ness is ever reached. The conclusion stands; the mechanism in the code comments has been corrected, and #384 has been updated.)*

## Behaviour change, stated plainly

The anchor change is user-visible in one narrow case: a child FK carrying an exact integer above 2^53 — e.g. a text-typed FK attribute — pointing at a bigint parent-ID attribute whose stored value was rounded at write will now stop enriching, where the old `[]float64` bind found the parent by sharing the write path's rounding error. That is the intended contract and it is what the equivalent filter already does, but it is a behaviour change rather than pure hygiene.

## Verification

- **Mutation-verified.** With the anchor's `int64` arm reverted to append `float64(parsed)`, `exact_operand_above_ceiling_matches_nothing` fails with `unexpected row`; restored, it passes. The DuckDB half was mutation-checked the same way via `go test -overlay`: deleting the `ToDuckDBParam` passthrough fails its own test plus five characterization rows.
- `make fmt-check && make lint && make test` green.
- `make test-e2e-production`: 231 PASS / 0 FAIL.
- Federated e2e smoke green.

**What the e2e deliberately does not cover.** The probe calls the repository rather than relation enrichment, because enrichment cannot exhibit the contract difference: `fetchParents` derives its operands from stored child records, which have already been through the write path's float64 hop, so no non-float64-exact operand can reach the anchor that way.

## Follow-ups (none closed by this PR)

- **#384** — EAV-only integer/smallint above 2^31: NULL on every DuckDB tier vs matched in Postgres; updated during this work with the corrected mechanism (the predicate side *errors* rather than silently diverging).
- **#385** — CI provisions a Postgres service the `internal` integration suite never connects to; the suite silently skips.
- **#386** — `conditionexpr.ParseNumeric` is a dead fifth numeric parse rule that diverges from `TryParseNumber` on `"42.0"`.

### DIFF
--- a/internal/e2e_harness/production/eav_numeric_batch_lookup_e2e_test.go
+++ b/internal/e2e_harness/production/eav_numeric_batch_lookup_e2e_test.go
@@ -1,0 +1,88 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/lychee-technology/forma/internal"
+	"github.com/lychee-technology/forma/internal/model"
+)
+
+// TestBatchAttrValueLookupBindsEAVNumericExactly is the real-Postgres half of
+// the #355 verification. It calls the batch anchor directly rather than through
+// relation enrichment for a reason recorded in the design: fetchParents derives
+// its operands from stored child records, which have already been through the
+// write path's float64 hop, so no non-float64-exact operand can ever reach the
+// anchor that way and the contract difference is invisible end to end.
+//
+// The contract mirrors TestEAVBigintFilterBoundaryBothDialects: the write path
+// rounded 2^53+1 down to 2^53 (#205 ceiling), so an exact operand of 2^53+1
+// must NOT find the row by colliding with that same rounding error, while the
+// value actually stored must stay addressable. The mixed integral/fractional
+// probe additionally proves what pgxmock structurally cannot: that pgx encodes
+// a per-element-typed []any into numeric[].
+func TestBatchAttrValueLookupBindsEAVNumericExactly(t *testing.T) {
+	cluster := SharedCluster(t)
+	env := NewEnv(t, cluster)
+	ctx := context.Background()
+	wide := DefaultSchemaFixtures()[1]
+
+	// total (attr 15) is EAV-only bigint; ratio (attr 16) is EAV-only numeric.
+	aboveCeiling := CreateEvent(wide, map[string]any{
+		"title": "batch-2p53p1", "total": int64(1)<<53 + 1, "ratio": 2.5,
+	})
+	small := CreateEvent(wide, map[string]any{
+		"title": "batch-small", "total": int64(7), "ratio": float64(8),
+	})
+	mustApplyEvents(ctx, t, env, "batch anchor seed", aboveCeiling, small)
+
+	repo := internal.NewDBPersistentRecordRepository(env.Pool, env.Metadata)
+
+	lookup := func(t *testing.T, attr string, values []string) []*model.PersistentRecord {
+		t.Helper()
+		page, err := repo.QueryPersistentRecordsByAttrValues(
+			ctx, env.Tables, wide.ID, attr, values, len(values))
+		if err != nil {
+			t.Fatalf("batch lookup on %s%v: %v", attr, values, err)
+		}
+		return page.Records
+	}
+
+	t.Run("exact_operand_above_ceiling_matches_nothing", func(t *testing.T) {
+		// 9007199254740993 was never stored — the write rounded it to 2^53.
+		// Pre-fix, ParseFloat rounded the operand the same way and "found" the
+		// row anyway. No groups means: the result must be empty.
+		assertRowIDSet(t, "total=2^53+1",
+			lookup(t, "total", []string{"9007199254740993"}))
+	})
+
+	t.Run("stored_rounded_value_stays_addressable", func(t *testing.T) {
+		assertRowIDSet(t, "total=2^53",
+			lookup(t, "total", []string{"9007199254740992"}),
+			[]uuid.UUID{aboveCeiling.RowID})
+	})
+
+	t.Run("exponent_spelling_addresses_the_same_row", func(t *testing.T) {
+		assertRowIDSet(t, "total=9.007199254740992e15",
+			lookup(t, "total", []string{"9.007199254740992e15"}),
+			[]uuid.UUID{aboveCeiling.RowID})
+	})
+
+	t.Run("multi_operand_integral_set", func(t *testing.T) {
+		assertRowIDSet(t, "total IN (2^53, 7)",
+			lookup(t, "total", []string{"9007199254740992", "7"}),
+			[]uuid.UUID{aboveCeiling.RowID, small.RowID})
+	})
+
+	t.Run("mixed_integral_and_fractional_operands_encode", func(t *testing.T) {
+		// The []any carries an int64 and a float64 in one array. If pgx cannot
+		// encode that into numeric[], this is where it fails — loudly, with an
+		// encode error rather than a wrong row set.
+		assertRowIDSet(t, "ratio IN (2.5, 8)",
+			lookup(t, "ratio", []string{"2.5", "8"}),
+			[]uuid.UUID{aboveCeiling.RowID, small.RowID})
+	})
+}
--- a/internal/postgres_persistent_repository_by_attr_values.go
+++ b/internal/postgres_persistent_repository_by_attr_values.go
@@ -3,10 +3,10 @@ package internal
 import (
 	"context"
 	"fmt"
-	"strconv"
 
 	"github.com/lychee-technology/forma"
 	"github.com/lychee-technology/forma/internal/model"
+	"github.com/lychee-technology/forma/internal/numutil"
 )
 
 // QueryPersistentRecordsByAttrValues fetches full records whose attribute
@@ -85,17 +85,29 @@ func buildAttrValuesAnchor(attr string, meta forma.AttributeMetadata, values []s
 
 	switch meta.ValueType {
 	case forma.ValueTypeNumeric, forma.ValueTypeInteger, forma.ValueTypeBigInt, forma.ValueTypeSmallInt:
-		numeric := make([]float64, 0, len(values))
+		// Per-element typing, mirroring ConvertPgMainValue and
+		// parseDuckDBRawParam (#281, #357): one literal must mean one Go value
+		// on every binder, or the same operand can match under a filter and
+		// miss under relation enrichment (#355). Integral literals in any
+		// accepted spelling bind as exact int64; genuinely fractional ones stay
+		// float64, which is what the write path's own float64 hop stored.
+		// Collapsing the slice to a single element type is not equivalent:
+		// []float64 loses exactness above 2^53, and rendering exact decimals
+		// into a ::numeric[] cast would stop matching stored fractional values.
+		operands := make([]any, 0, len(values))
 		for _, v := range values {
-			parsed, err := strconv.ParseFloat(v, 64)
-			if err != nil {
+			switch parsed := numutil.TryParseNumber(v).(type) {
+			case int64:
+				operands = append(operands, parsed)
+			case float64:
+				operands = append(operands, parsed)
+			default:
 				return "", nil, false, fmt.Errorf(
-					"invalid value %q for %s attribute %q: not parseable as float64 (expected a decimal value_numeric operand): %w",
-					v, meta.ValueType, attr, err)
+					"invalid value %q for %s attribute %q: not a numeric-family literal (expected a decimal value_numeric operand)",
+					v, meta.ValueType, attr)
 			}
-			numeric = append(numeric, parsed)
 		}
-		return "t.attr_id = $2 AND t.value_numeric = ANY($3)", []any{meta.AttributeID, numeric}, false, nil
+		return "t.attr_id = $2 AND t.value_numeric = ANY($3)", []any{meta.AttributeID, operands}, false, nil
 	default:
 		return "", nil, false, fmt.Errorf(
 			"unsupported value_type %s for attribute %q: no batch value column (expected text, uuid, or a numeric family type)",
--- a/internal/postgres_persistent_repository_by_attr_values_test.go
+++ b/internal/postgres_persistent_repository_by_attr_values_test.go
@@ -31,6 +31,14 @@ func newByAttrValuesTestCache(t *testing.T) *schemameta.MetadataCache {
 			AttributeID: 31,
 			ValueType:   forma.ValueTypeInteger,
 		},
+		"total": {
+			AttributeID: 32,
+			ValueType:   forma.ValueTypeBigInt,
+		},
+		"ratio": {
+			AttributeID: 33,
+			ValueType:   forma.ValueTypeNumeric,
+		},
 	}))
 	return cache
 }
@@ -126,7 +134,7 @@ func TestQueryPersistentRecordsByAttrValuesNumericValues(t *testing.T) {
 	columns, values := optimizedQueryFixtureColumnsAndValues(rowID, int64(1))
 
 	mock.ExpectQuery(`t\.attr_id = \$2 AND t\.value_numeric = ANY\(\$3\)`).
-		WithArgs(int16(1), int16(31), []float64{7, 9}, 2, 0).
+		WithArgs(int16(1), int16(31), []any{int64(7), int64(9)}, 2, 0).
 		WillReturnRows(pgxmock.NewRows(columns).AddRow(values...))
 
 	page, err := repo.QueryPersistentRecordsByAttrValues(ctx, tables, 1, "rank", []string{"7", "9"}, 2)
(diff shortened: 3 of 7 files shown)


iceboundrock:
## Verification record

Executed via subagent-driven development: a fresh implementer per task, a spec+quality review after each, and a whole-branch review at the end. What the reviews actually caught, for the record:

**Task 1** — spec ✅, quality approved first pass. The reviewer independently traced pgx v5.8.0's `encodePlanArrayCodec`, which re-plans per element off `reflect.TypeOf(elem)`, and predicted the mixed `[]any` would encode into `numeric[]`. Task 2 then confirmed it empirically.

**Task 2** — spec ✅, quality approved. Mutation-verified: reverting the anchor's `int64` arm to `float64(parsed)` makes `exact_operand_above_ceiling_matches_nothing` fail with `unexpected row`; restored, it passes. `make test-e2e-production`: 231 PASS / 0 FAIL. The reviewer's own judgement on why only one of five subtests is mutation-sensitive is worth preserving: since the write path pushes every stored EAV numeric through a float64 hop, *any* probe that must **match** a stored row is structurally incapable of discriminating int64 from float64 binding — only "must match nothing" probes can.

**Task 3** — two fix rounds.
- Round 1 corrected an Important finding: the rationale the plan had prescribed verbatim ("the column is `TRY_CAST` to NULL above 32-bit, so the parameter's type cannot matter") is wrong on the parameter side. The parameter cast is a strict `CAST`, so an out-of-range operand raises a Conversion Error and the column is never reached. Verified directly against DuckDB: `Type INT64 … out of range for INT32` vs `Failed to cast … to type INT32` — same class, different text. The conclusion (no query result changes) survives; the mechanism was restated at all four sites and #384 was updated.
- Round 2 caught a false completion claim: round 1 had amended only the commit *message*, leaving the three file edits uncommitted, so the commit contradicted itself. Staged and amended in.

**Final whole-branch review** — clean, no Critical or Important. It verified by mutation probe (`go test -overlay`) that both the anchor bind and the `ToDuckDBParam` passthrough are nailed down, and confirmed `numutil.TryParseNumber` is now the parse rule for all four live numeric binders (`ConvertPgMainValue`, `parsePgEavValue`, `parseDuckDBRawParam`/`ToDuckDBParam`, `buildAttrValuesAnchor`). It also established that `optimizedQueryShapeKey` hashes the arg *count*, not arg types, so the `[]float64` → `[]any` switch causes no plan-cache divergence.

## Deferred findings and their triage

Two findings against the new e2e were triaged **leave**, with reasoning worth recording rather than just the verdict:

- *`exponent_spelling_addresses_the_same_row` uses the stored, float64-exact value, so it is not sensitive to a regression of #357's integral-exponent handling.* Left: the proposed swap to the never-stored value trades one discriminator for another rather than dominating it — with an empty expectation, a regression that parses the exponent into *some* wrong number would pass silently. The float64 axis is already pinned twice at unit level. If both axes are wanted, the right move is to **add** a sixth subtest, not replace this one.
- *`limit = len(values)` leaves no headroom and `TotalRecords` is discarded.* Left: `len(values)` is exactly what production passes (`entity_relation_service.go:126`), so widening it would reduce fidelity. Asserting `page.TotalRecords == len(want)` would buy the headroom without touching the call shape, if that is wanted later.

Comment-accuracy findings were fixed in `9c06ce2` (behaviour-free — the only non-comment line in that commit is a deleted duplicate test). Two residual inaccuracies that commit itself introduced are listed in the PR body rather than silently shipped.

## Standards

Files ≤ 500 lines and functions ≤ 100 lines hold across the diff (`predicate_characterization_test.go` is at 497 — worth noting for whoever touches it next). Both error classes in the diff are deliberate: the anchor's is a plain read-path `fmt.Errorf`, `parseDuckDBRawParam`'s stays `forma.InvalidInputf`. `gofmt` clean on all seven changed files.

iceboundrock:
## CI: 5/5 green

| check | result |
| --- | --- |
| Lint | pass (42s) |
| Test | pass (1m5s) |
| E2E Smoke Tests | pass (7m18s) |
| Benchmark Smoke | pass (58s) |
| k6 Smoke | pass (2m33s) |

Run: https://github.com/Lychee-Technology/forma/actions/runs/31302916006

Awaiting review — not merged, per AGENTS.md.

iceboundrock:
## Re-review of #387 — independent verification against `main` @ `c819301`

Independent second-pass review of the four commits (`dfa2243` → `9c06ce2`), re-run without relying on the author's own verification record. Everything below was re-derived from the branch and the issue thread, and the two central mutation claims were re-executed locally.

---

## Findings

No **Critical** or **Important** findings.

**Minor 1 — the two known residual comment inaccuracies are shipped rather than fixed.** `internal/sqlgen/duckdb_type_mapper.go`:
- `:71` "Three binders depend on this contract" — the live consumer count is one (`predicate_normalizer.go:397`); `param_cast` (`template_renderer.go:89`) is registered but has **zero** template call sites (verified by grep — no template body invokes it).
- `:76` "an exact-precision decimal string, never a number" — true only for int/int64/string inputs; a float64 input exits through `decimalString`'s `%.15g` (`duckdb_type_mapper.go:220`).

The PR body documents these rather than fixing them. Since `9c06ce2` was *itself* a comment-accuracy commit, and both are one-liners, the cleaner close is to fix them in place. **Non-blocking** — the record-and-ship choice is consistent with the issue's decision-record rule, and neither line affects behavior. **Suggested fix** for `:71`: "The predicate normalizer binds through this contract" (or name the actual two call sites); for `:76`: "an exact decimal string for int/int64/string; float64 goes through %.15g".

**Minor 2 — `predicate_characterization_test.go` sits at 497 lines** (3 from the 500 limit). Pre-existing on `main` (also 497), and this PR only rewrites in-place lines, so no new violation — but the PR touches it, so worth the headroom note already carried in the PR body. If a future change adds a row there, split the builder functions first.

No spec deviations, no scope creep, no suspicious implementations found.

---

## Spec

The originating spec is issue #355 (audit in comment `5229859768`, plan in comment `5229990260`). Cross-checked every element of the plan against the diff:

| Plan item | Status |
| --- | --- |
| Task 1: `buildAttrValuesAnchor` parses per element via `numutil.TryParseNumber`, binds `[]any` | ✅ exact match (`postgres_persistent_repository_by_attr_values.go:98-113`) |
| Task 1: error message drops "not parseable as float64", names the numeric family, stays a plain read-path error (not a 4xx carrier) | ✅ — now `fmt.Errorf("…not a numeric-family literal…")`, no `%w` of a strconv error (matches the design's note that `TryParseNumber` returns a string, not an error) |
| Task 2a: `parseDuckDBRawParam` smallint/integer arm merges into the numeric/bigint arm (int64-first) | ✅ `dualpath_sql_helpers.go:123-153` |
| Task 2b: `ToDuckDBParam` smallint/integer arm accepts `int64` directly (else 2a is a silent no-op one hop later at `predicate_normalizer.go:397`) | ✅ narrow `int64` passthrough at `duckdb_type_mapper.go:126-131`; int/int16/int32 still funnel to float64 |
| §3.1 unit tests: pgxmock pins bound Go types, incl. 2^53+1 → exact int64 and fractional → float64 | ✅ three new tests + updated `WithArgs(…, []any{int64(7), int64(9)}, …)` |
| §3.2 real-Postgres probe via the production harness, hand-supplied operands (the only way to feed non-float64-exact spellings, per audit §1.4) | ✅ `eav_numeric_batch_lookup_e2e_test.go`, 5 subtests covering the probe table exactly |
| §3.2 fallback (per-element exact rendering + `$3::numeric[]` cast) only if pgx refuses `[]any` | ✅ **not needed** — probe confirms pgx v5 encodes the mixed `[]any` into `numeric[]` with no error; the documented fallback was correctly not taken |
| §3.3 gates + mutation check | ✅ mutation check independently re-run (see Verification) |
| Task 2 stated as fixing **no observable defect**, not as a bug fix | ✅ carried in commit `63faf4c` message and PR body |
| Out-of-scope: #384 (integer/smallint >2^31), #385 (integration suite silently skipped) filed separately | ✅ present on the issue, none closed by this PR |

One deliberate deviation from the design draft worth restating as a *correct* deviation: the design's original justification for Task 2 ("column is TRY_CAST to NULL above 32-bit, so the parameter type can't matter") was **wrong on the parameter side** — the parameter cast is a strict `CAST`, so out-of-range operands raise a conversion error before the column's NULL-ness matters. The commit comments and the PR body restate the corrected mechanism, and #384 was updated. Good catch by the review wave; the conclusion ("no query result changes") survives.

---

## Standards

Against `coding-standard.md` + `AGENTS.md`:

- **File sizes** — all seven changed files ≤ 500 lines (max 497, see Minor 2).
- **Function sizes** — `buildAttrValuesAnchor` 48 lines, `parseDuckDBRawParam` 59 lines, all < 100.
- **Error handling** — no bare `return err`; all wraps carry context (`fmt.Errorf("failed to query records by attr %q for schema_id %d: %w", …)`). The anchor's new error correctly stays a plain read-path error, matching `docs/error-handling.md`; `parseDuckDBRawParam` correctly keeps `forma.InvalidInputf` (write-path validation semantics). No `TestNoBareSentinelWraps` risk introduced.
- **Error messages** — the anchor error names the logical value type (`%s` attribute, `%q`), the offending value, and the expected state. Conforms.
- **gofmt** — clean on all seven files (re-verified).
- **Comments** — the "no behaviour change" caveat for Task 2 is stated at every site (arm comment, `ToDuckDBParam` doc, parity test doc, characterization row name), so the change can't be mistaken for a bug fix. See Minor 1 for the two residual inaccuracies.
- **Structure** — early returns, flat switch-based dispatch, ≤ 3 nesting levels. No code duplication introduced; the per-element parse in the anchor mirrors `ConvertPgMainValue`/`parseDuckDBRawParam` by construction (shared `TryParseNumber`), which is the point.

---

## Assessment

The PR does what it says and nothing more. The correctness argument is structural: with `TryParseNumber` as the shared parse rule across `ConvertPgMainValue`, `parsePgEavValue`, `parseDuckDBRawParam`/`ToDuckDBParam`, and now `buildAttrValuesAnchor`, the same input string yields the same Go value on every binder, so a filter and relation enrichment can no longer disagree about what a numeric literal means. The two rejected alternatives (`[]int64`/`[]float64` whole-slice typing, and exact-decimal `::numeric[]` rendering) are correctly argued against in the code comments.

The riskiest step — proving pgx encodes a per-element-typed `[]any` into `numeric[]` — is de-risked empirically rather than assumed, and the design's fallback was honestly available had it failed. The e2e deliberately probes the repository rather than `fetchParents`, and the audit's §1.4 reasoning (operands reaching enrichment are already float64-exact, so the contract difference is invisible end-to-end) is sound — an enrichment-level e2e could only have proven "no regression".

Plan-cache safety is confirmed: `optimizedQueryShapeKey` hashes the arg **count**, not arg types (`postgres_persistent_repository_query.go:127`), and the `[]float64`→`[]any` switch preserves count, so no cache divergence.

Readiness: ready to merge once Minor 1 is resolved or explicitly accepted. No blockers.

---

## Observations

- The one genuine behaviour change is stated plainly and correctly in the PR body: an exact integer above 2^53 in a text-typed FK pointing at a rounded bigint parent will now stop enriching — matching what the equivalent filter already did. This is the intended contract, not a regression.
- `TestToDuckDBParamIntegerArmPassesInt64Through` guards exactly the hop that would otherwise make Task 2a a no-op; its narrowness (int/int16/int32 keep the float64 funnel) is correctly pinned by the pre-existing `_FromInt16`/`_FromInt32` tests.
- The `numeric_bind_parity_test.go` doc comment's claim that `parsePgEavValue` "is pinned by the characterization matrix's PgArgs expectations" checks out — `parsePgEavValue` was already int64-first pre-PR and the PgArgs in `predicate_characterization_test.go` confirm it.
- `9c06ce2`'s test-drop (the duplicated `TestToDuckDBParamIntegerArmStillWidensOtherIntegerTypes`) is verified as a genuine value-for-value duplicate of `_FromInt16`/`_FromInt32`; the intent it recorded now lives in the sibling test's comment. Correct call.
- The comment "parameter cast is a strict CAST, so the query errors before the column's NULL-ness is ever reached" is accurate for the DuckDB parameter path; the accompanying `TRY_CAST` column-side note correctly routes the residual divergence to #384.

---

## Verification (re-executed, not taken on faith)

- `gofmt -l` on all seven changed files → empty.
- Unit: `go test ./internal -run TestQueryPersistentRecordsByAttrValues` → all 8 pass.
- Unit: `go test ./internal/sqlgen -run 'TestNumericBinderTypeParity|TestToDuckDBParam'` → all pass, including the new `IntegerArmPassesInt64Through`.
- Full `go test ./internal/...` → clean (no failures).
- `go vet -tags=e2e ./internal/e2e_harness/production/` → clean (the new probe compiles under the e2e tag).
- **Mutation 1** (anchor): changed the anchor's `int64` arm to `append(operands, float64(parsed))` → `TestQueryPersistentRecordsByAttrValuesBindsIntegralOperandsExactly` **fails** (operand arrives as `9.007199254740992e+15` float64, not exact int64); reverted → passes. Confirms the e2e `exact_operand_above_ceiling_matches_nothing` claim has a working unit-level guard too.
- **Mutation 2** (`ToDuckDBParam` passthrough): deleted the `int64` passthrough → `TestToDuckDBParamIntegerArmPassesInt64Through` **fails**; reverted → passes. Confirms Task 2a is not a silent no-op.
- Author's claimed gates (lint, full e2e-production 231/0, federated smoke) are corroborated by the CI run linked in the thread; I did not re-run the Docker e2e locally (time cost), but the mutation probes cover the load-bearing assertions that no other test could.

---

## Summary

**Standards: pass** (one minor comment-accuracy nit). **Spec: pass** — the diff matches the plan item-for-item, including the one deviation, which was a correction, not a drift. **Verdict: approve**, no blockers. Total: 2 minor findings (both already known to the author, one of them fixable in a line). The single worst issue is the shipped-known-inaccurate comment at `duckdb_type_mapper.go:71` — cosmetic, and already recorded.

iceboundrock:
## Response to the re-review

Thanks — both findings check out. I re-derived each before acting rather than taking the review at face value; notes below.

### Minor 1 — fixed in `b88700f`

Verified both claims first:

- `grep -rn 'param_cast'` across the tree returns exactly one hit — its own registration at `template_renderer.go:87`. No template body invokes it, so the live consumer count is **one** (`predicate_normalizer.go:397`), as the review says. The two Postgres-side binders never reach this function at all.
- `decimalString` is `fmt.Sprintf("%.15g", v)` (`duckdb_type_mapper.go:220`), so "exact-precision" holds only for the int/int64/string arms of `toDuckDBDecimalParam`.

Fixed in place rather than shipped, per your suggestion. I took the substance of your proposed wording but went slightly further on the second one, naming *which* inputs are exact instead of only flagging that float64 is not:

```
// ToDuckDBParam converts a Go value to the form expected by DuckDB drivers for
// the given value type. The predicate normalizer binds this result, not
// parseDuckDBRawParam's, so each numeric arm's output type is load-bearing:
//   - uuid.UUID -> string
//   - time.Time -> int64 epoch-ms (BIGINT)
//   - smallint/integer -> an int64 passes through unchanged (#355); every other
//     accepted input widens to float64
//   - bigint/numeric -> a decimal string, never a number (see
//     toDuckDBDecimalParam): exact for int/int64/string inputs, while a float64
//     input is rendered by decimalString's %.15g
```

The commit also drops the orphaned `// Examples:` line, which `9c06ce2` had left stranded above a sentence written in authoring-instruction voice ("…so state each numeric arm exactly") — a note to the writer that had ended up addressed to the reader.

Comment-only; `make fmt-check && make lint && make test` green, no FAIL lines.

Worth naming the pattern, since it is the second instance in this PR: **a comment-accuracy commit is itself a place where inaccurate comments get written.** `9c06ce2` corrected four sites and introduced two new errors while doing it. The mechanism is that a corrective edit is drafted with the *old* claim in mind and inherits its framing — here, "several binders share this rule" was true of `TryParseNumber` (the sibling comment's subject) and got carried onto `ToDuckDBParam`, which is a different function with a different consumer set.

### Minor 2 — no code change here; escalated to #320 instead

Confirmed `predicate_characterization_test.go` is 497 lines on **`main` as well as this branch** — this PR only rewrites lines in place, so it adds no growth and there is no new violation to fix. Splitting the builders inside this PR would bury a comment-and-binding diff in unrelated churn, which is the same reasoning #314 used when it waived the two oversized `internal/` test files.

Recorded on **#320** (comment `5232708243`) as a near-miss, along with something the review surfaced indirectly: #320's own text says the 500-line cap "cannot be linted", but the repo already has two hand-rolled guards that enforce it as ordinary tests — `internal/e2e_harness/federated/query_file_size_test.go` and `internal/httpapi/server_file_size_test.go`. The mechanism exists and is proven; only the coverage is missing. A guard in `internal/sqlgen` globbing `*.go` would have caught this file at 497 before it crossed, and it composes with #369, which is already about widening one of those globs. That is a better close than a one-off split.

### PR body updated

The "Known residuals" section is removed, since there are no longer any.

---

No open items from this review. CI will re-run on `b88700f`; still not merged, per AGENTS.md.

iceboundrock:
## CI on `b88700f`: 5/5 green

| check | result |
| --- | --- |
| Lint | pass (49s) |
| Test | pass (1m4s) |
| E2E Smoke Tests | pass (7m18s) |
| Benchmark Smoke | pass (1m8s) |
| k6 Smoke | pass (2m32s) |

Run: https://github.com/Lychee-Technology/forma/actions/runs/31325569922 · `MERGEABLE / CLEAN`

Both re-review findings closed: Minor 1 fixed in `b88700f`, Minor 2 escalated to #320 (no change warranted here — the file is 497 on `main` too and this PR only rewrites lines in place). No open items. Not merged, per AGENTS.md.

iceboundrock:
## Verification

Verified locally against PR head `b88700f6f01f2ac23d51ebc5183b9b4e7f5abe93`:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal -run 'TestQueryPersistentRecordsByAttrValues' -v` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/sqlgen -run 'TestNumericBinderTypeParity|TestToDuckDBParamIntegerArm' -v` passed.
- `make fmt-check && make lint && make test` passed.
- `go test ./internal/e2e_harness/production -tags=e2e -run TestBatchAttrValueLookupBindsEAVNumericExactly -v -timeout=10m` passed, including all five subtests and real Postgres mixed-array encoding.
- `make test-e2e-production` passed.
- `go test -v ./internal/e2e_harness/federated/... -tags=e2e -short -timeout=30m` passed.
- `git diff --check` passed.

iceboundrock:
## Observations

- The user-visible change is intentionally narrow: an exact integral operand above 2^53 no longer aliases a value rounded at write time in the batch lookup path.
- The new e2e calls `QueryPersistentRecordsByAttrValues` directly. It does not exercise relation enrichment end to end because enrichment derives operands from already-stored float64 values; the PR documents this limitation correctly.
- The DuckDB integer/smallint change is parity-only and is not presented as fixing the separate #384 projection/range issue.
- Existing follow-ups #384, #385, and #386 remain visible and are not incorrectly closed by this PR.

iceboundrock:
## Spec

Source reviewed: issue #355 and its approved design/implementation plan:

- https://github.com/Lychee-Technology/forma/issues/355#issuecomment-5229859768
- https://github.com/Lychee-Technology/forma/issues/355#issuecomment-5229990260

Spec assessment:

- The batch numeric anchor now applies `TryParseNumber` per operand and binds a mixed `[]any`, preserving exact integral literals while retaining float64 semantics for fractional values.
- The real-Postgres probe verifies both the observable 2^53 contract and mixed `int64`/`float64` encoding into `numeric[]`.
- The DuckDB integer/smallint parser and `ToDuckDBParam` passthrough are changed together, so the new parse type is not converted back to float64 downstream.
- The characterization expectations are updated for the intended DuckDB argument-type change.
- The intentionally direct repository-level e2e, rather than relation enrichment, matches the approved plan and its stated observability constraint.
- No missing requirement, incorrect implementation, or unrequested production scope was found. The #384/#385 follow-ups remain correctly out of scope.

iceboundrock:
## Standards

Pass.

Against `AGENTS.md` and `coding-standard.md`:

- Changed source files remain within the 500-line limit and changed functions remain within the 100-line limit.
- New error paths include attribute/value/type context; no new bare error return was introduced.
- Control flow remains flat and does not introduce excessive nesting.
- Naming and package structure follow the surrounding Go code.
- No material duplication or unrelated refactor was introduced.

The format and lint gates also pass.

iceboundrock:
## Summary

Review complete: no findings.

Spec: pass.
Standards: pass.
Verification: pass, including unit, lint, format, targeted real-Postgres e2e, full production e2e, and federated e2e smoke.

The PR is ready for merge from the code-review perspective.

iceboundrock:
## Findings

No actionable findings identified.

I reviewed the complete diff from `c819301d6944b1fb289e247cdf45d4291f1890a3` to `b88700f6f01f2ac23d51ebc5183b9b4e7f5abe93`, including all seven changed files. No blocking, high-, medium-, or low-severity correctness issue was found in the changed behavior or tests.

iceboundrock:
## Assessment

Recommendation: approve.

The PR removes the remaining numeric binder contract fork without widening the change into the separately tracked EAV integer/smallint range defect. The implementation, tests, and stated behavior change are consistent with the approved #355 plan.

Issue #402, opened 2026-08-13
Numeric literal outside int64 and float64 range (e.g. 1e400) answers a redacted 500 for caller input


iceboundrock:
Follow-up from #322, found during its final branch review.

## The gap

A caller-supplied numeric literal that fits neither `int64` nor `float64` — e.g. an HTTP body of `{"score": 1e400}` — answers an operator-visible, redacted **500**, though it is pure caller input.

Chain (verified end to end on the #322 branch):

1. `internal/httpapi/server.go` decodes request bodies with `UseNumber`, so the literal survives as a `json.Number` carrying the exact text `1e400`.
2. `Validator.Validate` marshals the payload — `json.Marshal` of a `json.Number` re-emits the text verbatim, so the marshal **succeeds** (the #322 carrier at the marshal site never fires).
3. The round-trip redecode hands the literal to `exactNumberInstance` (`internal/schemavalidate/number_instance.go`), where both `Int64()` and `Float64()` fail (`value out of range`), and the resulting error is a **plain** `fmt.Errorf` (`validator.go`, "failed to decode payload for schema %d").
4. `validateWritePayload` returns any non-carrier error regardless of enforcement (report-only mode rightly refuses to absorb it — the document was never validated), and the HTTP boundary's default arm answers 500 with a redacted body.

So: valid JSON, one attribute out of representable range, and the caller gets a 500 they cannot act on, in both strict and report-only modes.

## Why #322 did not fix it

#322 classified the *marshal* failure of the payload (NaN/Inf, reachable only from a Go embedder) as `forma.ErrInvalidInput`. The `1e400` case sails through the marshal and fails one step later, in the exact-number rewrite that exists to keep >2^53 constraint checks exact (#282). It is HTTP-reachable, which the marshal case is not.

Note the shared wrap text: the truly-internal case (redecode of the marshaller's own output failing) and this caller-input case both surface as "failed to decode payload for schema %d". Any fix should split them rather than blanket-classify the wrap — the redecode failure is genuinely an internal fault.

## Direction

Classify the `exactNumberInstance` out-of-range failure as `forma.InvalidInputf`, naming the offending literal (the error already carries it: `numeric literal %q fits neither int64 nor float64`). The literal is the caller's own value, so publishing it follows the same reasoning as #322's marshal-site message. The genuinely-internal redecode branch stays plain.

## Also worth deciding here (from #322's review, deferred)

[...]

Issue #318, opened 2026-07-26
StripComputedFields strips relation roots by exact key, so the dotted spelling is persisted and the nested one is not


iceboundrock:
Follow-up from #314 (PR #316).

## The inconsistency

`StripComputedFields` (`internal/relation_index.go`) removes relation roots — properties whose `$ref` carries `x-relation` — by **exact key match**. So the two spellings of the same attribute are treated differently:

| payload | outcome |
| --- | --- |
| `{"contactSnapshot": {"name": "Ada"}}` | stripped, **not persisted** |
| `{"contactSnapshot.name": "Ada"}` | survives the strip, **persisted** |

`contactSnapshot.name` is a registered attribute (`cmd/server/schemas/visit_attributes.json`), so the dotted spelling writes a computed/relation field that the nested spelling deliberately discards.

This predates #314 — the strip has always been exact-key.

## Why #314 did not fix it

#314 surfaced it. Normalization would have rebuilt the stripped root from the surviving dotted descendant, producing a 400 the caller could not fix by sending the field (it gets stripped first). The fix there was to skip expansion beneath a relation root, which was chosen **specifically to avoid changing what gets persisted**: prefix-stripping dotted descendants would have stopped persisting a field that is persisted today, which is a behaviour change well outside that issue's scope.

So the current state is deliberate and documented: values beneath a relation root persist as they always have, and are simply not schema-validated (`docs/error-handling.md`, "Relation roots are never validated").

## The decision to make

Which spelling is right?

- **If the nested behaviour is correct** (a computed field should never be caller-writable), then `StripComputedFields` should prefix-match — `k == root || strings.HasPrefix(k, root+".")` — and the dotted spelling should stop persisting. That is a write-behaviour change and needs a migration thought: rows that already hold caller-written `contactSnapshot.*` values keep them.
- **If the dotted behaviour is correct** (these are ordinary attributes the caller may set), then the nested strip is over-aggressive, and relation roots should be strippable only for the computed sub-fields that are genuinely derived.

Either way the validation gap closes as a side effect, because a consistent rule makes the spellings agree.

## Scope note

Only `visit.contactSnapshot` carries `x-relation` today, so the blast radius is one attribute family. There is a related hazard already recorded in `internal/entity_crud_service.go`: if a future schema marks a relation root **required**, every create and update to that entity is rejected with a missing-required error the caller cannot fix, because the field is stripped before validation. Whichever direction this issue takes should resolve that too.

[...]

PR #403, opened 2026-08-13
fix: #322 classify non-finite payload numbers as invalid input and reject them at transform

### LINKED ISSUE #322: NaN/Inf in an embedder payload is classified as an operator fault (5xx) instead of invalid input
Follow-up from #314 (PR #316).

## The misclassification

A payload carrying `math.NaN()` or `math.Inf()` in a numeric attribute cannot be marshalled, so `Validator.Validate` (`internal/schemavalidate/validator.go`) fails at `json.Marshal` and returns a **plain** error. Plain means operator-visible: 500 on create, and on the report-only update path it is returned rather than absorbed, because #314 deliberately gates the `Warn`-and-proceed on `errors.Is(err, forma.ErrInvalidInput)`.

But NaN in a request body *is* caller input. It should be a 400, not a 500.

## Reachability

**Not reachable over HTTP** — `encoding/json` cannot decode `NaN` or `Infinity`, so no request body can produce one. It is reachable only for a **Go embedder** calling `EntityManager.Create`/`Update` directly with a Go-constructed map.

The misclassification predates #314 (the marshal error was always plain). What #314 changed is that it now *fails the write* on the report-only update path, where previously the write proceeded.

## Direction

Classify a marshal failure of the caller's own payload as `forma.ErrInvalidInput`, so it surfaces as 4xx for embedders and can be absorbed by report-only mode like any other violation. The message should name the offending path if `encoding/json`'s error carries it.

Worth deciding at the same time whether the transform layer should reject non-finite floats earlier with a message naming the attribute — that would be more actionable than a marshal error either way, and it is the layer that already rejects other unrepresentable values.

iceboundrock:
Closes #322.

## What this fixes

A write payload carrying `math.NaN()`/`math.Inf()` failed `Validator.Validate` at `json.Marshal` with a **plain** error — surfacing as an operator-visible 5xx, though it is the caller's own input. Worse, the planning investigation found the issue's "not reachable over HTTP" premise is false for the string spelling: `numutil.Float64` coerces strings via `strconv.ParseFloat`, which accepts `"NaN"`/`"Inf"`/`"Infinity"`, so an HTTP update body `{"score": "NaN"}` under the shipped report-only default was warn-absorbed as a schema violation and then **stored a NaN** — a poison row that 500s on every subsequent read (the response `json.Marshal` cannot represent it).

## The original branch (7 commits, ordered so every intermediate commit is safe)

1. **`406a48f` — transform-layer guard (the load-bearing half).** `finiteForEAV` rejects NaN/±Inf after numeric coercion at both write-side `ValueNumeric` funnels: `populateTypedValue` (publishes `invalid value for attribute '<name>' (attrID=<id>): non-finite number <v> is not storable; a finite value is required`) and `ToEAVRecord`'s numeric case (plain, per that converter's existing class). Deliberately **not** inside `numutil.Float64`, which also serves the DuckDB query path. Landing this first is what makes step 2 safe. Includes a pure move of `toFloat64ForEAV`/`parseTrimmedFloat64` into `numeric_conversion.go` to keep `attribute_converter.go` under the 500-line standard.
2. **`903ddc9` — validator reclassification (the issue's headline).** A `json.Marshal` refusal of the caller's payload now returns a published `forma.InvalidInputf` — 4xx for embedders, absorbable by report-only mode like any carrier, and safe to absorb only because of (1). Missing resolved schema and post-marshal failures stay plain.
3. **`6d2a441` + `f21a705` — manager-level pins.** New `entity_write_nonfinite_test.go`: embedder `math.NaN()` on Create → `ErrInvalidInput`; report-only update rejects **both** the float64 and the HTTP-reachable `"NaN"` string spelling naming the attribute, and the stored finite value survives.
4. **`d4fbddb` + `9c14bed` + `571254b` — prose sync.** Every site claiming "marshal failures stay plain" corrected; the class split is stated everywhere on **sentinel evidence**, matching the code.

## Review waves (7 further commits, each responding to review findings on this PR)

- **`f17a54f` — bool funnels closed (review P1).** `toBool`/`toBoolForEAV` silently coerced non-finite floats — and disagreed (NaN → true vs false). Finite check now guards every float/`json.Number` branch of both; `TestReportOnlyUpdateRejectsNonFiniteBool` pins the report-only path. Side effect, blessed as intended: a stored bool `value_numeric` holding NaN (corrupt storage — every writer produces only 0/1) now reads as a loud plain error instead of silently `false`.
- **`cbd5f47` + `ad74370` — attrID context (review P2).** `ToEAVRecord`'s conversion errors name `attrID` (that layer holds no attribute cache; the name-bearing carrier is `populateTypedValue`'s, which runs first).
- **`e9f1377` — published marshal message names the attribute (review P2).** `encoding/json` carries no path, so after a refusal the validator walks the payload (`nonFinitePaths`) and publishes e.g. `payload cannot be encoded as JSON: attribute "profile.score" holds non-finite number +Inf; a finite number is required`.
- **`077e252` + `d81e3ac` — walk hardened (round-2 P1).** Depth cap 1000; cycles and >1000-deep payloads **abandon** the walk (nil, never a truncated list) and publish the library text. Abandon-over-truncate is mutation-pinned.
- **`f1848ec` — walk taught the remaining refusing shapes (round-2 P2).** `*float64`/`*float32` and `json.Number` non-finite spellings; `json.Number("1e400")` is deliberately NOT flagged (valid JSON grammar, marshals fine — ParseFloat's ErrRange is the probe-verified tell).
- **`1b36ecc` — bool wording (round-3):** bool rejections publish `non-finite value <v> has no truth value; a finite value is required` instead of the numeric-flavored prose.

**Known boundary (deliberate):** the diagnostic walk is shape-bound to the payload shapes JSON decoding produces (`map[string]any`, `[]any`, floats, their pointers, `json.Number`). An embedder handing concrete Go containers (`[]float64{NaN}`, `map[string]float64`) still gets the 4xx carrier, but with the library's text and no attribute named — documented in `nonfinite_paths.go` rather than type-switching every numeric container.

## Found along the way

- **#402** — `{"score": 1e400}` (numeric literal fitting neither `int64` nor `float64`) answers a redacted 500 for pure caller input; documented honestly at every touched site, out of scope here.
- **#404** — the two bool funnels embody different truth tables for finite values they both accept (0.3/negatives/strings).
- **#405** — read-path consistency errors name attrID but never the row.
- **#406** — cyclic embedder payloads crash at transform normalization, before any guard can reject them (pre-existing; the round-2 depth cap is precedent for the fix shape).

## Verification

- TDD throughout; red-first and mutation evidence per wave (disabling any guard reddens exactly its pins; `-overlay` runs re-executed with `-count=1` after one `ok (cached)` non-result).
- `make fmt-check`, `make lint` (pinned golangci v1.64.8), full unit suite — clean at every wave head. The only local failure is `internal/e2e_harness`'s Docker-dependent smoke (daemon down locally — environmental); CI runs it green.
- Full review artifacts and per-finding responses are in the PR comments; execution record on #322.

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -135,11 +135,13 @@ payload.
   it at `Warn` and proceeding — safely even for the marshal case, because the
   transform layer independently rejects non-finite numbers with the attribute
   name before anything reaches storage.
-- **Everything else is returned regardless of enforcement** — in practice a
-  missing resolved schema. That is a plain error, therefore `500`, therefore
-  operator-visible. Absorbing it into report-only would write the document with
-  *zero* validation while a log line claimed it had merely failed a check, and
-  would blame the caller for a server fault.
+- **Everything else is returned regardless of enforcement** — a missing
+  resolved schema, plus the one caller-input case misfiled here (#402), both
+  enumerated under "Public HTTP error surface". Those are plain errors,
+  therefore `500`, therefore operator-visible. Absorbing them into report-only
+  would write the document with *zero* validation while a log line claimed it
+  had merely failed a check — and, for the missing schema, would blame the
+  caller for a server fault.
 
 ### Startup fails closed
 
@@ -805,9 +807,11 @@ Since #314 there is a **second** write-path validator on the same footing,
 for a payload `json.Marshal` refuses to encode (`NaN`/`Inf` from a Go
 embedder), which is caller input just the same. It is independent of the
 `required_policy` row above: the two check different things and both run. Its
-*remaining* errors deliberately stay plain, so a `500` — a missing resolved
-schema, and a numeric literal that fits neither `int64` nor `float64`; see
-"JSON Schema enforcement on write" for the split.
+*remaining* errors currently stay plain, so a `500` — a missing resolved
+schema, and a numeric literal that fits neither `int64` nor `float64`. Only the
+first of those is a deliberate `500`; the numeric literal is caller input
+misfiled on the operator side, a known gap tracked in #402. See "JSON Schema
+enforcement on write" for the split.
 
 Its published message deliberately includes the third-party `jsonschema-go`
 violation prose (decision recorded at the wrap site, `validator.go`): that text
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -129,14 +129,17 @@ payload.
 
 `validateWritePayload` splits on sentinel evidence, not on the enforce flag:
 
-- A genuine violation wraps `forma.ErrInvalidInput` → `400`. Report-only mode
-  absorbs exactly this case, logging it at `Warn` and proceeding.
-- **Everything else is returned regardless of enforcement** — a missing resolved
-  schema, or a payload that will not marshal (`NaN`/`Inf`). Those are plain
-  errors, therefore `500`, therefore operator-visible. Absorbing them into
-  report-only would write the document with *zero* validation while a log line
-  claimed it had merely failed a check, and would blame the caller for a server
-  fault.
+- A genuine violation wraps `forma.ErrInvalidInput` → `400`. So does a payload
+  `json.Marshal` refuses — `NaN`/`Inf` from a Go embedder (#322); no HTTP body
+  can encode one. Report-only mode absorbs exactly this carrier class, logging
+  it at `Warn` and proceeding — safely even for the marshal case, because the
+  transform layer independently rejects non-finite numbers with the attribute
+  name before anything reaches storage.
+- **Everything else is returned regardless of enforcement** — in practice a
+  missing resolved schema. That is a plain error, therefore `500`, therefore
+  operator-visible. Absorbing it into report-only would write the document with
+  *zero* validation while a log line claimed it had merely failed a check, and
+  would blame the caller for a server fault.
 
 ### Startup fails closed
 
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -801,10 +801,13 @@ stay in `Error()` and in the chain for `errors.Is`/`As`, and never in
 Since #314 there is a **second** write-path validator on the same footing,
 `internal/schemavalidate`'s `Validator.Validate`, which builds an
 `InvalidInputf` carrier for a JSON Schema violation — `enum`, `pattern`,
-`type`, `minimum`/`maximum`, and the schema's own `required`. It is independent
-of the `required_policy` row above: the two check different things and both
-run. Its *non*-violation errors deliberately stay plain, so a `500`; see "JSON
-Schema enforcement on write" for the split.
+`type`, `minimum`/`maximum`, and the schema's own `required` — and, since #322,
+for a payload `json.Marshal` refuses to encode (`NaN`/`Inf` from a Go
+embedder), which is caller input just the same. It is independent of the
+`required_policy` row above: the two check different things and both run. Its
+*remaining* errors deliberately stay plain, so a `500` — a missing resolved
+schema, and a numeric literal that fits neither `int64` nor `float64`; see
+"JSON Schema enforcement on write" for the split.
 
 Its published message deliberately includes the third-party `jsonschema-go`
 violation prose (decision recorded at the wrap site, `validator.go`): that text
--- a/internal/entity_write_nonfinite_test.go
+++ b/internal/entity_write_nonfinite_test.go
@@ -1,0 +1,114 @@
+package internal
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/lychee-technology/forma/internal/schemavalidate"
+	"github.com/lychee-technology/forma/internal/transform"
+	"github.com/stretchr/testify/require"
+)
+
+// numericValidationRegistry mirrors validationRegistry but declares a numeric
+// attribute, which the shared fixture deliberately lacks. File-local so the
+// shared fixture's assertions stay untouched.
+type numericValidationRegistry struct{}
+
+const numericSchemaJSON = `{
+  "type": "object",
+  "properties": {
+    "name":  {"type": "string"},
+    "score": {"type": "number"}
+  },
+  "required": ["name"]
+}`
+
+func (numericValidationRegistry) GetSchemaAttributeCacheByName(name string) (int16, forma.SchemaAttributeCache, error) {
+	if name != "test" {
+		return 0, nil, fmt.Errorf("schema %s not found", name)
+	}
+	return 100, forma.SchemaAttributeCache{
+		"name":  {AttributeID: 1, ValueType: forma.ValueTypeText},
+		"score": {AttributeID: 2, ValueType: forma.ValueTypeNumeric},
+	}, nil
+}
+
+func (r numericValidationRegistry) GetSchemaAttributeCacheByID(int16) (string, forma.SchemaAttributeCache, error) {
+	_, cache, err := r.GetSchemaAttributeCacheByName("test")
+	return "test", cache, err
+}
+
+func (numericValidationRegistry) ListSchemas() []string { return []string{"test"} }
+
+func (numericValidationRegistry) GetSchemaByName(string) (int16, forma.JSONSchema, error) {
+	return 100, forma.JSONSchema{ID: 100, Name: "test", Schema: numericSchemaJSON}, nil
+}
+
+func (numericValidationRegistry) GetSchemaByID(int16) (string, forma.JSONSchema, error) {
+	return "test", forma.JSONSchema{ID: 100, Name: "test", Schema: numericSchemaJSON}, nil
+}
+
+func newNumericValidatingManager(t *testing.T, strict bool) (forma.EntityManager, *mockPersistentRecordRepository) {
+	t.Helper()
+	registry := numericValidationRegistry{}
+
+	validator, err := schemavalidate.New(registry, t.TempDir())
+	require.NoError(t, err)
+
+	config := createTestConfig()
+	config.Entity.ValidateUpdatesStrict = strict
+
+	transformer := transform.NewPersistentRecordTransformer(registry)
+	repo := newMockPersistentRecordRepository()
+	manager := mustNewEntityManager(t, transformer, repo, nil, registry, config, validator)
+	return manager, repo
+}
+
+// TestCreateClassifiesEmbedderNaNAsInvalidInput is issue #322's headline: a Go
+// embedder handing math.NaN() to Create gets invalid input (4xx), not an
+// operator fault (5xx). The validator's marshal step fires first on create, so
+// this exercises the schemavalidate reclassification end to end.
+func TestCreateClassifiesEmbedderNaNAsInvalidInput(t *testing.T) {
+	manager, _ := newNumericValidatingManager(t, false)
+
+	_, err := manager.Create(context.Background(),
+		createOp(map[string]any{"name": "x", "score": math.NaN()}))
+
+	require.ErrorIs(t, err, forma.ErrInvalidInput)
+}
+
+// TestReportOnlyUpdateStillRejectsNonFinite pins the interaction that makes
+// #322's reclassification safe. Report-only mode now absorbs the validator's
+// marshal carrier like any violation, so the transform guard is all that
+// stands between an absorbed NaN and a stored row that 500s on every
+// subsequent read (the response json.Marshal cannot represent it). The string
+// spelling is the HTTP-reachable shape: strconv.ParseFloat accepts "NaN", so
+// before the guard a report-only update could poison a row over HTTP.
+func TestReportOnlyUpdateStillRejectsNonFinite(t *testing.T) {
+	manager, repo := newNumericValidatingManager(t, false)
+	created, err := manager.Create(context.Background(),
+		createOp(map[string]any{"name": "x", "score": 1.5}))
+	require.NoError(t, err)
+
+	for name, value := range map[string]any{
+		"embedder float64": math.NaN(),
+		"HTTP string":      "NaN",
+	} {
+		t.Run(name, func(t *testing.T) {
+			_, err := manager.Update(context.Background(),
+				updateOp(created.RowID, map[string]any{"score": value}))
+
+			require.ErrorIs(t, err, forma.ErrInvalidInput)
+			require.Contains(t, err.Error(), "score",
+				"the transform rejection names the attribute; a marshal error cannot")
+		})
+	}
+
+	stored, err := transform.NewPersistentRecordTransformer(numericValidationRegistry{}).
+		FromPersistentRecord(context.Background(), repo.records[100][created.RowID])
+	require.NoError(t, err)
+	require.EqualValues(t, 1.5, stored["score"], "the finite value must have survived both rejected updates")
+}
--- a/internal/entity_write_nonfinite_test.go
+++ b/internal/entity_write_nonfinite_test.go
@@ -20,8 +20,9 @@ type numericValidationRegistry struct{}
 const numericSchemaJSON = `{
   "type": "object",
   "properties": {
-    "name":  {"type": "string"},
-    "score": {"type": "number"}
+    "name":   {"type": "string"},
+    "score":  {"type": "number"},
+    "active": {"type": "boolean"}
   },
   "required": ["name"]
 }`
@@ -31,8 +32,9 @@ func (numericValidationRegistry) GetSchemaAttributeCacheByName(name string) (int
 		return 0, nil, fmt.Errorf("schema %s not found", name)
 	}
 	return 100, forma.SchemaAttributeCache{
-		"name":  {AttributeID: 1, ValueType: forma.ValueTypeText},
-		"score": {AttributeID: 2, ValueType: forma.ValueTypeNumeric},
+		"name":   {AttributeID: 1, ValueType: forma.ValueTypeText},
+		"score":  {AttributeID: 2, ValueType: forma.ValueTypeNumeric},
+		"active": {AttributeID: 3, ValueType: forma.ValueTypeBool},
 	}, nil
 }
 
@@ -114,4 +116,26 @@ func TestReportOnlyUpdateStillRejectsNonFinite(t *testing.T) {
 		FromPersistentRecord(context.Background(), repo.records[100][created.RowID])
 	require.NoError(t, err)
 	require.EqualValues(t, 1.5, stored["score"], "the finite value must have survived both rejected updates")
+}
+
+// TestReportOnlyUpdateRejectsNonFiniteBool pins the P1 from PR #403's review:
+// a bool attribute's converter must not silently coerce an absorbed non-finite
+// (pre-#322 this write failed as a plain error; the reclassification made it
+// absorbable, so the transform guard has to cover bool funnels too).
+func TestReportOnlyUpdateRejectsNonFiniteBool(t *testing.T) {
+	manager, repo := newNumericValidatingManager(t, false)
+	created, err := manager.Create(context.Background(),
+		createOp(map[string]any{"name": "x", "active": true}))
+	require.NoError(t, err)
+
+	_, err = manager.Update(context.Background(),
+		updateOp(created.RowID, map[string]any{"active": math.NaN()}))
+
+	require.ErrorIs(t, err, forma.ErrInvalidInput)
+	require.Contains(t, err.Error(), "active")
+
+	stored, err := transform.NewPersistentRecordTransformer(numericValidationRegistry{}).
+		FromPersistentRecord(context.Background(), repo.records[100][created.RowID])
+	require.NoError(t, err)
+	require.Equal(t, true, stored["active"], "the stored bool must be untouched")
 }
--- a/internal/entity_write_nonfinite_test.go
+++ b/internal/entity_write_nonfinite_test.go
@@ -78,6 +78,9 @@ func TestCreateClassifiesEmbedderNaNAsInvalidInput(t *testing.T) {
 		createOp(map[string]any{"name": "x", "score": math.NaN()}))
 
 	require.ErrorIs(t, err, forma.ErrInvalidInput)
+	// Pins the mechanism, not just the class: the 4xx must come from the
+	// validator's marshal step, which is the reclassification #322 made.
+	require.ErrorContains(t, err, "cannot be encoded as JSON")
 }
 
 // TestReportOnlyUpdateStillRejectsNonFinite pins the interaction that makes
--- a/internal/entity_write_relation_diagnosis_test.go
+++ b/internal/entity_write_relation_diagnosis_test.go
@@ -171,7 +171,7 @@ func TestWriteDiagnosisIsSilentWithoutRelationRoots(t *testing.T) {
 // gate, and it is not a special case in the code: forma.WithOperatorDetail
 // returns its input unchanged when that input publishes nothing
 // (client_error.go), and schemavalidate.Validate returns a plain error rather
-// than an ErrInvalidInput carrier for everything that is not a caller violation.
+// than an ErrInvalidInput carrier for everything that is not caller input.
 //
 // So a missing resolved schema — an operator fault that must surface as a 500,
 // not a 4xx — cannot pick up a diagnosis about relation roots on its way out.
--- a/internal/entity_write_relation_diagnosis_test.go
+++ b/internal/entity_write_relation_diagnosis_test.go
@@ -171,7 +171,7 @@ func TestWriteDiagnosisIsSilentWithoutRelationRoots(t *testing.T) {
 // gate, and it is not a special case in the code: forma.WithOperatorDetail
 // returns its input unchanged when that input publishes nothing
 // (client_error.go), and schemavalidate.Validate returns a plain error rather
-// than an ErrInvalidInput carrier for everything that is not caller input.
+// than an ErrInvalidInput carrier for every error outside the sentinel class.
 //
 // So a missing resolved schema — an operator fault that must surface as a 500,
 // not a 4xx — cannot pick up a diagnosis about relation roots on its way out.
--- a/internal/entity_write_validation.go
+++ b/internal/entity_write_validation.go
@@ -72,12 +72,13 @@ type writeValidation struct {
 // Anything that is not caller input is returned regardless of enforce. Validate
 // distinguishes the two by wrapping forma.ErrInvalidInput for caller input —
 // a genuine violation, or a payload json.Marshal refuses (NaN/Inf, #322) — and
-// returning a plain error otherwise, i.e. a missing resolved schema. The plain
-// case must not be absorbed by report-only mode: the document would be written
-// with *zero* validation while a log line claimed it had been checked and
-// merely failed (docs/error-handling.md). The absorbed marshal case cannot
-// write a non-finite row: transform's finiteForEAV independently rejects
-// NaN/Inf with the attribute name before anything is staged for storage.
+// returning a plain error otherwise: a missing resolved schema, or a numeric
+// literal that fits neither int64 nor float64. Those must not be absorbed by
+// report-only mode: the document would be written with *zero* validation while
+// a log line claimed it had been checked and merely failed
+// (docs/error-handling.md). The absorbed marshal case cannot write a
+// non-finite row: transform's finiteForEAV independently rejects NaN/Inf with
+// the attribute name before anything is staged for storage.
 //
 // A nil validator means validation is unconfigured and both steps are skipped.
 // Validate on a nil validator returns an error rather than doing nothing, so
--- a/internal/entity_write_validation.go
+++ b/internal/entity_write_validation.go
@@ -64,19 +64,22 @@ type writeValidation struct {
 // map it built.
 //
 // enforce is true on create and follows Entity.ValidateUpdatesStrict on update.
-// It governs *caller input only*. With enforcement off a violation is logged
-// and the write proceeds: rows written before #314 may already violate their
-// schema, and rejecting on update would leave them un-updatable over an
-// unrelated field.
+// It governs *sentinel carriers only* — the split is on sentinel evidence, not
+// on the enforce flag (docs/error-handling.md). With enforcement off a
+// violation is logged and the write proceeds: rows written before #314 may
+// already violate their schema, and rejecting on update would leave them
+// un-updatable over an unrelated field.
 //
-// Anything that is not caller input is returned regardless of enforce. Validate
-// distinguishes the two by wrapping forma.ErrInvalidInput for caller input —
-// a genuine violation, or a payload json.Marshal refuses (NaN/Inf, #322) — and
-// returning a plain error otherwise: a missing resolved schema, or a numeric
-// literal that fits neither int64 nor float64. Those must not be absorbed by
-// report-only mode: the document would be written with *zero* validation while
-// a log line claimed it had been checked and merely failed
-// (docs/error-handling.md). The absorbed marshal case cannot write a
+// Anything without the sentinel is returned regardless of enforce. Validate
+// decides that by wrapping forma.ErrInvalidInput for the caller input it
+// recognises — a genuine violation, or a payload json.Marshal refuses (NaN/Inf,
+// #322) — and returning a plain error otherwise: a missing resolved schema, or
+// a numeric literal that fits neither int64 nor float64. That last one is
+// caller input too and still answers a 500 for want of the sentinel: a known
+// gap, tracked in #402, not a claim that it belongs on this side. Those must
+// not be absorbed by report-only mode: the document would be written with
+// *zero* validation while a log line claimed it had been checked and merely
+// failed (docs/error-handling.md). The absorbed marshal case cannot write a
 // non-finite row: transform's finiteForEAV independently rejects NaN/Inf with
 // the attribute name before anything is staged for storage.
 //
@@ -195,8 +198,8 @@ func validateWritePayload(v writeValidation) error {
 // One narrowing does come for free, and is not a special case here:
 // forma.WithOperatorDetail returns its input unchanged when that input publishes
 // nothing (client_error.go), and schemavalidate.Validate answers a plain error
-// rather than an ErrInvalidInput carrier for everything that is not caller
-// input — a missing resolved schema — and it passes through untouched,
+// rather than an ErrInvalidInput carrier for every error outside the sentinel
+// class — a missing resolved schema — and it passes through untouched,
 // which is the surviving require.Same in
 // TestExplainStrippedRelationRootsNamesEveryRootOnce.
 func explainStrippedRelationRoots(err error, schemaName string, relationRoots []string) error {
--- a/internal/entity_write_validation.go
+++ b/internal/entity_write_validation.go
@@ -64,18 +64,20 @@ type writeValidation struct {
 // map it built.
 //
 // enforce is true on create and follows Entity.ValidateUpdatesStrict on update.
-// It governs *violations only*. With enforcement off a violation is logged and
-// the write proceeds: rows written before #314 may already violate their schema,
-// and rejecting on update would leave them un-updatable over an unrelated field.
+// It governs *caller input only*. With enforcement off a violation is logged
+// and the write proceeds: rows written before #314 may already violate their
+// schema, and rejecting on update would leave them un-updatable over an
+// unrelated field.
 //
-// Anything that is not a violation is returned regardless of enforce. Validate
-// distinguishes the two by wrapping forma.ErrInvalidInput for a genuine
-// violation (→4xx) and returning a plain error otherwise — a missing resolved
-// schema, or a payload that will not marshal (NaN/Inf). Those must not be
-// absorbed by report-only mode: the document would be written with *zero*
-// validation while a log line claimed it had been checked and merely failed, and
-// the message would blame a caller violation for what is an operator fault
-// (docs/error-handling.md).
+// Anything that is not caller input is returned regardless of enforce. Validate
+// distinguishes the two by wrapping forma.ErrInvalidInput for caller input —
+// a genuine violation, or a payload json.Marshal refuses (NaN/Inf, #322) — and
+// returning a plain error otherwise, i.e. a missing resolved schema. The plain
+// case must not be absorbed by report-only mode: the document would be written
+// with *zero* validation while a log line claimed it had been checked and
+// merely failed (docs/error-handling.md). The absorbed marshal case cannot
+// write a non-finite row: transform's finiteForEAV independently rejects
+// NaN/Inf with the attribute name before anything is staged for storage.
 //
 // A nil validator means validation is unconfigured and both steps are skipped.
 // Validate on a nil validator returns an error rather than doing nothing, so
@@ -192,10 +194,9 @@ func validateWritePayload(v writeValidation) error {
 // One narrowing does come for free, and is not a special case here:
 // forma.WithOperatorDetail returns its input unchanged when that input publishes
 // nothing (client_error.go), and schemavalidate.Validate answers a plain error
-// rather than an ErrInvalidInput carrier for everything that is not a caller
-// violation — a missing resolved schema, a payload that will not marshal. Those
-// are operator faults bound for a 500 and they pass through untouched, which is
-// the surviving require.Same in
+// rather than an ErrInvalidInput carrier for everything that is not caller
+// input — a missing resolved schema — and it passes through untouched,
+// which is the surviving require.Same in
 // TestExplainStrippedRelationRootsNamesEveryRootOnce.
 func explainStrippedRelationRoots(err error, schemaName string, relationRoots []string) error {
 	return forma.WithOperatorDetail(err, fmt.Errorf(
--- a/internal/schemavalidate/nonfinite_paths.go
+++ b/internal/schemavalidate/nonfinite_paths.go
@@ -1,0 +1,103 @@
+package schemavalidate
+
+import (
+	"fmt"
+	"math"
+	"slices"
+	"strings"
+
+	"github.com/lychee-technology/forma"
+)
+
+// nonFinitePath is one non-finite float found in a payload, with the dotted
+// path it sits at.
+type nonFinitePath struct {
+	path  string
+	value float64
+}
+
+// marshalRefusalError classifies a json.Marshal refusal of the caller's own
+// payload (#322): caller input, so it carries forma.ErrInvalidInput and
+// publishes.
+//
+// encoding/json's own text names the offending value ("json: unsupported
+// value: NaN") but never where it sits, and an error the caller cannot locate
+// in their payload is barely actionable. So the payload is walked here to
+// derive the path. Only the first offender is named — alphabetically first, so
+// the same payload always produces the same message — with a count of the
+// rest, matching the precedent in transform's missing-required error.
+//
+// If the walk finds nothing the refusal came from something else (a cycle, a
+// failing json.Marshaler, an unsupported type such as a channel or func), and
+// the library's text is published unchanged: it is the only description of the
+// fault that exists.
+func marshalRefusalError(doc any, marshalErr error) error {
+	found := nonFinitePaths(doc)
+	if len(found) == 0 {
+		return forma.InvalidInputf("payload cannot be encoded as JSON: %v", marshalErr)
+	}
+	more := ""
+	if len(found) > 1 {
+		more = fmt.Sprintf(" (and %d more)", len(found)-1)
+	}
+	return forma.InvalidInputf(
+		"payload cannot be encoded as JSON: attribute %q holds non-finite number %v; a finite number is required%s",
+		found[0].path, found[0].value, more)
+}
+
+// nonFinitePaths walks a document and returns every non-finite float in it,
+// sorted by path. It exists for one error path — json.Marshal has just refused
+// the payload — and is never on a successful write.
+//
+// Only the shapes a payload is made of are walked: map[string]any, []any,
+// float64 and float32. Anything else either cannot hold the non-finite that
+// made Marshal fail, or is exotic enough (a custom Marshaler, a struct) that
+// the caller is better served by the library's own text, which the empty
+// result falls back to. A non-finite at the document root is likewise left to
+// the library: there is no attribute to name.
+func nonFinitePaths(doc any) []nonFinitePath {
+	var found []nonFinitePath
+	collectNonFinite("", doc, &found)
+	slices.SortFunc(found, func(a, b nonFinitePath) int {
+		return strings.Compare(a.path, b.path)
+	})
+	return found
+}
+
+func collectNonFinite(path string, node any, found *[]nonFinitePath) {
+	switch v := node.(type) {
+	case map[string]any:
+		keys := make([]string, 0, len(v))
+		for key := range v {
+			keys = append(keys, key)
+		}
+		slices.Sort(keys)
+		for _, key := range keys {
+			collectNonFinite(joinAttributePath(path, key), v[key], found)
+		}
+	case []any:
+		for i, elem := range v {
+			collectNonFinite(fmt.Sprintf("%s[%d]", path, i), elem, found)
+		}
+	case float64:
+		appendIfNonFinite(path, v, found)
+	case float32:
+		appendIfNonFinite(path, float64(v), found)
+	}
+}
+
+func appendIfNonFinite(path string, value float64, found *[]nonFinitePath) {
+	if path == "" {
+		return
+	}
+	if math.IsNaN(value) || math.IsInf(value, 0) {
+		*found = append(*found, nonFinitePath{path: path, value: value})
+	}
+}
+
+func joinAttributePath(prefix, key string) string {
+	if prefix == "" {
+		return key
+	}
+	return prefix + "." + key
+}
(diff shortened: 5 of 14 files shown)


iceboundrock:
**Review artifacts for this branch** (posted per repo PR rules; produced during subagent-driven execution — per-task reviews ran after each commit and are summarized in the issue's execution record; below are the two branch-level artifacts in full).

---

# Final whole-branch review (base 5fedafd, head 9c14bed — before the fix wave)

### Strengths

- **Commit ordering is load-bearing and correct.** 406a48f (transform guard) lands before 903ddc9 (validator reclassification), so no intermediate commit lets report-only mode absorb a marshal carrier without the storage-side guard in place.
- **Funnel coverage is complete.** Independent sweep of every `ValueNumeric = &` producer: the only write-path funnels fed by caller values are `populateTypedValue` (including list-element recursion) and `ToEAVRecord`'s numeric case — both guarded. The `persistent_record.go` sites are read-direction or system-column conversions; the hot-field write copies `*attr.ValueNumeric`, i.e. downstream of the guard. The main-table path is covered transitively.
- **Carrier hygiene is right.** `forma.InvalidInputf(...: %v, err)` — flattened, not a bare `%w` of the sentinel; published text carries encoding/json's value name and no schema id. `numutil.Float64` untouched — the shared-helper trap (#301) explicitly avoided and documented at `finiteForEAV`.
- **Tests pin cause and storage, not just class.** 4 value types × 6 spellings incl. `json.Number` and the HTTP-reachable string forms; stored 1.5 survival asserted against the repo end-state. Mutation evidence recorded for both guards.
- **The semantic sweep worked.** Commit 9c14bed caught two prose sites beyond the plan's enumeration; grep at HEAD finds no residual site claiming marshal failures stay plain.
- **Read paths untouched.**

### Issues

**Critical 1 — `internal/schemavalidate/validator.go:249-251`:** the new doc-comment sentence "…that is an internal fault, not something the caller handed in" is affirmatively false: an HTTP body `{"score": 1e400}` is decoded with `UseNumber`, the `json.Number` survives `json.Marshal` verbatim, and `exactNumberInstance` fails both `Int64` and `Float64` → plain error → 500 **for caller input**. This branch exists to make the classification story true; merging with a known-false classification claim in the very function it fixes recreates the #318 failure mode.

**Important 2 — the "caller input" axis, three sites** (`entity_write_validation.go:67`, `:72-73`, `entity_write_relation_diagnosis_test.go:174`): enumerations behaviorally accurate, but the axis sentence files the 1e400 case (caller input) on the "not caller input" side. Restate on **sentinel evidence**, as `docs/error-handling.md:130` already does.

**Important 3 — `docs/error-handling.md:808`:** "Its *remaining* errors **deliberately** stay plain" claims a decision where one of the two enumerated items (the numeric literal) is an acknowledged accident.

**Minor 4:** the create pin was class-only (`ErrorIs`); one `ErrorContains("cannot be encoded as JSON")` pins the mechanism. **Minor 5:** the absorption-safety claim is narrower than the class it covers (embedder-side `MarshalerError` etc. absorb with zero schema validation; exotic, embedder-only — folded into #402's scope note).

**Deferred-minor triage:** all per-task ledger minors ruled deferrable (BigInt case redundancy, guard chaining shape, test-idiom mix, file-length headroom, NotContains schema-id guard, relation-root note on the NaN carrier per the #318 wide-gate ruling, dead `strict` param, map-range order). Fix-before-merge set = Critical 1 + Important 2/3.

**Verdict: With fixes** — nothing behavioral blocks merge; one docs-only commit (paired with filing the 1e400 follow-up) closes all of it.

---

# Scoped re-review of the fix wave (9c14bed → f21a705)

- **C1 — ADDRESSED.** `validator.go:249-257` keeps the true `dec.Decode` claim and splits out `exactNumberInstance` as caller input misfiled as 500, citing #402. Every clause verified against source; the 1e400 chain re-run empirically (UseNumber decode → marshal re-emits verbatim → `Int64` invalid syntax, `Float64` out of range).
- **I2 — ADDRESSED (all four sites** — the implementer additionally fixed `explainStrippedRelationRoots:201`, which carried identical phrasing): the split is now stated on sentinel evidence everywhere; "for the caller input it **recognises**" makes room for the gap, named and cited to #402. Residual sweep for the retired frame: no production site remains on it.
- **I3 — ADDRESSED.** "currently stay plain… Only the first of those is a deliberate 500; the numeric literal is caller input misfiled on the operator side, a known gap tracked in #402."
- **Fold-in A — ADDRESSED.** `ErrorContains` added and `-overlay`-mutation-verified (rewriting the marshal message turns exactly the new line red).
- **Fold-in B — ADDRESSED.** The class-list bullet defers to the authoritative "Public HTTP error surface" enumeration; one backwards clause ("would blame the caller for a server fault") correctly re-scoped to the missing-schema case.
- **New breakage: none** (prose-only verified mechanically for 571254b). Non-blocking residuals parked with rulings in the execution record: an appositive single-example at `entity_write_validation.go:202`; enumerations omit the internal `dec.Decode` case (named explicitly at `validator.go:249-251`, so nowhere false); pre-existing `entity_write_validation_test.go:290` "forgives *violations*" now under-inclusive (pre-#322 file; one-word candidate for #402's pass).

**Verdict: all findings addressed, no new Critical/Important breakage.**

iceboundrock:
## Findings

**P1: non-finite floats still bypass the write guard for boolean attributes.**

`internal/transform/typed_value.go:58-64` calls `toBool` without `finiteForEAV`, and `internal/transform/attribute_converter.go:107-113` does the same through `toBoolForEAV`. A Go embedder can therefore send `math.NaN()` or `math.Inf()` for a bool attribute: the validator returns the new `ErrInvalidInput` carrier, report-only mode absorbs it, and the transform converts the value to a boolean (`NaN != 0` in `toBool`; `float64ToBool` maps the result to 0/1) instead of rejecting it. This silently changes and persists caller input.

This conflicts with issue #322 and the plan requirement that a non-finite write value “can never be absorbed into storage by report-only mode.” Add the finite check before boolean conversion at both write funnels and add a manager-level report-only regression test.

iceboundrock:
## Spec

The core #322 requirements are implemented for numeric, bigint, integer, and smallint attributes: the validator classifies marshal refusal as `ErrInvalidInput`, the transform layer rejects coerced non-finite numeric values, report-only safety is pinned, and the documentation is synchronized.

There is one partial requirement: the plan states the write path “can never be absorbed into storage by report-only mode,” and the issue goal says a non-finite float in a write payload is rejected. The bool funnels at `internal/transform/typed_value.go:58-64` and `internal/transform/attribute_converter.go:107-113` violate that broader statement as described in the Findings comment. No unrelated scope creep was found.

iceboundrock:
## Standards

Two error-message standards findings remain:

- `internal/transform/attribute_converter.go:77-79` returns `convert to numeric: non-finite number ...` without the offending attribute or `attrID`. The new error should preserve conversion context and identify the logical field, consistent with `AGENTS.md` error semantics and `coding-standard.md` error-context rule.
- `internal/schemavalidate/validator.go:287` publishes `payload cannot be encoded as JSON: ...`, which identifies the non-finite value but not the logical value type, offending attribute/column, or expected state. This does not meet `AGENTS.md` error-message requirements. If `encoding/json` cannot provide a path at this stage, either derive the path or explicitly route the missing path to operator detail while keeping the public message actionable.

File/function sizes, nesting, and the no-bare-error-wrapping rule otherwise pass.

iceboundrock:
## Assessment

**Changes requested.** The main numeric-path fix is sound, but the bool write-path hole is a behavioral gap in the stated safety invariant and can silently persist non-finite embedder input under report-only updates. The two error-context issues should also be corrected before merge, or explicitly justified against the repository error-semantics standard.

iceboundrock:
## Observations

- The known `{"score": 1e400}` path remains caller input that reaches `exactNumberInstance`, returns a plain error, and can surface as a 500. The PR documents this accurately and tracks it in #402; I am not counting it as a separate #322 scope blocker.
- The PR correctly leaves `numutil.Float64` unchanged, avoiding a read/query-path behavior change.
- The existing CI checks reported success, including lint, unit, E2E smoke, k6 smoke, and benchmark smoke.

iceboundrock:
## Verification (corrected) | Base: `5fedafd1d9cd2ea9ab50291d417ade3a32e7b433`; head: `f21a70585746997588b2ad9209ff11472d5fbd5`. | Passed: targeted non-finite/validation tests; `go test ./... -count=1`; `git diff --check`; GitHub lint, test, E2E smoke, k6 smoke, benchmark smoke. | The suite does not cover non-finite values through bool funnels; that gap is the P1 finding.

iceboundrock:
## Summary

The PR correctly fixes the reported NaN/Inf classification and numeric EAV poisoning path, with strong tests and documentation. Review found **1 behavioral P1** (bool conversion bypasses non-finite rejection) and **2 P2 standards issues** (missing attribute/type context in new error messages). Spec coverage is otherwise complete, and all executed tests plus CI checks pass. Overall review status: **changes requested**.

iceboundrock:
## Review Status | The requested formal review state could not be submitted because the authenticated reviewer is the PR author. The comments above contain the complete findings and explicitly recommend changes before merge.

iceboundrock:
Response to the review findings — all three resolved, in `f17a54f`..`ad74370` (4 commits).

## P1 — bool funnels bypass the non-finite guard: confirmed and fixed (`f17a54f`)

Verification turned up one detail beyond the finding: the two funnels didn't just coerce — they **disagreed**. `toBool`'s float case is `v != 0`, so NaN persisted as **true**; `toBoolForEAV` routes through the `float64ToBool` threshold, so NaN persisted as **false**. Silent, and inconsistent between entry points.

Fix as requested: finite check before boolean conversion at both funnels — `toBool`'s `float64`/`json.Number` branches and `toBoolForEAV`'s `float32`/`*float32`/`float64`/`*float64`/`json.Number` branches (the int/string/bool branches cannot carry a non-finite; `toBool`'s string case already rejects via `ParseBool`). `populateTypedValue` wraps the rejection in the attribute-naming `ErrInvalidInput` carrier; `ToEAVRecord` stays plain per that converter's class. The manager-level report-only regression test is in (`TestReportOnlyUpdateRejectsNonFiniteBool`: create with `active: true`, report-only update with `math.NaN()` → 4xx naming `active`, stored bool asserted untouched). Red-first, plus `-overlay` mutation: disabling the guard reddens exactly the three new tests.

Two scope notes from the same sweep:

- **The shared funnel is also the read path** (`extractValueFromEAVRecord` → `toBoolForEAV(record.ValueNumeric)`), so a stored bool `value_numeric` holding NaN now reads as a **plain** operator-visible error instead of silently `false`. We traced every writer of bool `ValueNumeric` (both funnels, hot-field encodings): all produce 0/1 only — and did before this branch too — so the guard can only fire on genuinely corrupt storage (direct SQL, or a numeric→bool type change over a pre-#322 NaN row, which already 500s on numeric reads). Loud-on-corrupt matches `docs/error-handling.md`'s read-path class; blessed as intended behavior. The chain names the attrID but not the row — filed **#405** for the missing row context (pre-existing ceiling on that path, not introduced here).
- **Text attributes deliberately keep accepting non-finite floats** (`toString`'s `%v` fallback stores the spelling `"NaN"`). That is the universal any-value→text coercion (1.5 stores `"1.5"` by the same rule), the stored value is a faithful, finite string, and there is no storage/read-path integrity risk — unlike bool, where the coerced value was semantically arbitrary. On create/strict paths the validator's marshal guard still 400s the whole payload regardless of attribute type. If you want text attrs to reject non-finite floats for invariant uniformity, say so and it's a two-line extension of the same guard.

## P2a — `convert to numeric:` names no field: fixed (`cbd5f47`)

Now `convert to numeric for attrID %d:` (and the new bool case matches: `convert to bool for attrID %d:`). AttrID rather than name, deliberately: `ToEAVRecord` holds no attribute cache, attrID is the EAV key, and it matches the repo precedent (`missing required attribute '%s' (attrID=%d)`). The name-bearing rejection remains `populateTypedValue`'s carrier, which runs first on the write path — a follow-on doc commit (`ad74370`) scopes that claim precisely after review caught it overstated.

## P2b — published message names no attribute: fixed via path derivation (`e9f1377`)

`encoding/json`'s error carries no path, so after a marshal refusal the validator now walks the normalized payload (`nonFinitePaths`, new file — runs only on the already-failed error path) and publishes:

> `payload cannot be encoded as JSON: attribute "profile.score" holds non-finite number +Inf; a finite number is required (and 1 more)`

Offending attribute (dotted spelling — the caller's own key shape either way they sent it), logical value type, expected state. Multiple offenders name the sorted-first plus a count; a marshal failure with no non-finite in the payload (chan/func/cycle/MarshalerError — embedder-only exotica) falls back to the previous library-text form byte-for-byte, so nothing regresses to a redacted body. Tested at the walker level (flat/nested/array/float32/none) and at `Validate` level via `ResolvePublicMessage`; the existing pins (`NaN`/`+Inf`/`-Inf` in the published text, `cannot be encoded as JSON` at the manager seam) still pass non-vacuously.

## Follow-ups filed from this wave

- **#404** — the two bool funnels embody different truth tables for values they both still accept (finite floats: `!= 0` vs threshold; negatives; unparseable strings: reject vs silent-false). Latent divergence of exactly the shape that hid this P1; needs a single ruled truth table.
- **#405** — read-path consistency errors name attrID but never the row (includes `FromEAVRecord`'s pre-existing bare `return err`, the hop that could carry it).

Gates re-run at `ad74370`: fmt-check, golangci-lint v1.64.8, full unit suite — clean.

iceboundrock:
## Summary

Review status: **changes requested**.

- **1 P1:** `nonFinitePaths` has no cycle detection and can recurse until stack exhaustion on a cyclic caller payload.
- **1 P2:** path derivation does not cover `json.Number("NaN")` or non-finite float pointers, so the published error can omit the offending attribute.
- **Spec:** core #322 behavior is implemented; diagnostic-path coverage is partial.
- **Standards:** no additional hard violations found.
- **Verification:** local gates and all reported GitHub checks pass.

The authenticated reviewer is also the PR author, so GitHub does not allow a formal approval/request-changes review submission. These comments are the complete review record.

iceboundrock:
## Observations

- The guard coverage now includes both numeric and bool `ValueNumeric` funnels, and the manager-level report-only tests verify that the existing finite value survives rejected updates.
- Leaving `numutil.Float64` unchanged is the correct boundary because that helper is also used by query/read-related code.
- The documentation accurately records the known `{"score": 1e400}` gap in #402 rather than silently claiming every caller-input failure is classified consistently.
- The new path walker is on the marshal-error path only, so the normal successful-write path does not pay its traversal cost.

iceboundrock:
## Verification

Review base: `5fedafd1d9cd2ea9ab50291d417ade3a32e7b433`
Review head: `ad743702960af303d3a490b641fe1d0d2bd07290`

Local verification on the PR head passed:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/schemavalidate ./internal/transform ./internal -run 'NonFinite|EmbedderNaN|NamesNonFinite|FallsBackToLibraryText' -count=1 -v`
- `make fmt-check`
- `make lint`
- `make test`
- `git diff --check`

GitHub checks are also green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke. Existing tests do not cover cyclic payloads or `json.Number`/pointer non-finite path reporting, which is why the Findings remain open.

iceboundrock:
## Spec

Compared with issue #322 and `docs/superpowers/plans/2026-08-12-issue-322-nonfinite-invalid-input.md`.

The core requirement is implemented: marshal refusals carry `forma.ErrInvalidInput`; numeric and bool write funnels reject non-finite values before `ValueNumeric` is staged; report-only update safety is pinned; and the numeric/bool tests preserve the prior stored value.

Coverage is partial in the diagnostic path. The plan goal says non-finite write input is rejected with a message naming the offending attribute, while `nonFinitePaths` does not identify `json.Number("NaN")` or non-finite float pointers. More importantly, the intended marshal-failure fallback is not safe for cyclic maps or slices: the walker can recurse indefinitely instead of returning the `InvalidInput` error. No unrelated scope creep was found; the bool and AttrID follow-ups are directly required to preserve the report-only safety invariant and error context.

iceboundrock:
## Standards

No additional hard standards violations were found after reviewing `AGENTS.md`, `coding-standard.md`, and the complete diff.

- Changed source files remain within the 500-line limit and changed functions remain within the 100-line limit.
- The new AttrID context in `ToEAVRecord` and attribute-path context in the validator align with the repository error-message rules.
- The PR keeps `numutil.Float64` unchanged, avoiding an unintended read/query-path behavior change.
- The P1 in the Findings comment is a robustness violation in the changed error path: the code comment promises a safe fallback for cycles, but the implementation has no cycle protection.

iceboundrock:
## Assessment

**Changes requested.** The main classification and storage-poisoning fix is sound, including the later bool-funnel correction. However, a caller-controlled cyclic payload can cause unbounded recursion in the newly added error diagnostic path, and supported non-finite forms can still lose the offending attribute path. Fix the P1 before merge and add regression coverage; the P2 should be fixed as part of completing the published error contract.

iceboundrock:
## Findings

**P1: the diagnostic walker can recurse until stack exhaustion on cyclic payloads.**

`internal/schemavalidate/nonfinite_paths.go:60,67-86` recursively traverses maps and slices without tracking visited containers. `json.Marshal` rejects a self-referential map or slice with a cycle error, but `marshalRefusalError` then calls this walker before returning the documented fallback at lines 30-33. A Go embedder can therefore turn the intended `forma.ErrInvalidInput` response into process instability instead of receiving the marshal error. Add cycle detection, or make the diagnostic walk abort and return the original marshal text when a cycle is encountered.

**P2: attribute-path diagnostics miss valid non-finite payload shapes.**

`nonFinitePaths` handles only `map[string]any`, `[]any`, `float64`, and `float32` (`internal/schemavalidate/nonfinite_paths.go:48-57,67-86`). A `json.Number("NaN")` or a pointer such as `*float64(math.NaN())` can also make `json.Marshal` refuse the payload, but the walker finds nothing and publishes only the generic marshal error without the offending attribute. This leaves the PR short of the plan goal to reject non-finite input with a message naming the offending attribute. Either support the representable pointer/`json.Number` shapes or explicitly document and test the fallback boundary.

iceboundrock:
Response to the round-2 findings — both resolved, in `077e252`..`d81e3ac` (3 commits).

## P1 — walker recursion on cyclic payloads: confirmed and fixed (`077e252`, hardened in `d81e3ac`)

Fixed with the depth cap you offered as an alternative: `maxNonFiniteWalkDepth = 1000`; exceeding it **abandons the entire walk** (returns nil, never a truncated list that could name a wrong offender) and publishes the library text — which for a cycle ("encountered a cycle via …") is the only truthful description. The cap covers both cycles and pathologically deep acyclic payloads. RED evidence was grim enough to make the point: pre-fix, the cyclic test had to be killed after 36s of runaway stack growth. The abandon-over-truncate invariant is mutation-pinned (`d81e3ac`): a payload with a shallow NaN *plus* a >1000-deep subtree must return nil, and deleting the abort-return reddens exactly that test.

The comment falsity you flagged in Standards was real twice over — the first rewrite claimed encoding/json's nesting limit bounds documents "long before" the cap, and the decoder limit is actually **10000**, ten times above it. That premise is now removed rather than softened; what actually bounds exposure is provenance (JSON has no non-finite syntax, so this whole error path is embedder-only).

One reachability note for the record: on the production write path a cyclic embedder payload today crashes **before** the walker — `transform.NormalizeDottedKeys`/`flattenToAttributes` recurse unbounded one layer earlier, so the crash predates this branch. This fix removes the second site; the pre-existing first one is filed as **#406** (with this cap as the precedent for the fix shape).

## P2 — missing shapes: confirmed and fixed (`f1848ec`)

Added exactly the shapes that make `json.Marshal` refuse:

- `*float64` / `*float32` — nil skipped (marshals as `null`), non-nil dereferenced and checked;
- `json.Number` — flagged when `strconv.ParseFloat` succeeds **and** the value is non-finite. That gate is load-bearing, verified by probe: `ParseFloat("1e400")` returns `+Inf` *with* `ErrRange`, yet `1e400` is valid JSON grammar and **marshals fine** — flagging it would fabricate a failure cause. The flagged set was probe-verified (16 literals + 5 container shapes, including lowercase `nan`/`inf`) to be a strict subset of the marshal-refused set, so every skip falls to the library text and every flag is genuine.

Boundary documented and pinned per your "or explicitly document and test" clause: garbage literals (`json.Number("abc")`) fall back to the library's `invalid number literal "abc"` — which already names the value — and structs, custom Marshalers, and typed containers (`[]float64`, `map[string]float64`) likewise fall back rather than lie. Tests cover the named-path cases (`*float64` NaN, `*float32` +Inf, nested `json.Number("NaN")`), both boundary pins (`1e400` not flagged, `abc` → library text via `ResolvePublicMessage`), cyclic map/slice, 1001-deep abort, and a depth-50 control that still names the attribute.

Gates re-run at `d81e3ac`: fmt-check, golangci-lint v1.64.8, full unit suite — clean. One process note from this wave, for the record: an `-overlay` mutation run answered `ok (cached)` once — cached green is no evidence, and all mutation runs here were re-executed with `-count=1`.

iceboundrock:
## Review — PR #403 (`fix/322-nonfinite-invalid-input`, HEAD `d81e3ac`)

Reviewed against `main` (merge-base `5fedafd`). Two axes: **Spec** (does it do what #322 asked?) and **Standards** (does it conform to `coding-standard.md` + `AGENTS.md` error semantics?). Verification at the end.

---

### Spec

The originating issue is **#322** (`NaN/Inf in an embedder payload is classified as an operator fault (5xx) instead of invalid input`). Its "Direction" asks for exactly two things:

1. Classify a marshal failure of the caller's own payload as `forma.ErrInvalidInput` → 4xx, absorbable by report-only mode. **Done** — `internal/schemavalidate/validator.go` now returns `marshalRefusalError(doc, err)` on the `json.Marshal` failure (previously `fmt.Errorf("failed to marshal payload for schema %d")`, a plain error → 5xx).
2. "Worth deciding … whether the transform layer should reject non-finite floats earlier with a message naming the attribute." **Done, and done well** — `finiteForEAV` sits at both write funnels (`populateTypedValue` in `typed_value.go`, and `ToEAVRecord`'s numeric case), and the caller-facing carrier in `populateTypedValue` names the attribute.

**Two scope findings, both justified:**

- **The issue's "not reachable over HTTP" premise is false.** `numutil.Float64` coerces `string` via `strconv.ParseFloat`, which accepts `"NaN"`/`"Inf"`/`"Infinity"`. So an HTTP body `{"score": "NaN"}` *is* reachable, and under the shipped report-only default it was warn-absorbed as a schema violation and then stored a NaN — a poison row that 500s on every subsequent read. I confirmed this trace in `TestReportOnlyUpdateStillRejectsNonFinite/HTTP_string` (the `WARN` line shows the string spelling being absorbed, then rejected by the transform guard). This is a *correct* expansion of the issue, not scope creep, and the PR documents it as such.
- **Bool funnels** (`toBool` / `toBoolForEAV`) were added in the review fix-wave because a bool attribute reaches `ValueNumeric` through the same column, and pre-#322 a NaN bool write failed plain but the reclassification made it absorbable. Correct follow-on; `TestReportOnlyUpdateRejectsNonFiniteBool` pins it.

**Deliberate, correctly-scoped non-goal:** the `1e400` numeric literal (`exactNumberInstance`) still answers a redacted 500 for caller input. Filed as **#402** and named honestly in the code comment and prose rather than silently folded in. Good scope boundary.

---

### Standards

`coding-standard.md` + `AGENTS.md` error semantics. All pass:

- **Line limits.** `attribute_converter.go` is 491 lines after the pure move of `toFloat64ForEAV`/`parseTrimmedFloat64` into `numeric_conversion.go` (86 lines) — the move that motivated the refactor is real (the guard would have pushed it past 500). `nonfinite_paths.go` 185, its test 296. `validator_test.go` sits at exactly 500 (the ceiling), which the PR acknowledges and worked around by landing new validator tests in `nonfinite_paths_test.go`.
- **Error-context wrapping.** Every new site wraps with context (`convert to numeric for attrID %d: %w`, `payload cannot be encoded as JSON: %v`). No bare `return err`.
- **Message style.** New messages name value type + attribute (or AttrID) + expected state: `non-finite number %v is not storable; a finite value is required`, wrapped by `invalid value for attribute 'score' (attrID=2)`. The `ToEAVRecord` plain errors name `attrID` (the EAV key) because that layer holds no attribute name — the doc comment records *why* that's operator-visible by design.
- **Prose sync is exemplary.** `docs/error-handling.md`, `validateWritePayload`, `explainStrippedRelationRoots`, and `validator.go` were all corrected to the *sentinel-evidence* split — including two sites only a semantic sweep found (the PR's commit `9c14bed` explains the `{"score": 1e400}` plain-case enumeration). This matches AGENTS.md's "error messages must name the class" discipline.
- **Commit sequencing is safe.** The transform guard (`406a48f`) lands *before* the reclassification (`903ddc9`), so no intermediate commit makes the marshal carrier absorbable without a storage backstop.

---

### Assessment

This is a high-quality, well-sequenced change. The author did the hard part correctly: the load-bearing insight is that the reclassification is only *safe* because the transform layer independently rejects non-finite values, and the commit order proves that invariant holds at every intermediate state. The defensive depth-cap + abandon (not truncate) semantics in the walk, with the "partial result is discarded, not published" pin, is the kind of edge-case rigor this repo's `docs/error-handling.md` standardizes. No blocking issues.

---

### Observations / findings

1. **(Judgement call — documented, not closed.)** `nonFinitePaths` walks only the JSON-ish shapes (`map[string]any`, `[]any`, `float64/32`, `*float64/32`, `json.Number`). A Go embedder handing a concrete container — `map[string]any{"score": []float64{math.NaN()}}` or `map[string]float64` — gets the library text with **no attribute named**. The doc comment (`nonfinite_paths.go:373-378`) states this as a deliberate limit ("type-switching every concrete numeric container is a losing game"). Reasonable, but it means the issue's "name the offending path" goal is only partially met for embedders using concrete Go types rather than the `any`-map shapes. Consider a one-line note in the PR body that the walk's coverage is shape-bound, since embedders are exactly the audience of this fix.

2. **(Minor wording.)** `finiteBoolInput` reuses `finiteForEAV`'s message verbatim, so a bool rejection publishes `non-finite number NaN is not storable; a finite value is required`. "not storable" is numeric-flavored prose for a boolean. Harmless (the wrapper names the attribute and it reads coherently), but a `non-finite value %v has no truth value; a finite value is required` for the bool path would be more precise.

3. **(Stale PR description.)** The PR body enumerates "7 commits" and stops at the prose sync (`571254b`). The actual branch is **15 commits**, and the last 8 — the review-driven fix-wave (bool funnels `f17a54f`, attrID naming `cbd5f47d`, marshal-message path derivation `e9f13775`, doc scoping `ad74370`, cycle depth-cap `077e252`, pointer/json.Number spellings `f1848ec`, cap-comment truth `d81e3ac`) — are absent from the description. Worth a one-line edit so reviewers aren't surprised by `nonfinite_paths.go` existing.

4. **(Understanding note, not a defect.)** In `TestReportOnlyUpdateStillRejectsNonFinite/HTTP_string`, the string `"NaN"` is caught by *two* layers: the schema validator first reports a type violation (string vs `number`), which report-only mode absorbs; only then does the transform guard reject it. Under `ValidateUpdatesStrict` the schema validator alone would catch it. The transform guard is the load-bearing backstop specifically for report-only mode — the test is correct and pins the right thing, but the two-layer mechanism is worth knowing when reading the log output.

5. **(Confirmation.)** The `json.Number` gate (`ParseFloat` err == nil) correctly distinguishes `"1e400"` (ParseFloat overflows → +Inf/ErrRange, but valid JSON grammar that marshals fine) from `"NaN"`/`"Inf"` (ParseFloat succeeds → non-finite, marshals fail). Both boundaries are pinned (`TestNonFinitePathsIgnoresNonRefusingNumberLiterals`, `TestValidateFallsBackOnInvalidNumberLiteral`).

---

### Verification

Ran at HEAD `d81e3ac` (clean worktree, detached):

- `go build ./...` — clean.
- `gofmt -l` over the changed `.go` files — clean (no output).
- `go vet ./internal/schemavalidate/... ./internal/transform/...` — clean.
- `go test ./internal/schemavalidate/... ./internal/transform/...` — **ok**.
- `go test ./internal` (full package) — **ok**.
- Targeted pins (`TestCreateClassifiesEmbedderNaNAsInvalidInput`, `TestReportOnlyUpdateStillRejectsNonFinite`, `TestReportOnlyUpdateRejectsNonFiniteBool`, `TestWriteDiagnosisIsSilentOnOperatorClassErrors`, `TestExplainStrippedRelationRootsNamesEveryRootOnce`) — **all pass**.

I did not run `make lint` (golangci-lint v1.64.8) or the `internal/e2e_harness` suite — the latter's Docker-dependent host resolution is exactly the environmental failure the PR body already calls out. The PR's mutation-check claims (red-first, `finiteForEAV` removal reddens transform + manager pins, reclassification revert reddens the create pin) are credible but I did not independently reproduce them.

---

### Summary

**Spec: pass** (both direction points met, plus a correctly-justified scope widening). **Standards: pass** (no hard violations; two nitpicks + one stale-description note). Worst single issue: none blocking — the shape-bound attribute-naming walk (observation 1) is the only substantive gap, and it is deliberately documented rather than accidental.

iceboundrock:
Response to the round-3 review (approving, five observations) — one implemented, two documentation actions taken, two acknowledged.

**Obs 2 (bool message wording) — implemented (`1b36ecc`).** Bool rejections now publish `non-finite value <v> has no truth value; a finite value is required` (your suggested wording, adjusted to also read correctly on the read path, where the same converter describes a persisted corrupt value). Structural detail: the two guards now share an extracted `isNonFinite` predicate but deliberately not the prose — they cannot drift on *what* they reject while staying free to differ on how they say it. Both bool pins were re-anchored to the distinct text (`"has no truth value"`), mutation-verified: reverting the delegation reddens all 10 bool subtests and nothing else.

**Obs 1 (shape-bound walk) — documented as requested.** The PR body now carries a "Known boundary (deliberate)" paragraph stating the walk is shape-bound to JSON-decode shapes, and that concrete Go containers (`[]float64{NaN}`, `map[string]float64`) still get the 4xx carrier but with the library text and no attribute named. Agreed it stays a judgment call rather than a gap to close: the flagged-set-⊂-refused-set property is what keeps the walker honest, and type-switching arbitrary numeric containers would grow the surface without changing any classification.

**Obs 3 (stale PR description) — fixed.** The body now describes all 15 commits: the original 7 plus the three review waves, each mapped to the finding it answers.

**Obs 4 (two-layer catch on the string spelling) — acknowledged, and it is the intended mechanism.** Under report-only the schema violation is absorbed first (that WARN line is the operator's repair signal per #314's design) and the transform guard is the storage backstop; under strict the validator alone rejects. The test pins the backstop because that is the layer the safety invariant rests on.

**Obs 5 (`json.Number` gate) — confirmed on both sides,** and the `1e400`-not-flagged half is pinned precisely so a future "obvious fix" can't widen the gate into fabricating causes.

One residual noted for the record: the anti-drift pinning is one-directional (bool prose can't regress into numeric prose; the numeric pins still assert only the shared `"non-finite"` substring). Left as-is — the numeric message is otherwise pinned by content, and the asymmetry has no behavioral face.

Gates re-run at `1b36ecc`: fmt-check, golangci-lint v1.64.8, full unit suite — clean.

iceboundrock:
## Findings

No blocking issues. Four minor/nit findings, ordered by significance.

**Minor 1 — the attribute-naming walk is shape-bound, leaving #322's "name the offending attribute" direction partial for concrete Go containers.** `nonfinite_paths.go` walks only `map[string]any`, `[]any`, `float64/32`, `*float64/32`, `json.Number`. An embedder passing `map[string]float64{"score": NaN}` or `[]float64{NaN}` (or a typed struct) still gets the 4xx carrier, but with the library's text and **no attribute named** — the exact hole the walk exists to close, and the embedder is precisely this fix's audience. Documented as deliberate ("type-switching every concrete numeric container is a losing game", nonfinite_paths.go:373-378) — a defensible call, and the storage-safety invariant still holds there (transform's `finiteForEAV` names the attribute on the write path). Recording, not requesting a change.

**Minor 2 — prose-must-match-code (#318 shape): `isNonFinite`'s doc says "the shared predicate behind both guards below".** One of the two guards, `finiteForEAV`, sits *above* it (`numeric_conversion.go:69` vs `:79`); only `finiteBoolInput` is below. One-word fix.

**Nit 1 — `TestValidateClassifiesUnmarshallablePayloadAsInvalidInput` has a stale inline comment.** It asserts `require.Contains(msg, name)` under the original plan's rationale ("encoding/json names the offending value; publish it"). The assertion now passes only because the synthesized message re-prints the value via the walk, not because the library text is published. The pin is still a fine seam-level check; the comment no longer describes what it pins.

**Nit 2 — vocabulary drift between the two guards.** The marshal path publishes "a finite **number** is required" while the transform guard publishes "a finite **value** is required" (the plan's contract for `finiteForEAV`, plan:46). The bool/marshal split is deliberate ("has no truth value"), but the numeric/marshal split is accidental.

iceboundrock:
## Spec

Compared against issue #322 and the plan `docs/superpowers/plans/2026-08-12-issue-322-nonfinite-invalid-input.md`.

Core direction fully met:

- Marshal refusal of the caller's payload → `forma.ErrInvalidInput` (4xx), absorbable by report-only mode. Done (`validator.go` now calls `marshalRefusalError` on the `json.Marshal` failure).
- Transform rejects non-finite floats after numeric coercion, naming the attribute. Done at both write funnels — `populateTypedValue` (carrier, names the attribute) and `ToEAVRecord` (plain, names attrID) — plus list-element recursion.
- Report-only safety invariant holds for numeric + bool + list; the finite stored value survives rejected updates (pinned at manager level).

One divergence from the plan's stated contract: Task 2 specified `InvalidInputf("payload cannot be encoded as JSON: %v", err)` publishing the library text *verbatim*. The implementation instead synthesizes a message naming the attribute via the depth-capped walk. This is aligned with the plan's Goal line ("message naming the offending attribute") and is an improvement — but it is a contract change the plan did not prescribe, worth a one-line note in the execution record rather than silent divergence.

One plan evidence error, correctly corrected by the PR: plan line 19 asserted bool "cannot store a non-finite (`float64ToBool` yields 0/1)". The bool-funnel sweep proved `toBool`'s `!= 0` coerced NaN→true while `toBoolForEAV`'s threshold coerced the same NaN→false. A real hole the plan missed, closed as a justified extension (not creep).

`finiteForEAV`'s message matches plan:46 exactly. The validator split is correct: marshal reclassified; both decode-side failures and the missing-schema branch stay plain (`validator_test.go` `TestValidateUnknownSchemaIDIsNotClientError` intact).

Residual gap: the shape-bound walk (see Findings Minor 1) means "message naming the offending attribute" is partial for concrete Go containers.

iceboundrock:
## Standards

No hard violations against `coding-standard.md`, `AGENTS.md`, or `docs/error-handling.md`.

- **File/function limits.** All changed files ≤500 lines. `attribute_converter.go` is 491 after the pure move of `toFloat64ForEAV`/`parseTrimmedFloat64` into `numeric_conversion.go` — the move that kept the guard from pushing it over the ceiling. `nonfinite_paths.go` 185, its test 296. No changed function exceeds 100 lines.
- **Error wrapping.** No bare `return err` on new error-producing paths. The two bool float cases return `finiteBoolInput`'s error via `return false, err`, but context is added at the callers (`convert to bool for attrID %d`, `populateTypedValue`'s carrier) — matching the existing `derefPointer` pattern, acceptable.
- **Message quality.** New messages name value + attribute/attrID + expected state. `ToEAVRecord`'s attrID-instead-of-name choice is explicitly justified in its new doc comment (no metadata cache at this layer), and the follow-on `ad74370` scopes that claim to only the two cases that earn it.
- **Carrier hygiene.** `marshalRefusalError` uses `InvalidInputf` (publishes); no bare `%w` sentinel wrap; the class split is stated on sentinel evidence consistently across `entity_write_validation.go` and `docs/error-handling.md`.
- **Doc-comment accuracy (#318 discipline).** Verified: `maxNonFiniteWalkDepth`=1000 vs the decoder's 10000 ("ten times" holds); sort-by-path matches `strings.Compare`; root-skip matches the "left to the library" claim; the `json.Number` `1e400`/`abc` gate matches code. One miss: `isNonFinite`'s "both guards below" (see Findings Minor 2).

**Note:** `validator_test.go` is at exactly **500 lines** — zero headroom; the next test added there violates the standard. The PR already worked around this by landing new validator tests in `nonfinite_paths_test.go`. A follow-up split of `validator_test.go` is worth filing.

iceboundrock:
## Assessment

High-quality, well-sequenced change. The load-bearing insight is that the reclassification is only *safe* because the transform layer independently rejects non-finite values, and the commit order (`406a48f` guard before `903ddc9` reclassification) proves that invariant at every intermediate state — no commit lets report-only mode absorb a marshal carrier without a storage backstop. The depth-capped walk with abandon-over-truncate semantics (a partial result is discarded, never published) is exactly the edge-case rigor `docs/error-handling.md` standardizes, and it is mutation-pinned.

**Recommend: approve.** No blocking issues. The single substantive gap (shape-bound attribute naming, Findings Minor 1) is deliberate and documented. The remaining three findings are follow-up-sized and none require a merge blocker.

iceboundrock:
## Observations

- **The reachability finding is correct and load-bearing.** The PR body corrects #322's "not reachable over HTTP" premise: `numutil.Float64` → `strconv.ParseFloat` accepts `"NaN"`/`"Inf"`/`"Infinity"`, so an HTTP update body `{"score": "NaN"}` under the shipped report-only default was absorbed as a schema violation and then stored a poison row. `TestReportOnlyUpdateStillRejectsNonFinite/HTTP_string` pins exactly this trace. This correctly upgrades the transform guard from "worth deciding" to a required backstop.
- **`numutil.Float64` deliberately untouched.** The #301 trap (hardening a shared helper for one caller's write semantics) is avoided and documented, since that helper also serves the DuckDB read/query path. Correct boundary.
- **The `json.Number` gate is subtle and right.** It flags a literal only when `ParseFloat` returns no error, separating `1e400` (valid JSON grammar, marshals fine, ParseFloat gives +Inf with ErrRange) from `"NaN"`/`"Inf"` (parse clean, marshal fails). Both boundaries pinned — a future "obvious fix" can't widen the gate into fabricating a cause.
- **Read-path consequence, blessed as intended.** A stored bool `value_numeric` holding NaN (corrupt storage; every writer produces 0/1) now reads as a loud plain error instead of silent `false` — matching the read-path class in `docs/error-handling.md`. #405 filed for the missing row context.
- **The two-layer catch on the string spelling is the intended mechanism:** report-only absorbs the schema type violation (its WARN is the operator repair signal, #314), and the transform guard is the storage backstop. The test pins the backstop because that is where the safety invariant rests.
- **PR body is comprehensive** and matches the actual 15 commits, including per-review-wave mapping.

iceboundrock:
## Verification

Review base `5fedafd`, head `1b36ecc`. Ran in a clean worktree at head:

- `go test ./... -count=1` — **all pass**, including `internal/e2e_harness` smoke (Docker available in this environment; the PR body's "daemon down" local failure did not reproduce here).
- Targeted pins (`NonFinite|EmbedderNaN|Unmarshallable|NamesNonFinite|FallsBack|Aborts|CoversPointer`) — pass across `internal/schemavalidate`, `internal/transform`, `internal`.
- `gofmt -l` over changed `.go` files — clean; `git diff --check` — clean.
- `make lint` (golangci-lint v1.64.8) — exit 0.
- Line counts: `validator_test.go` = 500 (ceiling), `attribute_converter.go` = 491, `nonfinite_paths.go` = 185 — within limits.

I did not independently re-run the mutation/overlay checks the PR body claims; they are credible, and the abandon-over-truncate pin specifically was reviewed for shape (a shallow NaN at "a" plus a >1000-deep branch must return nil — a truncating walk would publish `attribute "a"`, reddening the pin).

iceboundrock:
## Summary

Review status: **approve**. (No formal review state — the authenticated reviewer is the PR author; these comments are the complete review record.)

- **Spec:** core #322 direction fully met; one justified scope widening (bool funnels) that corrected a plan evidence error; one documented residual gap (shape-bound attribute naming).
- **Standards:** no hard violations; four minor/nit findings (`isNonFinite` "below" wording, stale test comment, "number" vs "value" wording, `validator_test.go` at the 500-line ceiling).
- **Verification:** full unit suite, gofmt, diff-check, and lint all clean at head.
- **Worst single issue:** the shape-bound walk (Findings Minor 1) — deliberate and documented, not a merge blocker.

iceboundrock:
Response to the round-4 review (approve) — three items fixed in `0fc6e15`, the rest routed to follow-ups and the execution record.

**Minor 2 (`isNonFinite` "both guards below") — fixed.** "below" → "in this file"; `finiteForEAV` sits above the predicate.

**Nit 1 (stale assertion rationale) — fixed.** The comment now states what the pin actually guarantees post-walk: the published message names the offending value on *either* `marshalRefusalError` branch — synthesized when the walk finds the path, library text when it doesn't. Assertion untouched.

**Nit 2 (vocabulary drift) — fixed.** The marshal message's tail is now `a finite value is required`, matching the transform guard's plan-pinned contract (the bool "has no truth value" split stays, per round 3 — that one is deliberate). Three sites updated (one production, two pins), mutation-verified.

**Minor 1 (shape-bound walk) — filed as #407** rather than left only in a doc comment. The issue records the boundary, the audience argument you made (embedders are exactly this fix's targets), the honesty constraint any extension must preserve (flagged set ⊂ marshal-refused set — the `1e400` lesson), and the cheap candidate shapes (`[]float64`/`map[string]float64`) versus the reflection territory where the current "losing game" judgment likely still stands.

**Standards note (`validator_test.go` at the 500 ceiling) — filed as #408**, with the natural three-way split (`TestResolve*`/`TestLoader*`, `TestNew*`, `TestValidate*`) as a pure move. The round-4 fix itself demonstrated the problem live: the first draft of the Nit-1 comment pushed the file to 503, and nothing in the toolchain flags the ceiling — it was caught only by a manual `wc -l` habit. A file at the ceiling turns every comment edit into a size decision, which is exactly why the split is worth doing now.

**Spec note (Task-2 contract divergence) — recorded on #322** as an execution-record addendum ([comment](https://github.com/Lychee-Technology/forma/issues/322#issuecomment-5296152104)): the plan specified publishing the library text verbatim; the shipped code synthesizes the attribute-naming message with verbatim text as the fallback branch. Divergence introduced by this review's own P2b, serving the plan's Goal line — now written where the plan lives instead of left silent.

Gates re-run at `0fc6e15`: fmt-check, golangci-lint v1.64.8, full unit suite — clean.

iceboundrock:
## Findings

No blocking issues. One new P3 observation, and confirmation that the round-4 items are correctly landed.

**P3 (new) — the attribute-naming walk can mis-attribute a mixed-cause refusal.** `nonFinitePaths` names the alphabetically-first *non-finite*, but `json.Marshal` refuses on the *first offending value it reaches* while iterating map keys in sorted order. When a payload holds a non-finite *and* another refusing value whose key sorts earlier — probe: `map[string]any{"ch": make(chan int), "score": math.NaN()}` marshals to `json: unsupported type: chan int` (the channel is reached first) — the walk still reports `attribute "score" holds non-finite number NaN`. The message is truthful (the NaN would independently refuse), but it names a value that was never the proximate cause, and a caller who fixes `score` will resend and hit the channel next. Embedder-only (a channel/func cannot arrive over HTTP) and requires two distinct refusal causes; the `marshalRefusalError` doc's "first offender" prose describes the walk's output, not the marshal's cause. Recording, not requesting a change — if ever tightened, the honest fix is to fall back to the library text when the marshal error type is not one the walk explains.

Round-4 items verified at HEAD `0fc6e15`: (a) the marshal message tail now reads `a finite value is required` (aligned with the transform guard's plan-pinned contract); (b) `isNonFinite`'s comment says "both guards in this file" (not "below"); (c) `TestValidateClassifiesUnmarshallablePayloadAsInvalidInput`'s rationale now states the pin holds on either `marshalRefusalError` branch. All three match their round-4 descriptions.

Prior tracked residuals remain accurate: #407 (shape-bound walk over concrete Go containers) and #408 (`validator_test.go` at the 500-line ceiling).

iceboundrock:
## Spec

Compared against issue #322 and the plan `docs/superpowers/plans/2026-08-12-issue-322-nonfinite-invalid-input.md`.

Both of #322's direction points are met at HEAD:

- Marshal refusal of the caller's payload → `forma.ErrInvalidInput` (4xx), absorbable by report-only mode (`validator.go` → `marshalRefusalError`).
- Transform rejects non-finite floats after numeric coercion at both write funnels, naming the attribute (`populateTypedValue` carrier) or attrID (`ToEAVRecord` plain), plus list-element recursion.

The plan's safety invariant ("can never be absorbed into storage by report-only mode") holds for numeric, bool, and list; the finite stored value surviving rejected updates is pinned at manager level (`TestReportOnlyUpdateStillRejectsNonFinite`, `TestReportOnlyUpdateRejectsNonFiniteBool`).

Two plan divergences, both already on the record:

- The bool funnels were a genuine plan evidence error (plan line 19 asserted bool "cannot store a non-finite"; the sweep proved `toBool`'s `!= 0` coerced NaN→true while `toBoolForEAV`'s threshold coerced it→false). Closed as a justified extension.
- Task 2 specified publishing the library text verbatim; the shipped code synthesizes an attribute-naming message with verbatim text as the fallback. Recorded as an execution-record addendum on #322.

Residual gap (tracked, not a blocker): the shape-bound walk means "name the offending attribute" is partial for concrete Go containers — filed #407.

iceboundrock:
## Standards

No hard violations against `coding-standard.md`, `AGENTS.md`, or `docs/error-handling.md`.

- **File/function limits** (verified at HEAD): `validator_test.go` = 500 (the ceiling, tracked #408), `attribute_converter.go` = 491, `nonfinite_paths.go` = 185, `nonfinite_paths_test.go` = 296, `numeric_conversion.go` = 102. No changed function exceeds 100 lines.
- **Error wrapping**: no bare `return err` on new error-producing paths; `finiteBoolInput`'s error is returned via `return false, err` but context is added at both callers (`convert to bool for attrID %d`, `populateTypedValue`'s carrier). Carrier hygiene correct: `handleConversionError` flattens with `%v`, no bare `%w` of the sentinel.
- **Doc-comment accuracy** (#318 discipline): `maxNonFiniteWalkDepth` = 1000 vs `encoding/json`'s decoder limit of 10000 ("ten times" holds); sort-by-path uses `strings.Compare`; the `json.Number` `1e400`/`abc` gate matches code. The `isNonFinite` "in this file" wording is now correct (round 4).
- **Prose sync**: `docs/error-handling.md`, `validateWritePayload`, `explainStrippedRelationRoots`, and `validator.go` all state the split on sentinel evidence, with the `1e400` gap named and cited to #402.

iceboundrock:
## Assessment

**Recommend: approve.** The load-bearing insight — the reclassification is only safe because the transform layer independently rejects non-finite values — is preserved by commit ordering (guard `406a48f` before reclassification `903ddc9`), so no intermediate commit lets report-only mode absorb a marshal carrier without a storage backstop. The depth-capped walk with abandon-over-truncate semantics (a partial result is discarded, never published) is exactly the edge-case rigor `docs/error-handling.md` standardizes, and it is mutation-pinned. The four review waves each closed their findings with red-first tests and mutation evidence. No merge blocker.

iceboundrock:
## Observations

- **Mixed-cause probe (see Findings P3):** `json.Marshal(map[string]any{"ch": make(chan int), "score": math.NaN()})` fails on the channel first, yet the walk names `score`. Truthful but not the proximate cause; embedder-only.
- **`numutil.Float64` deliberately untouched** — the correct boundary, since it also serves the DuckDB read/query path (#301 trap avoided).
- **The `json.Number` gate is right and pinned** — `1e400` (ParseFloat `ErrRange`, valid JSON grammar, marshals fine) is not flagged; `"NaN"`/`"Inf"` are. A future "obvious fix" can't widen the gate into fabricating a cause.
- **Read-path consequence, blessed as intended:** a stored bool `value_numeric` holding NaN (corrupt storage; every writer emits 0/1) now reads as a loud plain error, matching the read-path class. #405 tracks the missing row context.
- **The two-layer catch on the string spelling is the intended mechanism:** report-only absorbs the schema type violation (WARN = operator repair signal, #314) and the transform guard is the storage backstop; strict mode rejects at the validator alone.
- **PR body matches the actual 15 commits** with per-review-wave mapping and an explicit "Known boundary (deliberate)" paragraph.

iceboundrock:
## Verification

Review base `5fedafd`, head `0fc6e15`. Ran in a clean worktree at head:

- `go build ./...` — clean.
- `go vet ./internal/schemavalidate/... ./internal/transform/...` — clean.
- `gofmt -l` over all changed `.go` files — clean (no output).
- `git diff --check` base...head — clean.
- `go test ./internal/... -count=1` — all packages pass, including `internal/e2e_harness` (Docker available in this environment; the PR body's "daemon down" note did not reproduce).
- Line counts re-checked: `validator_test.go`=500, `attribute_converter.go`=491, `nonfinite_paths.go`=185.

`golangci-lint` is not installed locally, so I did not run `make lint`; the GitHub **Lint** check on the head commit is green. GitHub checks are green across the board: Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke.

I did not independently re-run the mutation/overlay checks the PR body claims; they are credible, and the abandon-over-truncate pin was reviewed for shape.

iceboundrock:
## Summary

Review status: **approve**. (No formal review state — the authenticated reviewer is the PR author; these comments are the complete review record.)

- **Spec:** #322 direction fully met; one justified scope widening (bool funnels) that corrected a plan evidence error; one documented residual (shape-bound walk, #407).
- **Standards:** no hard violations; the round-4 fixes are correctly landed and `validator_test.go`'s ceiling is tracked (#408).
- **Verification:** full unit suite (including e2e harness), `go vet`, `gofmt`, `git diff --check` clean at head; all five GitHub checks green.
- **Worst single issue:** the new mixed-cause attribution (Findings P3) — embedder-only, truthful-but-not-proximate, not a merge blocker.

Issue #410, opened 2026-08-15
CI lint gate never loads e2e-tagged files; 6 pre-existing findings hide in internal/e2e_harness/federated/


iceboundrock:
Found during #345's full sweep.

`make lint` (and the CI gate that mirrors it) runs golangci-lint without `--build-tags=e2e`, so every `//go:build e2e` file is invisible to the linter. Running `golangci-lint run --build-tags=e2e ./internal/e2e_harness/...` today reports 6 pre-existing findings (errcheck/unused/staticcheck) in `internal/e2e_harness/federated/`; `internal/e2e_harness/production/` is clean.

Two decisions to make:
1. Whether the lint gate should carry the `e2e` tag (or a second tagged invocation should be added to `make lint`/CI) so tagged test files stop being a blind spot.
2. Whether to fix or waive the 6 existing findings when the gate lands (fail-closed would need them fixed in the same change).

Related note from the same sweep: CI races only the unit suite; the production e2e suite passed under `-race` during #345 but nothing keeps that green automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

iceboundrock:
Recording the disposition per #452's Phase 1 carry-out:

**Decisions 1–2 (lint gate + the 6 findings):** already settled by #436 / PR #483 — the two issues are the same finding from different sweeps. `make lint` runs `golangci-lint run --build-tags e2e` fail-closed (the findings were fixed in the same change; the real baseline turned out to be 83 violations, the 6 here were an output-cap artifact), CI runs the recipe via `make lint` (#454), and `TestLintRecipeCoversE2ETag` guards the flag. Verified at current main: `make lint` exits 0.

**The -race ruling this issue's related note asked for:** the production e2e suite runs under `-race` **nightly, not in the PR gate** — the race detector's slowdown would invalidate the wall-clock tripwires #434 calibrated for race-free runs, while the nightly workflow already has the budget and failure-issue reporting. Implemented in PR #492 (new `production-race` job on its own `nightly-e2e-race` label, guarded by `TestNightlyRacesProductionSuite`). The ruling's verification run also caught and deflaked a 2s clock-ahead budget in `TestRecreateBetweenExportAndMark` that `-race` slowdown blows — details in the PR.

PR #492 carries `Closes #410`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

iceboundrock:
[...]

Issue #409, opened 2026-08-15
BuildParquetScanSource renders duplicate identically-named columns when a stale cold-missing skeleton is served


iceboundrock:
Spun out of #345's mutation verification (see #345 issue comments for the full finding).

When a compiled DuckDB skeleton built with a cold-missing set goes stale (served after the column has landed — reachable only if the #255 scope-key component were removed or broken), the scan source renders

```
SELECT * REPLACE(...), NULL::INTEGER AS score FROM read_parquet(..., union_by_name=true)
```

where `*` now also expands a REAL `score` column: the name appears twice in the projection. The pinned DuckDB binds the FIRST (real) occurrence, so the stale NULL is shadowed and the answer stays correct — but correctness is resting on DuckDB's duplicate-name binding order, which is not a contract.

Observed empirically in #345's overlay-mutated run: with the cold-missing scope lines deleted from `duckPlanScopeParts`, the stale skeleton was reused (`plan_cache=hit`), its SQL still carried `NULL::INTEGER AS score`, and the total was still correct (3, not 0) purely because of binding order.

Candidate hardening: make `BuildParquetScanSource` collision-proof — e.g. render missing-column NULLs through the same `REPLACE`/explicit-projection mechanism so a duplicate name is structurally impossible, or fail loudly on a name already present in the scan schema. This is defense-in-depth on top of the #255 scope keying (which `TestPlanCacheColdMissingRekeyEndToEnd` and `federated.TestEngineColdMissingSetRekeysPlanCache` already guard).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #411, opened 2026-08-15
test(e2e): #345 wire the factory-shape plan cache into the production harness engine

### LINKED ISSUE #345: Production e2e harness engine omits the plan cache that factory wires

iceboundrock:
Closes #345

The production e2e harness engine (`internal/e2e_harness/production/engine.go`) built the federated engine without `WithPlanCache`, unlike the real composition root — so `serveFromPlanCache` returned `ok=false` on every harness query and the entire production e2e suite never exercised the compiled-plan serve path. This branch wires it, guards it, restores #255's poisoning probe as a genuine end-to-end proof, and retires the stale prose. Harness-only: no production (non-test) code is touched.

## Commits

- `75afb92` — `Engine()` now builds one `queryplan.NewCache(4096)` shared between its repository and the engine options, exactly the `factory.newRepositoryAndEngine` pairing (#142). The cache is a local of the assembly, so `EvolveSchema`/`ReopenDuckDB` (which nil `e.engine`) discard it with the engine, matching the cold restart both model — `evolve.go`'s existing comment claimed this and is now true. Guarded by `TestEnginePlanCacheServesRepeatedShape`: red-first (no `plan_cache=` note of either kind existed before the wiring), and a deletion spot-check confirmed removing the `WithPlanCache` engine option turns it red again.
- `502b13c` — `TestPlanCacheColdMissingRekeyEndToEnd`: glob Env (`WithoutManifest()`, constant path component) with both measured reads on an empty dirty set, so between the cached pre-flush serve and the post-flush read the ONLY scope-key component that moves is the #255 cold-missing set. Asserts miss → hit → (flush) → miss.
- `cb386f5` — retires the now-false "built without WithPlanCache" scope note in `schema_evolution_never_flushed_e2e_test.go`; a tree-wide sweep confirmed it was the only stale claim.
- `9e51371` — final-review fixes: `post2` re-run asserting `plan_cache=hit` after the post-flush miss (pins that the miss produced a usable recompiled entry AND that the cache survived the flush — a standing guard against the probe going silently vacuous again), plus a one-word comment grammar fix.

## Finding: the #255 scope component is defense-in-depth, not a reachable wrong answer

The issue (and the plan) predicted that a stale cold-absent skeleton served post-flush would keep projecting NULL and answer 0. Mutation verification **disproved the "answer 0" half** (full write-up: [#345 execution note](https://github.com/Lychee-Technology/forma/issues/345#issuecomment-5300696457)): with the cold-missing scope lines deleted (via `go test -overlay`, work tree untouched), the stale skeleton IS reused and still renders `NULL::INTEGER AS score` — but `SELECT *` under `union_by_name` also expands the real post-flush `score`, the name appears twice, and the pinned DuckDB binds the first (real) occurrence. Total stayed 3; only the `plan_cache=hit` note betrayed the staleness. The probe's load-bearing assertion is therefore the post-flush `plan_cache=miss` (keying truthfulness), and the shipped test comment states this true mechanism. The mutated run doubles as positive proof that every other scope component is stable across the flush. Follow-up filed: #409 (collision-proof `BuildParquetScanSource` instead of relying on binding order).

## Scope decision

`EntityManager()`'s repository deliberately keeps its repository-local default cache; only the `Engine()` repo/engine pair shares one cache instance. The OLTP path already caches either way — unifying the harness's two repo constructions into factory's single-instance shape would be a fidelity refactor with no observable behavior difference, and is left out.

## Full sweep (the issue's condition for this change; run at `cb386f5`)

| # | Suite | Command | Result | Counts | Wall |
|---|---|---|---|---|---|
| 1 | Format gate | `make fmt-check` | PASS | — | ~1s |
| 1 | Lint gate | `make lint` | PASS | 0 findings | 1.4s (4.7s cold cache) |
| 1 | Lint, `e2e` tag (extra) | `golangci-lint --build-tags=e2e ./internal/e2e_harness/...` | PASS for `production/` | 0 in changed pkg; 6 pre-existing in `federated/` → #410 | 4.9s |
| 2 | Unit | `make test` | PASS | 35 pkg ok, 0 fail | 16.0s |
| 3 | **Production e2e** | `make test-e2e-production` | **PASS** | 129 tests + 104 subtests, 0 fail, 0 skip | 2:18 |
| 4a | Harness smoke | `go test -v ./internal/e2e_harness/... -timeout=5m` | PASS | 162 pass, 0 fail | 8.2s |
| 4b | Federated e2e short | `go test -v ./internal/e2e_harness/federated/... -tags=e2e -short` | PASS | 187 + 41 subtests, 0 fail, 38 explicit `-short` skips | 2:00 |
| extra | Unit `-race` | `go test -race . ./cdc ./cmd/... ./factory ./internal/...` | PASS | 32 pkg ok, 0 races | 36.2s |
| extra | Production e2e `-race` | `go test -race ./internal/e2e_harness/production/ -tags=e2e` | PASS | 0 races, 0 fail | 2:45 |

Every one of the 129 production e2e tests now routes through `serveFromPlanCache` and answers identically with cached skeletons + per-request arg binding; no harness assertion needed loosening. The `-race` runs were added because a newly shared cache under concurrent queries is exactly where a race would hide. Post-sweep commit `9e51371` touches only the two harness test/comment files; both affected e2e tests were re-run green (`TestPlanCacheColdMissingRekeyEndToEnd` 4.64s, `TestEnginePlanCacheServesRepeatedShape` 4.21s) plus `make fmt-check`.

## Deferred (triaged by the final whole-branch review; none block merge)

- Repo-side `WithPlanCache` injection is unguarded — deleting it degrades only to the repo-local default cache (fidelity, not coverage).
- `assertPlanCacheNote` is membership-only — hit+miss coexistence in Notes is unreachable today (`executionPlanMark.rewind` truncates on retry; #348 tracks the Timings variant).

## Follow-ups filed

- #409 — `BuildParquetScanSource` duplicate-column collision-proofing (from the mutation finding above).
- #410 — CI lint gate never loads e2e-tagged files (6 pre-existing findings in `federated/`); also notes the `-race` e2e gap.

### DIFF
--- a/internal/e2e_harness/production/engine.go
+++ b/internal/e2e_harness/production/engine.go
@@ -8,14 +8,16 @@ import (
 	"github.com/lychee-technology/forma/internal"
 	fedengine "github.com/lychee-technology/forma/internal/federated"
 	"github.com/lychee-technology/forma/internal/manifest"
+	"github.com/lychee-technology/forma/internal/queryplan"
 	"github.com/lychee-technology/forma/internal/schemavalidate"
 	"github.com/lychee-technology/forma/internal/transform"
 )
 
 // Engine returns the Env's real federated query engine, assembling it on
 // first use: the production DBPersistentRecordRepository over the per-test
-// pool, the Postgres dirty-ID fetcher, the per-test DuckDB client, an
-// optional circuit breaker (WithBreaker), and the loaded metadata cache.
+// pool, the Postgres dirty-ID fetcher, the per-test DuckDB client, the
+// factory-shape shared plan cache (#345), an optional circuit breaker
+// (WithBreaker), and the loaded metadata cache.
 // This mirrors the production assembly used by the benchmark runner
 // (internal/e2e_harness/federated/benchmark/execute.go).
 func (e *Env) Engine() *fedengine.DBFederatedQueryEngine {
@@ -23,7 +25,13 @@ func (e *Env) Engine() *fedengine.DBFederatedQueryEngine {
 		return e.engine
 	}
 
-	repo := internal.NewDBPersistentRecordRepository(e.Pool, e.Metadata)
+	// One plan cache per engine assembly, shared between the repository and
+	// the engine exactly as factory.newRepositoryAndEngine pairs them
+	// (#142/#345). Building it here rather than on the Env ties its lifetime
+	// to the memoized engine: EvolveSchema and ReopenDuckDB drop e.engine, so
+	// the cache is discarded with it, matching the cold restart both model.
+	planCache := queryplan.NewCache(4096)
+	repo := internal.NewDBPersistentRecordRepository(e.Pool, e.Metadata, internal.WithPlanCache(planCache))
 	if e.breaker == nil && e.opts.breakerFailures > 0 {
 		// Built once per Env, not per engine assembly: breaker state must
 		// survive ReopenDuckDB/RestartPostgres handle rebuilds so #185
@@ -33,7 +41,10 @@ func (e *Env) Engine() *fedengine.DBFederatedQueryEngine {
 
 	// The engine's logger carries the #256 stamp/footer cross-check warning,
 	// which has no other outlet — the read it observes succeeds.
-	opts := []fedengine.EngineOption{fedengine.WithLogger(e.logger)}
+	opts := []fedengine.EngineOption{
+		fedengine.WithLogger(e.logger),
+		fedengine.WithPlanCache(planCache),
+	}
 	if src := e.parquetSource(); src != nil {
 		if e.ParquetSourceWrap != nil {
 			src = e.ParquetSourceWrap(src)
--- a/internal/e2e_harness/production/engine.go
+++ b/internal/e2e_harness/production/engine.go
@@ -29,7 +29,7 @@ func (e *Env) Engine() *fedengine.DBFederatedQueryEngine {
 	// the engine exactly as factory.newRepositoryAndEngine pairs them
 	// (#142/#345). Building it here rather than on the Env ties its lifetime
 	// to the memoized engine: EvolveSchema and ReopenDuckDB drop e.engine, so
-	// the cache is discarded with it, matching the cold restart both model.
+	// the cache is discarded with it, matching the cold restart that both model.
 	planCache := queryplan.NewCache(4096)
 	repo := internal.NewDBPersistentRecordRepository(e.Pool, e.Metadata, internal.WithPlanCache(planCache))
 	if e.breaker == nil && e.opts.breakerFailures > 0 {
--- a/internal/e2e_harness/production/plan_cache_e2e_test.go
+++ b/internal/e2e_harness/production/plan_cache_e2e_test.go
@@ -1,0 +1,57 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"testing"
+
+	"github.com/lychee-technology/forma/internal/model"
+)
+
+// assertPlanCacheNote pins the compiled-plan cache outcome recorded in the
+// execution plan (recordPlanCache, #142). Notes are internal-plan-only, so
+// asserting on them observes the engine, not the public API surface.
+func assertPlanCacheNote(t *testing.T, res *QueryResult, want string) {
+	t.Helper()
+	if res == nil || res.Plan == nil {
+		t.Fatalf("plan-cache note %q: result carries no execution plan", want)
+	}
+	for _, note := range res.Plan.Notes {
+		if note == want {
+			return
+		}
+	}
+	t.Fatalf("plan notes lack %q, got %v", want, res.Plan.Notes)
+}
+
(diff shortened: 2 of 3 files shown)


iceboundrock:
## Review — #345 plan-cache wiring (re-review)

### Spec
Issue #345 asks for three things; all three are delivered, and only these three:

1. **Wire `WithPlanCache` into the harness engine** — `engine.go:26-33` builds one `queryplan.NewCache(4096)` and injects it into both the repository (`internal.WithPlanCache`) and the engine options (`fedengine.WithPlanCache`), the exact `factory.newRepositoryAndEngine` pairing (`factory/factory.go:305-313`). ✓
2. **Restore #255's poisoning probe as a genuine e2e proof** — `TestPlanCacheColdMissingRekeyEndToEnd` uses `WithoutManifest()` so the glob path component of the scope key is a constant string across the flush, and asserts miss → hit → (flush) → miss → hit. ✓
3. **Retire the stale "built without WithPlanCache" note** in `schema_evolution_never_flushed_e2e_test.go`. ✓

No scope creep. The two mid-branch commits (`502b13c`, `9e51371`) are refinements squarely inside the issue's own "consider also…" list, not new work.

### Standards
- File/function limits: `plan_cache_e2e_test.go` is 146 lines (<500); the larger test body is ~45 lines (its long doc comment doesn't count against the 100-line function cap). No violations.
- Error handling: test code uses the idiomatic `t.Fatalf("%v", err)` rather than error wrapping — correct for `*testing.T` (the wrap-with-context rule targets production errors).
- Style consistency: `assertPlanCacheNote` uses `t.Fatalf` and matches `assertUsesDuckDB`/siblings; the `production` harness package uses no `testify/require` at all (verified: 0 files), so the helper is consistent.
- Verified: `gofmt -l` clean on all three changed files; `go vet -tags=e2e ./internal/e2e_harness/production/` exits 0 (test files compile under the e2e tag).

### Findings / Observations (non-blocking)
1. **Asymmetric repo construction (documented, deliberate).** `Engine()` now builds a cache-wired repo, but `EntityManager()` (`engine.go:138`) still builds its own repo with only the repository-local default cache. The harness thus keeps two repo instances where `factory.newRepositoryAndEngine` returns one shared. The PR's "Scope decision" section calls this a fidelity refactor with no observable behavior difference — I confirm that reasoning: the OLTP read path caches either way, so unification would be invisible to tests.
2. **The e2e probe proves keying, not rendering — correctly so.** The harness test's load-bearing assertion is the post-flush `plan_cache=miss` note, *not* `NotContains(NULL::INTEGER AS score)`. The engine-seam twin `TestEngineColdMissingSetRekeysPlanCache` asserts the rendering; the harness cannot, because under `union_by_name` the stale `NULL::INTEGER AS score` is shadowed by the real post-flush `score` (first-name binding order). The test comment and the #345 execution note state this exactly, and #409 is filed for the collision-proofing. Honest and correct — no change required.
3. **`assertPlanCacheNote` is membership-only.** Coexistence of hit+miss in Notes is unreachable today: `recordPlanCache` appends exactly one note per serve, and the plan context is rebuilt per query. Matches the deferral note; #348 tracks the Timings variant.
4. Cosmetic: the top `Engine()` doc comment still names only the benchmark runner as the mirror, while the new inline comment (lines 26-29) names factory. Both statements are true; no action needed.

### Verification
- Independent: `gofmt` clean; `go vet -tags=e2e ./internal/e2e_harness/production/` exits 0.
- I did not re-run the full e2e matrix (requires Docker, 2min+). The PR's sweep table reports `make test-e2e-production` (129 tests + 104 subtests), harness smoke, federated short, and two `-race` runs all green — consistent with a harness-only change.

### Assessment
**Approve.** Harness-only, mirrors the factory shape exactly, both new tests are red-first, and the probe's load-bearing assertion is honest about what it can and cannot observe. Deferred items and follow-ups (#409, #410) are appropriately triaged.

PR #412, opened 2026-08-15
#348 post-#251 polish: HTTP-visible partial marker, Timings rewind, two test pins

### LINKED ISSUE #348: Post-#251 polish: Timings rewind on retry, two missing test pins, HTTP-visible partial marker decision

iceboundrock:
Closes #348

## What this delivers (all four issue items, single PR per user adjudication)

**Item 4 — HTTP-visible partial marker (first priority per issue comment):**
- Engine records the #251 corrupt-exclusion set as `model.PartialScan` — an out-parameter on `FederatedQueryOptions`, overwritten per DuckDB pass (truth-by-construction: it describes the pass that produced the page), deliberately NOT gated on `include_execution_plan`.
- `PersistentRecordPage.Partial` carries it; the service projects it into the new public field `QueryResult.partial` = `{"reason": "corrupt_parquet_excluded", "excluded_object_count": N}` (`omitempty`).
- **Security (#301/#306):** only reason + count cross the HTTP boundary; the full storage keys stay on the internal plan Notes for embedders/operators. A test marshals the projection and asserts no `s3://` on the wire.
- Per the PR #349 review constraint, this is deliberately **not** `Routing.Reason` — the route does not change on a partial read. Degraded fallback keeps its existing `Routing.Reason` signal (adjudicated scope: #251 exclusion only). `degradeToPostgresOnly` also clears the out-parameter so opts never contradicts a postgres-only page.
- Coverage: engine seam tests (marker present after retry WITHOUT execution plan requested; nil on clean scans; nil after degradation), service wiring pin (delete `Partial: toPartialResult(page.Partial)` → red, per the #318 delete-it-goes-red standard), JSON key/omitempty pins, harness `assertPartialMarker` wired into all four corrupt-exclusion e2e sites — both `TestManifestConsistency_OneGoodOne*` scenarios ran green in real containers.

**Item 1 — Timings rewind:** `markExecutionPlan` now snapshots the pre-pass `Timings` entries and `rewind` restores them wholesale (a merged map cannot express drop-by-truncation). Red test was deterministic via the #255 scope-hash property (resolved path set is in the plan-cache key): warm the cache with a clean full-set request → the failed pass HITs, the retry over the remainder MISSes → pre-fix the request publicly reported both `plan_cache_hit` and `plan_cache_miss`; post-fix only the retry's stamp survives.

**Item 2 — pre-pass plan survival pin:** seeds a caller Source/Note/Timing before `Query` and requires all three survive the rewind, plus a retry-actually-ran odometer. Mutation-verified via `go test -overlay`: the issue's named mutation (`markExecutionPlan` → `return executionPlanMark{}`) kills all three assertions (source count 1→0, note absent, timing 7→0).

**Item 3 — half-open probe accounting pin:** unit test pins the full trace (open → half-open, probe hits confirmed corruption → `ReleaseProbe` not `RecordFailure`, retry's `Allow` re-reserves, `RecordSuccess` closes; final `Allow` returns the zero token only the closed state hands out). Both required mutations killed via overlay: deleting `ReleaseProbe` and swapping it for `RecordFailure` each fail the test at the retry admission.

**Docs:** `docs/federated-query/design.md` §7.3 Loudness contract updated for both the Timings restore and the public marker.

## Verification

- `make fmt-check`, `make lint`, `make test` clean; `go vet -tags=e2e ./internal/e2e_harness/production/` clean.
- Targeted Docker e2e: `TestManifestConsistency_OneGoodOneBadFile` (3.74s) + `OneGoodOneTruncatedFile` (0.70s) PASS.
- All mutation kill-checks ran via `go test -overlay` (working tree never mutated) and were independently re-run by the final re-reviewer.

## Known residuals (deliberately not in this PR)

- The breaker mutation `ReleaseProbe` → `RecordSuccess` survives the item-3 pin (behaviorally identical trace; distinguishing needs mid-query breaker introspection).
- `computeFederatedCount` uses fresh internal opts, so corruption first confirmed during a recount pass shrinks `total_records` without a marker (requires corruption manifesting between two adjacent passes of one request — exotic).

A follow-up issue for these is being filed and will be linked here.


iceboundrock:
Follow-up issue for the two known residuals named in the PR body: #413

iceboundrock:
## Re-review of PR #412 (second pass)

Compared `main` @ `49fccec` against `feat/348-post-251-polish` @ `082e9bf` (14 files, +429/−5). Re-reviewed against issue #348, the implementation plan recorded on the issue, `coding-standard.md`, and `AGENTS.md`.

### Spec

All four #348 items are delivered, matching the plan's recorded adjudications:

1. **Item 1 — Timings rewind** ✅ `executionPlanMark` gains a `timings` snapshot; `markExecutionPlan` copies the pre-pass `Timings` map; `rewind` restores it wholesale (`duckdb_query.go`). Correct mechanism: `Timings` *is* projected through `toExecutionPlan` (unlike `Notes`), so snapshot-restore is the only way to drop the failed pass's `plan_cache_hit` next to the retry's `plan_cache_miss`.
2. **Item 2 — pre-pass plan survival pin** ✅ `TestCorruptRetryRewindPreservesPrePassPlan` seeds caller Source/Note/Timing before `Query` and requires all three to survive the rewind. The issue's named mutation (`markExecutionPlan → return executionPlanMark{}`) kills it.
3. **Item 3 — half-open probe accounting pin** ✅ `TestBreakerHalfOpenProbeReleasedOnConfirmedCorruption` pins the full trace: probe reserved at `Allow`, released (not failure-recorded) on confirmed corruption, re-reserved by the retry, closed by `RecordSuccess`, closed-state `Allow` hands out the zero token.
4. **Item 4 — HTTP-visible partial marker** ✅ `PartialScan` out-param → `PersistentRecordPage.Partial` → `toPartialResult` → top-level `QueryResult.Partial` = `{reason, excluded_object_count}`, not gated on `include_execution_plan`, deliberately not `Routing.Reason`.

Adjudications respected: top-level field; reason enum + count only (no storage keys — marshal test asserts no `s3://` on the wire); #251-scope only (degraded fallback keeps its `Routing.Reason` signal); single PR.

The two residuals are correctly documented in the PR body and tracked in #413 (breaker `ReleaseProbe→RecordSuccess` indistinguishable at this seam; count-pass corruption invisible to the marker). Neither blocks.

### Standards

No hard violations against `coding-standard.md` / `AGENTS.md`:

- File sizes: `duckdb_query.go` 486, `engine.go` 477, `entity_query_service.go` 493, `types.go` 486 — all ≤500.
- Function sizes: largest new function is a test (~40 lines); no production function approaches 100.
- godoc present on all new exported symbols (`PartialScan`, `PartialResultInfo`, `PartialReasonCorruptParquetExcluded`, `QueryResult.Partial` field).
- Error wrapping: no new error-return sites; existing wraps unchanged.
- Early returns / flat structure maintained in `rewind`, `markExecutionPlan`, `toPartialResult`, `partialScanFromOpts`.

### Findings / Observations (non-blocking)

1. **Out-param clearing asymmetry (minor, likely acceptable).** `degradeToPostgresOnly` clears `opts.PartialScan = nil` so the out-param agrees with the postgres-only page it returns — justified, because a DuckDB pass actually ran (and set the marker) before degrading. The hot-only gate and routed-postgres paths do *not* clear `opts.PartialScan`, but they never set it in the same call (no DuckDB pass runs) and never set `page.Partial` either (only the DuckDB success literal calls `partialScanFromOpts`). So the only observable divergence is an embedder that reuses a single `FederatedQueryOptions` across calls and inspects `opts.PartialScan` after a postgres-only answer. Not a bug for the current service path (which builds fresh opts per request); flagging for completeness only.
2. **`rewind` now always reassigns `plan.Timings`.** On the only path it runs (the retry), `plan.Timings` is replaced by a freshly built map even when the snapshot is empty, converting a nil map to a non-nil empty map. Harmless given the early return on `opts.ExecutionPlan == nil`, but a subtle nil-vs-empty change from the pre-#348 truncation-only behaviour.

Neither is blocking.

### Assessment

Well-scoped and self-documenting. The test pins are genuinely load-bearing: `internal/entity_query_partial_wiring_test.go` closes the exact gap the #318 delete-it-goes-red standard targets — it asserts the `Partial:` field is actually wired into the `QueryResult` literal (deleting it would otherwise leave every other test green), plus the `partial,omitempty` tag on the wire. The #301/#306 security boundary is enforced by marshalling assertions in two independent places (unit projection + service result). The truth-by-construction argument for both the Timings restore and the out-param is coherent and matches the §7.3 design-doc update.

### Verification (independently reproduced on the PR branch)

- `make fmt-check` — clean.
- `go test . ./cdc ./factory ./internal/...` — all packages pass.
- `go vet -tags=e2e ./internal/e2e_harness/production/` — clean.
- Targeted: `go test ./internal/federated -run 'TestCorruptRetry|TestBreakerHalfOpenProbeReleasedOnConfirmedCorruption|TestCleanQueryPageHasNoPartialMarker|TestDegradedFallbackClearsPartialMarker'` — PASS; `go test ./internal -run 'TestToPartialResultProjectsCountOnly|TestQueryResultCarriesPartialMarkerFromEngine|TestQueryResultOmitsPartialMarkerForCleanPage'` — PASS.

Not re-run locally: `make lint` (golangci-lint v1.64.8 not installed in this environment) and the Docker e2e (`TestManifestConsistency_OneGoodOne*`). The four `assertPartialMarker` call sites and the author's reported green runs are consistent with the reproduced unit coverage.

### Summary

Spec: 4/4 items implemented correctly. Standards: no violations (two minor non-blocking observations). Verification: reproduced for fmt/unit/vet. **Approve** (no auto-merge per AGENTS.md); observations 1–2 are optional candidates to fold into #413.

iceboundrock:
Response to the re-review's two observations:

**Observation 1 (out-param clearing asymmetry) — fixed in 0eb9851.** Verified the asymmetry as described: the hot-only gate and the routed postgres-only path never touch `opts.PartialScan`, so a caller reusing one options value across queries would read the previous call's marker after a postgres-only answer — the same opts-only-reader class the degraded-path clear protects. Rather than clearing per postgres-only branch, `Query` now resets the out-parameter at entry, which makes the contract uniform: after any `Query` call, `opts.PartialScan` describes that call. Pinned by `TestPostgresOnlyAnswerClearsReusedPartialMarker` (corrupt-retry first call sets the marker on shared opts; hot-only second call must leave both `page.Partial` and `opts.PartialScan` nil). Kill-check via `go test -overlay` with the entry reset removed: fails exactly on the reuse assertion. `make fmt-check` + `./internal/federated ./internal .` green. The `FederatedQueryOptions.PartialScan` doc comment now states the entry-reset contract. Nothing left of this observation to fold into #413.

**Observation 2 (`rewind` reassigns `plan.Timings`, nil→non-nil empty map) — no change, intentional.** The wholesale reassignment IS the restore mechanism (a merged map cannot express drop-by-truncation), and the nil→empty conversion only occurs for a caller that passed `IncludeExecutionPlan` with a non-nil plan whose `Timings` was nil — a state in which every pass's `Timings[...]` writes (`recordTranslation`, `recordPlanCache`, `finalizeDuckDBExecutionPlan`) would panic on the nil map, before and after this PR alike. Post-rewind the retry pass writes into a non-nil map, so the change strictly reduces that pre-existing hazard on the one path it touches. `newDuckDBExecutionPlanContext` initializes `Timings` only when it creates the plan itself, so the nil-`Timings` state is reachable only by caller construction; making that state fully safe (init `Timings` for caller-provided plans too) would be a separate hardening beyond this PR's scope — noting it here rather than in #413 since no real caller constructs it.

PR #424, opened 2026-08-19
fix: demote and memoize the dropped-attr EAV skip log (#343)

### LINKED ISSUE #343: Demote/memoize the per-row Warn for skipped dropped-attr EAV records

iceboundrock:
Closes #343.

`transform.FromEAVRecords` logged a Warn once per call — i.e. once per row on the read path — whenever it skipped attrIDs removed by schema evolution, a condition #294 defines as supported. A 1000-row page over such a schema attempted 1000 Warn emissions, and Warn reads as an incident to on-call.

Changes:
- The skip log is demoted to **Info** and **memoized per (schemaID, sorted skipped-attrID set)**: one emission per process per schema shape. The memo is package-level because the read path builds a fresh `AttributeConverter` per row (`persistentRecordTransformer.newConverter`), so instance state would dedupe nothing.
- Message text now covers the #342 nuance ("removed by schema evolution **or retired**") — shipped ledgers retiring e.g. `visit_full.logs` (29) route those ids through this same skip path.
- Field rename `rowID` → `exampleRowID`: under memoization the logged row is only the first one observed with that shape. Grep found no doc/test/alerting reference to the old message text or field.
- The previously untested emission behavior is pinned: level, once-per-shape across fresh converters, re-log on a distinct skipped set, and exactly-once under 8-way concurrency (runs under CI's `-race`). A mutation check (memo disconnected → both tests red) was performed during development.
- `docs/error-handling.md` records the log contract next to the #294 tolerate-and-preserve paragraph.

Operator note: any log-based alerting keyed on the old Warn text/level must move to the new Info message (`skipped_attr_log.go`'s `skippedAttrLogMessage` const; stable fields `schemaID`, `attrIDs`).

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -597,11 +597,13 @@ deployment, and should be treated as operator-visible consistency failures.
 EAV records whose attribute id is no longer present in the schema metadata are
 not an error: they are skipped on read and preserved on update (#294
 tolerate-and-preserve), so removing an attribute from a schema is
-non-destructive and re-adding it restores the stored values. An attributeID
-freed by removing an attribute must never be reused for a different attribute:
-preserved EAV rows would silently bind to the new attribute's name (same value
-type) or make the row unreadable with a storage type mismatch (different value
-type).
+non-destructive and re-adding it restores the stored values. The read-path skip
+is logged at Info, once per process per (schema, skipped attributeID set),
+naming the first row observed as `exampleRowID` — a supported state's audit
+trail, not an incident signal (#343). An attributeID freed by removing an
+attribute must never be reused for a different attribute: preserved EAV rows
+would silently bind to the new attribute's name (same value type) or make the
+row unreadable with a storage type mismatch (different value type).
 
 Since #342 that rule is enforced. `generate-attributes` keeps the removed
 attribute's entry in `<schema>_attributes.json` marked `"retired": true` and
--- a/internal/transform/attribute_converter.go
+++ b/internal/transform/attribute_converter.go
@@ -215,8 +215,7 @@ func (c *AttributeConverter) FromEAVRecords(records []model.EAVRecord) ([]model.
 			ids = append(ids, id)
 		}
 		slices.Sort(ids)
-		zap.S().Warnw("skipped EAV records for attribute ids not in metadata cache (removed by schema evolution; rows preserved, #294)",
-			"schemaID", schemaID, "rowID", records[0].RowID, "attrIDs", ids)
+		logSkippedAttrIDs(schemaID, records[0].RowID, ids)
 	}
 
 	if err := c.checkRequiredAttributes(schemaID, cache, presentAttrIndices); err != nil {
--- a/internal/transform/skipped_attr_log.go
+++ b/internal/transform/skipped_attr_log.go
@@ -1,0 +1,55 @@
+package transform
+
+import (
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+)
+
+// skippedAttrLogMessage is a const so the tests pinning the once-per-shape
+// contract filter on the exact string production emits.
+const skippedAttrLogMessage = "skipped EAV records for attribute ids not in metadata cache (removed by schema evolution or retired; rows preserved, #294)"
+
+// skippedAttrLogSeen dedupes the tolerated-skip log across the process (#343).
+// FromEAVRecords runs once per row on the read path, and the skipped condition
+// is a property of the schema shape, not of the row, so per-call logging turned
+// a #294-supported state into per-row noise. Package-level deliberately:
+// production builds a fresh AttributeConverter per row
+// (persistentRecordTransformer.newConverter), so converter-scoped state would
+// dedupe nothing. Keys are schemaID plus the sorted skipped attrID set; the
+// population is bounded by registered schemas times the few dropped/retired
+// sets each can accumulate, so the map needs no eviction.
+var skippedAttrLogSeen = struct {
+	mu   sync.Mutex
+	keys map[string]struct{}
+}{keys: make(map[string]struct{})}
+
+// logSkippedAttrIDs emits the skip log at Info once per process per (schemaID,
+// skipped attrID set). Info, not Warn: #294 defines the state as supported, and
+// Warn reads as an incident to an on-call operator. ids must already be sorted
+// so equal sets build equal keys; exampleRowID is the first row observed with
+// this shape, recorded as a lookup pointer, not as the condition's scope.
+func logSkippedAttrIDs(schemaID int16, exampleRowID uuid.UUID, ids []int16) {
+	var key strings.Builder
+	key.WriteString(strconv.Itoa(int(schemaID)))
+	for _, id := range ids {
+		key.WriteByte(':')
+		key.WriteString(strconv.Itoa(int(id)))
+	}
+
+	skippedAttrLogSeen.mu.Lock()
+	_, dup := skippedAttrLogSeen.keys[key.String()]
+	if !dup {
+		skippedAttrLogSeen.keys[key.String()] = struct{}{}
+	}
+	skippedAttrLogSeen.mu.Unlock()
+	if dup {
+		return
+	}
+
+	zap.S().Infow(skippedAttrLogMessage,
+		"schemaID", schemaID, "exampleRowID", exampleRowID, "attrIDs", ids)
+}
--- a/internal/transform/skipped_attr_log_test.go
+++ b/internal/transform/skipped_attr_log_test.go
@@ -1,0 +1,117 @@
+package transform
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/lychee-technology/forma"
+	"github.com/lychee-technology/forma/internal/model"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+// resetSkippedAttrLogMemo clears the process-wide memo so each test observes
+// first-emission behavior regardless of what earlier tests in the package
+// (e.g. TestAttributeConverterFromEAVRecords_SkipsUnknownAttributeIDs) fed it.
+func resetSkippedAttrLogMemo() {
+	skippedAttrLogSeen.mu.Lock()
+	defer skippedAttrLogSeen.mu.Unlock()
+	skippedAttrLogSeen.keys = make(map[string]struct{})
+}
+
+func skipLogRegistry(schemaID int16) *stubSchemaRegistry {
+	return &stubSchemaRegistry{
+		schemaID:   schemaID,
+		schemaName: fmt.Sprintf("skip_log_schema_%d", schemaID),
+		cache: forma.SchemaAttributeCache{
+			"name": {AttributeID: 1, ValueType: forma.ValueTypeText},
+		},
+	}
+}
+
+func skipLogRecords(schemaID int16, staleAttrID int16) []model.EAVRecord {
+	rowID := uuid.Must(uuid.NewV7())
+	known := "alive"
+	stale := "mystery"
+	return []model.EAVRecord{
+		{SchemaID: schemaID, RowID: rowID, AttrID: 1, ValueText: &known},
+		{SchemaID: schemaID, RowID: rowID, AttrID: staleAttrID, ValueText: &stale},
+	}
+}
+
+// TestSkippedAttrLogInfoOncePerShape pins #343: the #294 tolerated-skip log is
+// Info (not Warn — the state is supported, not an incident) and fires once per
+// process per (schemaID, skipped-attrID-set), surviving converter churn — the
+// read path builds a fresh converter per row (persistentRecordTransformer's
+// newConverter), which is exactly why the memo must be package-level.
+func TestSkippedAttrLogInfoOncePerShape(t *testing.T) {
+	resetSkippedAttrLogMemo()
+	registry := skipLogRegistry(441)
+
+	core, logs := observer.New(zap.InfoLevel)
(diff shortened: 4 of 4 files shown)


iceboundrock:
**Final whole-branch review** (pre-PR verification artifact; per-task spec+quality reviews also passed — working copies lived in the gitignored SDD workspace, now recorded here).

Range reviewed: `bcf9a3a..2e4d9fb`. Verdict: **Ready to merge** — 0 Critical, 0 Important.

### Strengths

- **The memo's placement argument is verified, not asserted.** Both production call sites (`persistent_record.go:153` via per-row `newConverter`, `transformer.go:97`) route through the package function, so package-level state is the only level that dedupes — and the code comment records *why*, citing the per-row converter churn.
- **Key encoding is injective.** Decimal integers joined by `:`, and no component can contain `:` — `(1, [23])` → `"1:23"` vs `(12, [3])` → `"12:3"` cannot collide, including for negative int16 values. The empty-ids degenerate key is unreachable from the only caller (guarded by `len(skippedAttrIDs) > 0`) and harmless if reached.
- **Lock discipline is right.** Check-and-insert under one lock, emission outside it — a slow sink can't serialize readers, and exactly one goroutine passes the memo. Race-clean under `-race` (CI runs unit tests with `-race`).
- **The mutation check was real:** disconnecting the memo turned both new tests red (evidence retained in the task report); revert re-verified green. The wiring-must-turn-red lesson (#318) applied.
- **Test hygiene:** memo reset at test start; no `t.Parallel` anywhere in `internal/transform` (verified), so `zap.ReplaceGlobals` cannot race a sibling test; the new tests are order-independent from the pre-existing skip tests that also feed the memo.
- **Message-as-const** (`skippedAttrLogMessage`) eliminates production/test string drift; tests filter on the exact string production emits.
- **Doc sentence re-verified against shipped code** (level, cadence, field name), honoring the grep-back rule.
- File-size constraint respected: `attribute_converter.go` 491 → 490 lines; new logic in a new 55-line file.

### Minor observations (none actionable)

1. Memo marks "seen" even if the emission was filtered by a Warn-level logger later reconfigured to Info — theoretical only; logger level is fixed at startup here.
2. E2E harnesses spinning multiple isolated environments in one process share the memo; a second environment reusing schemaID+set stays silent. Log-only impact.
3. "The first row observed" is, under concurrency, the memo-winner's row, not necessarily temporally first. The field is a lookup pointer either way.

### Deferred-minor triage (from per-task reviews)

| Finding | Verdict |
|---|---|
| Memo key's schemaID component not load-bearing under test (same id-set on a second schema would pin it) | ship-as-is; fold a ~6-line test into any future touch of this file |
| `exampleRowID` field rename unpinned by tests | ship-as-is; pinning field names would be over-pinning an informational field |
| `key.String()` called twice inside the lock | ship-as-is; `strings.Builder.String()` shares the buffer — zero cost |
| Doc sentence names no filter anchor for alert migration | ship-as-is; the anchor is the `skippedAttrLogMessage` const and the PR body carries the operator note — a doc copy could drift |
| "Once per process" is one-way (memo never evicts; absence ≠ condition ended) | ship-as-is; the phrase states the eviction-free contract accurately |
| Doc sentence interleaves between two semantic rules in the paragraph | ship-as-is; editorial preference |

### Standing note

If `internal/transform` tests ever adopt `t.Parallel`, the two `zap.ReplaceGlobals` tests (and `resetSkippedAttrLogMemo`) must stay serial — the package-wide absence of `t.Parallel` is the invariant protecting them, and nothing pins it.

Gate evidence (from execution): `make test`, `make fmt-check`, `make lint` (golangci-lint pinned v1.64.8) all exit 0.

iceboundrock:
## Review: findings (#343)

Second whole-branch review of PR #424 (range `bcf9a3a..2e4d9fb`, i.e. `origin/main...origin/fix/343-memoize-skip-log`; commits `f12c644`, `2e4d9fb`). Severity-ranked per `docs/ai_review_guideline.md` (P0–P3).

### Findings

**No material findings.** 0 × P0, 0 × P1, 0 × P2, 0 × P3.

Both review axes were run independently over the diff and neither produced a defect-class finding:

- **Key injectivity holds:** keys are `Itoa(schemaID)` + colon-joined `Itoa(id)`; no segment can contain `:` (int16 decimal is digits and an optional leading `-`), so `(1,[23])` → `"1:23"` vs `(12,[3])` → `"12:3"` cannot collide, negatives included. The empty-set key is unreachable (guarded by `len(skippedAttrIDs) > 0`) and injective anyway.
- **Lock discipline is correct:** check-and-insert under one mutex, emission outside it → exactly one winner emits, and a slow log sink never serializes readers.
- **House rules respected:** no bare error returns; `attribute_converter.go` stays ≤500 lines (490); `logSkippedAttrIDs` ≈ 25 lines; no >3-level nesting added.

Residual P3 gaps (tracked under *observations*, not blocking): `exampleRowID` field value asserted nowhere; no negative-attrID/empty-set key boundary test; the package-wide no-`t.Parallel` invariant that protects the global memo + `zap.ReplaceGlobals` swaps is unpinned.

iceboundrock:
## Review: spec (#343)

Spec sources: issue #343 (body) + implementation plan `docs/superpowers/plans/2026-08-18-343-demote-memoize-skip-log.md` (gitignored; decision record lives on the issue).

| Spec requirement | Status | Evidence |
|---|---|---|
| Issue (a): demote to Info **or** memoize per (schemaID, skipped-attrID-set) | **Met — both done** (plan decision 1) | `attribute_converter.go` no longer emits Warn; `logSkippedAttrIDs` → `zap.S().Infow` at `skipped_attr_log.go:53` |
| Issue (b): pin the previously untested emission | Met | `TestSkippedAttrLogInfoOncePerShape` + `TestSkippedAttrLogConcurrentCallsEmitOnce`, zaptest/observer precedent |
| Plan 2: key = schemaID + sorted skipped-attrID set | Met | caller sorts via `slices.Sort` (`attribute_converter.go:217`); key built from schemaID + sorted ids → equal sets, equal keys |
| Plan 3: message covers the #342 retired-attr nuance | Met | const text `"…removed by schema evolution or retired; rows preserved, #294"` |
| Plan 4: field rename `rowID` → `exampleRowID` | Met | field renamed; grep confirms no doc/test/alerting reference to the old text or field |
| Plan 5: message string as package const | Met | tests filter on `skippedAttrLogMessage` — exact string production emits |
| Constraint: shrink `attribute_converter.go` (was 491, at the 500 cap) | Met | 491 → 490 |

- **Missing / partial:** none.
- **Scope creep:** none — the diff touches exactly the four planned files.
- **Wrong implementation:** none. Findings-candidate points all checked out sound: `mark-before-emit` still honors "≤ once per process per shape" (missing an emission because the level filters it is within contract); `fmt.Sprint(fields["attrIDs"]) == "[999]"` holds because zap's observer preserves the `[]int16` field type (the plan's Known-risk note anticipated this); `t.Errorf` inside a goroutine is legal (only `FailNow`/`Fatal` are prohibited off the test goroutine) and `wg.Wait()` precedes the count assert.

iceboundrock:
## Review: standards (#343)

Sources: AGENTS.md / CLAUDE.md, `coding-standard.md`, `docs/ai_review_guideline.md`.

**Hard violations:** none.

- File/function sizes: `skipped_attr_log.go` 55 lines, `skipped_attr_log_test.go` 117, `logSkippedAttrIDs` ≈ 25 lines; `attribute_converter.go` stays at/below 500.
- No bare `return err`; any errors carry context.
- Comment style (prose, `#N` references, rationale-before-code) consistent with the surrounding `attribute_converter.go` / `persistent_record.go` files.
- Docs edit in `docs/error-handling.md` states level / once-per-(schema,set) / `exampleRowID` accurately against the shipped code (Info, not Warn).

**Judgement calls:**

1. **Package-level global memo** `skippedAttrLogSeen` — `coding-standard.md` §7 and `docs/ai_review_guideline.md` §3.5 flag "global variables for application state". Here it is **forced, not a style choice**: `persistentRecordTransformer.newConverter()` builds a fresh `AttributeConverter` per row (`persistent_record.go:30-34`, reached per row via `entity_crud_service.go:130`, `entity_batch_service.go:242`, `entity_manager.go:251`), so converter-scoped state would dedupe nothing. The code comment records this why. One nuance: the write-path `transformer` (`transformer.go:36`) *does* hold a long-lived converter, so the "dedupe nothing" rationale is strictly true only on the read path that motivates the memo — acceptable.
2. **No-eviction memo** — "bounded by registered schemas × dropped/retired sets" is a practical, not strict, bound: #294 explicitly supports attribute re-add/remove, so a schema's skipped set is not monotonic and a long-lived process could accumulate keys over churn. Practically negligible; residual risk, not a defect.
3. **Test global-state coupling** (`resetSkippedAttrLogMemo` wipes and repopulates the process memo with no restore) — acknowledged in the comment that names the sibling test feeding it (`TestAttributeConverterFromEAVRecords_SkipsUnknownAttributeIDs`); safe today because no `t.Parallel` exists in `internal/transform`.

iceboundrock:
## Review: assessment (#343)

**Ready to merge.**

The change does exactly what issue #343 and the plan specify — demote *and* memoize, with the anti-counter-argument verified rather than asserted:

- The **package-state is necessary** claim is validated against the actual call sites (per-row `newConverter` on the read path; fresh converter per `FromEAVRecords` call in the tests).
- The **key encoding is injective**, so two distinct schema-shapes can never share a memo slot and silently silence a first emission.
- **Exact-once semantics** (single winner under a mutex, emission outside the lock) and **once-per-shape across converter churn** are both pinned by tests, including an 8-way concurrent case that runs under CI's `-race`.
- The **memo is load-bearing proved, not claimed**: disconnecting the dedupe turns *both* new tests red (reproduced in this review; see *verification*).
- The **doc sentence** re-reads the shipped code (grep-back rule honored) and records an operator-actionable contract (`skippedAttrLogMessage` const, stable fields `schemaID`, `attrIDs`).

No findings block this PR. Nothing here needs a change before merge.

iceboundrock:
## Review: observations (#343)

Deferred / minor observations — none actionable:

1. **Mark-before-emit:** the memo marks a shape "seen" even if the emission is filtered later (e.g. a Warn-level logger reconfigured down to Info). Theoretical only here — logger level is fixed at startup.
2. **Memo is one-way (never evicts):** absence of a later skip does not "clear" the once-per-process status. The "once per process" phrasing in doc/code states the eviction-free contract accurately.
3. **`exampleRowID` under concurrency:** the logged row is the memo-winner's row, not necessarily temporally first. It is a lookup pointer either way.
4. **P3 test gaps (ship-as-is):** `exampleRowID` field value is asserted nowhere (level/count/`attrIDs`/`schemaID` are); no negative-attrID or empty-set key boundary test (unreachable, but a ~6-line addition would pin the injectivity argument in-register for the next touch of this file).
5. **Shared process, multiple environments:** an E2E harness spinning several isolated environments in one process shares the memo; a second environment reusing (schemaID, set) stays silent. Log-only impact.
6. **Standing note:** nothing pins the package-wide no-`t.Parallel` invariant that protects the two `zap.ReplaceGlobals` tests and `resetSkippedAttrLogMemo`. If `internal/transform` ever adopts `t.Parallel`, these tests (and the reset helper) must stay serial.

iceboundrock:
## Review: verification (#343)

Gates run on worktree `../forma-343` @ `2e4d9fb`, mirroring the Makefile env (`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false`):

| Check | Command | Result |
|---|---|---|
| Target tests | `go test ./internal/transform -run 'TestSkippedAttrLog\|TestAttributeConverterFromEAVRecords' -v` | all PASS, incl. `TestAttributeConverterFromEAVRecords_SkipsUnknownAttributeIDs` |
| Race | `go test -race ./internal/transform -run 'TestSkippedAttrLog' -count=1` | ok |
| Full unit suite (make test body) | `go test . ./cdc ./cmd/... ./factory ./internal/...` | all ok |
| gofmt | `gofmt -l internal/transform/` | clean |
| File sizes | `wc -l` | `attribute_converter.go` 490 (was 491), `skipped_attr_log.go` 55, `skipped_attr_log_test.go` 117 |
| **Mutation check (reproduced this review)** | `if dup { return }` → `_ = dup` | both new tests RED — `TestSkippedAttrLogInfoOncePerShape` got 3 entries, `TestSkippedAttrLogConcurrentCallsEmitOnce` got 8, instead of 1 each; reverted after |

The reproduced mutation red confirms the memo is load-bearing: the tests catch a silently disconnected dedupe, matching the development-time record in the PR body. `golangci-lint` (pinned v1.64.8) was not re-run in this review — the binary is absent from PATH here — but the PR records `make lint` exit 0 at development time, and tooling-enforced rules are out of scope for this review axis.

iceboundrock:
## Review: summary (#343)

Second whole-branch review of #424: **Standards — 0 findings; Spec — 0 findings.** No critical, important, or minor defects; no spec gaps, no scope creep, no wrong implementations.

Verified rather than assumed: the memo's package-level placement argument matches the real call sites; the schema-shape key is injective; lock discipline yields exactly-once emission under 8-way parallelism; both requirements (Info level + once-per-process-per-shape) are pinned by tests; the mutation check was reproduced and turned both tests red. The doc sentence states the shipped contract accurately.

**Verdict: Ready to merge.**

PR #425, opened 2026-08-19
fix: #331 evict and close superseded Runner cache entries on credential rotation

### LINKED ISSUE #331: cdc: Runner S3-runtime/exporter caches never evict — each credential rotation strands an open DuckDB instance

iceboundrock:
Closes #331.

## What

`Runner.s3Runtimes` and `Runner.duckExporters` were append-only: every
credential rotation minted a new cache key and stranded the old entry —
one open DuckDB instance per rotation — until process exit (#329 review
finding O2). Both caches now key on the non-credential config group and
hold exactly one entry per group:

- **s3Runtimes**: rotation replaces the entry; no close needed (s3.Client
  holds no OS resources), GC reclaims it.
- **duckExporters**: rotation replaces the entry and closes the superseded
  DuckDB handle — immediately when idle, else when its last in-flight
  RunOnce frame releases it (refcount). A held exporter is never closed
  underneath its holder.
- **Runner.Close** keeps hard-shutdown semantics: everything closes at
  once; a straggling release after Close is an idempotent sql.DB re-Close.

Adjudication (issue #331 listed three un-adjudicated directions): user
chose evict-on-supersede with refcounting over bounded LRU and
accept-and-document — see the plan comment on #331.

Accepted trade-off (documented on Runner): two configs sharing a group but
alternating different static credential triples rebuild per alternation —
a rebuild cost, never a correctness issue; rotation itself never alternates.

## Tests

- Red anchor: rotation closes the superseded idle exporter and leaves one
  map entry (`TestRunnerDuckExporterRotationClosesSuperseded`).
- Refcount: in-flight supersession defers close to the last release; the
  live entry's release does not close it.
- Leak assertion from the issue: 5 rotations leave exactly 1 live handle.
- Boundary: distinct endpoints/groups coexist, neither evicted.
- `-race` stress: 8 goroutines alternating two triples; held exporters
  always answer Ping; storm settles to one live handle.
- Existing #329 cache-key tests re-purposed to rotation-replacement tests
  (assertions extended, not weakened); #302/#326 provider-resolution tests
  untouched and green.

Gates: `make fmt-check`, `make lint` (pinned v1.64.8), `make test` all
green; full `internal/cdc` package green under `-race`.

### DIFF
--- a/cdc/flush.go
+++ b/cdc/flush.go
@@ -17,7 +17,11 @@ type S3ObjectClient = internalcdc.S3ObjectClient
 // S3FullClient extends S3ObjectClient with manifest read/write operations.
 type S3FullClient = internalcdc.S3FullClient
 
-// FlushRunner reuses AWS/S3/DuckDB initialization across multiple flush passes.
+// FlushRunner reuses AWS/S3/DuckDB initialization across multiple flush
+// passes. Safe for long-lived use under credential rotation: superseded
+// cached exporters are evicted and closed once idle (#331), so rotations do
+// not accumulate open DuckDB instances. Call Close at shutdown to release
+// whatever is still cached.
 type FlushRunner struct {
 	runner *internalcdc.Runner
 }
--- a/internal/cdc/runner.go
+++ b/internal/cdc/runner.go
@@ -30,7 +30,7 @@ type Runner struct {
 	logger        *zap.Logger
 	mu            sync.Mutex
 	s3Runtimes    map[s3RuntimeGroupKey]*cachedS3Runtime
-	duckExporters map[duckExporterKey]*DuckExporter
+	duckExporters map[duckExporterGroupKey]*cachedDuckExporter
 }
 
 type cachedS3Runtime struct {
@@ -64,20 +64,50 @@ type s3RuntimeGroupKey struct {
 	usePath  bool
 }
 
-type duckExporterKey struct {
-	dbPath      string
-	threads     int
-	memLimit    string
-	region      string
-	endpoint    string
-	useSSL      bool
-	usePath     bool
-	accessKeyID string
-	// Same rule as s3RuntimeKey: the exporter bakes the token into
-	// SET s3_session_token at construction, so a key without it returns a
-	// stale-token exporter after a rotation (#329).
+// duckExporterGroupKey identifies a cache slot by everything *except* the
+// credential triple. Credentials live on the cached entry and are compared on
+// lookup, so a rotation replaces the slot's occupant instead of stranding it
+// under a sibling key for process lifetime (#331).
+type duckExporterGroupKey struct {
+	dbPath   string
+	threads  int
+	memLimit string
+	// The raw cfg region, not s3Runtime.region: the exporter is configured
+	// from cfg alone, and an empty cfg.S3Region suppresses SET s3_region
+	// entirely. Keying on the chain-resolved region claims a distinction
+	// the exporter never makes, so two runs producing byte-identical
+	// exporters would miss the cache (#329). It also cut the other way: an
+	// empty-region cfg whose chain resolved to some region could collide
+	// with an explicitly-configured cfg of that same region, two configs
+	// that build *different* exporters, so the second silently reused the
+	// first's — a latent wrong cache hit the raw cfg region rules out.
+	region   string
+	endpoint string
+	useSSL   bool
+	usePath  bool
+}
+
+// cachedDuckExporter is one cache slot's occupant. refs counts in-flight
+// RunOnce frames currently holding the exporter; doomed marks an entry that
+// has been superseded (or hard-closed via Close) and must not be handed out
+// again. A doomed entry's DB closes when refs reaches zero — immediately at
+// supersession if idle, else at the last release (#331).
+type cachedDuckExporter struct {
+	exporter        *DuckExporter
+	accessKeyID     string
 	secretAccessKey string
 	sessionToken    string
+	refs            int
+	doomed          bool
+}
+
+// credsEqual: same rule as cachedS3Runtime.credsEqual — the exporter bakes
+// the token into SET s3_session_token at construction, so an entry built from
+// a different triple is a stale-token artifact, never a cache hit (#329).
+func (c *cachedDuckExporter) credsEqual(accessKeyID, secretAccessKey, sessionToken string) bool {
+	return c.accessKeyID == accessKeyID &&
+		c.secretAccessKey == secretAccessKey &&
+		c.sessionToken == sessionToken
 }
 
 func NewRunner(logger *zap.Logger) *Runner {
@@ -87,7 +117,7 @@ func NewRunner(logger *zap.Logger) *Runner {
 	return &Runner{
 		logger:        logger,
 		s3Runtimes:    make(map[s3RuntimeGroupKey]*cachedS3Runtime),
-		duckExporters: make(map[duckExporterKey]*DuckExporter),
+		duckExporters: make(map[duckExporterGroupKey]*cachedDuckExporter),
 	}
 }
 
@@ -97,20 +127,23 @@ func (r *Runner) Close() error {
 	}
 
 	r.mu.Lock()
-	exporters := make([]*DuckExporter, 0, len(r.duckExporters))
-	for _, exporter := range r.duckExporters {
-		exporters = append(exporters, exporter)
+	entries := make([]*cachedDuckExporter, 0, len(r.duckExporters))
+	for _, entry := range r.duckExporters {
+		// Hard shutdown: doomed here + idempotent sql.DB.Close means a release
+		// arriving after Close is harmless.
+		entry.doomed = true
+		entries = append(entries, entry)
 	}
 	r.s3Runtimes = make(map[s3RuntimeGroupKey]*cachedS3Runtime)
-	r.duckExporters = make(map[duckExporterKey]*DuckExporter)
+	r.duckExporters = make(map[duckExporterGroupKey]*cachedDuckExporter)
 	r.mu.Unlock()
 
 	var errs []error
-	for _, exporter := range exporters {
-		if exporter == nil || exporter.DB == nil {
+	for _, entry := range entries {
+		if entry.exporter == nil || entry.exporter.DB == nil {
 			continue
 		}
-		if err := exporter.DB.Close(); err != nil {
+		if err := entry.exporter.DB.Close(); err != nil {
 			errs = append(errs, err)
 		}
 	}
@@ -144,12 +177,13 @@ func (r *Runner) RunOnce(ctx context.Context, cfg CDCConfig, s3Client S3ObjectCl
 	}
 	defer db.Close()
 
-	duck, err := r.getOrCreateDuckExporter(ctx, cfg, s3Runtime)
+	duck, releaseDuck, err := r.getOrCreateDuckExporter(ctx, cfg, s3Runtime)
 	if err != nil {
 		// getOrCreateDuckExporter carries the "new duck exporter:" prefix; this
 		// layer adds the run-level step so the two never duplicate.
 		return fmt.Errorf("prepare duck exporter: %w", err)
 	}
+	defer releaseDuck()
 
 	flushCtx := newSchemaFlushContext(flushContextParams{
 		cfg:            cfg,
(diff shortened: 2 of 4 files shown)


iceboundrock:
## Findings

Independent second review of the full branch (merge-base `4ffdb1d` → `e953496`). No Critical or Important defects found. The two readonly pointers below are **Minor test-hardening gaps**, not bugs — both claims hold by code reasoning and/or empirical verification, but neither is directly pinned by a test.

**Minor 1 — "doomed-but-in-flight entry is never handed out" is not directly asserted.**
`TestRunnerDuckExporterInFlightSupersedeDefersClose` proves an in-flight superseded exporter survives to its last release, but never asserts that a *subsequent* acquire with the superseded (now-stale) triple rebuilds instead of receiving the doomed entry. The design guarantees this: supersession replaces the map value under `r.mu` (runner.go:315-323), so a doomed entry is never the map value outside `Close`'s locked iteration — `credsEqual` can therefore never match a doomed entry. A two-line extension to the in-flight test (acquire again with `old` after `exp2` is cached; assert `NotSame(exp1, expired)` and map length stays 1) would pin this against any future refactor of the refcount protocol.

**Minor 2 — the "straggling release after Close is an idempotent re-Close" claim is not directly exercised.**
`TestRunnerClose_ClearsCachesAndClosesExporters` seeds `refs: 1` (an in-flight hold at shutdown) but never invokes a release hook after `Close`. The second `runner.Close()` returns nil only because the maps are already empty (it iterates zero entries), so the test does not actually drive the double-`Close` the doc comments claim is harmless. I verified empirically that `sql.DB.Close()` on go-duckdb `:memory:` is idempotent — second call returns nil, `PingContext` after close returns `"sql: database is closed"` — so the claim holds; pinning it by calling the straggling release post-`Close` (asserting no error, no panic, DB still closed) would remove the reliance on (correct) code-reasoning alone.

iceboundrock:
## Spec

Fidelity to issue #331 (and the persistence of #329's cache-key contract).

- **Adjudication honored.** #331 listed three un-adjudicated directions; the plan comment on #331 records the user's choice of direction 1 (evict-on-supersede + refcount) over bounded LRU and accept-and-document. This PR implements exactly that, and the rejected directions are not half-implemented.
- **s3Runtimes** — evict-on-supersede via bare map replacement. The issue's "needs care re: in-flight use — refcount or drain" is answered with pointer semantics: an in-flight `RunOnce` holds the runtime struct (not the map slot), nothing is closed on replacement, so GC reclaims after the frame. Sound; no refcount needed (runner.go:268-270).
- **duckExporters** — evict-on-supersede + refcounted deferred close. Covers precisely the in-flight window the issue flagged (close immediately when idle, else at last release). Superseded-credential exporters are never handed out (see Findings Minor 1).
- **Runner.Close** keeps hard-shutdown semantics (close everything at once, in-flight or not) — unchanged behavior, now documented as doomed + idempotent re-Close (runner.go:140-142).
- **Issue leak arithmetic** ("~24 open DuckDB instances/day under hourly STS rotation") maps directly to `TestRunnerDuckExporterRotationLeavesSingleLiveHandle` (5 rotations → exactly 1 live handle).
- **Accepted trade-off** (two configs sharing a group but alternating different static triples rebuild per alternation) is documented on both `Runner` (runner.go:34-36) and `FlushRunner` (cdc/flush.go:20-24), as the plan required.
- **No scope creep.** Diff is confined to `runner.go`, `runner_test.go`, the new `runner_eviction_test.go`, and a doc-comment-only change to `cdc/flush.go`. The plan's constraint that the package-level `RunOnce` in `flusher.go` (fresh exporter per call) stays out of scope is honored.

Verdict: implemented faithfully; no missing requirement, no wrong behavior found. `#329`'s required rebuild-on-token-rotation is preserved (the same-credentials build-race loser is closed, runner.go:306-312, and the rotated-triple acquire still misses the cache).

iceboundrock:
## Standards

Against `coding-standard.md` and the AGENTS.md rules. No hard violations found.

- **File/function size** — `runner.go` 362 lines, `runner_test.go` 362, `runner_eviction_test.go` 235 (all ≤500); `getOrCreateDuckExporter` ~56 lines, `getOrCreateS3Runtime` ~54 (all ≤100). Within budget with headroom — the plan's escape hatch (split into `runner_duck_cache.go`) was not needed.
- **Error wrapping** — every wrap carries context (`new duck exporter: %w`, `prepare duck exporter: %w`, `prepare s3 runtime: %w`, `load aws config: %w`); no bare `return err`; no sentinel issues (n/a here). `closeExporterDB` logs instead of returning a close error — a *sanctioned* deviation, explicitly justified in its doc comment (failing the new run because an old handle failed to close would be wrong, runner.go:352-354).
- **Structure** — early returns preserved; hot paths stay at ≤3 levels of nesting. The three long key/entry comments are justified: each carries load-bearing rationale (#329's token-staleness rule, the raw-vs-chain region decision, the group vs credential distinction).
- **Doc-comment gofmt trap (#276)** — ASCII quotes/periods only in the new prose; `make fmt-check` clean.
- **Seam-stubbing discipline** — new tests correctly avoid `t.Parallel()` (process-global seam, `runner_test.go:17-19`); the stub comment restates it. The `-race` storm uses `t.Errorf`, not testify, inside goroutines (no `FailNow` off the test goroutine), and the stub returns open errors instead of `require`-ing them — both the mid-execution amendments recorded on #331.
- **Commit hygiene** — 6 commits, each `type: #331 <subject>` with topic-scoped content (fix×2, test×1, docs×3); the two `docs:` commits touch only comments/log strings; multi-line bodies with `Co-Authored-By:` trailer, per the plan's prepare-commit-msg constraint.
- **Minor DRY (non-blocking)** — `credsEqual` is structurally identical on `cachedS3Runtime` and `cachedDuckExporter`. Acceptable given the two structs have different lifetimes and eviction semantics; a shared helper would cost more in comment conflation than it saves.

iceboundrock:
## Assessment

The design is sound. The three invariants the PR depends on all hold by construction:

1. **A held exporter is never closed underneath its holder** — `closeExporterDB` is reachable only when `doomed && refs == 0` (runner.go:319-320, 339-348); the immediate-close supersede path guards `superseded.refs == 0` under the same lock (runner.go:327).
2. **A superseded entry is never handed out again** — supersession atomically replaces the map value under `r.mu` (runner.go:315-323), so `credsEqual` can't match a doomed entry; only `Close` dooms in place, inside its own locked iteration.
3. **Close hard-shutdown + straggling release is harmless** — the release path's re-`Close` is idempotent on go-duckdb (empirically verified: second `sql.DB.Close()` returns nil).

I scrutinized the one superficially alarming corner — the comment "a slower build can install an older triple over a fresher one" (runner.go:313-314). Under concurrent acquisition straddling a rotation boundary this can evict a *fresh* idle exporter in favor of a *stale* one. I traced it end-to-end and it **cannot regress #329**: an exporter is only ever returned to an acquirer whose own resolved triple passed `credsEqual`, so a stale map occupant is never actually handed to a fresh-credential caller — the next acquire simply misses and rebuilds. The cost is bounded rebuild churn correctable by the next acquire, which is exactly the documented, user-accepted alternation trade-off (and unobserved in the actual rotation pattern, which is monotonic).

Both Minor findings are test-hardening suggestions, not defects. Independent verdict: **ready to merge**. This corroborates the pre-PR review's "0 Critical/Important findings" claim rather than relying on it.

(Not merging — 请不要自动合并 PR.)

iceboundrock:
## Observations

- **S3 in-flight semantics are intentionally per-frame.** A `RunOnce` frame that resolved an *old* triple keeps using that runtime (client + credProvider) for its entire frame even if rotation lands mid-frame — correct, because the frame already committed to its signing identity for both the S3 client and the Postgres connection (`setupPostgresConnection` uses `s3Runtime.credProvider`). The next frame re-resolves. This is implicit in the pointer-based design (no close on the S3 path today); only worth re-examining if a future change ever adds OS resources to the runtime.
- **Use-after-Close is not defended.** A builder mid-flight when `Close` resets the maps installs its fresh exporter into the empty map (runner.go:315) and it is never closed until a later rotation or another `Close`. Documented hard-shutdown assumes no post-Close use. A `closed` flag rejecting new acquires after `Close` would harden this if it ever matters; out of scope today, per contract.
- **The storm test pin is the leak, not the survivor.** `TestRunnerDuckExporterConcurrentRotationIsSafe` correctly asserts one live handle but, under last-writer-wins, cannot assert *which* triple survives — so it pins leak-freedom, not freshness. If "the freshest token must win" ever becomes a requirement, the test would need to assert the survivor's `sessionToken`; today that would be flaky by construction.
- **Comment accuracy** — the adapted `TestRunnerDuckExporterCacheKeyIgnoresChainResolvedRegion` comment ("pointlessly churning (#331)") and the re-purposed rotation tests read accurately and do not overclaim; the "Last writer wins" honesty on runner.go:313-314 is the right call for this subtlety.
- **Coverage map against the plan** — the plan's Task 1/2/3/4 deliverables all landed: red-anchor, refcount deferral, leak assertion, distinct-group coexistence, `-race` storm, doc comments on both `Runner` and `FlushRunner`. The only plan items not converted to standalone tests are the two in Findings, which is why they read as gaps rather than regressions.

iceboundrock:
## Verification

Re-ran the gates independently in the worktree (`../forma-331`, branch `fix/331-runner-cache-eviction`, base `4ffdb1d`):

- `make test` — full unit suite green (all packages).
- `make fmt-check` — clean.
- `make lint` (golangci-lint pinned v1.64.8) — clean, exit 0.
- `go test ./internal/cdc -race -count=1` — green.
- `go test ./internal/cdc -run TestRunnerDuckExporterConcurrentRotationIsSafe -race -count=5 -v` — PASS ×5, no race reports.
- **Empirical** — `sql.DB.Close()` on go-duckdb `:memory:` is idempotent: first and second close both return nil; `PingContext` after close returns `"sql: database is closed"`. This independently confirms the tests' `require.Error` ping assertions and the PR's "idempotent sql.DB re-Close" claim (which, as noted in Findings Minor 2, the existing test does not itself exercise on the straggling-release path).
- **Diff/commit integrity** — `git diff origin/main...HEAD` is exactly the PR's 4 files / 432+/97−; the 6-commit structure (fix×2, test, docs×3) matches the plan's task breakdown, and the last three `docs:` commits are comment/log-string-only.
- Review-path correctness cards I traced (all under `-race` reasoning, not just reading): same-credentials build-race → redundant build closed; idle supersede → immediate close; in-flight supersede → deferred to last release; live-entry release → no close; post-Close straggling release → idempotent double-close; stale-triple install over fresher one → next acquire rebuilds, stale-triple never handed out.

Not run: the Docker e2e suites (`internal/e2e_harness`, `tests/e2e`) — out of scope for this change (unit-isolated, `internal/cdc` only), and CI covers them. Note on the production harness (`internal/e2e_harness/production/cdc.go:83-84`): it creates a fresh Runner per flush and `Close`s it via defer, so the refcount/eviction surface is exercised sequentially there, not under rotation.

iceboundrock:
## Summary

**Verdict: Approve — ready to merge. No blocking findings.**

- **Findings:** 0 Critical / 0 Important; 2 **Minor** test-hardening gaps (no test pins "doomed entry is never handed out"; no test drives the post-`Close` straggling release). Neither is a bug — both invariants hold by construction and/or empirical verification.
- **Spec:** faithfully implements #331's chosen direction (evict-on-supersede + refcount), preserves #329's rebuild-on-rotation, keeps `Close` hard-shutdown, and documents the accepted alternation trade-off. No scope creep.
- **Standards:** sizes, error wrapping, nesting, seam-stubbing, doc-comment hygiene (#276), and commit hygiene all clean under `coding-standard.md` + AGENTS.md.
- **Assessment:** the two highest-risk invariants — held exporter never closed underneath its holder, superseded entry never handed out again — hold by construction; the "older triple can win under concurrency" corner is bounded churn, not a #329 regression.
- **Observations:** per-frame S3 semantics are correct; use-after-`Close` of a Runner is unguarded (documented contract, optional hardening); the storm test pins leak-freedom but not survivor freshness.
- **Verification:** `make test` / `make fmt-check` / `make lint` green; `./internal/cdc` green under `-race`; storm test PASS ×5; double-`sql.DB.Close()` idempotence confirmed empirically; diff and commit structure match the plan.

Per repo rules the PR was not merged (请不要自动合并). After merge: clean up `../forma-331` and the local branch, fast-forward `main`, and add the closing comment on #331 linking this PR.

PR #426, opened 2026-08-19
test: close false-negative paths in TestInitSchemaScopedIsolation (#248)

### LINKED ISSUE #248: [E2E] Close false-negative paths in TestInitSchemaScopedIsolation (#186 follow-up)
Follow-up to #186 / PR #247, from the post-merge review ([comment](https://github.com/Lychee-Technology/forma/pull/247#issuecomment-4966536457)). Two false-negative paths keep the test from proving the claims it makes, plus two helper renames.

## 1. Sibling isolation is blind to overwrites (P2)

`multi_schema_isolation_e2e_test.go` (`TestInitSchemaScopedIsolation`) proves "scoped init never touches the sibling schema" via:

- `InitReport.NewObjects` — a **key diff** (`diffKeys`, production/cdc.go): an overwrite of an existing sibling object reuses its deterministic `min_max` key and never appears;
- `buildBasePaths` set comparison — a sibling manifest rewritten with the **same paths** (different stats/etag/order) also passes.

**Fix:** snapshot the sibling's observable state before the target-scoped init and compare after:

- S3 objects under the sibling partition with key + size + ETag + LastModified — reuse `snapshotS3Inventory` (`dryrun_immutability_e2e_test.go`, same package; filter by `{S3Prefix}/{siblingID}/`);
- sibling manifest raw bytes + ETag — reuse `snapshotManifest` (same file, already per-schema).

With that, the no-write contract fails on overwrite and same-path mutation, not just create/delete. (The snapshot helpers may be worth hoisting out of the dryrun test file once they gain a second consumer.)

## 2. Base-tier visibility assertion can pass without reading Parquet (P2)

The final oracle check clears `change_log` and asserts `len(result.Records) == 5`, but a routing regression to the PostgreSQL-only path would still read `entity_main` and return 5 rows.

**Fix:** assert `result.Plan.Routing.UseDuckDB` for both schemas before accepting the count as cold-tier coverage (precedent: `assertFederatedRowCount`, `cdc_concurrent_flush_e2e_test.go`).

Note: #176's `TestInitRerunIdempotency` / `TestInitRerunAfterChangesReconcilesManifest` use the same clear-and-count pattern without a routing assertion — same weakness, optional to strengthen in the same PR.

## 3. Verb-first helper renames (standards)

`coding-standard.md` §4 requires verb-first private helpers; `cdc_fault_helpers_e2e_test.go` added noun-first `schemaKeyPrefix` and `finalsForSchema`. Rename (e.g. `buildSchemaKeyPrefix`, `filterFinalsForSchema`) and update the call sites in the two #186 test files.

## Acceptance

- Overwriting a sibling base object or rewriting its manifest in place (fault-injected or hand-mutated) turns the test red.
- The final parity check fails if the query does not route to DuckDB.
- New/renamed helpers are verb-first; suite, `make lint`, `make test` green.

iceboundrock:
Closes #248

Test-only hardening of the production e2e harness, per the plan on the issue ([plan](https://github.com/Lychee-Technology/forma/issues/248#issuecomment-5347935727), [execution evidence](https://github.com/Lychee-Technology/forma/issues/248#issuecomment-5348124521)).

## What changed

1. **Sibling snapshot closes the overwrite blind spot (issue item 1).** `TestInitSchemaScopedIsolation` proved "scoped init never touches the sibling" only via the `InitReport.NewObjects` key diff and a `buildBasePaths` set comparison — both blind to in-place overwrites (deterministic `min_max` keys) and to a manifest rewritten with the same paths. The test now captures the sibling's full observable S3 surface (every partition object + the manifest object, each with size/ETag/LastModified, plus manifest raw bytes + ETag) before the target-scoped rerun and requires exact identity after. `snapshotS3Inventory`/`snapshotManifest`/`s3ObjectStat` are hoisted from the dry-run test into `state_snapshot_helpers_e2e_test.go`, with `snapshotS3Inventory` gaining a prefix parameter (both pre-existing consumers — `captureState`, `compaction_e2e_test.go` — pass the old hardcoded prefix verbatim).
2. **Routing assertion guards the base-tier oracle (issue item 2).** The final clear-and-count check now goes through `assertFederatedRowCount`, which additionally asserts `result.Plan.Routing.UseDuckDB` — a PostgreSQL-only routing regression would still count 5 `entity_main` rows. The issue's optional note is included: the two #176 rerun tests (`TestInitRerunIdempotency`, `TestInitRerunAfterChangesReconcilesManifest`) got the same strengthening.
3. **Verb-first renames (issue item 3).** `schemaKeyPrefix` → `buildSchemaKeyPrefix`, `finalsForSchema` → `filterFinalsForSchema`, swept across the whole package (call sites now include the #347-era consumers, not just the two #186 files).

## Mutation verification (acceptance criteria)

- **Sibling overwrite turns the test red.** The plan's original probe (garbage-bytes PutObject) was replaced by ruling — it would corrupt the base tier and fail the final oracle for the wrong reason. The shipped probe was the exact regression class: `sleep 1.1s` + `env.RunInit(ctx, second)` (in-place sibling re-export, valid data, same keys). Old test with probe: **PASS** (false negative proven). New assertions with probe: **FAIL**, both legs firing — `schema manifest rewritten in place` (ETag + raw bytes, version 1→2) and `schema objects mutated` (parquet caught by LastModified only). Probe removed: green.
- **Empirical finding en route:** the init re-export is **byte-deterministic** (same size, same ETag) — the plan's comment rationale claiming a re-export "also changes bytes" was falsified by measurement and the committed comments state the observed behavior instead: the object leg may rest on LastModified (1s granularity) alone; the manifest leg (version increments on every save through the manifest store) is the unconditional detector.
- **Non-DuckDB routing turns the parity check red.** Temporary `PreferHot: true` (forces the PostgreSQL-only path per `EvaluateRoutingPolicy`) → **FAIL** with `did not route to duckdb`; reverted → green.

## Verification

- `make fmt-check`, `make lint`, `make test`, `make test-e2e-production` all green at the review head (5381131; 529fb38 on top is a comment-only polish, fmt+lint re-verified) (no flakes, no re-runs needed).
- Whole-branch review (0 critical/important findings): deferred follow-ups — manifest-key derivation is duplicated in 4 places (two pre-existing) and could route through `manifest.PathResolver`; `assertFederatedRowCount` (pre-existing) lacks a nil-`Plan` guard; `snapshotManifest`'s "seed must have created it" hint misattributes a hypothetical deletion regression. Follow-up: #427.

### DIFF
--- a/internal/e2e_harness/production/cdc_fault_helpers_e2e_test.go
+++ b/internal/e2e_harness/production/cdc_fault_helpers_e2e_test.go
@@ -196,22 +196,22 @@ func assertSameRowIDs(t *testing.T, label string, got, want map[string]bool) {
 	}
 }
 
-// schemaKeyPrefix returns the Env-scoped S3 key prefix of one schema's
+// buildSchemaKeyPrefix returns the Env-scoped S3 key prefix of one schema's
 // parquet partition (e.g. "e2e/<run>/env3/22/"). Matching on the full prefix
 // rather than a bare "/22/" substring keeps the filter unambiguous no matter
 // what the run ID or env sequence look like.
-func schemaKeyPrefix(env *Env, schema SchemaRef) string {
+func buildSchemaKeyPrefix(env *Env, schema SchemaRef) string {
 	return fmt.Sprintf("%s/%d/", env.S3Prefix, schema.ID)
 }
 
-// finalsForSchema filters keys down to one schema's final parquet objects
+// filterFinalsForSchema filters keys down to one schema's final parquet objects
 // (multi-schema passes legitimately mix partitions in one report, so the
 // global splitKeys alone cannot attribute objects to a schema).
-func finalsForSchema(env *Env, keys []string, schema SchemaRef) []string {
+func filterFinalsForSchema(env *Env, keys []string, schema SchemaRef) []string {
 	finals, _ := splitKeys(keys)
 	var got []string
 	for _, k := range finals {
-		if strings.HasPrefix(k, schemaKeyPrefix(env, schema)) {
+		if strings.HasPrefix(k, buildSchemaKeyPrefix(env, schema)) {
 			got = append(got, k)
 		}
 	}
@@ -225,12 +225,12 @@ func finalsForSchema(env *Env, keys []string, schema SchemaRef) []string {
 // legitimately create objects in the same report.
 func assertSchemaUntouchedByFault(ctx context.Context, t *testing.T, env *Env, report *FlushReport, schema SchemaRef, wantDirty int) {
 	t.Helper()
-	if finals := finalsForSchema(env, report.NewObjects, schema); len(finals) != 0 {
+	if finals := filterFinalsForSchema(env, report.NewObjects, schema); len(finals) != 0 {
 		t.Errorf("schema %d must promote no final parquet after its fault, got %v", schema.ID, finals)
 	}
 	_, tmps := splitKeys(report.NewObjects)
 	for _, k := range tmps {
-		if strings.HasPrefix(k, schemaKeyPrefix(env, schema)) {
+		if strings.HasPrefix(k, buildSchemaKeyPrefix(env, schema)) {
 			t.Errorf("schema %d must leave no tmp residue after its copy fault (#226), got %s", schema.ID, k)
 		}
 	}
@@ -256,7 +256,7 @@ func assertSchemaFullyFlushed(ctx context.Context, t *testing.T, env *Env, newKe
 		t.Fatalf("schema %d must be fully flushed: flushed=%v dirty=%v, want %d flushed",
 			schema.ID, flushed, dirty, wantRows)
 	}
-	finals := finalsForSchema(env, newKeys, schema)
+	finals := filterFinalsForSchema(env, newKeys, schema)
 	if len(finals) != 1 {
 		t.Fatalf("schema %d must promote exactly one final delta, got %v", schema.ID, finals)
 	}
--- a/internal/e2e_harness/production/compaction_e2e_test.go
+++ b/internal/e2e_harness/production/compaction_e2e_test.go
@@ -153,7 +153,7 @@ func TestCompactionPromotionEquivalence(t *testing.T) {
 	mBefore := loadSchemaManifest(ctx, t, env, wide)
 	assertDeltaSizesPopulated(t, mBefore)
 	baseBefore, deltaBefore := countTier(mBefore, "base"), countTier(mBefore, "delta")
-	invBefore := snapshotS3Inventory(t, ctx, env)
+	invBefore := snapshotS3Inventory(t, ctx, env, env.S3Prefix+"/")
 
 	result := assertCompactionEquivalence(ctx, t, env, wide,
 		compactionEquivalenceQueries(wide), CompactionOverrides{TargetBaseSizeBytes: 1}, "promotion")
@@ -180,7 +180,7 @@ func TestCompactionPromotionEquivalence(t *testing.T) {
 
 	// Promotion is a manifest-only relabel: zero parquet objects created,
 	// deleted, or modified.
-	assertParquetInventoryUnchanged(t, "promotion", invBefore, snapshotS3Inventory(t, ctx, env))
+	assertParquetInventoryUnchanged(t, "promotion", invBefore, snapshotS3Inventory(t, ctx, env, env.S3Prefix+"/"))
 
 	// The relabeled entries still serve reads: the post-compaction query must
 	// route through DuckDB over the manifest-listed objects.
--- a/internal/e2e_harness/production/concurrent_flush_multi_schema_e2e_test.go
+++ b/internal/e2e_harness/production/concurrent_flush_multi_schema_e2e_test.go
@@ -85,7 +85,7 @@ func TestConcurrentFlushDifferentSchemas(t *testing.T) {
 		t.Fatalf("runner2 must not touch the locked schema %d: flushed=%v dirty=%v",
 			simple.ID, flushedA, dirtyA)
 	}
-	if finals := finalsForSchema(env, r2.NewObjects, simple); len(finals) != 0 {
+	if finals := filterFinalsForSchema(env, r2.NewObjects, simple); len(finals) != 0 {
 		t.Errorf("runner2 must not promote finals for the locked schema %d, got %v", simple.ID, finals)
 	}
 
--- a/internal/e2e_harness/production/dryrun_immutability_e2e_test.go
+++ b/internal/e2e_harness/production/dryrun_immutability_e2e_test.go
@@ -9,10 +9,6 @@ import (
 	"fmt"
 	"reflect"
 	"testing"
-	"time"
-
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/service/s3"
 
 	"github.com/lychee-technology/forma/internal/manifest"
 )
@@ -85,12 +81,6 @@ type changeLogRow struct {
 	DeletedAt int64 // 0 when NULL
 }
 
-type s3ObjectStat struct {
-	Size         int64
-	ETag         string
-	LastModified time.Time
-}
-
 // stateSnapshot captures the three mutation surfaces #180 guards.
 type stateSnapshot struct {
 	changeLog    map[string]changeLogRow
@@ -104,7 +94,7 @@ func captureState(t *testing.T, ctx context.Context, env *Env, schema SchemaRef)
 	raw, etag := snapshotManifest(t, ctx, env, schema)
 	return stateSnapshot{
 		changeLog:    snapshotChangeLog(t, ctx, env),
-		s3:           snapshotS3Inventory(t, ctx, env),
+		s3:           snapshotS3Inventory(t, ctx, env, env.S3Prefix+"/"),
 		manifestRaw:  raw,
 		manifestETag: etag,
 	}
@@ -134,51 +124,6 @@ func snapshotChangeLog(t *testing.T, ctx context.Context, env *Env) map[string]c
 		t.Fatalf("iterate change_log rows: %v", err)
 	}
 	return snap
-}
-
-func snapshotS3Inventory(t *testing.T, ctx context.Context, env *Env) map[string]s3ObjectStat {
-	t.Helper()
-	inv := make(map[string]s3ObjectStat)
-	var token *string
-	for {
-		out, err := env.Cluster.S3.ListObjectsV2(ctx, &s3.ListObjectsV2Input{
-			Bucket:            aws.String(env.Cluster.Bucket),
-			Prefix:            aws.String(env.S3Prefix + "/"),
-			ContinuationToken: token,
-		})
-		if err != nil {
-			t.Fatalf("snapshot s3 inventory: %v", err)
-		}
-		for _, obj := range out.Contents {
-			stat := s3ObjectStat{Size: aws.ToInt64(obj.Size), ETag: aws.ToString(obj.ETag)}
-			if obj.LastModified != nil {
-				stat.LastModified = obj.LastModified.UTC()
-			}
-			inv[aws.ToString(obj.Key)] = stat
-		}
-		if out.NextContinuationToken == nil {
-			return inv
-		}
-		token = out.NextContinuationToken
-	}
-}
-
-// snapshotManifest returns the manifest's raw bytes and ETag. The seed phase
-// runs a real flush + init first, so the manifest must exist by the time any
-// snapshot is taken — a load failure here is a test bug, not a skip.
-func snapshotManifest(t *testing.T, ctx context.Context, env *Env, schema SchemaRef) ([]byte, string) {
-	t.Helper()
-	store := &manifest.S3Store{Client: env.Cluster.S3, Bucket: env.Cluster.Bucket}
-	resolver := manifest.PathResolver{Prefix: env.CDC.ManifestPrefix, PathTemplate: env.CDC.ManifestTemplate}
-	path, err := resolver.Resolve(schema.ID)
-	if err != nil {
-		t.Fatalf("resolve manifest path for schema %d: %v", schema.ID, err)
-	}
-	raw, etag, err := store.Load(ctx, path)
-	if err != nil {
-		t.Fatalf("load manifest %s (seed must have created it): %v", path, err)
-	}
-	return raw, etag
 }
 
 func assertStateUnchanged(t *testing.T, label string, before, after stateSnapshot) {
--- a/internal/e2e_harness/production/init_rerun_e2e_test.go
+++ b/internal/e2e_harness/production/init_rerun_e2e_test.go
@@ -65,10 +65,8 @@ func TestInitRerunIdempotency(t *testing.T) {
 
 	// Same federated results after the rerun (base-only source).
 	env.ExecSQL(ctx, "DELETE FROM change_log WHERE schema_id = $1", wide.ID)
-	result := env.AssertQueryMatches(ctx, Query{Schema: wide, Limit: 100})
-	if result != nil && len(result.Records) != 20 {
-		t.Fatalf("federated result has %d rows after rerun, want 20", len(result.Records))
-	}
+	assertFederatedRowCount(ctx, t, env, "post-rerun base tier",
+		Query{Schema: wide, Limit: 100}, 20)
 }
 
 func buildBasePaths(m *manifest.Manifest) map[string]bool {
@@ -152,8 +150,6 @@ func TestInitRerunAfterChangesReconcilesManifest(t *testing.T) {
 	if flush.UnflushedAfter != 0 {
 		t.Fatalf("flush left %d unflushed rows", flush.UnflushedAfter)
 	}
-	result := env.AssertQueryMatches(ctx, Query{Schema: wide, Limit: 100})
-	if result != nil && len(result.Records) != 23 {
-		t.Fatalf("federated result has %d rows after changed rerun, want 23", len(result.Records))
-	}
+	assertFederatedRowCount(ctx, t, env, "post-rerun parity",
+		Query{Schema: wide, Limit: 100}, 23)
 }
--- a/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
+++ b/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
@@ -303,14 +303,14 @@ func TestInitSchemaScopedIsolation(t *testing.T) {
 	}
 
 	// Onboarding contract: init does not clear change_log; clear it so the
-	// base tier alone must answer, then oracle-check both schemas.
+	// base tier alone must answer, then oracle-check both schemas. The count
+	// is only cold-tier evidence if the query actually routed to DuckDB —
+	// a PG-only routing regression would still count 5 entity_main rows (#248).
 	for _, ref := range []SchemaRef{simple, second} {
 		env.ExecSQL(ctx, "DELETE FROM change_log WHERE schema_id = $1", ref.ID)
 	}
 	for _, ref := range []SchemaRef{simple, second} {
-		result := env.AssertQueryMatches(ctx, Query{Schema: ref, Limit: 100})
-		if result != nil && len(result.Records) != 5 {
-			t.Errorf("schema %d federated result has %d rows, want 5", ref.ID, len(result.Records))
-		}
+		assertFederatedRowCount(ctx, t, env,
+			fmt.Sprintf("schema %d base tier", ref.ID), Query{Schema: ref, Limit: 100}, 5)
 	}
 }
--- a/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
+++ b/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
@@ -272,6 +272,12 @@ func TestInitSchemaScopedIsolation(t *testing.T) {
 		t.Fatal("schema 22 init must produce base entries (positive control)")
 	}
 
+	// #248: the NewObjects key diff is blind to overwrites that reuse the
+	// deterministic min_max keys, and the base-path comparison below is blind
+	// to a manifest rewritten with the same paths. Require exact stat/byte
+	// identity of the sibling's whole S3 surface across the rerun.
+	siblingBefore := captureSchemaS3State(t, ctx, env, second)
+
 	// Rerun schema 20's init: the sibling's manifest and objects must be
 	// byte-for-byte uninvolved (deterministic keys make the rerun itself a
 	// pure overwrite, so any sibling-partition key here is a leak).
@@ -280,6 +286,8 @@ func TestInitSchemaScopedIsolation(t *testing.T) {
 		t.Fatalf("rerun init schema %d: %v", simple.ID, err)
 	}
 	assertInitTouchesOnly(rerun, simple, second)
+	assertSchemaS3StateUnchanged(t, "schema 20 rerun vs sibling 22",
+		siblingBefore, captureSchemaS3State(t, ctx, env, second))
 	manifests, err = env.loadManifests(ctx)
 	if err != nil {
 		t.Fatalf("reload manifests: %v", err)
--- a/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
+++ b/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
@@ -262,11 +262,18 @@ func TestInitSchemaScopedIsolation(t *testing.T) {
 		t.Fatalf("schema %d must have no manifest before its own init, got %+v", second.ID, manifests[second.ID])
 	}
 
+	// #248: schema 20's surface already exists here, so schema 22's init is
+	// held to the same overwrite-proof contract as the rerun below — the key
+	// diff alone would miss an in-place re-export of the sibling.
+	simpleBefore := captureSchemaS3State(t, ctx, env, simple)
+
 	secondInit, err := env.RunInit(ctx, second)
 	if err != nil {
 		t.Fatalf("init schema %d: %v", second.ID, err)
 	}
 	assertInitTouchesOnly(secondInit, second, simple)
+	assertSchemaS3StateUnchanged(t, "schema 22 init vs sibling 20",
+		simpleBefore, captureSchemaS3State(t, ctx, env, simple))
 	siblingPaths := buildBasePaths(secondInit.Manifest)
 	if len(siblingPaths) == 0 {
 		t.Fatal("schema 22 init must produce base entries (positive control)")
--- a/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
+++ b/internal/e2e_harness/production/multi_schema_isolation_e2e_test.go
@@ -81,7 +81,7 @@ func testMultiSchemaCopyFault(ctx context.Context, t *testing.T) {
 	simple, second := seedTwoSchemas(ctx, t, env)
 
 	faulty := &FaultInjectingS3{Inner: env.Cluster.S3,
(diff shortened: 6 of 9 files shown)


iceboundrock:
**Whole-branch review** (pre-PR final review, dispatched on the strongest available model over the full 6f2ba4b..5381131 diff; posted per PR rules — review artifacts live on the PR). Verdict: **Ready to merge — Yes**, 0 critical / 0 important. The two comment-wording minors it flagged were fixed in 529fb38 (scoped re-review: both ADDRESSED); the remaining deferred minors became #427.

---

### Strengths

- **The two false-negative closures are real and correctly targeted.** The sibling-state snapshot covers exactly the two blind spots named in the issue: `NewObjects` key-diff blindness to deterministic-key overwrites, and `buildBasePaths` blindness to same-path manifest rewrites. The capture is taken at the only correct point — after `secondInit` (so the sibling surface exists) and immediately before the rerun — matching `snapshotManifest`'s Fatalf-on-missing behavior.
- **Empirical falsification was absorbed into comments, not papered over.** The plan's original claim ("a re-export also changes bytes and therefore ETag/Size") was measured false, and the committed `s3ObjectStat` comment now says the opposite — parquet leg is best-effort LastModified, manifest leg is the unconditional detector. The comment matches the mutation evidence (parquet leg went red on LastModified only). This is exactly the discipline #355 demands.
- **No flakiness introduced.** The committed code has no sleeps or timing dependence; LastModified second-granularity only weakens detection (a miss), never produces a false red — nothing rewrites the sibling between the two captures on a healthy run. The env's `S3Prefix` is per-env unique, so the prefix-scoped listings are immune to the package's `t.Parallel()` siblings. RustFS/S3 listings are strongly consistent.
- **Rename swept completely.** Zero leftover `schemaKeyPrefix`/`finalsForSchema` occurrences repo-wide; doc comments renamed along with identifiers and still read grammatically. All later commits use the new names — clean cross-task ordering.
- **The prefix-parameter compile-break worked as designed.** The plan listed only the dry-run test as a consumer, but `compaction_e2e_test.go` was a second pre-existing consumer; the parameter addition forced it to surface, and both call sites were updated with the semantics-preserving `env.S3Prefix+"/"`.
- **The optional #176 strengthening is meaningful at both sites**: hot tier is empty at each (change_log cleared / fully drained), so the count is only cold-tier evidence with the routing check. The nil-tolerant `if res == nil { return }` in the helper is safe: `AssertQueryMatches` only returns nil after `t.Fatalf`.

### Issues

**Critical:** none. **Important:** none — looked specifically for residual false negatives and healthy-run flakiness in the new assertions; found none beyond the documented, comment-acknowledged LastModified-granularity gap (which the manifest leg backstops for any store-mediated rewrite).

**Minor (status):**
1. Hoist-rationale comment factually off (second consumer pre-existed) — **fixed in 529fb38**.
2. Manifest-key exact-prefix listing would also match hypothetical `…/22.json.*` keys — harmless today, noted only.
3. `assertFederatedRowCount` (pre-existing) dereferences `res.Plan` without a nil check — **deferred to #427**.

### Deferred-minor triage (from task reviews)

1. Task-report line-number discrepancy — defer (documentation-only, no code artifact).
2. Manifest key derived two ways (hardcoded Sprintf vs `manifest.PathResolver`) — defer; divergence only possible if a test overrides `ManifestTemplate` (default renders exactly `%s/manifest/%d.json`); on divergence the stat leg degrades silently but the resolver-based raw-bytes/ETag leg still fires → **#427**.
3. Manifest-key literal exists in four places (two pre-existing, one outside the e2e build tag) — defer; a single DRY pass is the right shape → **#427**.
4. "any rewrite increments version" over-claim — **fixed in 529fb38** ("any save through the manifest store").
5. Deleted-sibling-manifest failure misattributed to seed — defer; loud failure, one minute of reader confusion on a failure mode no known regression produces → **#427**.

### Assessment

**Ready to merge? Yes.** All three issue items plus the optional #176 strengthening are implemented, mutation-verified, and internally coherent; the one plan deviation (probe shape) is a documented ruling, and the empirically-falsified rationale was corrected in the committed comments rather than shipped. None of the deferred minors can produce a false green on a realistic regression or a false red on a healthy run.

iceboundrock:
## Findings

### P2 — Sibling-overwrite guard covers the rerun leg only; the `secondInit` leg keeps the same blind spot

In `TestInitSchemaScopedIsolation`, the new `captureSchemaS3State`/`assertSchemaS3StateUnchanged` guard is placed around the **rerun** of schema 20's init (sibling = 22, surface created by `secondInit`). That is exactly right for that leg.

However, the issue states the fix as "snapshot the sibling's observable state before the target-scoped init and compare after" — and the test has **two** target-scoped inits whose sibling already has an observable surface:

- the **rerun** of schema 20 (sibling 22 — now guarded), and
- **`secondInit`** of schema 22 (sibling 20, which already has objects + manifest from `first`) — **still guarded only by `assertInitTouchesOnly`'s `NewObjects` key diff and `buildBasePaths` path-set comparison**, both of which the issue explicitly identifies as blind to in-place same-key overwrites and same-path manifest rewrites.

So a buggy init of schema 22 that re-exports schema 20's deterministic `min_max` keys in place (or rewrites schema 20's manifest with the same paths) would still pass this test today. The regression the issue set out to close remains open on the `secondInit` direction.

**Suggestion:** mirror the rerun guard — take `captureSchemaS3State(t, ctx, env, simple)` before `env.RunInit(ctx, second)` and assert unchanged after (the schema-20 surface is fully established by `first` at that point, so it is a meaningful snapshot). If it is intentionally left key-diff-only, a comment explaining why that leg is exempt would help — right now the asymmetry is not self-documenting.

Severity kept at P2 because the exercised path shares production wiring, but this remains a genuine false-negative path, which is the entire point of #248.

### P3 — `assertFederatedRowCount` escalates count mismatch to `t.Fatalf` inside a two-schema loop

Previously the final oracle in `TestInitSchemaScopedIsolation` reported both schemas with `t.Errorf`; through `assertFederatedRowCount` (which uses `t.Fatalf` on count mismatch and `t.Errorf` on routing), a count failure on schema 20 now aborts the test before schema 22's routing/count is ever checked. A dual-schema regression yields half the diagnostics. Consider `t.Errorf` for the count leg as well, or accept the loss of diagnostic completeness deliberately.

iceboundrock:
## Spec

Issue #248 asks for three things; here is how the PR maps to each:

1. **Item 1 — sibling isolation blind to overwrites.** Implemented for the **rerun** leg of `TestInitSchemaScopedIsolation`: full observable S3 surface (partition objects with size/ETag/LastModified + manifest raw bytes + ETag) captured before the rerun and asserted byte-identical after. This is exactly the "stat + byte identity beats key-diff and path-set comparison" fix the issue prescribes. One gap remains: the `secondInit` leg (init of schema 22 with schema 20 as its sibling) is not covered — see Findings P2. Acceptance criterion "overwrite of a sibling base object turns the test red" is demonstrated via the mutation probe for the rerun leg; the equivalent probe on the `secondInit` leg was not demonstrated.

2. **Item 2 — base-tier visibility assertion could pass without reading Parquet.** Fully implemented: final clear-and-count goes through `assertFederatedRowCount`, which additionally requires `result.Plan.Routing.UseDuckDB`. The issue's optional strengthening of the two #176 rerun tests was also done (nice-to-have, keeps the parity oracles meaningful). Mutation-verified both ways (PreferHot forces the PG path → red).

3. **Item 3 — verb-first renames.** `schemaKeyPrefix` → `buildSchemaKeyPrefix`, `finalsForSchema` → `filterFinalsForSchema`, swept across the whole package (all 9 touched files), not just the two #186 files. Call-site sweep verified: no stale noun-first references remain.

Acceptance criteria are otherwise met: mutation verification evidence is documented on the issue, and the shipped assertions are mutation-tested.

Note: the helper hoisting (`snapshotS3Inventory` gaining a `prefix` parameter) is a spec-neutral refactor that preserves the two pre-existing consumers' behavior verbatim (`env.S3Prefix + "/"`), verified against the old hardcoded prefix.

iceboundrock:
## Standards

`coding-standard.md` compliance:

- **Verb-first private helpers** (§4): the two renamed helpers now read `buildSchemaKeyPrefix` / `filterFinalsForSchema`; the new helpers `captureSchemaS3State`, `assertSchemaS3StateUnchanged`, `snapshotS3Inventory`, `snapshotManifest` are all verb-first. Good.
- **File/function size limits**: all touched files are well under 500 lines (new helper file is 118); no function approaches 100 lines.
- **Error wrapping**: no bare `return err` — test helpers use `t.Fatalf`/`t.Errorf`. The hoisted `snapshotS3Inventory`/`snapshotManifest` follow the pre-existing Fatalf style.
- **Comment accuracy**: the final commit (529fb38) explicitly scopes the two comments to measured evidence rather than hand-wavy claims — this is exactly the right call given the PR's empirical finding that re-export is byte-deterministic. The `min_max`-key blind-spot rationale in `schemaS3State` is precise.
- **Format**: `gofmt -l` clean on the production e2e package at the review head.
- **No scope creep**: it's a test-only PR; the hoisting to a new shared helpers file is justified by a second consumer (not gratuitous).
- **Local standards context**: pre-existing long lines (122–175 chars) in `cdc_fault_helpers_e2e_test.go` are untouched by this PR (verified unchanged vs base), so no new lint debt is introduced.

Minor: `snapshotManifest`'s Fatalf message "(seed must have created it)" is now slightly stale for the new consumer (`captureSchemaS3State` snapshots a schema that was initialized via `RunInit`, not "seeded") — a drive-by wording fix would keep the message honest. This is cosmetic, not blocking.

iceboundrock:
## Assessment

**Overall: approvable.** This is a tightly-scoped, test-only hardening PR that does what it claims: it closes the two false-negative paths #248 identified, and it renames two helpers to meet the verb-first standard. The mutation verification evidence on the issue is reproducible in structure and the final comments were corrected to match measured behavior (byte-deterministic re-export → LastModified-only object leg; manifest version is the unconditional detector).

**Strength** — the design of `assertSchemaS3StateUnchanged` is sound: it no longer trusts key diffs or path-set comparisons, which is precisely where the false negatives lived. Requiring exact identity of size/ETag/LastModified plus raw manifest bytes + ETag is the right observable surface.

**The one substantive gap** is the asymmetry in `TestInitSchemaScopedIsolation` (Findings P2): the sibling snapshot only wraps the rerun leg, not `secondInit`. The test still cannot catch an in-place overwrite of schema 20's deterministic base objects by schema 22's init. Since the entire purpose of #248 is closing exactly this class, I'd recommend addressing it before merge — the fix is ~4 lines (a mirror snapshot around `secondInit`).

**Scope sanity**: 159 additions / 92 deletions over 9 files, all `_e2e` test code. No production code touched. The hoisting keeps the two existing consumers' behavior identical.

No critical or high findings. One P2 (the `secondInit` leg) and one P3 (Fatalf-in-loop diagnostic regression).

iceboundrock:
## Observations

1. **The `before` snapshot is taken after `secondInit`, and that's correct.** In `TestInitSchemaScopedIsolation`, `captureSchemaS3State(second)` runs only after schema 22's own init so the sibling surface is fully established; the rerun of schema 20 then must not perturb it. Snapshot placement is deliberate and right.

2. **`buildBasePaths` assertion is now partially redundant but kept as a fast-fail path.** The set-comparison loop after the rerun overlaps with `assertSchemaS3StateUnchanged`; keeping both is fine — the path-set check gives a terse, specific message while the S3-surface check gives full detail. No issue.

3. **`snapshotS3Inventory` prefix semantics are subtle but correct.** `captureSchemaS3State` first lists the whole partition prefix, then lists an object whose key exactly equals the manifest key. Listing with a prefix equal to a full key correctly returns exactly that object (ListObjectsV2 prefix = key → includes it). Verified the manifest key format matches `artifacts.go`'s `{S3Prefix}/manifest/{schemaID}.json` and the `manifest.PathResolver` used by `snapshotManifest`. Consistent.

4. **The memoized LastModified-leg claim (1s granularity) is honestly scoped.** The comment on `s3ObjectStat` correctly states the object leg may rest on LastModified alone for byte-deterministic re-exports, and the manifest leg is the unconditional detector (version increments on every save — verified in `internal/manifest/manifest.go:71-74`). Good documentation discipline.

5. **Small comment drift**: `snapshotManifest`'s Fatalf says "(seed must have created it)" but `captureState` (dryrun) and `captureSchemaS3State` operate on schema surfaces produced by `RunInit`/flush, not "seed". Pre-existing wording, now surfaced by the hoist. Cosmetic.

6. **The compaction-consumer update preserves hardcoded prefix exactly.** `snapshotS3Inventory(t, ctx, env, env.S3Prefix+"/")` matches the old hardcoded `env.S3Prefix + "/"`; flag-level behavior of `TestCompactionPromotionEquivalence` is unchanged.

7. **The rename is complete but touched a wider surface than strictly required** (7 files beyond the two #186 files). That's better than a partial sweep — it prevents `schemaKeyPrefix` from lingering as dead/confusing naming in sibling tests. No concern.

iceboundrock:
## Verification

I did not use the Docker e2e suite (testcontainers, `-tags=e2e -timeout=30m`) in this review environment, but I did run repo-native static gates against the PR head (529fb38):

- `go vet -tags e2e ./internal/e2e_harness/production/...` → clean, no output. Confirms the new `state_snapshot_helpers_e2e_test.go`, the hoists, and the call-site sweep all compile and type-check under the e2e build tag.
- `gofmt -l internal/e2e_harness/production/` → clean (no files listed).
- Call-site sweep: grepped the whole branch for all six names (`buildSchemaKeyPrefix`, `filterFinalsForSchema`, `schemaKeyPrefix`, `finalsForSchema`, `snapshotS3Inventory`, `s3ObjectStat`) — every consumer updated, no stale noun-first references, and the two pre-existing `snapshotS3Inventory` consumers pass `env.S3Prefix+"/"` verbatim.
- Long-line scan: the 122–175-char lines in `cdc_fault_helpers_e2e_test.go` are pre-existing on base `main` (identical line numbers); the PR adds no new long lines.
- Manifest version-increment claim cross-checked against `internal/manifest/manifest.go` (`Version` bumped on every `Save`) — supports the "manifest leg is the unconditional detector" comment.

The PR's own claim that `make fmt-check`, `make lint`, `make test`, `make test-e2e-production` are green at the review head is consistent with the mutation evidence posted on #248; I could not re-run the Docker e2e suite here. Recommended: re-run `make test-e2e-production` once after addressing the P2 finding, since it changes `TestInitSchemaScopedIsolation`.

iceboundrock:
## Summary

Test-only hardening PR for the production e2e harness, driven by issue #248. All three issue items are addressed: (1) sibling S3-surface snapshot closes the overwrite blind spot in `TestInitSchemaScopedIsolation`'s rerun leg, (2) `assertFederatedRowCount` routing check guards the base-tier oracle (plus the two #176 rerun tests), and (3) verb-first renames swept package-wide. The mutation verification and honest comment-scoping on the issue are a model example of evidence-driven test work.

**Recommend finding:** the sibling snapshot only wraps the rerun leg; `secondInit` (init of schema 22 vs sibling 20, which already has objects+manifest) retains the exact false-negative class #248 exists to close. A mirrored snapshot around `secondInit` would complete the fix (~4 lines). See Findings P2.

**Minor:** `assertFederatedRowCount`'s `t.Fatalf` on count mismatch reduces diagnostics in the two-schema loop (P3).

Static gates (`go vet -tags e2e`, `gofmt -l`) pass at the head; no new lint debt; no production code touched. Approve after the P2 is resolved or explicitly documented.

iceboundrock:
Response to the review comments, item by item.

## P2 — secondInit leg blind spot: confirmed and fixed in b772215

Verified against the test flow: after `first`, schema 20 has a full surface (objects + manifest), so `secondInit` is indeed a target-scoped init whose sibling is only key-diff-guarded — the exact class #248 exists to close. My earlier placement ruling ("overwrite is only possible at the rerun") was wrong.

Fix is the suggested ~4-line mirror: `captureSchemaS3State(t, ctx, env, simple)` before `secondInit`, `assertSchemaS3StateUnchanged(t, "schema 22 init vs sibling 20", ...)` right after `assertInitTouchesOnly(secondInit, second, simple)`.

Mutation-verified on this leg (probe = `sleep 1.1s` + `env.RunInit(ctx, simple)` between capture and `secondInit`, same shape as the rerun-leg probe):
- **Red:** FAIL at `schema 22 init vs sibling 20` — manifest ETag `ea6d8bf8…` → `304975f9…`, raw bytes `version: 1` → `version: 2`, LastModified `02:51:54.652` → `02:51:55.861`.
- **Green:** probe removed → PASS (2.12s), no probe residue.
- Per the Verification comment's recommendation, `make test-e2e-production` re-run at the new head: **all green (140s)**; `fmt-check` and `lint` green.

## P3 — Fatalf in the two-schema loop: accepted deliberately, not changing

Pushing back on modifying the helper. `assertFederatedRowCount`'s Fatalf-on-count is the shared contract at its 8 pre-existing call sites, and this package has a standing ruling (from the #215 review cycle) that the Fatalf/Errorf asymmetry among assert helpers is deliberate and should not be unified in passing. Within the helper, continuing past a count mismatch would also run the routing check against an already-broken result — noise, not diagnostics. The cost is real but small: on a dual-schema regression the first failing schema is where investigation starts anyway, and a loop-local Errorf variant would duplicate the helper for one site. Accepting the diagnostic-completeness loss as the documented trade-off.

## snapshotManifest "(seed must have created it)" wording (Standards minor / Observation 5)

Already captured as item 3 of #427 (filed before this review round) with the same suggested generalization ("caller guarantees it exists by now"). Deferred there rather than a drive-by here, since #427 also touches the surrounding helper wording.

## Spec-comment note on the undemonstrated secondInit probe

Now demonstrated — evidence above. With b772215, both target-scoped inits that run against an existing sibling surface (secondInit and the rerun) carry the same overwrite-proof contract, each red/green-verified.

Branch head is b772215; the PR diff remains test-only.

iceboundrock:
## Findings (round 2 — review at head `b772215`)

### P2 — secondInit leg blind spot: RESOLVED
The round-1 P2 is fixed as suggested: `captureSchemaS3State(t, ctx, env, simple)` now runs before `env.RunInit(ctx, second)` and `assertSchemaS3StateUnchanged(t, "schema 22 init vs sibling 20", ...)` immediately after `assertInitTouchesOnly`. Both target-scoped inits that run against an existing sibling surface (secondInit and the rerun) now carry the same overwrite-proof guard. Verified in the code, and the mutation evidence posted on the issue (probe between capture and `secondInit`: FAIL on manifest ETag + raw bytes + LastModified; probe removed: PASS) is internally consistent with the rerun-leg evidence. No residual false-negative on this leg: a buggy init of 22 that re-exports 20's deterministic bits in place, or rewrites 20's manifest with the same paths, now turns the test red.

### P3 — `t.Fatalf` inside the two-schema loop: accepted deliberately, no change
The pushback is sound and I accept it: `assertFederatedRowCount`'s Fatalf-on-count is the shared contract at its pre-existing call sites, the #215 ruling already governs the Fatalf/Errorf asymmetry, and continuing past a count mismatch would run the routing check against a broken result. The diagnostic-completeness loss on a dual-schema regression is a documented trade-off, not a defect.

### snapshotManifest "(seed must have created it)" wording
Deferred to #427 as item 3 — appropriate placement; no action needed on this PR.

### No new findings
I looked specifically for residual false negatives and healthy-run flakiness in the new secondInit guard: the capture sits after `first` (schema 20's surface fully established) and nothing rewrites schema 20 between the two captures on a healthy run; the LastModified 1s-granularity gap only weakens the object leg, and the manifest leg (version increments through the store) backstops it. No new issues at the head.

iceboundrock:
## Spec (round 2 — at head `b772215`)

Issue #248's three items now all hold at the current head:

1. **Sibling isolation blind to overwrites** — now closed on **both** target-scoped inits. The rerun leg (covered since round 1) and the `secondInit` leg (added in `b772215`) each snapshot the sibling's full observable S3 surface (partition objects with size/ETag/LastModified + manifest raw bytes + ETag) around the scoped init and require exact identity after. The acceptance criterion "overwriting a sibling base object or rewriting its manifest in place turns the test red" is now demonstrated by mutation probe on both legs (evidence on the issue).
2. **Base-tier visibility assertion could pass without reading Parquet** — fully implemented: the final clear-and-count goes through `assertFederatedRowCount`, which requires `result.Plan.Routing.UseDuckDB`; the optional #176 strengthening (`TestInitRerunIdempotency`, `TestInitRerunAfterChangesReconcilesManifest`) is included. Mutation-verified both directions at Task 3.
3. **Verb-first renames** — `schemaKeyPrefix` → `buildSchemaKeyPrefix`, `finalsForSchema` → `filterFinalsForSchema` swept across all 9 touched files; zero stale references repo-wide (re-verified at the head).

Acceptance criteria otherwise met: helper hoist is the issue's own suggestion and preserves the pre-existing consumers' behavior verbatim; new/renamed helpers are verb-first; suite + lint green.

One plan deviation remains documented as a ruling rather than drifted: the sibling mutation probe uses an in-place re-init (exact regression class) instead of garbage-byte PutObject, because the latter would corrupt the base tier and fail the final oracle for the wrong reason. That ruling is sound.

iceboundrock:
## Standards (round 2 — at head `b772215`)

`coding-standard.md` compliance at the head:

- **Verb-first private helpers (§4)**: all helpers are verb-first — `buildSchemaKeyPrefix`, `filterFinalsForSchema`, `captureSchemaS3State`, `assertSchemaS3StateUnchanged`, `snapshotS3Inventory`, `snapshotManifest`. The one noun-adjacent name, `s3ObjectStat`, is a data type, not a function — no violation.
- **Size limits**: `state_snapshot_helpers_e2e_test.go` is 118 lines (≤500); `TestInitSchemaScopedIsolation` is 91 lines (≤100) even with the added guard. No function approaches limits.
- **Error handling**: no bare `return err`; helpers use `t.Fatalf`/`t.Errorf` in the established fixture style.
- **Comment accuracy**: post-`529fb38`, the two evidence-scoped comments (`s3ObjectStat` byte-deterministic/LastModified caveat; `schemaS3State` "any save through the manifest store" wording) state measured behavior — confirmed accurate against the mutation tables on the issue and against `internal/manifest/manifest.go` (version increments on every `Save`). The new `// #248:` block on the secondInit leg correctly names the earlier-in-the-test blind spot.
- **No scope creep**: still a test-only PR (9 files, all `_e2e`); the hoist is justified by real multiple consumers.
- **No new lint debt**: no new lines >120 chars in any touched file; the pre-existing long lines in `cdc_fault_helpers_e2e_test.go` are untouched vs base.
- **Format/lint**: `gofmt -l` clean, `make fmt-check` OK, `make lint` exit 0 at the head (re-run this session).

iceboundrock:
## Assessment (round 2 — at head `b772215`)

**Approvable — ready to merge.** This is a tightly-scoped, test-only hardening PR that now fully closes the two false-negative paths #248 identified, plus the verb-first renames.

The round-1 P2 was the only substantive gap and it is now closed: both target-scoped inits run against an existing sibling surface carry the same overwrite-proof, mutation-verified guard. The P3 was pushed back on with a defensible, precedent-backed rationale that I accept. The remaining deferred items are tracked in #427 and none can produce a false green or a false red.

**Strength of the design** (unchanged from round 1, now complete): the guard trusts stat + byte identity of the sibling's whole observable S3 surface rather than the key-diff/path-set comparisons that hid the false negatives; the manifest leg is the unconditional detector and the object leg's LastModified-only caveat is honestly scoped in comments rather than papered over.

**Scope sanity**: 166 insertions / 92 deletions over 9 `_e2e` files; no production code touched. The hoist keeps the pre-existing consumers' behavior bit-for-bit identical.

No critical, high, or remaining P2/P3 findings at the head. The author's verification claims — including the e2e re-run at the new head — were re-confirmed this session (see Verification).

iceboundrock:
## Observations (round 2 — at head `b772215`)

1. **The secondInit guard placement is exact.** `simpleBefore := captureSchemaS3State(..., simple)` runs after `first` (so schema 20's surface — 5 base objects + manifest — fully exists) and immediately before `env.RunInit(ctx, second)`. The `snapshotManifest` Fatalf-on-missing precondition is satisfied for both legs: schema 20's manifest exists from `first`, schema 22's from `secondInit`.

2. **The object-map identity check is stronger than `assertInitTouchesOnly` in one more dimension than the issue named.** `reflect.DeepEqual` on the object map fails not only on in-place overwrites and same-path rewrites but also on **deletion** of a sibling object — which the `NewObjects` key diff could never see (it only reports new keys). Worth noting as a bonus for the same fixture cost.

3. **Symmetry in labels is clean.** `"schema 22 init vs sibling 20"` and `"schema 20 rerun vs sibling 22"` each name the init action and the guarded sibling; failure output is self-explanatory.

4. **The `t.Errorf` (not `t.Fatalf`) choice in `assertSchemaS3StateUnchanged` keeps both detection legs reporting** — a mutation that hits both object and manifest legs shows both messages (as the issue's red-run table did), which is exactly the diagnostic value the sibling snapshot was built for. It composes fine with the accepted P3 Fatalf on the count oracle.

5. **Env isolation keeps the prefix-scoped listings parallel-safe.** The env's `S3Prefix` is per-env unique, so `captureSchemaS3State` under `buildSchemaKeyPrefix` never collides with the package's `t.Parallel()` siblings. RustFS/S3 listings are strongly consistent.

6. **The manifest-key literal in `captureSchemaS3State` stays consistent with the resolver-based `snapshotManifest`.** The hardcoded `{S3Prefix}/manifest/{id}.json` matches the default `ManifestTemplate` rendering used by `PathResolver`, so the object-map manifests entry and the raw-bytes/ETag legs can't silently diverge under the default config (the divergence-only-if-overridden caveat remains tracked in #427).

7. **Probe hygiene verified at the head.** No `time.Sleep`, no `probe`, and no temporary imports remain in the committed test files; the mutation probes that produced the red/green evidence exist only in the issue record.

iceboundrock:
## Verification (round 2 — independently re-run at head `b772215`)

I re-ran the gates this session in the `forma-248` worktree at `b772215`:

- `make test-e2e-production` (full Docker production harness) → **`ok ... 147.298s`**, `--- PASS: TestInitSchemaScopedIsolation (2.09s)`. This independently confirms the author's "all green (140s)" claim at the fix commit.
- `make fmt-check` → OK (gofmt gate clean).
- `make lint` → exit 0 (golangci-lint v1.64.8, no findings).
- `go vet -tags e2e ./internal/e2e_harness/production/` → clean.
- `gofmt -l internal/e2e_harness/production/` → no files listed.
- Rename sweep: `grep` for `schemaKeyPrefix|finalsForSchema` excluding the new names → zero hits repo-wide.
- Long-line scan: no line >120 chars in `multi_schema_isolation_e2e_test.go` or `state_snapshot_helpers_e2e_test.go`.
- Manifest version-increment claim cross-checked at `internal/manifest/manifest.go` (`Version` bumped on every `Save`).
- Probe hygiene: no `Sleep`/`probe` residue in the committed test files.

Nothing to caveat: the e2e suite, fmt, lint, vet, and the author's mutation evidence on the issue are all mutually consistent at the current head.

iceboundrock:
## Summary (round 2 — at head `b772215`)

**Ready to merge — Yes.** Round 2 confirms round 1's approval now holds across the whole issue: the round-1 P2 (`secondInit` leg blind spot) is fixed in `b772215` with the same mutation-verified overwrite-proof guard, the P3 was defended on precedent and accepted, and all three #248 items plus the optional #176 strengthening are in place. No critical, high, or remaining P2/P3 findings.

Independent re-verification at the head this session: `make test-e2e-production` green (147s, including `TestInitSchemaScopedIsolation` PASS), `make fmt-check`/`make lint`/`go vet -tags e2e` clean, rename sweep complete, no probe residue, no new long lines, no production code touched. The mutation evidence structure on the issue is reproducible and the committed comments match the measured behavior.

Per AGENTS.md: not merging automatically — the branch is ready for merge whenever the operator pulls the trigger; after merge, follow the routine (clean up local branch/worktree, fast-forward `main`, update/close #248).

PR #428, opened 2026-08-20
test: close two guards that cannot fail for the regressions they name (#321)

### LINKED ISSUE #321: Two guards cannot fail for the regressions their comments name (Covers dot boundary, lambda entry-point scan)
Follow-up from #314 (PR #316). Two small test-hardening gaps, both demonstrated by mutation rather than inferred.

## 1. `RelationRoots.Covers` dot boundary is unpinned

`internal/transform/relation_roots.go` decides whether a dotted attribute name lies beneath a relation root. The distinction that matters is the separator: `contactSnapshot.name` is beneath the root `contactSnapshot`, while `contactSnapshotExtra.x` merely *starts with* the same characters and must still be expanded and validated.

`Covers` implements that correctly — it scans dot positions and does a prefix-map lookup, verified by construction. But **replacing it with the naive `strings.HasPrefix(name, root) && name != root` leaves the entire suite green.** The only outside-root test uses `RelationRoots{"other": {}}` against `contact.email`, which no prefix mutant can fail.

One case closes it: a relation root plus a sibling attribute sharing its prefix, asserting the sibling still expands.

## 2. `cmd/lambda/main_test.go` entry-point guard cannot catch what it advertises

`TestEntityConfigFromEnvWiredInEveryWriteEntryPoint` was added so a *future* entry point that builds an `EntityManager` and forgets `bootstrap.EntityConfigFromEnv` fails the build. Its comment says exactly that. But the `entryPoints` map is hardcoded to `cmd/lambda` and `cmd/server`, so a new entry point is never read — confirmed by adding a scratch `cmd/apiworker/main.go` that calls the factory with no overlay: the test still passed.

Two further holes:

- The gate substring is `NewEntityManagerWithConfigContext`, which does not match the **already-existing** `cmd/sample/main.go` `factory.NewEntityManagerWithConfig(` wrapper. So the "differently-named wrapper" evasion the test's own comment acknowledges is present in-repo today. (`cmd/sample` serves no HTTP routes, so there is no production exposure.)
- Replacing the call with a comment mentioning the symbol also passes — inherent to a source scan, and acknowledged.

Direction: walk `cmd/*/` and match `factory.NewEntityManager` rather than enumerating files and one exact call form. That makes the comment true. Alternatively narrow the comment to what it really is — a two-file removal tripwire — but the walk is barely more code.

## Why these are grouped

Both are the same failure shape and it recurred repeatedly through #314: **a test whose comment claims it guards a specific regression, which the regression does not break.** Three others of this kind were found and fixed during that PR. The cheap discipline is to introduce the regression the comment names and watch the test fail before trusting it.

iceboundrock:
Closes #321.

Two guards whose comments named a regression they could not fail, plus the
one-line wiring gap the second guard's widening exposed.

## What changed

**`internal/transform/relation_root_required_policy_test.go`** — pins
`RelationRoots.Covers` on the separator rather than the characters. The fixture
gains `contactSnapshotExtra.note` / `contactSnapshotExtra.status` alongside the
`contactSnapshot` root, and the new test asserts the sibling's required policy is
still enforced. The two attributes carry `required_if_parent_present` rather than
`required_always` so they stay inert in the four tests that share this registry
and supply only `id`. This is the twin of `TestStripKeepsSamePrefixSibling`,
which already pins the same boundary on `RelationIndex.coversRelationSubtree`.

**`cmd/sample/main.go`** — applies `bootstrap.EntityConfigFromEnv`. The CSV bulk
importer is a real write path but built its manager from `DefaultConfig` alone,
so `VALIDATE_UPDATES_STRICT` was silently ignored there. `EntityConfigFromEnv`
copies its input and touches only `ValidateUpdatesStrict`, so the
`SchemaDirectory` assigned just above survives.

**`cmd/lambda/main_test.go`** — the entry-point guard now walks every directory
under `cmd/` and matches the `factory.NewEntityManager` prefix, per directory
rather than per file. Both wrappers count, and a directory added later is read
without anyone updating the test. Comment rewritten to what the walk actually
does.

An **anti-vacuity floor** was added beyond the issue's ask: after the walk,
assert `lambda`, `server` and `sample` were all classified as manager builders.
It is a floor, not a ceiling — new entry points are still discovered, never
enumerated. Without it, a wrong root, a renamed directory, or a scan that matches
nothing passes while reading zero files, which is this issue's failure shape
relocated rather than fixed.

## Verification — every guard was watched failing

Both issue premises were re-checked against `main` at `eb1302c` first, not
inherited from #314.

| Mutation | Expected | Result |
|---|---|---|
| `Covers` → naive `strings.HasPrefix(name, root) && name != root`, whole unit tier | red | **before this PR: green** (28 packages, zero failures) |
| same mutant, after the new pin | red | red — `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling`, and only it |
| scratch `cmd/scratchworker` calling `NewEntityManagerWithConfigContext` | red | red |
| same probe using the `NewEntityManagerWithConfig` wrapper | red | red (the spelling the old gate substring missed) |
| walk root pointed at a tree with no `cmd` directories | red | red — all three anti-vacuity assertions fire |
| `cmd/sample` overlay line deleted | red | red |

The `Covers` mutations ran under `go test -overlay=` so the working tree was
never modified; the `cmd/` probes need real files, and the tree was verified
clean after each.

**One incidental finding, worth recording.** The first `cmd/scratchworker` probe
passed. The cause was the probe's own doc comment, which contained the words
`bootstrap.EntityConfigFromEnv` — so the source scan counted the entry point as
wired. That is the scan's known, documented blind spot (a comment naming the
symbol satisfies it), and this is the first time it has actually been observed
rather than assumed. It stays acknowledged in the test comment rather than fixed;
closing it means parsing the AST for a call expression, which is more machinery
than a two-substring gate deserves. Filed as nothing — noting it here so the next
reader of that comment knows it is measured, not theoretical.

## Not covered

- No e2e tier. None of the three changes touches the federated, CDC or compaction
  paths.
- `make fmt-check`, `make lint` (v1.64.8) and `make test` all clean.

### DIFF
--- a/cmd/lambda/main_test.go
+++ b/cmd/lambda/main_test.go
@@ -3,7 +3,9 @@ package main
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -20,42 +22,102 @@ func TestBootstrapLambda_CanceledContext(t *testing.T) {
 	}
 }
 
+// knownManagerBuilders are the cmd/ directories that build an EntityManager
+// today. The walk below discovers directories rather than reading this list, so
+// this is a floor and never a ceiling: it exists only so a walk that finds
+// nothing — wrong root, renamed directory, a scan that matches no files — fails
+// loudly instead of passing vacuously.
+var knownManagerBuilders = []string{"lambda", "server", "sample"}
+
 // TestEntityConfigFromEnvWiredInEveryWriteEntryPoint guards that every entry point
-// serving the write API overlays the entity config from the environment, so the
-// #314 staged rollout (VALIDATE_UPDATES_STRICT) is available on all deployment
-// targets rather than only the ones someone remembered to wire.
+// building an EntityManager overlays the entity config from the environment, so
+// the #314 staged rollout (VALIDATE_UPDATES_STRICT) is available on all
+// deployment targets rather than only the ones someone remembered to wire.
+//
+// It walks every directory under cmd/ and matches the "factory.NewEntityManager"
+// prefix, so an entry point added later is read without anyone updating this
+// test, and the WithConfig / WithConfigContext wrappers — and any future one —
+// all count. Before #321 the directories were a hardcoded pair and the match was
+// the exact WithConfigContext spelling, so the regression this comment names
+// could not fail it: a scratch cmd/apiworker went unread, and the already
+// existing cmd/sample used the unmatched WithConfig wrapper.
+//
+// Classification is per directory, not per file, so an entry point that builds
+// its manager outside main.go is still covered.
 //
 // This is a source-level guard rather than a wiring test on purpose.
 // bootstrapLambda takes no injection seams and connects to a real database before
 // reaching the config assignment, so exercising the wiring would need a live
-// Postgres/DSQL — a heavyweight harness for a one-line assignment. Scanning the
-// source costs nothing and catches the regression that actually matters: a third
-// entry point added later that builds an EntityManager and forgets the overlay.
+// Postgres/DSQL — a heavyweight harness for a one-line assignment. What a source
+// scan cannot see remains: replacing the call with a comment naming the symbol
+// still passes.
 //
-// It lives in this package because cmd/ has no shared test package; it
-// deliberately covers cmd/server too rather than being duplicated in both.
+// It lives in this package because cmd/ has no shared test package.
 func TestEntityConfigFromEnvWiredInEveryWriteEntryPoint(t *testing.T) {
 	// Relative to this package directory, which is the cwd for `go test`.
-	entryPoints := map[string]string{
-		"cmd/lambda": "main.go",
-		"cmd/server": "../server/main.go",
+	const cmdDir = ".."
+
+	entries, err := os.ReadDir(cmdDir)
+	if err != nil {
+		t.Fatalf("failed to read cmd directory %s: %v", cmdDir, err)
 	}
 
-	for name, path := range entryPoints {
-		source, err := os.ReadFile(path)
-		if err != nil {
-			t.Fatalf("failed to read %s entry point %s: %v", name, path, err)
-		}
-		text := string(source)
-
-		// Only entry points that actually build an EntityManager need the overlay.
-		if !strings.Contains(text, "NewEntityManagerWithConfigContext") {
+	buildsManager := map[string]bool{}
+	for _, entry := range entries {
+		if !entry.IsDir() {
 			continue
 		}
-		if !strings.Contains(text, "bootstrap.EntityConfigFromEnv") {
-			t.Errorf("%s builds an EntityManager but never calls "+
+		builds, overlays, err := scanEntryPointDir(filepath.Join(cmdDir, entry.Name()))
+		if err != nil {
+			t.Fatalf("failed to scan cmd/%s: %v", entry.Name(), err)
+		}
+		if !builds {
+			continue
+		}
+		buildsManager[entry.Name()] = true
+		if !overlays {
+			t.Errorf("cmd/%s builds an EntityManager but never calls "+
 				"bootstrap.EntityConfigFromEnv, so VALIDATE_UPDATES_STRICT is silently "+
-				"ignored on that deployment target (#314)", name)
+				"ignored on that deployment target (#314)", entry.Name())
 		}
 	}
+
+	for _, name := range knownManagerBuilders {
+		if !buildsManager[name] {
+			t.Errorf("the walk did not classify cmd/%s as building an EntityManager; "+
+				"it is reading the wrong tree or matching nothing, so its green would be "+
+				"meaningless (#321)", name)
+		}
+	}
+}
+
+// scanEntryPointDir reports whether any non-test Go file directly in dir builds
+// an EntityManager, and whether any of them applies the environment overlay.
+func scanEntryPointDir(dir string) (builds, overlays bool, err error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return false, false, fmt.Errorf("failed to read %s: %w", dir, err)
+	}
+
+	for _, entry := range entries {
+		name := entry.Name()
+		if entry.IsDir() || !strings.HasSuffix(name, ".go") || strings.HasSuffix(name, "_test.go") {
+			continue
+		}
+		path := filepath.Join(dir, name)
+		source, err := os.ReadFile(path)
+		if err != nil {
+			return false, false, fmt.Errorf("failed to read %s: %w", path, err)
+		}
+		text := string(source)
+		// The prefix, not one exact wrapper: WithConfig and WithConfigContext both
+		// build a manager, and so will the next wrapper.
+		if strings.Contains(text, "factory.NewEntityManager") {
+			builds = true
+		}
+		if strings.Contains(text, "bootstrap.EntityConfigFromEnv") {
+			overlays = true
+		}
+	}
+	return builds, overlays, nil
 }
--- a/cmd/lambda/main_test.go
+++ b/cmd/lambda/main_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -25,8 +26,12 @@ func TestBootstrapLambda_CanceledContext(t *testing.T) {
 // knownManagerBuilders are the cmd/ directories that build an EntityManager
 // today. The walk below discovers directories rather than reading this list, so
 // this is a floor and never a ceiling: it exists only so a walk that finds
-// nothing — wrong root, renamed directory, a scan that matches no files — fails
-// loudly instead of passing vacuously.
+// nothing — wrong root, a scan that matches no files — fails loudly instead of
+// passing vacuously.
+//
+// Because it is indexed by directory name, renaming an entry point turns this
+// red until the name here is updated. That is deliberate: a rename is exactly
+// when the walk silently reading the wrong tree is most likely.
 var knownManagerBuilders = []string{"lambda", "server", "sample"}
 
 // TestEntityConfigFromEnvWiredInEveryWriteEntryPoint guards that every entry point
@@ -34,23 +39,33 @@ var knownManagerBuilders = []string{"lambda", "server", "sample"}
 // the #314 staged rollout (VALIDATE_UPDATES_STRICT) is available on all
 // deployment targets rather than only the ones someone remembered to wire.
 //
-// It walks every directory under cmd/ and matches the "factory.NewEntityManager"
-// prefix, so an entry point added later is read without anyone updating this
-// test, and the WithConfig / WithConfigContext wrappers — and any future one —
-// all count. Before #321 the directories were a hardcoded pair and the match was
-// the exact WithConfigContext spelling, so the regression this comment names
-// could not fail it: a scratch cmd/apiworker went unread, and the already
-// existing cmd/sample used the unmatched WithConfig wrapper.
+// For each child directory of cmd/ it reads every non-test Go file in that
+// child's whole subtree and matches the "factory.NewEntityManager" prefix, so an
+// entry point added later is read without anyone updating this test, and the
+// WithConfig / WithConfigContext wrappers — and any future one — all count.
+// Before #321 the directories were a hardcoded pair and the match was the exact
+// WithConfigContext spelling, so the regression this comment names could not fail
+// it: a scratch cmd/apiworker went unread, and the already existing cmd/sample
+// used the unmatched WithConfig wrapper.
 //
-// Classification is per directory, not per file, so an entry point that builds
-// its manager outside main.go is still covered.
+// The unit of judgement is that whole subtree, not a single file. One cmd/<name>
+// tree is one binary, so an entry point that builds its manager outside main.go,
+// or applies the overlay from a sibling file or a helper subpackage, is wired and
+// must pass. The cost is the converse: a tree where the overlay call sits in a
+// file that is not on the path to the manager also passes. Attributing the
+// overlay to the file that builds the manager would trade that for false
+// failures on the ordinary split-across-files wiring, which is the more common
+// shape — and a substring scan cannot tell a reachable call from a dead one
+// either way.
 //
 // This is a source-level guard rather than a wiring test on purpose.
 // bootstrapLambda takes no injection seams and connects to a real database before
 // reaching the config assignment, so exercising the wiring would need a live
 // Postgres/DSQL — a heavyweight harness for a one-line assignment. What a source
-// scan cannot see remains: replacing the call with a comment naming the symbol
-// still passes.
+// scan cannot see remains, for the call and the overlay alike: writing either
+// symbol in a comment satisfies the scan without wiring anything. Observed, not
+// assumed — the first #321 probe passed on its own doc comment. Closing that
+// needs an AST parse (#429).
 //
 // It lives in this package because cmd/ has no shared test package.
 func TestEntityConfigFromEnvWiredInEveryWriteEntryPoint(t *testing.T) {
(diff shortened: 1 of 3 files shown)


iceboundrock:
## Reviewed by autonomous review session

Review scope: diff `eb1302c`...PR head (3 commits), against issue #321, `AGENTS.md`/`CLAUDE.md`, `coding-standard.md`. All mutation claims in the PR body were independently re-run (see Verification).

## Summary

PR #428 closes the two "guards that cannot fail" gaps from #321 (`RelationRoots.Covers` dot boundary unpinned; lambda entry-point guard hardcoded to two files and one exact call form), plus the one-line wiring fix in `cmd/sample` the widened guard exposes. Both guards now actually fail under the mutations their comments name. The work is small, well-documented, and — with two minor observations below — I recommend merging.

## Spec

Against issue #321's stated requirements:

- **Following the issue's direction for the guard is faithful.** The test now walks `cmd/`, reads each subdirectory's non-test Go files, and matches the `factory.NewEntityManager` prefix rather than a hardcoded two-file map and one exact call spelling. Verified against real call sites: `lambda`, `server`, `sample` all classify as builders.
- **The Covers split is the one the issue asked for** — a sibling (`contactSnapshotExtra`) sharing the root's characters but not its dot boundary — and the new test fails under the naive `HasPrefix(name, root) && name != root` mutation and only under it (verified).
- **One nuance vs. the issue's wording:** the issue says "asserting the sibling still expands." The new test asserts the *validation* half — that `contactSnapshotExtra.status`'s required policy is still enforced — not the expansion output. That is the half `Covers` actually controls (`relation_roots.go` notes the payload strip never consults this set), and the expansion half is already pinned by the twin `TestStripKeepsSamePrefixSibling` on `RelationIndex.coversRelationSubtree` (relation_roots.go:11-14; index test in `internal/relation`). So the boundary is covered on both halves; the deviation from the issue's literal wording is justified, though it would be worth a one-line comment noting the new test deliberately pins the validation half only.
- **Scope creep worth flagging:** `cmd/sample/main.go` adds a production behaviour change (`config.Entity = bootstrap.EntityConfigFromEnv(config.Entity)`) that the issue never literally requested. It is the natural resolution of the widened guard — the issue itself acknowledged the sample wrapper wouldn't match the old gate, and a walk that classifies `sample` as a builder must either wire the overlay or exempt it; wiring it is the correct choice. Non-blocking.

## Standards

Read `AGENTS.md`, `CLAUD.md` (referenced), and `coding-standard.md` against the diff. No hard violations.

- `scanEntryPointDir` wraps errors with `%w` and context (`fmt.Errorf("failed to read %s: %w", ...)`) — matches coding-standard.md:22-33, AGENTS Coding Standards. `%v` in `t.Errorf`/`t.Fatalf` assertions carries context; those are assertions, not returned errors. Judgement call: fine.
- Sizes: files 123/312/248 lines, functions 27–37 lines — all comfortably under 500/100.
- Nesting max depth 2; flat early-return structure throughout.
- The `internal/bootstrap` import in `cmd/sample` follows the established cmd/ pattern (lambda, server, tools already import it).

Judgement calls, non-blocking:
- `var knownManagerBuilders` is a package-scoped mutable var (test fixture, not application state — coding-standard.md:89's "global variables for application state" trigger doesn't strictly apply). A `var` is needed here to keep it out of the function's flow; acceptable.
- Longest added lines run ~138 chars (registry entries in the transform test) and ~97 chars (lambda test). No documented line-length limit and lll is not an enforced gate — readability note only.

## Assessment

The PR does what it says: both guards now fail for the exact regressions their comments name, and the wiring gap in `cmd/sample` (a real write path silently ignoring `VALIDATE_UPDATES_STRICT`) is closed. The mutation discipline is exemplary and honest (each mutation was watched failing before the fix, and the tree was verified clean after). The observations below are all latent or edge-case; none block merge.

## Observations

1. **Per-directory overlay flag can be satisfied by any file.** `scanEntryPointDir` sets `overlays=true` if *any* non-test file in the directory calls `EntityConfigFromEnv`, while `builds` is likewise per-directory. Today every `cmd/*` is a single package with the call in `main.go`, so this is equivalent — but a future directory whose manager-building file lacks the overlay while another file (e.g. a config helper) calls it would pass spuriously. Consider attributing `overlays` to the same file that set `builds`, or at least noting the assumption in the comment.
2. **Non-recursive walk.** The walk reads files only directly inside each `cmd/*` child directory. `cmd/sample/schemas/`, `cmd/server/schemas/` etc. are skipped only by luck of being data, not Go. A future nested entry point under `cmd/foo/sub/` would again be unread — partially the exact gap the issue names. The comment ("walks every directory under `cmd/`") slightly overstates what it reads; tightening it to "reads the non-test Go files directly in each `cmd/*` child directory" would make it true.
3. **The floor is name-indexed.** `knownManagerBuilders` renames a directory (e.g. `cmd/server` → `cmd/gateway`) will go red until the list is updated. That is the intended tripwire semantics, but it means the list is both a floor and a rename index — worth one line acknowledging renames intentionally (and loudly) require the list update.
4. **The measured blind spot.** The PR records that a comment naming `bootstrap.EntityConfigFromEnv` satisfies the scan. This is inherent to a source-level guard and stays acknowledged — but note the same evasion applies to the *new* walk: a future entry point that writes the symbol in a doc comment instead of calling it still passes. Closing it would need an AST parse; the trade-off is reasonable and correctly documented.

## Verification

Commands run on the PR head (worktree at `forma-321`, `GOCACHE`/`GOFLAGS` mirrored from the Makefile):

- `go test ./internal/transform -run 'TestFromEAVRecords' -v` → all 6 pass, including `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling`.
- `go test ./cmd/lambda -run 'TestEntityConfigFromEnvWiredInEveryWriteEntryPoint' -v` → passes.
- **Mutation 1 — naive `Covers`:** replaced `Covers` with `strings.HasPrefix(name, root) && name != root`. Result: exactly `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling` fails, all others pass. Restoration verified via diff.
- **Mutation 2 — overlay line deleted from `cmd/sample/main.go`:** walk test goes red with `cmd/sample builds an EntityManager but never calls bootstrap.EntityConfigFromEnv`.
- **Mutation 3 — walk root pointed at a tree with no `cmd` children** (`../server`): all three anti-vacuity assertions fire (`lambda`, `server`, `sample` not classified).
- `go build ./...` and `go vet ./cmd/... ./internal/transform/...` → clean.

Every claim in the PR's verification table was reproducible. Working tree clean on main after review.

iceboundrock:
## Review response — f2c7650

Three items addressed in code, one partially with reasoning, one answered
without a change. Follow-up filed as #429.

### Obs 2 — non-recursive walk (fixed in code, not by rewording)

Correct, and the substantive fix was cheap, so the comment was not merely
tightened: `scanEntryPointDir` now walks each `cmd/<name>` child's **whole
subtree** via `filepath.WalkDir`. One `cmd/<name>` tree is one binary, so that is
the honest unit.

Verified, and the recursion is load-bearing rather than decorative — a nested
probe at `cmd/scratchworker/sub/main.go` calling
`factory.NewEntityManagerWithConfig` with no overlay anywhere in the tree:

| Scan | Result |
|---|---|
| previous non-recursive version (via `-overlay` of `git show HEAD:...`) | **green** — the hole was real |
| this commit | red: `cmd/scratchworker builds an EntityManager but never calls bootstrap.EntityConfigFromEnv` |
| probe removed | green |

The comment now says what it reads: "for each child directory of cmd/ it reads
every non-test Go file in that child's whole subtree."

### Obs 1 — overlay attribution (assumption stated; the same-file suggestion declined)

The observation is right that `overlays` is satisfied by any file in the unit.
The suggested remedy — attribute `overlays` to the file that set `builds` — I
don't think is an improvement, for one reason: wiring split across files in the
same binary is the ordinary shape, not the pathological one. An entry point that
assembles its config in `config.go` and builds the manager in `main.go` is
correctly wired, and same-file attribution would fail it. That trades a rare
false pass for a common false failure. And a substring scan cannot distinguish a
reachable call from a dead one under either rule, so the stricter rule buys less
than it looks.

What was missing was the admission, which the comment now carries explicitly:
the unit of judgement is the subtree, an overlay call off the path to the manager
also passes, and here is why that trade was taken. The real fix is #429.

### Obs 3 — the floor is a rename index (comment)

Accurate. `knownManagerBuilders` now says so, and says it is deliberate: a rename
is precisely when the walk is most likely to be reading the wrong tree, so it
should be loud rather than silently re-derived.

### Obs 4 — the evasion applies to the new walk too (answered, no change)

Agreed, and worth being blunt: it applies to **both** symbols, not just the
overlay. A file that names `factory.NewEntityManager` only in prose is classified
as an entry point; a file that names `bootstrap.EntityConfigFromEnv` only in prose
is classified as wired. The comment previously mentioned only the second
direction; it now states both, notes that this was observed rather than assumed
(the first #321 probe passed on its own doc comment), and points at **#429** for
the AST fix that closes both this and obs 1.

No behaviour change here — closing it inside this PR means a `go/parser` pass,
which is more machinery than the diff warrants and is worth doing on its own.

### Spec nuance — "asserting the sibling still expands"

Taken as a comment. The new test pins the **validation** half deliberately,
because that is the only half `Covers` governs: the payload strip never consults
this set. The expansion half is the strip predicate's, held by the twin
`TestStripKeepsSamePrefixSibling`. The test comment now names that split and
points at the twin's file (`internal/relation_index_strip_test.go` — the review
placed it in `internal/relation`).

### Re-verification after the change

Every red proof from the original PR re-run against f2c7650, none stale:

- `Covers` → naive `HasPrefix`: red, and still only `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling`.
- walk root pointed at a tree with no `cmd` children: all 3 anti-vacuity assertions fire.
- `cmd/sample` overlay line deleted: red.
- nested probe: table above.
- `make fmt-check`, `make lint`, `make test`: clean; working tree clean after each probe.

iceboundrock:
## Reviewed by autonomous review session (round 2)

Second-pass review of PR head `f2c7650` against base `eb1302c` (4 commits: 83f88de, be38235, b5d27f3, f2c7650), issue #321, AGENTS.md/CLAUDE.md/coding-standard.md, plus the prior review session's findings. All mutation claims independently re-run on the PR head — see Verification. Recommendation: **approve**. Two judgement-call observations below, none blocking.

## Summary

The 4th commit (`f2c7650`) resolves the previous round's substantive finding (non-recursive walk) with the right fix and verified load-bearing recursion, and rewords the guard's contract honestly. The spec is fully implemented; the one wording deviation is sound and now documented in code. Standards hold (gofmt/lint/sizes clean); the only recurring concern is comment proportionality.

## Spec

Against issue #321's two requirements:

1. **Covers dot-boundary pin — implemented, with a documented, justified deviation.** The issue asked to "assert the sibling still expands"; the new test pins the *validation* half instead (the sibling's required policy stays enforced). That is the only observable half of `Covers`: its sole production call site is the required-policy check (`internal/transform/attribute_converter.go:254`); the payload strip never consults this set (`internal/transform/relation_roots.go:11-14`). The expansion half is already pinned by the pre-existing twin `TestStripKeepsSamePrefixSibling` (`internal/relation_index_strip_test.go:43`). The test comment now names that split and points at the twin's file (`relation_root_required_policy_test.go:149-154`). Deviation is faithful, not scope-creep.
2. **Entry-point guard walks `cmd/` and matches the `factory.NewEntityManager` prefix — implemented exactly as the issue directed**, plus the plan ruling (overlay wired into `cmd/sample`, no exemption list; anti-vacuity floor naming lambda/server/sample; evacuation acknowledged for both symbols, AST fix tracked as #429). The recursion in `f2c7650` closes the nested-entry-point hole the previous review found; it is genuinely load-bearing (verified below).

**Scope note (non-blocking):** `cmd/sample/main.go:129` is a small *production* behaviour change (`config.Entity = bootstrap.EntityConfigFromEnv(config.Entity)`) the issue's bare text only implied. The plan comment on #321 explicitly ruled it in ("wire it"), and wiring is the correct resolution of the widened guard. Worth an operator eye purely because it flips `VALIDATE_UPDATES_STRICT` on for the CSV importer; that is the intent.

## Standards

No hard violations. Files 144/312/252 lines, funcs 27–37 lines, max nesting 3 — all within AGENTS.md limits. `%w` wrapping everywhere except the canonically idiomatic bare `return err` in the `WalkDir` callback (`main_test.go:119`), whose outer boundary wraps with context (`main_test.go:141`); compliant in spirit.

Judgement calls:

1. **Doc comment disproportionate to code.** `TestEntityConfigFromEnvWiredInEveryWriteEntryPoint` carries a 34-line comment for a ~36-line test (`main_test.go:37-70`). It documents the directory-vs-file ruling, verification observations, and the `#429` pointer — valuable, but AGENTS.md's "Specs and Design Docs" rule says rulings and verification notes belong on the issue, not load-bearing in code. Trimming to ~10 lines (what it does, the #321 regression, #429 pointer) with the trade-off analysis on issue #321 would read better. Non-blocking.
2. **Package-scoped `var knownManagerBuilders`** (`main_test.go:35`) — immutable test-fixture data, not application state; coding-standard.md §7's global-state trigger doesn't strictly apply. The "rename turns red on purpose" intent is documented. Acceptable.
3. **18-line comment on the new transform test** (`relation_root_required_policy_test.go:139-154`) — heavy, but mostly navigational (names the twin and explains the split). The HasPrefix mutation backstory could compress.

## Assessment

Both guards now fail for exactly the mutations their comments name, and only for them. The recursion fix is the honest unit (one `cmd/<name>` tree = one binary) and was verified load-bearing rather than decorative. The `cmd/sample` wiring gap is real and correctly closed. Mutation discipline throughout the PR is exemplary and was independently reproducible. No changes required before merge; observations below are all latent or aesthetic.

## Observations

1. **The pin is tight.** Mutating `Covers` to naive `HasPrefix(name, root) && name != root` turns red exactly `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling` and nothing else in the transform package — a precise mutation anchor. The only other `Covers` caller is a test (`internal/relation_index_strip_test.go:76`, `require.False`), so the boundary is held on both sides of the production/test split.
2. **Subtree scan false-pass remains acknowledged, not closed.** A future tree whose manager-building file lacks the overlay while a sibling file calls it (or names the symbol in a comment) still passes. The comment states this and why same-file attribution was rejected (`main_test.go:51-59`); the AST fix (#429) is the real closure. Reasonable.
3. **Verification claims trace to the PR body and commit messages** — the nested-probe red proof, the "first probe passed on its own doc comment" observation, and the anti-vacuity floor are all reproducible; nothing in the body was hand-waved.

## Verification

All commands run on the PR head worktree (`GOCACHE`/`GOFLAGS` mirrored from the Makefile), each mutation confirmed red-then-green with the tree verified clean after:

- `go test . ./cdc ./cmd/... ./factory ./internal/...` → all packages pass.
- `go test ./internal/transform -run TestFromEAVRecords -v` → 6/6 pass incl. `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling`.
- **Mutation 1 — naive `Covers`:** only `TestFromEAVRecordsEnforcesRequiredPolicyOnSamePrefixSibling` fails; all other `TestFromEAVRecords*` pass.
- **Mutation 2 — overlay line deleted from `cmd/sample/main.go`:** walk test red: `cmd/sample builds an EntityManager but never calls bootstrap.EntityConfigFromEnv`.
- **Mutation 3 — walk root pointed at a tree with no `cmd` children:** all three anti-vacuity assertions fire (`lambda`, `server`, `sample`).
- **Mutation 4 — nested probe at `cmd/scratchworker/sub/main.go`:** recursive scan red; **emulated non-recursive scan green** (the hole was real, not theoretical); probe removed → green.
- `gofmt -l` on the 3 changed files and `make fmt-check` → clean. CI on head (`f2c7650`): Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke all SUCCESS.

Issue #431, opened 2026-08-20
refactor(e2e): NewFederatedTestHarness is 131 lines, past the 100-line hard limit — blocks widening the function guard to harness*.go


iceboundrock:
`internal/e2e_harness/federated/harness.go`: `NewFederatedTestHarness` measures **131 lines**, past AGENTS.md's 100-line hard limit for functions (not merely past coding-standard.md §7's 80-line refactoring trigger). `initDatabaseSchema` in the same file is 70 lines, inside the limit but worth a glance while there.

This is filed separately from #220's note about `harness.go` being 501 lines: that one is about file size, this is a distinct rule and a distinct function.

## Why it is blocking something

#324 added an AST-based function-length guard at the 80-line trigger, scoped to `query*.go`. The natural scope would have been the existing `listGuardedSourceFiles()` set, which the 500-line file guard already uses — but that set also covers `harness*.go`, where this function would make the guard unsatisfiable on arrival. Including it would have dragged a `harness.go` refactor into a PR that owns neither.

So the guard's scope is currently narrower than the file guard's, with the reason recorded in `query_function_size_test.go`'s `listFunctionGuardedSourceFiles`. Once this function is under the trigger, the two guards can share one file list and that comment goes away.

## Suggested shape

`NewFederatedTestHarness` reads as several phases — config resolution, container/infra startup, database and DuckDB wiring, and harness assembly. PR #302 already moved some infra concerns into `harness_infra.go` (104 lines), which is the natural home for more of it.

Whoever takes this should also widen the function guard to `harness*.go` in the same change, so the fix and its guard land together.

Refs #324, #220

iceboundrock:
补注：本 issue 低估了自身的阻塞面 —— 把函数守卫收敛到裸 `*.go` 有**两个**越线函数，不止 `NewFederatedTestHarness`。

PR #446（已合并，`15cc13b`）把文件守卫拓宽到全包后，用 go/ast 扫描了 `internal/e2e_harness/federated` 全部 14 个非测试文件，越过 80 行触发线的函数：

| 行数 | 文件 | 函数 |
|-----:|------|------|
| 131 | `harness.go` | `NewFederatedTestHarness` |
| 84 | `helpers.go` | `loadExternalFederatedConfigFromEnv` |

含义：

[...]

Issue #430, opened 2026-08-20
refactor(e2e): dedupe ExecutePostgresQuery's scan loop against scanQueryResults; both miss rows.Err()


iceboundrock:
Found while planning #324 and deliberately left out of its scope (user ruling recorded in that issue's plan comment): the two items #324 names are pure string/control-flow refactors that untagged unit tests fully gate, whereas this one changes a scan destination type and can only be verified by the docker e2e suite.

## 1. The scan loop is duplicated

`internal/e2e_harness/federated/query.go`: `ExecutePostgresQuery` inlines a row-scanning loop that duplicates `scanQueryResults` almost line for line — roughly 50 similar lines, larger than the 18-line block #324 was filed for, and well over coding-standard.md §2's duplication gate.

The two are identical except for the `tradeTime` scan destination:

- `scanQueryResults` scans into `any` and passes it through `normalizeBenchmarkTradeTimeValue`, because DuckDB returns the column as several different Go types.
- `ExecutePostgresQuery` scans into `sql.NullInt64` directly.

`normalizeBenchmarkTradeTimeValue`'s `int64` branch already covers what the Postgres driver returns for BIGINT, so a shared helper looks feasible — but confirm that empirically rather than by reading, and note that `ExecutePostgresQuery` also builds its `PersistentRecord` with map literals where `scanQueryResults` uses `make`, so the merge is not purely mechanical.

Deduping this also drops `ExecutePostgresQuery` from 78 lines to roughly 35. That matters: #324 added an 80-line function guard over `query*.go`, and 78 leaves two lines of headroom, so any small addition to that function trips the guard until this is done.

## 2. Neither loop checks `rows.Err()`

Both iterate `for rows.Next()` and return without consulting `rows.Err()`. An error part-way through iteration therefore ends the loop silently and the caller receives a truncated result set as if it were complete. In a harness whose job is asserting record counts across tiers, a silent truncation is the worst possible failure mode — it reads as a data-correctness bug in the system under test.

This is a bug fix rather than a refactor, but it shares a fix site with item 1, which is why they are filed together.

## Verification

`make test` does not reach either loop. The gate is `go test ./internal/e2e_harness/federated/... -tags=e2e -timeout=30m` — and note that `-short` (what CI runs) skips `data_tier_test.go`, `merge_on_read_test.go`, `deduplication_test.go`, `soft_delete_test.go` and `performance_test.go`, so a `-short` pass is not sufficient evidence here.

Refs #324

[...]

PR #323, opened 2026-07-26
refactor(e2e): #220 split federated query builders

iceboundrock:
## Summary
- split federated query construction into focused hot-tier and PostgreSQL-only builder files
- keep public query entry points, execution, and scanning in `query.go`
- add a source-size regression guard for the query files

This is the federated-query slice of #220. The HTTP server slice remains deferred until #282 merges.

## Deliberate SQL Whitespace Deviation
`98d9665` removes trailing spaces from six lines in moved harness-only raw SQL. The change is non-semantic, kept in a separate `style:` commit, and recorded here because the #220 move guidance calls for byte-identical SQL text.

## Verification
- `make test`
- `make lint`
- `go test -v ./internal/e2e_harness/federated/... -tags=e2e -short -timeout=30m`

PR #432, opened 2026-08-20
refactor(e2e): #324 dedupe Postgres-only scan source, split ExecuteFederatedQuery, guard function size

### LINKED ISSUE #324: refactor(e2e): federated query cleanups deferred from PR #323 — dedupe Postgres-only select/count SQL; ExecuteFederatedQuery past the 80-line trigger
Recorded from PR #323's review (the federated-query slice of #220). Both items are pre-existing in the moved code; that PR was a pure move, so they were deferred per coding-standard.md's Incremental Improvement Mode rather than fixed in-flight.

## 1. Dedupe the Postgres-only select/count SQL

`internal/e2e_harness/federated/query_postgres_build.go`: `buildPostgresOnlySelectQuery` and `buildPostgresOnlyCountQuery` duplicate the entire FROM/JOIN/WHERE block — the `change_log`/`entity_main` joins, the `hot_vals` EAV pivot subquery with its 12 attr-ID format args, and the `flushed_at = 0` guard (~18 identical lines each, at roughly `:35-53` and `:66-84` on PR #323's head). That is over coding-standard.md §2's duplication gate (>15 similar lines).

Fix shape: extract a shared fragment builder (e.g. `buildPostgresOnlyFromWhere(attrIDs) string`) that both builders append after their respective SELECT lists; the trailing filter/order/limit handling stays per-builder. Behavior-preserving — the existing Postgres-only unit/e2e coverage is the guardrail.

## 2. `ExecuteFederatedQuery` is ~94 lines

`internal/e2e_harness/federated/query.go`: `ExecuteFederatedQuery` measures ~94 lines on PR #323's head, past coding-standard.md §7's >80-line refactoring trigger. Untouched by the split (it stayed in `query.go` as an entry point). Natural seam observed during the review: tier probing / plan assembly vs execution control flow.

## Not in scope here

- `harness.go` at 501 lines is already tracked by #220 ("trim opportunistically the next time that file is touched") — not duplicated into this issue.
- The size-guard test added in PR #323 intentionally covers only the `query*.go` family; whether it should widen is a separate conversation, not requested here.

Refs #220. Deferred from PR #323 (see its review comments for the original findings).

iceboundrock:
Closes #324

The two cleanups PR #323 deferred, plus the guard that keeps the second one from
coming back. Execution plan: https://github.com/Lychee-Technology/forma/issues/324#issuecomment-5358570578

## 1. Dedupe the Postgres-only select/count SQL

`buildPostgresOnlySelectQuery` and `buildPostgresOnlyCountQuery` each carried the
whole FROM/JOIN/WHERE block — the `change_log`/`entity_main` joins, the
`hot_vals` EAV pivot with its twelve attribute-ID format args, and the
unflushed-and-live guard. Sixteen identical SQL lines per copy, over
coding-standard.md §2's duplication gate.

Extracted `buildPostgresOnlyScanSource`; filter/order/limit handling stays
per-builder, as the issue proposed.

**Byte-identical output**, verified by diffing the rendered SQL and args across
the change for both a fully-filtered and a bare `QueryOptions` — the only
difference in the captured output was the `go test` timing line.

## 2. `ExecuteFederatedQuery`: 94 → 68 lines

Split along the seam the #323 review identified. Three extractions, all in
`query.go`:

- `executePreferHotQuery` — the `PreferHot` override is an alternative execution
  path, not a step in the federated one.
- `federatedTierState` + `probeFederatedTiers` — one probe of parquet
  availability plus the dirty set. `isEmpty()` names the Postgres-fallback
  condition that was a three-clause boolean; `executionPlan()` collapses three
  verbose `buildExecutionPlan` call sites. `buildExecutionPlan` itself is
  unchanged.
- `emptyPageResult` — the short-circuit answer for a page past the last row.
  This one is beyond what the plan listed: the first two extractions landed at
  75 lines, and the plan promised headroom rather than five lines of slack.

Same order of operations, same fallbacks, same plan notes.

## 3. A function-length guard, so it cannot regrow

`ExecuteFederatedQuery` reached 94 lines because the package guards file size and
nothing else. New `query_function_size_test.go` walks non-test `query*.go` with
`go/parser` and fails any function past coding-standard.md §7's 80-line trigger.

Scoped to `query*.go` rather than the existing `listGuardedSourceFiles()` set,
which also covers `harness*.go` — where `NewFederatedTestHarness` is 131 lines,
already past AGENTS.md's 100-line hard limit. Including it would have made the
guard unsatisfiable on arrival. Reason is recorded in the source; the function is
now #431.

**Commit 3 is intentionally red** and commit 4 greens it. A guard committed
alongside its own fix is a guard nobody has watched fail (#321).

## Evidence

Every claim below was run, not reasoned about.

| Check | Result |
|---|---|
| `make test` | pass |
| `make fmt-check` | pass |
| `make lint` (golangci-lint v1.64.8) | pass |
| `go test ./internal/e2e_harness/federated/... -tags=e2e -short` | pass |
| `go test ./internal/e2e_harness/federated/ -tags=e2e` (full, no `-short`) | 123 pass, 0 fail, 0 skip, 246s |
| Guard red before the fix | `query.go: ExecuteFederatedQuery has 94 lines, exceeds the 80-line refactoring trigger` |
| Guard green after | pass; `ExecuteFederatedQuery` measures 68 |

**Mutation proofs** (the two new tests can actually fail):

1. Item 1's guard: swapping `attrIDs.symbol`/`attrIDs.exchange` in
   `buildPostgresOnlyCountQuery` alone turns
   `TestPostgresOnlyCountSharesSelectScanSource` red while the pre-existing
   `TestBuildPostgresOnlyQueriesSupportBenchmarkFilters` stays **green** — the
   old test only does `strings.Contains` checks, so it never saw that class of
   drift. That is the coverage the new test adds.
2. Item 3's guard: an 86-line function added to `query_hot.go` is reported by
   name, confirming the guard scans the whole `query*.go` family and not just
   the file this PR shrank.

## Two things worth knowing

**`go test -overlay` cannot mutation-test a guard that reads source from disk.**
My first attempt at proof 2 used `-overlay` (per the #354 practice of mutating
without dirtying the tree) and the guard passed — a false green. The overlay
only affects compilation; `TestGuardedFunctionsStayWithinRefactoringTrigger`
calls `os.ReadFile`, so it read the unmutated file. Redone as a real on-disk edit
restored from a copy. The same trap applies to the existing file-size guard.

**`-tags=e2e -short`, which is what CI runs, skips `data_tier_test.go`,
`merge_on_read_test.go`, `deduplication_test.go`, `soft_delete_test.go` and
`performance_test.go`.** It still exercises `ExecuteFederatedQuery` through
`consistency_test.go` and `failure_modes_test.go`, so CI's coverage of item 2 is
partial rather than absent — but the full suite was run locally for this PR
because a `-short` pass alone would not have been honest evidence. Flagging the
gap as an observation; not filing it, since CI runtime is presumably the reason.

## Follow-ups filed

- **#430** — `ExecutePostgresQuery`'s inline scan loop duplicates
  `scanQueryResults` (~50 lines, larger than what this issue was filed for), and
  neither loop checks `rows.Err()`, so a mid-iteration error silently truncates
  the result set. Kept out of this PR by user ruling: it changes a scan
  destination type and only the docker e2e suite can verify it. It also drops
  `ExecutePostgresQuery` from 78 lines to about 35 — relevant, because 78 against
  the new 80-line guard is two lines of headroom.
- **#431** — `NewFederatedTestHarness` at 131 lines, which is what blocks
  widening the function guard to `harness*.go`.

Out of scope and unchanged, per the issue: `harness.go`'s file size (#220) and
the 500-line file guard's scope.

### DIFF
--- a/internal/e2e_harness/federated/query.go
+++ b/internal/e2e_harness/federated/query.go
@@ -29,13 +29,13 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 	tradeTimeOnlyProjection := usesTradeTimeOnlyBenchmarkProjectionForSelect(opts)
 	if needsBenchmarkDuckDBMacros(opts, benchmarkProjection, tradeTimeOnlyProjection) {
 		if err := prepareBenchmarkDuckDBMacros(ctx, h); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("prepare benchmark duckdb macros: %w", err)
 		}
 	}
 
 	tiers, err := h.probeFederatedTiers(ctx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("probe federated tiers: %w", err)
 	}
 	if tiers.isEmpty() {
 		return h.ExecutePostgresQuery(ctx, opts)
@@ -57,7 +57,8 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 		return nil, fmt.Errorf("count query: %w", err)
 	}
 	if opts.CountOnly {
-		return &QueryResult{TotalRecords: totalRecords, Duration: time.Since(start), Plan: tiers.executionPlan(time.Since(start))}, nil
+		countDuration := time.Since(start)
+		return &QueryResult{TotalRecords: totalRecords, Duration: countDuration, Plan: tiers.executionPlan(countDuration)}, nil
 	}
 	if shouldSkipFederatedSelect(totalRecords, opts.Offset) {
 		return tiers.emptyPageResult(totalRecords, time.Since(start)), nil
@@ -74,7 +75,7 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 
 	records, err := h.scanQueryResults(rows, benchmarkProjection)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("scan query results: %w", err)
 	}
 
 	duration := time.Since(start)
--- a/internal/e2e_harness/federated/query.go
+++ b/internal/e2e_harness/federated/query.go
@@ -94,7 +94,7 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 func (h *FederatedTestHarness) executePreferHotQuery(ctx context.Context, opts *QueryOptions, start time.Time) (*QueryResult, error) {
 	result, err := h.ExecutePostgresQuery(ctx, opts)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("prefer-hot postgres query: %w", err)
 	}
 	if result.Plan == nil {
 		result.Plan = &model.ExecutionPlan{Notes: []string{}, Timings: map[string]int64{}}
@@ -119,6 +119,10 @@ func (s federatedTierState) isEmpty() bool {
 	return !s.hasBaseFiles && !s.hasDeltaFiles && len(s.dirtyIDs) == 0
 }
 
+// executionPlan adapts the probe's three fields to buildExecutionPlan, which is
+// the plan vocabulary shared with buildPostgresOnlyExecutionPlan. It exists to
+// keep the four-argument unpacking out of three call sites in
+// ExecuteFederatedQuery, not to add a layer over the builder.
 func (s federatedTierState) executionPlan(duration time.Duration) *model.ExecutionPlan {
 	return buildExecutionPlan(len(s.dirtyIDs), s.hasBaseFiles, s.hasDeltaFiles, duration)
 }
@@ -142,7 +146,7 @@ func (s federatedTierState) emptyPageResult(totalRecords int64, duration time.Du
 func (h *FederatedTestHarness) probeFederatedTiers(ctx context.Context) (federatedTierState, error) {
 	hasBaseFiles, hasDeltaFiles, err := h.checkTierFiles(ctx)
 	if err != nil {
-		return federatedTierState{}, err
+		return federatedTierState{}, fmt.Errorf("check tier files: %w", err)
 	}
 	dirtyIDs, err := h.getDirtyIDs(ctx)
 	if err != nil {
--- a/internal/e2e_harness/federated/query.go
+++ b/internal/e2e_harness/federated/query.go
@@ -23,17 +23,7 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 	opts = normalizeQueryOptions(opts)
 	start := time.Now()
 	if opts.PreferHot {
-		result, err := h.ExecutePostgresQuery(ctx, opts)
-		if err != nil {
-			return nil, err
-		}
-		if result.Plan == nil {
-			result.Plan = &model.ExecutionPlan{Notes: []string{}, Timings: map[string]int64{}}
-		}
-		result.Plan.Notes = append(result.Plan.Notes, "prefer_hot_override", "postgres_only_execution")
-		result.Plan.Timings["total"] = time.Since(start).Milliseconds()
-		result.Duration = time.Since(start)
-		return result, nil
+		return h.executePreferHotQuery(ctx, opts, start)
 	}
 	benchmarkProjection := usesBenchmarkProjectionForSelect(opts)
 	tradeTimeOnlyProjection := usesTradeTimeOnlyBenchmarkProjectionForSelect(opts)
@@ -43,20 +33,11 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 		}
 	}
 
-	// Check which tiers have parquet files
-	hasBaseFiles, hasDeltaFiles, err := h.checkTierFiles(ctx)
+	tiers, err := h.probeFederatedTiers(ctx)
 	if err != nil {
 		return nil, err
 	}
-
-	// Get dirty IDs from change_log
-	dirtyIDs, err := h.getDirtyIDs(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("get dirty ids: %w", err)
-	}
-
-	// If no parquet files exist and no hot records, fall back to Postgres-only query
-	if !hasBaseFiles && !hasDeltaFiles && len(dirtyIDs) == 0 {
+	if tiers.isEmpty() {
 		return h.ExecutePostgresQuery(ctx, opts)
 	}
 
@@ -65,8 +46,8 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 	deltaPath := fmt.Sprintf("s3://%s/%s/%d/delta/*.parquet", h.S3Bucket, h.S3Prefix, h.SchemaID)
 
 	// Build and execute the federated query
-	query := h.buildFederatedQuerySQLDynamic(basePath, deltaPath, hasBaseFiles, hasDeltaFiles, dirtyIDs, opts)
-	countQuery := h.buildFederatedQueryCountSQLDynamic(basePath, deltaPath, hasBaseFiles, hasDeltaFiles, dirtyIDs, opts)
+	query := h.buildFederatedQuerySQLDynamic(basePath, deltaPath, tiers.hasBaseFiles, tiers.hasDeltaFiles, tiers.dirtyIDs, opts)
+	countQuery := h.buildFederatedQueryCountSQLDynamic(basePath, deltaPath, tiers.hasBaseFiles, tiers.hasDeltaFiles, tiers.dirtyIDs, opts)
 
 	var totalRecords int64
 	if err := h.Duck.DB.QueryRowContext(ctx, countQuery).Scan(&totalRecords); err != nil {
@@ -76,17 +57,10 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 		return nil, fmt.Errorf("count query: %w", err)
 	}
 	if opts.CountOnly {
-		return &QueryResult{TotalRecords: totalRecords, Duration: time.Since(start), Plan: buildExecutionPlan(len(dirtyIDs), hasBaseFiles, hasDeltaFiles, time.Since(start))}, nil
+		return &QueryResult{TotalRecords: totalRecords, Duration: time.Since(start), Plan: tiers.executionPlan(time.Since(start))}, nil
 	}
 	if shouldSkipFederatedSelect(totalRecords, opts.Offset) {
-		plan := buildExecutionPlan(len(dirtyIDs), hasBaseFiles, hasDeltaFiles, time.Since(start))
-		plan.Notes = append(plan.Notes, "empty_page_short_circuit")
-		return &QueryResult{
-			Records:      nil,
-			TotalRecords: totalRecords,
-			Duration:     time.Since(start),
-			Plan:         plan,
-		}, nil
+		return tiers.emptyPageResult(totalRecords, time.Since(start)), nil
 	}
 
 	rows, err := h.Duck.DB.QueryContext(ctx, query)
@@ -104,7 +78,7 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 	}
 
 	duration := time.Since(start)
-	plan := buildExecutionPlan(len(dirtyIDs), hasBaseFiles, hasDeltaFiles, duration)
+	plan := tiers.executionPlan(duration)
 
 	return &QueryResult{
 		Records:      records,
@@ -112,6 +86,69 @@ func (h *FederatedTestHarness) ExecuteFederatedQuery(ctx context.Context, opts *
 		Duration:     duration,
 		Plan:         plan,
 	}, nil
+}
+
+// executePreferHotQuery serves the PreferHot override, which is a separate
+// execution path rather than a step in the federated one: it answers straight
+// from Postgres and only annotates the plan to say why.
+func (h *FederatedTestHarness) executePreferHotQuery(ctx context.Context, opts *QueryOptions, start time.Time) (*QueryResult, error) {
+	result, err := h.ExecutePostgresQuery(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+	if result.Plan == nil {
+		result.Plan = &model.ExecutionPlan{Notes: []string{}, Timings: map[string]int64{}}
+	}
+	result.Plan.Notes = append(result.Plan.Notes, "prefer_hot_override", "postgres_only_execution")
+	result.Plan.Timings["total"] = time.Since(start).Milliseconds()
+	result.Duration = time.Since(start)
+	return result, nil
+}
+
+// federatedTierState is what one probe of the three tiers found: which parquet
+// tiers have files, and which row IDs are still unflushed in Postgres.
+type federatedTierState struct {
+	hasBaseFiles  bool
+	hasDeltaFiles bool
+	dirtyIDs      []uuid.UUID
+}
+
+// isEmpty reports that there is nothing for DuckDB to federate — no parquet in
+// either tier and no hot rows — so the query belongs on Postgres alone.
+func (s federatedTierState) isEmpty() bool {
+	return !s.hasBaseFiles && !s.hasDeltaFiles && len(s.dirtyIDs) == 0
+}
+
+func (s federatedTierState) executionPlan(duration time.Duration) *model.ExecutionPlan {
+	return buildExecutionPlan(len(s.dirtyIDs), s.hasBaseFiles, s.hasDeltaFiles, duration)
+}
+
+// emptyPageResult answers a page that starts past the last row: the count is
+// still real, so it is reported, but no select is worth running.
+func (s federatedTierState) emptyPageResult(totalRecords int64, duration time.Duration) *QueryResult {
+	plan := s.executionPlan(duration)
+	plan.Notes = append(plan.Notes, "empty_page_short_circuit")
+	return &QueryResult{
+		Records:      nil,
+		TotalRecords: totalRecords,
+		Duration:     duration,
+		Plan:         plan,
+	}
+}
+
+// probeFederatedTiers gathers everything the query shape depends on: parquet
+// availability per tier plus the dirty set that decides which S3 rows are
+// discarded in favour of their hot versions.
+func (h *FederatedTestHarness) probeFederatedTiers(ctx context.Context) (federatedTierState, error) {
+	hasBaseFiles, hasDeltaFiles, err := h.checkTierFiles(ctx)
+	if err != nil {
+		return federatedTierState{}, err
+	}
+	dirtyIDs, err := h.getDirtyIDs(ctx)
+	if err != nil {
+		return federatedTierState{}, fmt.Errorf("get dirty ids: %w", err)
+	}
+	return federatedTierState{hasBaseFiles: hasBaseFiles, hasDeltaFiles: hasDeltaFiles, dirtyIDs: dirtyIDs}, nil
 }
 
 // normalizeQueryOptions sets default values for query options.
--- a/internal/e2e_harness/federated/query_file_size_test.go
+++ b/internal/e2e_harness/federated/query_file_size_test.go
@@ -43,9 +43,12 @@ func countSourceLines(source []byte) int {
 	return lines
 }
 
-func listGuardedSourceFiles() ([]string, error) {
+// listNonTestSources collects the non-test files matching the given patterns.
+// The two size guards apply the same rule to different scopes, so the scope is
+// the argument and the rule lives here once.
+func listNonTestSources(patterns ...string) ([]string, error) {
 	var files []string
-	for _, pattern := range []string{"query*.go", "harness*.go"} {
+	for _, pattern := range patterns {
 		candidates, err := filepath.Glob(pattern)
 		if err != nil {
 			return nil, fmt.Errorf("glob guarded source files %q: %w", pattern, err)
@@ -57,9 +60,13 @@ func listGuardedSourceFiles() ([]string, error) {
 		}
 	}
 	if len(files) == 0 {
-		return nil, fmt.Errorf("no guarded source files matched query*.go or harness*.go")
+		return nil, fmt.Errorf("no guarded source files matched %s", strings.Join(patterns, " or "))
 	}
 	return files, nil
+}
+
+func listGuardedSourceFiles() ([]string, error) {
+	return listNonTestSources("query*.go", "harness*.go")
 }
 
 // TestGuardedSourcesStayWithinFileSizeLimit prevents query assembly and harness
--- a/internal/e2e_harness/federated/query_function_size_test.go
+++ b/internal/e2e_harness/federated/query_function_size_test.go
@@ -1,0 +1,143 @@
+package federated
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// maxGuardedFunctionLines is coding-standard.md §7's refactoring trigger, which
+// is deliberately tighter than AGENTS.md's 100-line hard limit: the point is to
+// catch a function on its way past 80, not once it is already unmaintainable.
+const maxGuardedFunctionLines = 80
+
+type guardedFunction struct {
+	name  string
+	lines int
+}
+
+// measureFunctionLines reports the span of every function body in one source
+// file, from its func keyword through its closing brace. A doc comment sits
+// before FuncDecl.Pos() and so is not counted — the guard measures code, not
+// the explanation of it.
+func measureFunctionLines(filename string, source []byte) ([]guardedFunction, error) {
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, filename, source, 0)
+	if err != nil {
+		return nil, fmt.Errorf("parse %s: %w", filename, err)
+	}
+
+	var functions []guardedFunction
+	for _, decl := range file.Decls {
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Body == nil {
+			continue
+		}
+		functions = append(functions, guardedFunction{
+			name:  fn.Name.Name,
+			lines: fset.Position(fn.End()).Line - fset.Position(fn.Pos()).Line + 1,
+		})
+	}
+	return functions, nil
+}
+
+// listFunctionGuardedSourceFiles deliberately does not reuse
+// listGuardedSourceFiles: that one also covers harness*.go, where
+// NewFederatedTestHarness is already past the 100-line hard limit, so pulling it
+// in would make this guard unsatisfiable without a harness.go refactor this
+// change does not own. Widening the scope is the follow-up's job (#324).
+func listFunctionGuardedSourceFiles() ([]string, error) {
+	candidates, err := filepath.Glob("query*.go")
+	if err != nil {
+		return nil, fmt.Errorf("glob function-guarded source files: %w", err)
+	}
+
+	var files []string
+	for _, name := range candidates {
+		if !strings.HasSuffix(name, "_test.go") {
+			files = append(files, name)
+		}
+	}
+	if len(files) == 0 {
+		return nil, fmt.Errorf("no function-guarded source files matched query*.go")
+	}
+	return files, nil
+}
+
+func TestMeasureFunctionLinesSpansFuncKeywordToClosingBrace(t *testing.T) {
+	source := []byte(`package fixture
+
+// Doc comments precede the func keyword and must not be counted.
+// This second line would inflate the measurement if they were.
+func measured() int {
+	x := 1
+	return x
+}
+
+func declared()
+`)
+
+	functions, err := measureFunctionLines("fixture.go", source)
+	if err != nil {
+		t.Fatalf("measure function lines: %v", err)
+	}
+	if len(functions) != 1 {
+		t.Fatalf("expected only the function with a body, got %v", functions)
+	}
+	if functions[0].name != "measured" {
+		t.Fatalf("measured the wrong function: %q", functions[0].name)
+	}
+	if functions[0].lines != 4 {
+		t.Fatalf("measured %d lines, want 4 (func keyword through closing brace)", functions[0].lines)
+	}
+}
+
+func TestListFunctionGuardedSourceFilesScopesToQuerySources(t *testing.T) {
+	files, err := listFunctionGuardedSourceFiles()
+	if err != nil {
+		t.Fatalf("list function-guarded source files: %v", err)
+	}
+	if len(files) == 0 {
+		t.Fatal("expected function-guarded source files")
+	}
+	for _, name := range files {
+		if strings.HasSuffix(name, "_test.go") {
+			t.Errorf("test file %s included in function guard", name)
+		}
+		if !strings.HasPrefix(name, "query") {
+			t.Errorf("%s is outside the query*.go scope this guard can hold", name)
+		}
+	}
+}
+
+// TestGuardedFunctionsStayWithinRefactoringTrigger keeps the query assembly
+// entry points from accumulating back into oversized functions. ExecuteFederatedQuery
+// reached 94 lines because nothing was watching; this is what watches (#324).
+func TestGuardedFunctionsStayWithinRefactoringTrigger(t *testing.T) {
+	names, err := listFunctionGuardedSourceFiles()
+	if err != nil {
+		t.Fatalf("list function-guarded source files: %v", err)
+	}
+
+	for _, name := range names {
+		source, err := os.ReadFile(name)
+		if err != nil {
+			t.Fatalf("read %s: %v", name, err)
+		}
+		functions, err := measureFunctionLines(name, source)
+		if err != nil {
+			t.Fatalf("measure %s: %v", name, err)
+		}
+		for _, fn := range functions {
+			if fn.lines > maxGuardedFunctionLines {
+				t.Errorf("%s: %s has %d lines, exceeds the %d-line refactoring trigger",
+					name, fn.name, fn.lines, maxGuardedFunctionLines)
+			}
+		}
+	}
+}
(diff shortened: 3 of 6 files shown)


iceboundrock (commented):
# Review — PR #432 (#324 federated query cleanups)

Reviewed against `main` (merge-base) on branch `refactor/324-federated-query-cleanups` @ `85fa611`. Two-axis review (Standards + Spec), plus independent verification.

## Summary

PR #432 faithfully implements every requirement in issue #324: the Postgres-only select/count SQL is deduped with **byte-identical** output, `ExecuteFederatedQuery` is down from 94 to 68 lines with all plan-named extractions present, and the new 80-line function guard is genuinely red-first (#321-compliant by construction). Zero spec gaps, zero wrong-looking implementations. **2 hard standards violations and 2 judgement calls, all minor** — both hard items are inherited error-return lines that the extraction happened to touch; neither has correctness impact. Ready to merge once the two wrapping comments are addressed (inline or as a tracked follow-up).

## Findings

- **F1 — Hard (Standards).** `query.go:97`, `executePreferHotQuery`: `return nil, err` is a bare error return with no context, violating AGENTS.md / coding-standard.md §3 ("never bare `return err`"). This is a pre-existing line the extraction moved into a new function; per AGENTS.md ("fix the immediate area") it should read `fmt.Errorf("execute postgres query for prefer-hot override: %w", err)` while being touched.
- **F2 — Hard (Standards).** `query.go:144-146`, `probeFederatedTiers`: the `checkTierFiles` error is returned bare (`return federatedTierState{}, err`) while the sibling `getDirtyIDs` error one line below is wrapped. An inconsistency introduced **inside one new 13-line function** — the strongest of the two, worth fixing.
- **F3 — Judgement (Standards).** `query.go:122`, `federatedTierState.executionPlan`: a thin 1:1 delegator (ai_review_guideline §4.5 Middle Man / heuristic 16) and the only method on the struct without a doc comment, while `isEmpty`/`emptyPageResult` have one. Defensible design (collocates the 4-field data clump, keeps `ExecuteFederatedQuery` flat); a one-line godoc would match the established pattern.
- **F4 — Judgement (Standards).** `query_function_size_test.go:61-73`, `listFunctionGuardedSourceFiles`: duplicates `listGuardedSourceFiles`' glob/exclude logic (~12 shared lines, near §2's ~15-line duplication trigger). Deliberate and documented; the exclusion rationale is verified accurate (`NewFederatedTestHarness` measures 131 lines, past the 80-line trigger, so widening would break the guard on arrival).

## Spec

- **Missing / Partial: none.** Item 1 — `buildPostgresOnlyScanSource` is a pure function of `attrIDs`, appended after each builder's own SELECT list; the count still owns its `SELECT COUNT(*)` line; filter/order/limit stay per-builder; byte-identity holds by construction (`"\n\t\tSELECT COUNT(*)"` + the `"\n\t\tFROM …"` fragment concatenate to exactly the prior bytes). Item 2 — all plan-named pieces present (`executePreferHotQuery`, `federatedTierState`, `probeFederatedTiers`, `isEmpty()`, `executionPlan()`); all three former `buildExecutionPlan` call sites replaced with identical argument order; `buildExecutionPlan` itself byte-unchanged. Item 3 — guard scans exactly the four non-test `query*.go` files, excludes `_test.go` and `harness*.go` by glob.
- **Scope additions (all documented).** The third extraction `emptyPageResult` (behavior-preserving, also removes a doubled `time.Since(start)`); the new `TestPostgresOnlyCountSharesSelectScanSource` plus two small aux tests in the guard file. The mid-execution ruling behind `emptyPageResult` was posted to issue #324 (execution-notes comment — the issue carries both plan and notes), so the AGENTS.md decision-record requirement is satisfied.
- **Wrong-looking: none.**
- **Test-quality note.** The prefix test is near-tautological against the current code (both builders invoke the same helper; it only guards future re-inlining) — but the comment says exactly that, the single-`Conditions` constraint correctly avoids the map-order flake, and the mutation claim is plausible (the legacy `strings.Contains` test cannot see reordered attr-ID args).

## Standards

Hard violations: **2** (F1, F2 — both bare error returns on lines moved/extracted by this PR). Judgement calls: **2** (F3, F4). No file/function-size breaches, no nesting depth issues, no error-message omissions on new messages, comments meet the intent-not-mechanics bar. gofmt/golangci-lint concerns: none found (see Verification).

## Assessment

A genuinely high-quality, evidence-driven refactor. The commit sequence is exactly as planned and the red-first construction satisfies the #321 fail-closed requirement by ordering rather than by argument. The new tests adopt the repo's existing guard-test conventions (diagnostics carry inputs, `t.Errorf` vs `t.Fatalf` used appropriately). The two hard findings are inherited-line omissions with **no correctness impact** (read-path harness errors, not sentinel carriers, so no `TestNoBareSentinelWraps` exposure) — F2 is the one I'd fix before merge since the inconsistency was created inside a brand-new function; F1 can go with it in the same 2-line edit. Neither blocks merge on its own.

## Observations

- `ExecutePostgresQuery` sits at 78 lines — **two lines of headroom** under the new 80-line guard; already tracked by #430 (which also drops it to ~35).
- CI's `-tags=e2e -short` skips five federated test files (`data_tier`, `merge_on_read`, `deduplication`, `soft_delete`, `performance`); `ExecuteFederatedQuery` is still exercised via `consistency_test.go`/`failure_modes_test.go`. Author flagged this gap; not filed — acceptable.
- I did **not** re-run the docker e2e full suite (246s per the author's evidence); item 2's runtime coverage rests on the author's documented run plus the untagged static checks.

## Verification

Run on PR head (`85fa611`), mirroring the Makefile env:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false go test ./internal/e2e_harness/federated/ -count=1` → **pass**.
- Filtered rerun of all 7 relevant tests (`TestPostgresOnlyCountShares|TestBuildPostgresOnly|TestGuarded|TestListGuarded|TestMeasure|TestCountSourceLines`) → **all pass**.
- `gofmt -l` on all four changed files → **clean** (no output, exit 0).
- **Red-first confirmed:** at the guard commit `eea08d9` (commit 3), `ExecuteFederatedQuery` still measures **94 lines** — the guard is genuinely red before the refactor; at `85fa611` it measures 68, green.

iceboundrock:
## Review response — round 1

Two commits on top of `85fa611`: `bc7bfb2` (F1, F2, F3) and `6f61db8` (F4). All four findings were checked against the code before acting; three are fixed as described, one is fixed in part with reasoning below.

### F1 — fixed

`executePreferHotQuery` now returns `fmt.Errorf("prefer-hot postgres query: %w", err)`. Shortened from the suggested wording to match the phrasing already used in this file (`list base files:`, `get dirty ids:`, `count query:`, `execute query:`).

### F2 — fixed

`probeFederatedTiers` now wraps the `checkTierFiles` error as `check tier files: %w`, symmetric with the `get dirty ids: %w` line below it.

Agreed this is the stronger of the two: the asymmetry was inherited from the old inline code, but extracting it into a 13-line function is what turned two unrelated lines into one visible inconsistency.

**Checked before wrapping** that neither change breaks the two tests asserting on this error text — `failure_modes_test.go:47` (`assert.Contains(t, err.Error(), "S3")`) and `:154` (`"No files found"`) both use `Contains`, and `%w` preserves the inner message. Full e2e re-run below confirms.

### F3 — fixed in part, with a reason for the other part

**The concrete half is fixed:** `executionPlan` now has a doc comment, matching `isEmpty` and `emptyPageResult`.

**On the Middle Man flag, I'd like to keep the delegation.** You're factually right that `buildExecutionPlan` now has exactly one caller — I verified that, it does look like a candidate for absorption. Two reasons not to:

1. `buildExecutionPlan` pairs with `buildPostgresOnlyExecutionPlan` in `query_postgres_build.go`. They are the two plan builders for the two execution paths, and they read as a pair. Absorbing one into a method on an unexported struct breaks that symmetry for a saving of one function.
2. The method is not a layer over the builder — it is a data-clump adapter that keeps a four-argument unpacking out of three call sites in `ExecuteFederatedQuery`. That is a different thing from a class that only forwards, which is what §4.5's Middle Man is aimed at.

The new doc comment says both of these, so the next reader gets the rationale rather than re-deriving the smell. If you'd still rather collapse them, say so and I'll do it — it is a small change either way.

### F4 — fixed

Extracted `listNonTestSources(patterns ...string)`; both `listGuardedSourceFiles` and `listFunctionGuardedSourceFiles` are now one-line calls that differ only in scope. Net −8 lines.

You're right that it was under §2's trigger, but introducing a near-duplicate in a PR whose subject is deduplication was a poor look regardless. It also means #431 can converge the two guards by editing one pattern list.

**One thing this surfaced:** the old comment on `listFunctionGuardedSourceFiles` said it "deliberately does not reuse `listGuardedSourceFiles`", which stopped being true the moment both routed through the shared lister. Rewrote it to explain the scope difference — which is the part that is actually deliberate — rather than a helper relationship that no longer exists.

### On the test-quality note

Agreed the prefix test is close to tautological *against the current code*, and the comment says so. Worth adding that the mutation was run rather than assumed: swapping `attrIDs.symbol`/`attrIDs.exchange` inside `buildPostgresOnlyCountQuery` only (via `go test -overlay`, so the tree stays clean) turns `TestPostgresOnlyCountSharesSelectScanSource` red while `TestBuildPostgresOnlyQueriesSupportBenchmarkFilters` stays green. What the test buys is that it compares two independently assembled strings — so re-inlining, a divergent fragment call, or a changed argument order are all caught, not only literal re-inlining.

### On the observations

- `ExecutePostgresQuery` at 78/80 — agreed, and #430 is where it gets fixed. Flagged in the PR body for exactly this reason.
- The `-short` CI gap — agreed, leaving it unfiled.
- You didn't re-run the docker e2e; I re-ran it in full after these two commits, since F1/F2 change runtime error paths. Result below.

### Verification of these two commits

| Check | Result |
|---|---|
| `make test` | pass |
| `make fmt-check` | pass |
| `make lint` | pass |
| `go test ./internal/e2e_harness/federated/ -tags=e2e` (full, no `-short`) | 123 pass, 0 fail, 0 skip, 265s — including `TestFailureMode_S3Unavailable`, which is the test asserting on the error text F2 now wraps |
| `ExecuteFederatedQuery` still under the guard | 68 lines |

**Both guards re-proved able to fail after the F4 plumbing change** — this mattered, because F4 refactored the guards' own file-listing, and a lister that quietly returned the wrong set would have left two green tests guarding nothing. An 86-line function written to `query_hot.go` on disk is still reported by name (`query_hot.go: overgrownProbeFunction has 86 lines`). Not done through `-overlay`: as noted in the PR body, that only substitutes compilation inputs and these guards read source from disk at runtime, so an overlay-based check here returns a false green.

iceboundrock (commented):
# Review — PR #432, round 2 (#324)

Re-review of the two commits added since round 1 (`bc7bfb2`, `6f61db8`), addressing findings F1–F4, against the author's round-1 response comment. Verified at head `6f61db8` (main unchanged at `51ee90c`).

## Summary

All four round-1 findings are resolved: **F1, F2, F4 fixed as recommended; F3's concrete half (missing godoc) fixed, and its Middle Man half declined with reasoning I verified to be factually grounded.** No new findings introduced by the fix commits. The PR is ready to merge.

## Findings

- **F1 — RESOLVED (verified).** `query.go:97` now returns `fmt.Errorf("prefer-hot postgres query: %w", err)`, matching this file's `verb-noun: %w` phrasing (`list base files:`, `get dirty ids:`, `count query:`, `execute query:`). Author's safety claim checks out: `failure_modes_test.go:47` asserts `assert.Contains(err.Error(), "S3")` and `:154` `"No files found"` — both `Contains`, and `%w` preserves the inner message. Nothing breaks.
- **F2 — RESOLVED (verified).** `query.go:148` now returns `fmt.Errorf("check tier files: %w", err)`, symmetric with the `get dirty ids: %w` line directly below it. The intra-function asymmetry that made this the stronger finding is gone.
- **F3 — RESOLVED, as a documented judgement call.** The missing godoc is added (`query.go:122-125`) and states the actual rationale. The author declined to absorb `buildExecutionPlan`; I independently confirmed **both** premises of the rebuttal: (1) `buildExecutionPlan` indeed has exactly one caller (`query.go:127`), so it *is* a single-caller function; but (2) `buildPostgresOnlyExecutionPlan` does exist as a real pairing partner (`query_postgres_build.go:197`, called at `query.go:380` and `:446`) — the two plan builders for the two execution paths. The pairing rationale is grounded, the data-clump-adapter framing is accurate, and the comment now records it. Acceptable as-is.
- **F4 — RESOLVED (verified).** `listNonTestSources(patterns...)` extracted in `query_file_size_test.go`; both guards are now one-line calls differing only in scope. The stale comment on `listFunctionGuardedSourceFiles` that said it "deliberately does not reuse `listGuardedSourceFiles`" had been made false by the very refactor — rewriting it to explain the actual scope difference was the right catch.

## Spec

No drift from issue #324. The two commits are review-driven additions to the PR, not new requirements, and change none of items 1–3's implementations: the Postgres-only SQL is untouched (zero SQL-token changes in this diff → byte-identity still holds by construction), `ExecuteFederatedQuery` still measures 68 lines, and the guard's file scope is unchanged. The wraps are behavior-preserving by construction, and the commit that touches runtime error paths (`bc7bfb2`) was followed by a full docker e2e re-run.

## Standards

Round 1's 2 hard violations (F1, F2) are both fixed; both judgement calls are resolved (F3 documented, F4 removed). **No new violations** in the fix commits: wrap messages are lowercase, context-carrying, and follow the file's convention; the shared lister is a clean parameter-object-shaped helper (scope as argument, rule in one place); comments state intent, not mechanics. gofmt/golangci-lint clean (see Verification).

## Assessment

This is a model review-driven revision. Each finding was checked against the code before acting (the F1/F2 `Contains` check is the right gate before wrapping error texts). The one finding declined was argued with two independently verifiable premises instead of hand-waving, and the doc comment captures the decision so the next reader doesn't re-derive the smell. The best moment is the mutation re-proof: the author re-ran the guard's fail-closed check *specifically because F4 had refactored the guards' own file-listing* — exactly the self-skepticism that keeps guards from guarding nothing. Nothing blocking; merge when ready.

## Observations

- `buildExecutionPlan` is now a single-caller free function. If the F3 pairing ever breaks (e.g., `buildPostgresOnlyExecutionPlan` gets absorbed or removed), absorption into `federatedTierState` becomes the obvious move — the comment already points at the trade-off.
- The shared lister's empty-match error now generalizes via `strings.Join(patterns, " or ")`; no test asserts on that text, so the message change is safe.
- The author re-ran the full docker suite on the new head (123 pass, 0 fail, 0 skip, 265s) after the error-path commits. I did not re-run the docker suite; this review's runtime evidence is the untagged unit layer plus the mutation proof, and the docker claim is taken from the author's documented run.

## Verification

Run independently on head `6f61db8` (Makefile env mirrored):

- `go test ./internal/e2e_harness/federated/ -count=1` → **pass** (full untagged package).
- Filtered rerun of all 7 guard/prefix tests (`TestPostgresOnlyCountShares|TestBuildPostgresOnly|TestGuarded|TestListGuarded|TestMeasure|TestCountSourceLines`) → **all pass**.
- `gofmt -l` on all 5 changed files (`query.go`, `query_postgres_build.go`, `query_function_size_test.go`, `query_file_size_test.go`, `query_unit_test.go`) → **clean**.
- **Guard fail-closed re-proved after F4's lister refactor:** on a throwaway worktree at `6f61db8`, appended an 88-line `overgrownProbeFunction` to `query_hot.go` → `query_hot.go: overgrownProbeFunction has 88 lines, exceeds the 80-line refactoring trigger` (test red). Restored and worktree removed. The guard still scans the whole `query*.go` family by name, not just the file this PR shrank.
- **F3 premises confirmed by search:** `buildExecutionPlan` has exactly one call site (`query.go:127`); `buildPostgresOnlyExecutionPlan` exists at `query_postgres_build.go:197` with callers at `query.go:380,446`.
- **F1/F2 wrap safety confirmed:** `failure_modes_test.go:47` (`S3`) and `:154` (`No files found`) both use `assert.Contains`, so `%w` wrapping cannot break them.

iceboundrock:
## Review response — round 2

Nothing to fix this round: all four findings were confirmed resolved and no new ones were raised. Recording what I verified independently rather than just accepting the round-2 verdict, plus one small correction to an observation.

### Re-verified the round-2 claims at head `6f61db8`

Every cited location checks out:

| Claim | Verified |
|---|---|
| `query.go:97` wrap | `return nil, fmt.Errorf("prefer-hot postgres query: %w", err)` |
| `query.go:122-125` godoc | present, four lines |
| `buildExecutionPlan` has exactly one caller | `query.go:127`, plus its definition at `:283` — no others repo-wide |
| `buildPostgresOnlyExecutionPlan` exists as the pairing partner | defined `query_postgres_build.go:197`, called `query.go:380` and `:446` |
| F2 wrap | `query.go:149` (the review cited `:148`, which is the `if err != nil {` line — same block) |

Also confirming the spec claim rather than leaving it as construction-based reasoning: **`query_postgres_build.go` has zero diff between `85fa611` and `6f61db8`**, so the byte-identity evidence from the original run still covers the current head. The two fix commits touch only `query.go` (two wraps plus a doc comment) and the two guard test files.

Branch is current with `main`: `origin/main` is `51ee90c` and that is also the merge-base, so no rebase is needed before merging.

### One correction to Observation 1

The observation says the doc comment "already points at the trade-off" for absorbing `buildExecutionPlan` if the pairing ever breaks. It's a little more modest than that — the comment records *why the pairing exists*, not what to do when it stops existing:

> `executionPlan` adapts the probe's three fields to `buildExecutionPlan`, which is the plan vocabulary shared with `buildPostgresOnlyExecutionPlan`. It exists to keep the four-argument unpacking out of three call sites in `ExecuteFederatedQuery`, not to add a layer over the builder.

A reader who finds the pairing gone can draw the conclusion, but the comment does not spell it out. I'd rather leave it that way than add "if `buildPostgresOnlyExecutionPlan` ever disappears, inline me" — speculative-future instructions in comments age badly, and this one would be describing a state that may never arrive. Flagging it so the round-2 record isn't slightly more generous to the comment than the comment deserves.

### No follow-up issues from this round

Checked each observation for whether it needs tracking, and none does:

- **`buildExecutionPlan` as a single-caller function** — conditional on a future change that may not happen; the rationale is in the code. An issue saying "if X changes, consider Y" would be backlog noise.
- **The shared lister's generalized empty-match message** — a verification note; I confirmed no test asserts on that text.
- **Docker suite not re-run by the reviewer** — process note. The full run on head `6f61db8` (123 pass, 0 fail, 0 skip, 265s) is the standing evidence.

The two follow-ups that this PR does owe are already filed: **#430** (`ExecutePostgresQuery`'s duplicated scan loop plus the missing `rows.Err()`, which also clears its 78/80 headroom under the new guard) and **#431** (`NewFederatedTestHarness` at 131 lines, which is what keeps the function guard's scope narrower than the file guard's).

CI is green on `6f61db8` across all five checks. Not self-merging, per the repo rule.

iceboundrock (commented):
# Review — PR #432, round 3 (fresh clean-slate pass at `6f61db8`)

A fresh, independent review of the **full 6-commit diff** (`origin/main...pull-432` at head `6f61db8`), run as two clean-slate sub-agent passes, plus my own line-level verification. The PR head is unchanged since round 2, so this is a verification-by-guessing-free second opinion rather than a delta review.

## Summary

**One new hard finding (F5)** — a bare `return nil, err` at the new `probeFederatedTiers` call site (`query.go:38`) that is the exact class the round-1 fix commit `bc7bfb2` was created to close, but which its sweep missed. Two pre-existing bare returns inside the same rewritten function are flagged as judgement/nits. Everything from rounds 1–2 is confirmed still resolved; the spec axis is fully clean. **Mergeable after the one-line F5 wrap.**

## Findings

- **F5 — Hard (Standards).** `query.go:38`, `ExecuteFederatedQuery`: `tiers, err := h.probeFederatedTiers(ctx); if err != nil { return nil, err }` returns the error **bare**. Verified against `origin/main`: this line is **new code written by commit 85fa611**, replacing two propagations — one that was already bare (`checkTierFiles`) and one that was **wrapped** (`return nil, fmt.Errorf("get dirty ids: %w", err)`). `bc7bfb2` wrapped the errors *inside* the helper (`check tier files:`, `get dirty ids:`) but left the new combined propagation bare, despite its commit message claiming to clear "the bare error returns the extraction moved into new functions". Fairness note: the internal wraps mean operator context *does* survive, so this is a standards/symmetry issue rather than a lost-context bug — but it is exactly the class this PR's own follow-up commit said it was eliminating. One-liner: `return nil, fmt.Errorf("probe federated tiers: %w", err)`.
- **Judgement (Standards).** `query.go:32` (`prepareBenchmarkDuckDBMacros`) and `query.go:77` (`scanQueryResults`) both propagate bare `return nil, err`. Verified **byte-identical in `origin/main`** (main lines ~41/~101) — pre-existing, not introduced here; defensible under "don't rewrite unrelated code", but AGENTS.md's "fix the immediate area" would justify wrapping both while this function is being reworked. Not blocking.

## Spec

**Fully clean at head.** All issue #324 requirements and every plan commitment are present. The one thing worth re-asserting after the round-2 lister refactor: I verified the six-commit head's **guard coverage is identical** to base — file guard still `{query*.go, harness*.go}` minus `_test.go` (`query_file_size_test.go:69`), function guard still `query*.go` minus `_test.go`. `bc7bfb2`'s wraps sit only on the prefer-hot and tier-probe paths; the Postgres-fallback branches the spec's execution flow cares about are untouched, and `failure_modes_test.go:47/:154/:206` assert via `Contains`/`ErrorIs`, so the new wrap context is safe. No missing, no scope creep beyond what the plan documented, no wrong-looking behaviour.

## Standards

Net from this fresh pass: **1 hard (F5), 2 pre-existing judgement calls**. Round-1's two hard violations remain fixed. No size breaches: `query.go` = 463 lines, `query_unit_test.go` = 496 (both < 500). Naming, comments, and error-message phrasing otherwise continue to meet the bar.

## Assessment

Still a high-quality, evidence-driven refactor — the fresh pass found essentially one two-line omission, and it happens to be in the very commit whose stated purpose was to eliminate that omission. That is worth fixing in the same follow-up (it is literally the same edit as round 1's F2). Everything else remains below the blocking bar. Ready to merge with the F5 wrap.

## Observations

- `query_unit_test.go` grew to **496 lines** — under the 500 cap, but the file-size guard only watches non-test files (`listNonTestSources` excludes `_test.go`), so test-file bloat is unwatched. A separate test file would buy headroom.
- `query.go:60`: the `CountOnly` branch computes `time.Since(start)` twice inside one return expression (pre-existing).
- `federatedTierState`'s name slightly undersells it — it carries the hot-tier `dirtyIDs` beside the two parquet-tier booleans; the godoc documents this, so acceptable.
- CI at head `6f61db8`: all five checks green, `mergeStateStatus: CLEAN`, `MERGEABLE`.

## Verification

Run fresh at head `6f61db8` (Makefile env mirrored):

- Full untagged package: `go test ./internal/e2e_harness/federated/ -count=1` → **pass**.
- All 7 guard/prefix tests (`TestPostgresOnlyCountShares|TestBuildPostgresOnly|TestGuarded|TestListGuarded|TestMeasure|TestCountSourceLines`) → **pass**.
- `gofmt -l` on the five changed files → **clean**.
- **F5 confirmed by diff:** `query.go:36-39` is new in `85fa611`; the replaced main code wrapped `getDirtyIDs` (`fmt.Errorf("get dirty ids: %w", err)`) and returned `checkTierFiles` bare — the merged call site is new and bare.
- **query.go:32/:77 confirmed pre-existing:** both `return nil, err` lines are byte-identical in `origin/main` (~:41 and ~:101).
- Line counts via `wc -l` (`query.go` 463, `query_unit_test.go` 496).

iceboundrock:
## Review response — round 3

Two commits on top of `6f61db8`: `12555d5` (F5 and the two judgement calls beside it) and `becd1af` (the 496-line observation). One new issue filed, one existing issue's scope widened.

### F5 — fixed, and the criticism lands

`query.go:38` now reads `fmt.Errorf("probe federated tiers: %w", err)`.

Worth saying plainly rather than burying: **`bc7bfb2`'s commit message overclaimed.** It said it cleared "the bare error returns the extraction moved into new functions", and it did wrap the two returns *inside* `probeFederatedTiers` — but the propagation *out of* the helper, which is the line the extraction actually created, was left bare. The message described a class; the change covered a subset. That is the kind of gap a fresh clean-slate pass is good at catching and an incremental one is not, which is presumably why round 3 found it and round 2 didn't.

The fairness note in the finding is accurate — the inner wraps mean no operator context was lost, so this was symmetry rather than a lost-context bug.

### The two pre-existing bare returns beside it — also fixed

`prepareBenchmarkDuckDBMacros` (`:32`) and `scanQueryResults` (`:77`) are now wrapped too, despite being flagged non-blocking and being byte-identical in `main`.

Reasoning: round 1 found two bare returns in this function, round 3 found a third, and leaving two known ones in place invites a fourth round to find them. `ExecuteFederatedQuery` now has **no bare error return at all**, which is a claim that can be checked in one grep rather than a claim about intent.

### The eight bare returns this does *not* fix

While confirming the above I grepped the rest of the file. `query.go` still has eight bare propagations, all outside the reworked function:

| Function | Lines |
|---|---|
| `getDirtyIDs` | 356, 364 |
| `ExecutePostgresQuery` | 378, 387, 403, 407 |
| `StreamFederatedQuery` | 455, 460 |

Left alone here under "don't rewrite unrelated code", but **added to #430's scope** ([comment](https://github.com/Lychee-Technology/forma/issues/430#issuecomment-5361626146)) rather than left to be rediscovered — four of the eight are inside the function #430 rewrites anyway. The pattern worth stopping is finding these one review round at a time.

### Observation: `query_unit_test.go` at 496 — fixed

This PR grew that file from 457 to 496, four lines from the limit, in a file the guard structurally cannot watch. Moved the two tests covering `query_postgres_build.go` into `query_postgres_build_test.go`, named after what they test.

- `query_unit_test.go` 496 → **441**; new file **65**.
- **Pure move, verified by diff** rather than by reading: the relocated block is byte-identical to what was removed (`git show HEAD:…query_unit_test.go | sed -n '151,204p'` against the new file's body → no differences). Both tests still run.

**Did not widen the guard to test files**, and filed **#433** with the reason: `performance_test.go` is 575 lines and `consistency_test.go` is 525, both already over the limit, so that guard would be red on arrival for reasons this PR does not own. #433 also asks the question underneath it — whether test files should be held to the 500-line rule at all, and if not, that the exemption belongs in `coding-standard.md` so the next reviewer doesn't re-raise it.

### Observation: doubled `time.Since(start)` in the `CountOnly` branch — fixed

Computed once into `countDuration`. This is the same fix `emptyPageResult` already applied to the branch directly below it, so leaving it was exactly the asymmetry F2 was about. Costs one line: `ExecuteFederatedQuery` is now **69**, still well under the 80 guard.

### Observation: `federatedTierState`'s name

Agreed it undersells the struct — the hot tier's `dirtyIDs` sits beside two parquet-tier booleans. I'd argue the name is still right: a dirty set *is* hot-tier state, so the struct is genuinely "what one probe of the three tiers found", which is what the godoc says. No change.

### Verification

At head `becd1af`:

| Check | Result |
|---|---|
| `make test` | pass |
| `make fmt-check` | pass |
| `make lint` | pass |
| `go test ./internal/e2e_harness/federated/ -tags=e2e` (full, no `-short`) | 122 pass, 1 environmental fail — see below |
| `ExecuteFederatedQuery` | 69 lines |
| `query.go` | 464 lines |
| Bare error returns in `ExecuteFederatedQuery` | 0 |

**Guard re-proved able to fail** after these commits, same as after the round-2 lister change: an 86-line function written to `query_hot.go` on disk is reported by name. Re-running this every time the guard's surroundings move is the only way the green means anything.

Wrap safety re-confirmed before touching the error paths: `failure_modes_test.go` asserts via `assert.Contains` (`:47`, `:154`) and `assert.ErrorIs` (`:206`), and `%w` preserves both the message text and the `errors.Is` chain — so the `context.DeadlineExceeded` assertion in the timeout test still holds through the new `probe federated tiers:` layer, which is the layer that error now travels through.

### The e2e run needs an honest caveat

Both full-suite runs at this head returned **122 pass, 1 fail**, and each failure was a *different* test. Neither is a regression, but I am not going to write "pass" over that.

- Run A: `TestMergeOnRead_DirtyIDExclusion` failed at `failed to create harness` — the RustFS container never published port 9000 and the docker socket call timed out. No query ran.
- Run B: `TestDeduplication_BulkPerformance` failed its `require.Less(duration, 5*time.Second)` ceiling at 5.22s.

Attribution for run B, since a 4% overshoot on a query I touched deserves more than a shrug: the failing assertion and the log line measure the **same variable** (the `ExecuteFederatedQuery` call). Run in isolation three times at this head it measures **222ms / 214ms / 221ms** — a 23× margin under the ceiling. A code regression would show in isolation; a 24× spread between isolated and in-suite is host contention after four minutes of container churn. The mechanism argument agrees: this round's changes are three `fmt.Errorf` wraps on paths not taken on success, one variable in the `CountOnly` branch this test does not enter, and a test-file move.

The decisive evidence is that **this same suite returned 123 pass / 0 fail / 0 skip twice on this branch** in rounds 1 and 2, including both of these tests.

Filed **#434** with both failure modes and the isolation measurements. Note that CI cannot see either one: `-short` skips `deduplication_test.go` and `merge_on_read_test.go`, so this whole class only appears in local full-suite runs.

One process note while I am at it: my background e2e wrapper ended in `; echo "exit=$?"`, so the harness reported "completed, exit code 0" for a run that had actually failed. The failure was caught by reading the output file, not by trusting the notification. Worth knowing if anyone else scripts these runs the same way.

iceboundrock (commented):
# Review — PR #432 · round 4 · `becd1af` (#324)

固定点 `origin/main` (`51ee90c`) → `becd1af`（8 commits），3 轮前置 review 已在 PR 上。全量复核新 head，重点验证 `12555d5`/`becd1af` 对 round 3 发现的修复。

## findings

- **F5 已关闭（hard → 0）**：`query.go:38` 新引入的 `probeFederatedTiers` 传播点现为 `fmt.Errorf("probe federated tiers: %w", err)`。`bc7bfb2` 只包裹了 helper 内部、遗漏该调用点的缺陷已修复。
- **Round 3 两个 judgement 已一并关闭**：`query.go:32` `prepareBenchmarkDuckDBMacros` → `prepare benchmark duckdb macros: %w`，`query.go:77` `scanQueryResults` → `scan query results: %w`。Commit 说明显式标注“顺手关闭同函数内的剩余裸返回，剩余 6—8 处裸返回位于 `getDirtyIDs`/`ExecutePostgresQuery`/`StreamFederatedQuery`，按 *don't rewrite unrelated code* 留给 #430”，与实测一致 —— `grep "return nil, err" query.go` 仅剩 6 处，均在 `ExecuteFederatedQuery` 之外（356/364/378/387/403/407 行）。
- **额外修复**：`CountOnly` 分支的 `time.Since(start)` 重复计算已合并为 `countDuration`（与 `emptyPageResult` 的写法一致）。
- **无新增 hard/judgement**：新增两 commit 未引入重复、超长函数、错误信息缺失或注释债务。

## spec

- **Issue #324 三项要求**：① Postgres-only 扫描源去重（`buildPostgresOnlyScanSource`）② `ExecuteFederatedQuery` 80 行触发器（实测 68→ 现 464 行文件内仍 68 行函数）③ 超出 scope 的 `harness.go`/500 行 guard 不动 —— 均保持满足。`becd1af` 的测试文件拆分是**纯移动**，`query_postgres_build_test.go:1-65` 与 `query_unit_test.go` 删除块字节一致。
- **Scope 控制**：`12555d5` 的三处 `fmt.Errorf` 包裹均为行为保持的上下文补充，`%w` 保留 `errors.Is/Contains` 链；`becd1af` 将 `_test.go` 排除在 `listNonTestSources` 之外的设计保持不变，未扩大 guard 范围（符合 commit 中“`performance_test.go 575`/`consistency_test.go 525` 已超限，另案处理”的说明）。

## standards

- **错误包裹**（`coding-standard.md §3` / `AGENTS.md`）：`ExecuteFederatedQuery` 内已**零裸返回**，全部 `return nil, fmt.Errorf("…: %w", err)`；其余裸返回均在未重写函数内，符合“修当前区域”原则。
- **重复与长度**：`query.go 464/500`，`query_unit_test.go 441/500`（原 496），新文件 `65` 行，均合规；`listNonTestSources` 去重逻辑保持。
- **命名/注释**：`executionPlan` 的 Middle-Man 取舍已在 `query.go:122-125` 留档，`query_postgres_build_test.go` 头部说明拆分原因。
- **gofmt / golangci-lint**：见 verification，`gofmt -l` 零输出。

## assessment

两轮“扫尾”commit 质量高：不仅补上 round 3 的 F5，还把同函数内剩余的同类债务一次性清零，并把测试文件过长的**观测**转为**结构修复**。PR 自 `85fa611` 起的演进路径清晰（去重 → 80 行拆分 → guard → 包裹修复 → 共享 lister → 扫尾），每步均有可复现的验证记录。

**结论：无阻塞项，满足合并条件。**

## observations

- `query.go` 仍有 6 处裸返回，集中在 `getDirtyIDs`/`ExecutePostgresQuery`/`StreamFederatedQuery` —— 与 commit 描述一致，留给 #430 合适。
- `query_unit_test.go` 从 496→441 释放了 `500` 行天花板压力；`query_postgres_build_test.go` 按文件名归属测试是更可发现的布局。
- `CountOnly` 分支现与 `emptyPageResult` 分支在 `Duration` 计算上一致，消除了细微的不对称。

## verification

在 `becd1af`（`GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false`）执行：

- `go test ./internal/e2e_harness/federated -count=1` → **pass**（全量 untagged，含 `TestBuildPostgresOnlyQueriesSupportBenchmarkFilters`/`TestPostgresOnlyCountSharesSelectScanSource`，两者在新文件下通过）。
- `gofmt -l` 5 文件 → **clean**。
- `grep`：`ExecuteFederatedQuery` 内 `return nil, err` 计数 **0**；`return nil, fmt.Errorf` 计数 **6**（32/38/57/72/78/98），与预期一致。
- `wc -l`：`query.go 464`、`query_unit_test.go 441`、`query_postgres_build_test.go 65`。
- 未重跑 ` -tags=e2e` 全量（需 Docker，约 4—5 min）；作者在 round 1—3 已多次全量验证且本次改动仅为错误包裹与测试移动，风险低。

## summary

- Round 3 的 1 hard + 2 judgement 已全部关闭，`ExecuteFederatedQuery` 达“零裸返回”。
- Spec/standards 均无新增问题，测试与格式门禁通过。
- 建议合并；后续跟踪仍为已立项的 **#430**（`ExecutePostgresQuery` 78 行与剩余裸返回）与 **#431**（`harness.go` guard 扩面）。

Issue #448, opened 2026-08-22
chore(build): Go 1.27 breaks local make lint (46 spurious typecheck errors) and make test; CI's 1.26.0 pin hides it


iceboundrock:
On a machine with Go 1.27.0 installed, both `make lint` and `make test` fail on a pristine `main` (verified at `c3c98ce`). CI does not see either failure because `.github/workflows/ci.yml` pins `GO_VERSION: '1.26.0'`, and `go.mod` declares `go 1.26.0` with no `toolchain` directive, so `GOTOOLCHAIN=auto` happily runs the newer local toolchain.

Discovered while running the gates for PR #446; unrelated to that change, which is green under the pinned toolchain.

## 1. `make lint` — 46 spurious typecheck errors

golangci-lint is pinned to v1.64.8 (per AGENTS.md: "do not upgrade the pin"). Its bundled `go/types` cannot read Go 1.27's export data. The root cause is stated outright in one of the diagnostics:

```
internal/e2e_harness/production/cluster.go: could not import sync/atomic
  (-: could not load export data: internal error in importing "sync/atomic"
  (cannot decode "sync/atomic", export data version 4 is greater than maximum
  supported version 2); please report an issue) (typecheck)
```

Everything else cascades from that failed import: 46 `(typecheck)` errors across `cmd/tools`, `internal/compaction`, `internal/e2e_harness`, `internal/federated`, `internal/reconcile` and others, reported as nonsense like `undefined: pgx`, `undefined: sqlmock`, `undefined: smithy`, and `h.Duck undefined (type *FederatedTestHarness has no field or method Duck)`. None are real.

## 2. `make test` — one real behaviour change

```
--- FAIL: TestValidateFallsBackOnInvalidNumberLiteral
    nonfinite_paths_test.go:294:
      "payload cannot be encoded as JSON: json: error calling MarshalJSONTo for
       type *json.Number: cannot parse \"abc\" as JSON number: invalid syntax"
      does not contain "invalid number literal"
```

`internal/schemavalidate` asserts on the error text produced when marshalling an invalid `json.Number`. Go 1.27's `encoding/json` changed both the wording and the calling convention named in it (`MarshalJSONTo`). This is a genuine assertion-on-stdlib-message fragility, not a lint artifact — it will need attention whenever the repo does move to 1.27, independently of the pin question.

## Verification that it is toolchain drift, not repo rot

Same worktree, same commit, only `GOTOOLCHAIN` differs:

[...]

PR #476, opened 2026-08-24
security(federated): gate + escape caller-supplied s3_parquet_path_template (#456)

### LINKED ISSUE #456: security(federated): caller-supplied s3_parquet_path_template is interpolated raw into DuckDB read_parquet — SQL injection + arbitrary-path read with server S3 credentials
## Problem

The caller-supplied `s3_parquet_path_template` field of the `POST /api/v1/advanced_query` request body is interpolated **raw** (no literal escaping, no allowlist) into DuckDB SQL on the federated read path. On any DuckDB-enabled deployment this is a caller-reachable SQL injection into DuckDB, and — even without breaking out of the string literal — lets the caller redirect `read_parquet` at an arbitrary path read with the server's own S3 credentials.

## Current behavior

`FederatedQueryRequest.S3ParquetPathTemplate` (`types.go`, `json:"s3_parquet_path_template"`) is a public field. `federated.enabled` is itself a request-body flag; the only server-side gate is that the DuckDB engine be wired (`entity_query_service.go` returns an error if `federatedQueryEngine == nil`). Past that gate there is no opt-in and no validation:

1. `entity_query_service.go` copies the field straight into `model.DuckDBRenderHints{S3ParquetPathTemplate: …}` — no validation.
2. `internal/federated/parquet_source.go` `resolveParquetPaths`: **"an explicit caller-supplied render hint always wins"** — a hinted path returns *before* manifest resolution and *before* `corruptPaths.Split`, so it is never checked against the configured bucket/prefix and never filtered.
3. `internal/sqlgen/duckdb_sql_generator.go` `RenderS3ParquetPath` runs the caller string through `text/template` and returns it.
4. `internal/sqlgen/duckdb_template_renderer.go` `formatDuckDBPathList` wraps each path as `fmt.Sprintf("'%s'", path)` — **no `sqlutil.EscapeLiteral`** — feeding `read_parquet(<caller text>, union_by_name=true)`.

The exact same `DESCRIBE SELECT * FROM read_parquet('%s')` is rendered elsewhere **with** `sqlutil.EscapeLiteral(uri)` (`internal/parquetcheck/describe.go`), so the rule exists and is simply not applied at the federated sites. `internal/sqlutil/literal.go` is documented (#307/#310) as "the single home of the rule."

Raw interpolation sites reachable from a caller-supplied path:
- `internal/sqlgen/duckdb_template_renderer.go` `formatDuckDBPathList` → `BuildParquetScanSource` → `read_parquet('…')`
- `internal/federated/parquet_schema_validation.go` — `glob('%s')` and `DESCRIBE SELECT * FROM read_parquet('%s')`
- `internal/federated/parquet_verify.go` — `SELECT * FROM read_parquet('%s')`
- `internal/federated/parquet_guard_identify.go` — `BuildParquetScanSource(fmt.Sprintf("'%s'", path), nil)`

## Reproduction

On a `DUCKDB_ENABLED=true` deployment:

1. `POST /api/v1/advanced_query` with body `{"schema_name":"…","condition":{…},"federated":{"enabled":true,"s3_parquet_path_template":"<payload>"}}`.
2. Quote-breakout payload: `s3://b/x.parquet') UNION ALL SELECT … FROM read_csv('/etc/passwd` renders to `read_parquet('s3://b/x.parquet') UNION ALL SELECT … FROM read_csv('/etc/passwd', union_by_name=true)`. With `httpfs`+`parquet` loaded, DuckDB gives local-file read (`read_csv`/`read_text`), arbitrary-URL fetch, and reach into the attached `postgres_scan` DSN in the same session.
3. No-quote payload: `s3://attacker-bucket/**` or `/etc/hosts` or `https://attacker/x.parquet` — the caller fully chooses the scan target, read with the *server's* S3 credentials (SSRF / cross-bucket read / exfil beacon).

(A live proof-of-exploit against the pinned DuckDB build was not executed — the reviewing environment could not link the driver — but every hop is read directly from source and the chain is unbroken.)

## Root cause

The path set is treated as trusted, operator-authored deployment configuration at the render sites, but one of its two producers is an unvalidated request field, and the escaping rule that guards the sibling `parquetcheck` site was never applied to the federated sites.

## Impact

Caller-reachable SQL injection into DuckDB and arbitrary-target reads with the server's storage credentials, on any DuckDB-enabled deployment — confidentiality (local files, cross-bucket, DSN reach) and integrity of query results. Also a cache/DoS amplifier: hinted paths participate in the plan-cache scope key, so many distinct templates force `queryplan.Cache`'s wholesale clear, and each distinct path adds a never-evicted entry to `parquetSchemaValidator.valid` (tracked separately).

- #311 (closed) asserts at the *package* level that the only raw-interpolation hole (`PG_CONN`) "was not a caller-reachable injection vector." That reasoning does not extend to this caller-supplied field — same class, different value, **not covered**.
- #310 consolidated the quote-doubler but never applied it to the path sites — this is a superset of #310's scope.
- #301/#306 (closed) redact S3 keys/creds from error bodies — a different leak surface.
- Not a duplicate of any open federated issue (searched `s3_parquet_path_template`, `injection`, `read_parquet`, `SSRF`, `path template`, `hint`).

## Acceptance criteria

- [ ] Caller-supplied `s3_parquet_path_template` is validated (allowlisted against the configured bucket/prefix scheme) and/or gated behind a server-side opt-in, not a request-body flag alone.
- [ ] Every `read_parquet`/`glob`/`DESCRIBE` render of a path routes through `sqlutil.EscapeLiteral` (audit all five sites above + `formatDuckDBPathList`).
- [ ] A single quote in a hinted path cannot alter SQL structure; an out-of-scope path cannot be scanned.

- [ ] Injection test: a hinted path containing `'` cannot break out of the literal (assert generated SQL and/or execution rejects it).
- [ ] Scope test: a hinted path outside the configured bucket/prefix is rejected (or the feature is gated off).
- [ ] Guard test enumerating the path-render sites so a new one cannot skip `EscapeLiteral`.

iceboundrock:
Closes #456.

## Problem

`federated.s3_parquet_path_template` (`POST /api/v1/advanced_query`) was interpolated **raw** into DuckDB SQL on the federated read path. On any DuckDB-enabled deployment this was a caller-reachable SQL injection into DuckDB **and** an arbitrary-path read with the server's own S3 credentials (local files via `read_csv`, cross-bucket, `https://` SSRF). The only server-side gate was that the engine be wired — past that, the field flowed unvalidated and unescaped into `read_parquet`/`glob`/`DESCRIBE`.

## Fix — three fail-closed layers

1. **Opt-in gate.** New `DuckDB.AllowCallerParquetPaths` flag (default **off**, env `DUCKDB_ALLOW_CALLER_PARQUET_PATHS`). A request carrying the hint is **rejected** (not silently ignored) unless the operator enabled it — so every existing deployment is closed by default.
2. **Bucket allowlist.** When enabled, every rendered path must sit under `s3://<S3Bucket>/` (trailing-slash exact, defeating the `s3://bucketX/` prefix collision), carry no `' " ; \n \r`, and the bucket must be configured. Closes local-file / cross-bucket / http-SSRF targets.
3. **Universal escaping.** `sqlutil.EscapeLiteral` at all five parquet-path render sites (`formatDuckDBPathList`, `drainParquet`, `drainGuardedParquet`, `globParquetPaths`, `describeParquetColumns`), matching the already-correct `parquetcheck` sibling. A source-level AST guard (`TestParquetPathRendersAreEscaped`, scanning `internal/federated`, `internal/sqlgen`, `internal/parquetcheck`) fails if a new render site skips escaping.

Error semantics: rejections are `forma.InvalidInputf` (4xx) echoing the caller's own path; the configured bucket name stays behind `WithOperatorDetail`, out of the published body (#313).

## Acceptance criteria

- [x] Caller template validated (bucket allowlist) **and** gated behind a server-side opt-in, not a request-body flag alone.
- [x] Every `read_parquet`/`glob`/`DESCRIBE` path render routes through `sqlutil.EscapeLiteral`.
- [x] A single quote in a hinted path cannot alter SQL structure; an out-of-scope path cannot be scanned.

## Tests

- Injection: pure-string (`formatDuckDBPathList`) + executor sites, asserting a `'`-bearing path renders doubled.
- Gate/scope: flag-off rejected, out-of-scope + prefix-collision + local + http rejected, empty-bucket rejected, in-scope (glob) honored.
- Guard: AST enumeration of render sites, mutation-verified.

## Verification

`make fmt-check` ✅ · `make lint` ✅ (0 issues) · unit suite ✅. The Docker-gated `TestE2EHarnessMinimal` smoke test is the only local red (no Docker daemon in the dev sandbox); it is environmental and independent of this change.

Design and implementation plan are recorded on #456.

https://claude.ai/code/session_0196sNgz4EfzUQ3xS667Fb2R


iceboundrock:
## Independent security review — APPROVED

A fresh-eyes security review was run over the full branch diff (`git diff main...HEAD`). Verdict: **APPROVED**, no bypass found.

### Bypass analysis — none found
- **Single funnel confirmed.** The only production ingress is `entity_query_service.go` copying `req.Federated.S3ParquetPathTemplate` into `fq.DuckDBHints` → `resolveParquetPaths` → `duckDBParquetPathsForQuery` (gate + scope). `grep` shows no other production caller of the resolver, `DuckDBHints`, or the render sites that skips the gate.
- **Both provenance branches escape.** Hint paths and manifest-source paths converge into the same slice → `formatDuckDBPathList` and the four executor sites, all now via `sqlutil.EscapeLiteral`. Manifest paths get the same escaping.
- Glob-expansion output is storage-authored, confined to the pattern's bucket, and re-escaped at every drain/describe site.

### Allowlist correctness — sound
- Trailing-slash prefix `s3://<bucket>/` defeats the `s3://bucketX/...` collision (pinned by `TestHintOutOfScopeRejected`).
- Local file, `https://` SSRF, and cross-bucket all fail the prefix. Comma-separated multi-paths are validated individually. `..` passes the prefix but is confined to the bucket by S3 key semantics (no traversal).

### Error semantics — correct
- Bucket name lives only in `WithOperatorDetail`; it never reaches the 4xx body, while the caller's own path is echoed (#313).
- The pre-existing `fmt.Errorf("render parquet path hint: %w", err)` wrap preserves both `ErrInvalidInput` classification and the `PublicMessage` body, and is not a bare-sentinel wrap.

### Guard robustness — mutation-verified
A temporary unescaped `read_parquet('%s')` site was injected and `TestParquetPathRendersAreEscaped` failed on it (then removed). The `scanned < N` floor prevents a vacuous pass.

### Follow-up applied
The review flagged that the guard did not scan `internal/parquetcheck` (whose `read_parquet('%s')` render is already escaped). Commit `91e39b5` adds that package to the guard's scan, closing the forward-coverage gap.

### Non-blocking observations (no change required)
- AST guard inspects only direct `fmt.Sprintf` with `%s`/exact-`'%s'` shapes; renders via `%v`/concatenation are a documented blind spot, mitigated by the behavioral escape tests.
- `plan_cache_parity_test.go` fixture `s3://b/bench/{schema_id}/base.parquet` uses single-brace `{schema_id}` (renders literally) — harmless; it passes scope + char checks.

## Verification
- `make fmt-check` ✅ · `make lint` ✅ (0 issues) · unit suite ✅.
- Only local red: `TestE2EHarnessMinimal` (testcontainers, needs a Docker daemon absent in the dev sandbox) — environmental, fails identically on `main`, independent of this change.

iceboundrock:
## Findings

### [P2] Document the new fail-closed opt-in and migration contract

The implementation changes established behavior for every request carrying `federated.s3_parquet_path_template`: it now returns `ErrInvalidInput` unless `DUCKDB_ALLOW_CALLER_PARQUET_PATHS=true` is enabled. The authoritative design still says the hint always wins at `docs/federated-query/design.md:118` and `:134`, while the server configuration table at `:141-148` and the E2E environment table at `tests/e2e/README.md:248-253` do not document the new flag.

A deployment following the current documentation will either fail existing hint-bearing clients after upgrade or have no discoverable way to restore the intended behavior. Please update the API/configuration documentation and migration notes with the default-off gate, the required `S3Bucket` scope, and the explicit rejection behavior. This is not a security bypass in the code; it is a release/operability contract gap.

iceboundrock:
## Observations

- `unverifiablePath` still skips quote-bearing paths because it was written for the pre-escaping behavior. The new implementation can safely render those paths, so verification/guard attribution still leaves a non-security coverage gap for manifest entries containing quotes. Consider allowing escaped quote-bearing paths and adding a regression test.
- `parquet_literal_guard_test.go` intentionally covers only direct `fmt.Sprintf` calls with literal format strings and direct `EscapeLiteral` arguments. That is reasonable as a forward guard, but concatenation, indirect format strings, or helper dataflow remain outside its scope and should stay covered by behavioral tests when new render paths are added.
- The PR keeps all changed files below the repository size limit; `internal/federated/parquet_source_test.go` is close to the 500-line ceiling at 495 lines, so future additions should avoid pushing it over the limit.

iceboundrock:
## Verification

- `make fmt-check`: passed.
- `git diff --check`: passed.
- Root package tests, including `TestParquetPathRendersAreEscaped`: passed.
- Local targeted `go test` for `internal/federated`, `internal/sqlgen`, and `cmd/server` could not link in this sandbox because the DuckDB CGO link requires the missing `libstdc++` development library; this is an environment limitation, not a test failure.
- PR CI: `Lint`, `Test`, `E2E Smoke Tests`, and `Benchmark Smoke` are passing in the current run. The `k6 Smoke` job failed in the run while the overall workflow was still in progress; it exercises the OLTP load path and does not send the caller Parquet hint.
- The latest successful `main` CI run passed all listed jobs, so the k6 result should be checked as a separate CI/infrastructure regression before merge.

iceboundrock:
## Findings

### [P2] Validate the opt-in and bucket combination at startup

`validateHintPathScope` rejects an empty `S3Bucket` only when the first request carrying a hint reaches `internal/federated/duckdb_query_build.go:335-340`. `forma.Config.Validate` / the factory startup validation do not reject `AllowCallerParquetPaths=true` with an empty bucket, so a misconfigured deployment can start successfully and return a caller-facing 4xx for every such request.

Please add a configuration validation rule for `AllowCallerParquetPaths && strings.TrimSpace(S3Bucket) == ""` and cover it with a config test, or document why request-time failure is intentional. Startup failure would be consistent with the existing manifest validation and would avoid turning an operator configuration error into invalid caller input.

iceboundrock:
## Assessment

The security fix is technically sound on the reviewed production paths: I found no caller-controlled route that bypasses the opt-in/scope gate, and no unescaped current `read_parquet`, `glob`, or `DESCRIBE` render site. The two findings are release-readiness issues rather than SQL-injection bypasses.

Recommendation: address the documentation/migration contract and startup validation before merge; after that, the code-level security assessment is approval-ready.

iceboundrock:
## Summary

PR #476 closes the reported caller-controlled Parquet SQL/path injection with a default-off server gate, exact configured-bucket allowlist, and centralized literal escaping across the federated read path. The implementation and added tests cover the main exploit surfaces and preserve explicit-hint precedence/error semantics.

Review findings: two P2 release-readiness items remain, namely documenting the new opt-in/migration contract and validating the enabled-flag/empty-bucket combination at startup. Non-blocking observations concern quote-bearing path verification and the intentional limits of the AST guard. No code-level injection bypass was found.

iceboundrock:
## Spec

### Security acceptance

- Caller-supplied path hints are rejected by default unless the deployment opts in with `AllowCallerParquetPaths`.
- When enabled, every rendered path is constrained to the exact configured `s3://<S3Bucket>/` prefix, including the trailing-slash prefix-collision case.
- Local paths, HTTP paths, cross-bucket paths, and SQL delimiter characters are rejected before execution.
- All current production render sites route path values through `sqlutil.EscapeLiteral`, including list rendering, guarded drains, verification drains, glob expansion, schema probing, and `internal/parquetcheck`.
- Invalid or empty explicit hints do not silently fall through to the manifest source.

The implementation satisfies these code-level acceptance points. The remaining spec delta is operational: the new opt-in and its migration behavior need to be added to the design/configuration docs.

iceboundrock:
## Standards

Review against `coding-standard.md`: the changed source files remain within the 500-line limit, the new resolver/validation helpers are small and flat, errors are wrapped with context where they cross service boundaries, and the new public configuration field has a descriptive comment.

The diff is gofmt-clean and `git diff --check` reports no whitespace errors. Tests cover default-off gating, bucket scope, prefix collision, local/HTTP rejection, empty bucket, quote escaping, executor render sites, environment wiring, and the AST forward guard.

iceboundrock:
## Findings

### [P1] Enable the caller-path opt-in in the black-box CI server

The current CI run shows a direct regression caused by this PR. The `k6 Smoke` job fails during `federated-check --schema log --require-duckdb`, before k6 starts, because the script sends `s3_parquet_path_template` and the server returns: `caller-supplied s3_parquet_path_template ... is not permitted on this deployment`.

`.github/workflows/ci.yml:146-169` starts the server with `DUCKDB_ENABLED=true` and the S3 settings but does not set `DUCKDB_ALLOW_CALLER_PARQUET_PATHS=true`. The PR updated the Go E2E harnesses to opt in, but missed this black-box harness. Please either add the explicit opt-in and bucket configuration to this server environment, or change that check to use the manifest-only path intentionally. The job must not be made to pass by silently ignoring the hint, because the new rejection is the intended provenance behavior.

iceboundrock:
## Summary Update

Additional blocking finding: the black-box CI server is missing `DUCKDB_ALLOW_CALLER_PARQUET_PATHS=true`, so the required `k6 Smoke` workflow fails in its federated-check setup. This is the concrete manifestation of the undocumented migration gap and should be fixed in the workflow/configuration before merge.

iceboundrock:
## Verification Correction

The workflow has now completed: `Lint`, `Test`, `E2E Smoke Tests`, and `Benchmark Smoke` pass; `k6 Smoke` fails. The failure is **PR-related**, not an unrelated OLTP/infrastructure failure: its log shows `federated-check --schema log --require-duckdb` sending `s3://forma-cdc/delta/{{.SchemaID}}/*.parquet`, followed by the new default-off rejection.

The earlier verification comment described k6 as independent; that statement is superseded by this correction. The P1 CI finding above should be resolved before merge.

iceboundrock:
## Assessment Update

Because the required CI workflow is red for the concrete caller-hint regression above, this review is not approval-ready yet. The security boundary itself remains correct, but the black-box test/deployment configuration must be migrated to the explicit opt-in before the PR can be considered merge-ready.

iceboundrock:
## Review response — findings addressed (commit `cdb9467`)

### [P1] k6 Smoke black-box server missing the opt-in — **fixed**
Confirmed real regression: `tests/e2e/scripts/federated-check.ts:143` sends `s3://forma-cdc/delta/{{.SchemaID}}/*.parquet`, and `.github/workflows/ci.yml` set `S3_BUCKET: forma-cdc` but not the opt-in, so the hint hit the new default-off rejection. Added `DUCKDB_ALLOW_CALLER_PARQUET_PATHS: "true"` to the k6 server env. The bucket is already `forma-cdc`, so the hint is in-scope and is now **honored** (not silently ignored / not switched to manifest-only) — matching the intended provenance behavior you called out.

### [P2] Validate opt-in + empty bucket at startup — **fixed**
`Config.Validate` (`config_duckdb.go`) now rejects `AllowCallerParquetPaths && strings.TrimSpace(S3Bucket) == ""` with a `ConfigError` on `duckdb.s3Bucket`, mirroring the manifest-read bucket rule. An operator misconfiguration now fails fast at startup instead of 4xx-ing every hint request. Covered by `TestValidateDuckDBConfig_AllowCallerParquetPathsRequiresBucket`.

### [P2] Document the fail-closed opt-in and migration contract — **fixed**
Updated:
- `docs/federated-query/design.md`: the `s3_parquet_path_template` field row now states it is disabled by default; §4.3.1 level 1 documents the opt-in, the required in-bucket scope, the explicit-rejection (vs fall-through) behavior, and the startup bucket requirement; the server-config table adds a `duckdb.allowCallerParquetPaths` / `DUCKDB_ALLOW_CALLER_PARQUET_PATHS` row.
- `tests/e2e/README.md`: env table adds `DUCKDB_ALLOW_CALLER_PARQUET_PATHS` (default off, must be `true` with `S3_BUCKET` for `federated-check`'s hint).

## Responses — non-blocking observations (no change)

- **`unverifiablePath` still skips quote-bearing paths.** Agreed this is a non-security coverage gap, not a bug: those manifest entries just fall to all-or-nothing verify/guard attribution rather than being mis-scanned, and the render path is now safely escaped regardless. I've left `unverifiablePath` as-is to keep this PR scoped to the security fix; happy to open a follow-up issue to let it verify escaped quote-bearing paths with a regression test.
- **AST guard covers only direct `fmt.Sprintf` with literal formats.** Intentional and documented in the guard's own comment as a blind spot; concatenation / indirect formats stay covered by the behavioral escape tests when new render paths are added. No change.
- **`parquet_source_test.go` near the 500-line ceiling (495).** Noted — the fixes in this commit add no lines to that file, and future additions there should split rather than push it over.

## Verification
`make fmt-check` ✅ · `make lint` ✅ (0 issues) · root-package tests incl. the new config test and `TestParquetPathRendersAreEscaped` ✅. The k6 Smoke regression should clear on the next CI run now that the server opts in.

iceboundrock (commented):
# PR #476 Review — security(federated): gate + escape caller-supplied s3_parquet_path_template (#456)

## Spec

Checked against #456's acceptance criteria and required regression coverage:

| Acceptance criterion | Status |
| --- | --- |
| Hint validated (allowlist) **and** gated behind a server-side opt-in, not a request-body flag alone | ✅ `AllowCallerParquetPaths` default-off (`config_duckdb.go:74`), env opt-in (`cmd/server/main.go:106`); flag-off rejects rather than ignores (`duckdb_query_build.go:288`); bucket allowlist with load-bearing trailing slash blocks the `s3://<bucket>X/` prefix collision (`validateHintPathScope`, `duckdb_query_build.go:341`) |
| Every `read_parquet`/`glob`/`DESCRIBE` path render routes through `sqlutil.EscapeLiteral` | ✅ All five sites escaped: `formatDuckDBPathList` (`duckdb_template_renderer.go:329`), `drainParquet` (`parquet_verify.go:73`), `drainGuardedParquet` (`parquet_guard_identify.go:74`), `globParquetPaths` + `describeParquetColumns` (`parquet_schema_validation.go:400,425`), matching the already-correct `parquetcheck.Describe` sibling. Grep sweep confirms no remaining caller-reachable raw interpolation inside `internal/federated` / `internal/sqlgen`. |
| A `'` cannot alter SQL structure; an out-of-scope path cannot be scanned | ✅ Escape at render + char blocklist + prefix gate; behavioral tests assert doubled quotes on captured SQL; AST guard enumerates render shapes so new sites are covered automatically |

Regression coverage requested by the issue is all present: injection tests (pure-string + executor sites), scope tests (flag-off, cross-bucket, prefix-collision, local file, http, empty bucket, in-scope glob honored), and the guard test.

Design-doc updates (`docs/federated-query/design.md` §4.2 table, §4.3.1 level 1, env-var table) and `tests/e2e/README.md` accurately describe the new behavior. Error semantics follow #313: rejections are `forma.InvalidInputf` carriers echoing only the caller's own path/template; the bucket name stays behind `WithOperatorDetail`.

## Findings

### P1 — The startup fail-fast is not actually wired to any startup path

The new check lives in `(Config).validateDuckDBConfig` (config_duckdb.go:348), which is reachable **only** through `Config.Validate()` — and no production entrypoint calls that method:

- `bootstrapServer` validates only `duckCfg.ValidateManifestRead()` (cmd/server/main.go:159);
- the factory likewise calls only `ValidateManifestRead()` (factory/factory.go:105);
- repo-wide grep: `Config.Validate()` has no non-test callers.

So an operator who sets `DUCKDB_ALLOW_CALLER_PARQUET_PATHS=true` without a bucket does **not** get the documented startup failure. Every hint-bearing request instead answers 4xx "not permitted on this deployment" at run time — precisely the outcome the new code comment says must be prevented ("an operator misconfiguration must not masquerade as invalid caller input"), and contradicting the doc text added here ("the combination is rejected at startup (`Config.Validate`)").

Security posture is unaffected (the runtime gate fails closed either way), but the fail-fast contract claimed in three places (config comment, config test name, design doc) is not enforced. Suggested fix, mirroring the manifest-read pattern this PR itself cites:

- extract a `DuckDBConfig.ValidateCallerParquetPaths()` (or fold the check into the same preflight block) and call it beside `ValidateManifestRead()` in both cmd/server/main.go and factory/factory.go;
- add a bootstrap-level test pinning that `bootstrapServer` refuses the bad combo — its absence is how this slipped through, since the current test exercises `Config.Validate()` directly.

### P2 — docs/error-handling.md "maintained list" not updated

`internal/httpapi/error_response.go:184` designates the carrier-site table in docs/error-handling.md ("Public HTTP error surface") as *the maintained list* of sites that answer a caller mistake. This PR adds four new published-4xx carrier sites in `internal/federated/duckdb_query_build.go` (flag-off rejection :290, disallowed character :344, out-of-scope :348, empty bucket :338) — none are in that table (the pre-existing unrenderable-template carrier isn't either). Since the table is the maintained contract for which caller mistakes surface as disclosed 400s, it should gain rows for these sites in this PR.

### Minor

1. **Char-blocklist rationale slightly overstated** (`duckdb_query_build.go:328-330`): after `EscapeLiteral` doubling, only `'` can alter SQL structure in a `'…'` literal — `"`, `;`, `\n`, `\r` cannot. Keeping them blocked is fine defense-in-depth, but note the trade-off: legitimate object keys containing those characters are now rejected even though escaping would have made them safe. The doc comment ("could break out of the SQL string literal") overstates what they can do.
2. **AST guard package-blindness** (`parquet_literal_guard_test.go`): `isEscapeLiteralCall` matches any selector named `EscapeLiteral`, regardless of receiver package — `otherpkg.EscapeLiteral(path)` would satisfy the guard. Cheap to tighten by checking the `sqlutil` selector's X identifier.
3. **Test gaps (low value)**: no case where a *rendered* template goes out of scope (e.g. `s3://other/{{.SchemaID}}/x`) — same code path as the literal cases, post-render; and no mixed in-scope/out-of-scope multi-segment template. Both hit the same validation loop.

## Standards (coding-standard.md / AGENTS.md)

- File/function size limits respected; error wrapping carries context throughout.
- Error semantics (#313) correctly applied: carriers publish only caller-owned text; operator detail (bucket name, render prose) withheld via `WithOperatorDetail`; verified the plain `%w` wrap at `parquet_source.go:68` preserves sentinel + published message through `ResolvePublicMessage`.
- Fail-closed ordering is right: flag gate → render → degenerate check → char blocklist → prefix gate; hint-gate errors return before any degraded-mode absorption (`parquet_source.go:66-72`), so they cannot be relabelled `ErrFederatedReadFailed`.
- AST guard follows the `TestNoBareSentinelWraps` idiom including the vacuous-scan floor; blind spots stated honestly.
- The P2 above is the one standards gap (maintained-list upkeep).

## Observations

1. **In-bucket glob breadth**: with the opt-in enabled, callers may target any prefix inside the configured bucket — including `_tmp/` staging objects and base-tier files — and `**` globs are honored in hints even though the level-3 fallback deliberately forbids widening to `**` (design.md §4.3.1). Impact is bounded to server-written intra-bucket data; consider scoping hints to `s3DataPrefix` as a hardening follow-up if the feature sees real use.
2. **Plan-cache/validator DoS amplifier** from many distinct hinted paths: unchanged by this PR; the issue records it as tracked separately. Fine to leave out here.
3. **Whitespace-only bucket asymmetry**: `Config.Validate` trims, but the runtime gate checks `== ""` — a padded bucket passes the runtime empty-check yet matches nothing, rejecting every hint as out-of-scope. Both paths fail closed; harmless, just inconsistent.
4. **CI posture**: the ci.yml server now runs with hint-honoring enabled so federated-check exercises the hint path; production default remains off and the flag-off path is covered by unit tests. Acceptable, but worth remembering that CI's DuckDB surface is wider than the default deployment's.
5. `hybridDuckConfig()` flipping several existing engine tests behind the feature is a reasonable parity choice and clearly commented.
6. Out-of-scope sighting: `internal/compaction/uncovered_rows.go:54` and `merge_sql.go:112` still interpolate operator-configured paths into `read_parquet('%s')` raw. Not caller-reachable and outside the guard's scanned dirs — fine for #456, but candidates for the same treatment whenever compaction grows any request-influenced input.
7. Cosmetic: `cap` shadows the builtin in `parquet_path_escape_test.go` (lint-clean).

## Verification

Run locally against cdb9467 with the pinned toolchain (go1.26.0):

- `go test . ./internal/federated ./internal/sqlgen ./cmd/server ./internal/parquetcheck` — **all pass** (local CGO link needed a libstdc++ shim in my sandbox; environmental).
- New tests individually green: `TestHintRejectedWhenFeatureOff`, `TestHintNoHintReturnsNil`, `TestHintInScopeAccepted`, `TestHintOutOfScopeRejected` (incl. `s3://bx/` prefix collision, `/etc/passwd`, `https://`), `TestHintQuoteCharRejected`, `TestHintEmptyBucketRejected`, `TestExecutorParquetSitesEscapeQuotes` (4 executor sites), `TestFormatDuckDBPathListEscapesQuotes`, `TestParquetPathRendersAreEscaped`.
- **Mutation check on the guard**: reverted `drainParquet`'s escape → `TestParquetPathRendersAreEscaped` fails naming `internal/federated/parquet_verify.go:73` exactly; working tree restored. The guard is real, not vacuous.
- Confirmed no production entrypoint calls `Config.Validate()` (basis of P1) and that hint-gate errors bypass degraded-mode absorption (basis of the error-semantics assessment above).
- CI on this head: Lint ✅, Test ✅; E2E Smoke / k6 Smoke / Benchmark Smoke pending at review time — worth a glance before merge since ci.yml itself changed.

## Assessment

The core fix is sound and well-layered: opt-in default-off gate → bucket allowlist with the trailing-slash subtlety handled → universal escaping with a self-enumerating source-level guard. Tests genuinely pin the behavior (the mutation check proves the guard bites). Two items should land before/in-this-PR: the P1 wiring gap (documented startup contract unenforced) and the P2 maintained-list update. Neither weakens the security closure; both are contract-honesty fixes.

**Recommendation: address P1/P2, then merge (do not auto-merge per repo rules).**

## Summary

Caller-supplied `s3_parquet_path_template` is now rejected by default, allowlisted to the configured bucket when opted in, and escaped at every render site with a forward guard against regressions — closing #456's injection and arbitrary-path-read chain. Follow-ups worth tracking: wire the startup validation into the real bootstrap path (P1), update the error-handling maintained list (P2), optionally scope hints below bucket granularity, and fold the compaction render sites under the escaping rule someday.

iceboundrock:
## Review response — round 2 (commit `28c448d`)

Thanks — the P1 catch is exactly right and important. Details below.

### [P1] Startup fail-fast was not wired to any startup path — **fixed**
Confirmed: `Config.Validate()` has no production caller (only benchmark harness), so the `validateDuckDBConfig` check I added last round never fired — the doc/comment/test all claimed a fail-fast that didn't exist. Fixed by mirroring the `ValidateManifestRead` pattern this PR already cites:
- Extracted `DuckDBConfig.ValidateCallerParquetPaths()` (`config_duckdb.go`).
- Called it beside `ValidateManifestRead()` in **both** production entrypoints: `bootstrapServer` (`cmd/server/main.go`) and `newEntityManagerWithConfigContext` (`factory/factory.go`), before any I/O. `Config.Validate` keeps the check too (now via the shared method).
- Added `TestBootstrapServer_RejectsCallerParquetPathsWithoutBucket` — a bootstrap-level test asserting the bad combo is refused before the pool connect (its absence is how this slipped through, as you noted).

### [P2] Maintained "Public HTTP error surface" list not updated — **fixed**
Added a row to the table in `docs/error-handling.md` covering `internal/federated/duckdb_query_build.go` (`duckDBParquetPathsForQuery`, `validateHintPathScope`): the flag-off / unrenderable / degenerate / disallowed-char / out-of-scope carrier sites, noting that the hint template and offending path are published while the bucket name stays operator-only.

### Minor — all addressed
1. **Char-blocklist rationale overstated** — corrected the comment: after `EscapeLiteral` doubling, only `'` can alter SQL structure in a `'…'` literal; `" ; \n \r` are defense-in-depth (kept, with the trade-off now stated honestly).
2. **AST guard package-blindness** — `isEscapeLiteralCall` now requires the `sqlutil` receiver, so `otherpkg.EscapeLiteral` no longer satisfies the guard.
3. **Test gaps** — added `TestHintRenderedOutOfScopeRejected` (a template whose `{{.SchemaID}}` render lands out of bucket) and `TestHintMixedScopeRejected` (in-scope + out-of-scope multi-segment template rejects the whole hint).

### Observations
- **[3] Whitespace-only bucket asymmetry** — fixed: `validateHintPathScope` now `TrimSpace`s the bucket, matching the startup check, so a padded bucket fails at startup and a hand-built engine stays consistent.
- **[7] `cap` shadows builtin** — renamed to `rec`.
- **[1] In-bucket glob breadth (`**`, `_tmp/`, base-tier)** — real breadth concern, out of scope for the injection fix. Tracked in #477.
- **[6] Compaction raw `read_parquet('%s')` sites** — not caller-reachable today, out of the guard's dirs. Tracked in #478.
- **[2] Plan-cache/validator DoS amplifier** — unchanged here, tracked separately as noted.
- **[4] CI posture / [5] hybridDuckConfig parity** — acknowledged, no change.
- (Prior round) **`unverifiablePath` skipping escaped quote-bearing paths** — non-security coverage gap, tracked in #479.

### Verification
`make fmt-check` ✅ · `make lint` ✅ (0 issues) · `go test . ./cmd/server ./factory ./internal/federated` ✅, including the new bootstrap, config, gate, and guard tests (local CGO link needs the libstdc++ shim; environmental).

Follow-ups opened: #477 (hint scope hardening), #478 (compaction escaping), #479 (`unverifiablePath` quote-bearing paths).

iceboundrock:
## Spec

Checked against #456 acceptance criteria and required regression coverage at HEAD `28c448d`:

| Criterion | Status | Evidence |
|-----------|--------|----------|
| Hint validated (allowlist) **and** gated behind server-side opt-in, not request-body flag alone | ✅ | `config_duckdb.go:81` `AllowCallerParquetPaths` default `false`; `cmd/server/main.go:106` / `factory/factory.go:111` fail-closed gate; `internal/federated/duckdb_query_build.go:288-290` rejects flag-off with `ErrInvalidInput`; never silently falls through to manifest source |
| Bucket allowlist with trailing-slash exactness defeats `s3://bucketX/` collision | ✅ | `internal/federated/duckdb_query_build.go:353` prefix `s3://<bucket>/`; pinned by `TestHintOutOfScopeRejected` (`internal/federated/parquet_hint_gate_test.go:51`) incl. `s3://bx/` case |
| Local / `http(s)` / cross-bucket rejected, disallowed chars rejected, empty bucket rejected | ✅ | `validateHintPathScope` (`duckdb_query_build.go:346-364`) checks `"'\";\n\r"` + prefix; `config_duckdb.go:361` startup bucket-required; `TestHintOutOfScopeRejected`, `TestHintQuoteCharRejected`, `TestHintEmptyBucketRejected` |
| Every `read_parquet`/`glob`/`DESCRIBE` path render routes through `sqlutil.EscapeLiteral` | ✅ | 5 sites: `internal/sqlgen/duckdb_template_renderer.go:329` `formatDuckDBPathList`, `internal/federated/parquet_verify.go:73` `drainParquet`, `internal/federated/parquet_guard_identify.go:74` `drainGuardedParquet`, `internal/federated/parquet_schema_validation.go:400` `globParquetPaths`, `:425` `describeParquetColumns`; sibling `internal/parquetcheck/describe.go:53` already correct. Behavioral tests `internal/sqlgen/duckdb_path_escape_test.go:8`, `internal/federated/parquet_path_escape_test.go:22` assert doubled `''` |
| `'` cannot alter SQL structure; out-of-scope cannot be scanned | ✅ | Escape at render + gate; `TestFormatDuckDBPathListEscapesQuotes` and `TestExecutorParquetSitesEscapeQuotes` pin `parquet'') UNION` vs `parquet') UNION` breakout |
| Invalid / degenerate explicit hint never falls through | ✅ | Unrenderable (`duckdb_query_build.go:292-301`), empty-render (`:310-319`) both `InvalidInputf`; `parquet_source_test.go:172` degenerate `", "` case |
| Guard test enumerates render sites so new site cannot skip escaping | ✅ | `parquet_literal_guard_test.go:37` scans `internal/federated`, `internal/sqlgen`, `internal/parquetcheck`; `isPathRenderFormat` (`:61`) matches `read_parquet('%s'`, `glob('%s'`, `'%s'`; `isEscapeLiteralCall` (`:40`) now requires `sqlutil` receiver; `scanned<10` floor prevents vacuous pass; mutation-verified in prior rounds |

Docs: `docs/federated-query/design.md:118` field row, `:134` level-1 text, `:149` config table, and `tests/e2e/README.md` env table now document default-off, in-bucket scope, explicit-rejection vs fall-through, and startup bucket requirement. `docs/error-handling.md:846` adds the four new carrier sites (`duckDBParquetPathsForQuery`, `validateHintPathScope`) to the maintained public-error-surface table.

No spec delta remains.

iceboundrock:
## Standards

Review against `coding-standard.md` / `AGENTS.md` / `docs/error-handling.md`:

- **File/function size:** all changed files stay ≤500 lines. Largest: `config_duckdb.go:368`, `internal/federated/duckdb_query_build.go:365`, `factory/factory.go:490`, `parquet_literal_guard_test.go:123`. `internal/federated/parquet_source_test.go:495` is at 495 lines — next addition there should split, but this PR adds no lines to it (hybridDuckConfig tweak is comment-only parity).
- **Error wrapping / classification:** gate, render, degenerate, char-block, scope all return `forma.InvalidInputf` carriers; bucket name and `text/template` prose are withheld via `forma.WithOperatorDetail` (`duckdb_query_build.go:300`, `:349`, `:359`) per #313. The `parquet_source.go:67` `fmt.Errorf("render parquet path hint: %w", err)` preserves `ErrInvalidInput` + `PublicMessage` without bare-sentinel publish (guarded by `TestNoBareSentinelWraps`). `config_duckdb.go:362` returns `*ConfigError` → operator-visible 500-class via caller, not a published 4xx — correct for startup validation.
- **Fail-closed ordering:** `AllowCallerParquetPaths` → `RenderS3ParquetPath` → empty-check → `validateHintPathScope` (`duckdb_query_build.go:278-325`). Hint-gate errors return from `resolveParquetPaths` (`parquet_source.go:65`) before degraded-mode absorption, so they cannot be relabelled `ErrFederatedReadFailed`.
- **Early returns / flat structure:** resolver and validator follow existing flat style; no new nesting >3.
- **gofmt / whitespace:** `make fmt-check` clean in CI; local `git diff --check` clean (no whitespace errors).
- **AST guard idiom:** follows `TestNoBareSentinelWraps` — explicit blind-spot disclosure (`parquet_literal_guard_test.go:35`), vacuous-scan floor (`:116` `scanned<10`), reports every offender (`:120`).
- **Maintained-list upkeep:** previously missing `docs/error-handling.md` row is now present (`:846`), closing the prior P2 standards gap.

No standards violation found.

iceboundrock:
## Findings

Reviewed `git diff main...HEAD` (25 files, +539/-27) at `28c448d`. Prior P1/P2 are now closed; **no blocking finding remains**.

### [Closed] P1 — Startup fail-fast not wired — FIXED
Previous round: `validateDuckDBConfig` check had no production caller (`Config.Validate` has 0 non-test callers). Now `DuckDBConfig.ValidateCallerParquetPaths` (`config_duckdb.go:360`) is extracted and called in both production entrypoints before any I/O — `cmd/server/main.go:165` (`bootstrapServer`, beside `ValidateManifestRead` and above pool creation) and `factory/factory.go:111` (`newEntityManagerWithConfigContext`, above `collectTables`). `Config.Validate` keeps the shared check (`config_duckdb.go:343`). Pinned by `cmd/server/main_test.go:214` `TestBootstrapServer_RejectsCallerParquetPathsWithoutBucket` (asserts `ConfigError` on `duckdb.s3Bucket` before pool) and `config_duckdb_test.go:346` `TestValidateDuckDBConfig_AllowCallerParquetPathsRequiresBucket`. Verified `rg Config.Validate` → only those three call sites + tests.

### [Closed] P2 — Black-box CI missing opt-in — FIXED
`tests/e2e/scripts/federated-check.ts:143` sends `s3://forma-cdc/delta/{{.SchemaID}}/*.parquet`; CI now sets `.github/workflows/ci.yml:170` `DUCKDB_ALLOW_CALLER_PARQUET_PATHS=true` with `S3_BUCKET: forma-cdc` already in scope, so the hint is **honored** in-scope rather than rejected. E2E harnesses likewise opt in (`internal/e2e_harness/federated/harness.go:230`, `internal/e2e_harness/production/env.go:258`). CI `k6 Smoke` now passes (was red at prior review).

### [Closed] P2 — Maintained HTTP error surface list — FIXED
`docs/error-handling.md:846` now rows `internal/federated/duckdb_query_build.go` (`duckDBParquetPathsForQuery`, `validateHintPathScope`) with the five carrier shapes (flag-off, unrenderable, degenerate, disallowed-char, out-of-scope), noting caller-owned publication vs `WithOperatorDetail` bucket name.

### [Closed] Minor — char-blocklist rationale, guard receiver, test gaps — FIXED
- `duckdb_query_build.go:335-340` correctly states only `'` can alter SQL after `EscapeLiteral`; `"` `;` `\n` `\r` kept as defense-in-depth with trade-off disclosed.
- `parquet_literal_guard_test.go:54` now requires `sqlutil` identifier, so `otherpkg.EscapeLiteral` no longer satisfies.
- Added `internal/federated/parquet_hint_gate_test.go:76` `TestHintRenderedOutOfScopeRejected` (rendered `{{.SchemaID}}` out-of-scope) and `:86` `TestHintMixedScopeRejected` (comma-separated in-scope + out-of-scope rejects whole hint); `validateHintPathScope` now `TrimSpace`d bucket (`duckdb_query_build.go:347`) for padded-bucket consistency; `cap` shadowing renamed.

### No new blocking findings

Sweep for bypass: `rg S3ParquetPathTemplate` shows single production ingress `internal/entity_query_service.go:236-237` → `model.DuckDBRenderHints` → `parquet_source.go:66` `duckDBParquetPathsForQuery(q, e.cfg)`; no other caller. Both hint and manifest-source paths converge into `formatDuckDBPathList` + four executor sites, all now via `sqlutil.EscapeLiteral`. Remaining raw `read_parquet('%s')` in `internal/compaction/`, `internal/e2e_harness/` are not caller-reachable; tracked separately as #478 (see Observations). Scope and escaping are defense-in-depth: a `'` in a hint is rejected at `validateHintPathScope` (`duckdb_query_build.go:355`) *and* doubled at render — intentional.

iceboundrock:
## Observations

Non-blocking, for follow-up or awareness:

1. **In-bucket glob breadth remains wide when opted in** — a hint `s3://<bucket>/**` or `s3://<bucket>/_tmp/**` or `s3://<bucket>/base/**` is accepted as in-scope (only bucket + trailing-slash is checked, `duckdb_query_build.go:353`). The level-3 fallback deliberately keeps `*` from widening to `**` (`design.md:136`), but the hint path does not enforce that. Impact is bounded to server-written intra-bucket data; tracked as #477 to optionally scope hints to `s3DataPrefix` or forbid `**`/`_tmp`.

2. **`unverifiablePath` still skips quote-bearing paths** — manifest entries containing `'` are marked unverifiable and fall to all-or-nothing verify/guard attribution rather than per-file drains. Render is now safely escaped, so this is a coverage gap, not a bypass. Tracked as #479.

3. **AST guard deliberate blind spot** — `parquet_literal_guard_test.go:35` only inspects direct `fmt.Sprintf` with a literal format string and a direct `sqlutil.EscapeLiteral` arg. Concatenation, indirect formats, or helper dataflow are invisible here and rely on behavioral tests (`internal/sqlgen/duckdb_path_escape_test.go`, `internal/federated/parquet_path_escape_test.go`). Acknowledged in comment; acceptable as forward guard.

4. **CI surface wider than default deployment** — `.github/workflows/ci.yml:170` runs the black-box server with the hint-honoring opt-in so `federated-check` exercises the hint path. Production default stays off; flag-off is covered by `TestHintRejectedWhenFeatureOff` (`parquet_hint_gate_test.go:20`). Worth remembering that CI's DuckDB surface is intentionally larger.

5. **`hybridDuckConfig()` parity choice** — `internal/federated/parquet_source_test.go:15` enables `AllowCallerParquetPaths=true, S3Bucket="b"` for hybrid tests, so existing engine tests that carry a hint run behind the same gate as production. Clearly commented; reasonable.

6. **`parquet_source_test.go` near ceiling** — `wc -l` 495/500. Future additions there should split; this PR does not grow it.

7. **Compaction harness raw interpolations** — `internal/compaction/uncovered_rows.go:54`, `merge_sql.go:112` still `Sprintf("read_parquet('%s')", path)` raw; `internal/e2e_harness/federated/query_build.go:115` and `internal/e2e_harness/production/artifacts.go:280` likewise. Not caller-reachable today; tracked as #478. Out of scope for #456.

8. **Design doc wording residual** — `docs/federated-query/design.md:134` says "rejected at startup (`Config.Validate`)". Actual enforcement is `DuckDBConfig.ValidateCallerParquetPaths` called from both `bootstrapServer` and the factory beside `ValidateManifestRead`; `Config.Validate` keeps it but has no production caller. Consider rephrasing to "rejected at startup (bootstrap + factory preflight, also `Config.Validate`)" for precision — not a behavior bug.

iceboundrock:
## Verification

Local (pinned toolchain `GOTOOLCHAIN=go1.26.0`, no `libstdc++-devel` so CGO link fails for DuckDB packages — environmental, matches prior review):

- `go test -run TestParquetPathRendersAreEscaped -v` — **PASS** (`parquet_literal_guard_test.go:37` covers 3 dirs, `scanned<10` floor holds)
- `go test -run TestValidateDuckDBConfig_AllowCallerParquetPathsRequiresBucket -v .` — **PASS** (`config_duckdb.go:360` whitespace-only bucket rejected, `ConfigError` field `duckdb.s3Bucket`)
- `go test -run TestNoBareSentinelWraps -v .` — **PASS** (no regression on #313 guard)
- `git diff --check` — clean; `make fmt-check` reported clean in CI

`cmd/server` / `internal/federated` / `internal/sqlgen` unit tests requiring DuckDB CGO link could not run in this sandbox (`cannot find -lstdc++` — missing `libstdc++-devel`, not a test failure), identical to prior review environment. Root-package tests that are link-independent all pass.

CI (actions `32711715434`, branch `fix/456-parquet-path-injection`):

- `Lint` ✅ (`golangci-lint v1.64.8`, 0 issues)
- `Test` ✅ (unit `go test ./... -race`)
- `k6 Smoke` ✅ (was **red** at prior review due to missing opt-in; now `DUCKDB_ALLOW_CALLER_PARQUET_PATHS=true` exercises the honored hint path `s3://forma-cdc/delta/{{.SchemaID}}/*.parquet` instead of default-off rejection — fix confirmed)
- `Benchmark Smoke` ✅
- `E2E Smoke Tests` ⏳ pending at time of review (Docker-gated `TestE2EHarnessMinimal` is environmental; no change to infra)

Mutation spot-check from prior round remains valid: reverting `drainParquet`'s `EscapeLiteral` makes `TestParquetPathRendersAreEscaped` fail at `internal/federated/parquet_verify.go:73`; guard is not vacuous.

No verification red remains attributable to this PR.

iceboundrock:
## Assessment

The fix is **sound and complete** for #456. Three fail-closed layers reinforce each other:

1. **Opt-in gate** — `AllowCallerParquetPaths` default-off (`config_duckdb.go:81`, `cmd/server/main.go:106`). Flag-off **rejects** (`duckdb_query_build.go:288` `InvalidInputf`) rather than silently ignoring, preserving provenance and preventing a manifest-source fallback from misreporting the answer.
2. **Bucket allowlist** — `validateHintPathScope` (`duckdb_query_build.go:346`) enforces exact `s3://<bucket>/` prefix with defense-in-depth char blocklist (`"'\";\n\r"`), defeating local-file / `https://` SSRF / cross-bucket / `s3://bucketX/` prefix-collision; comma-separated multi-paths validated per segment. Startup bucket-required (`config_duckdb.go:360`, wired in `cmd/server/main.go:165` + `factory/factory.go:111` before any I/O) turns operator misconfiguration into fail-fast rather than per-request 4xx.
3. **Universal escaping** — `sqlutil.EscapeLiteral` at all five production render sites plus sibling `parquetcheck`, with AST forward guard (`parquet_literal_guard_test.go:37`) that now also requires the `sqlutil` receiver and scans `internal/parquetcheck`.

Error semantics follow #313: caller-owned template/path published; bucket name stays behind `WithOperatorDetail`. Docs, maintained error-surface table, CI, and E2E harnesses have been migrated.

Recommendation: **approve and merge** (do not auto-merge per repo rules). No code-level injection bypass remains; residual items (#477 glob breadth, #478 compaction, #479 `unverifiablePath`) are correctly tracked as hardening follow-ups, not blockers.

iceboundrock:
## Summary

PR #476 closes #456's caller-reachable DuckDB SQL injection + arbitrary-path read with server S3 credentials. On any DuckDB-enabled deployment, `federated.s3_parquet_path_template` was interpolated raw into `read_parquet`/`glob`/`DESCRIBE`. The fix layers:

- **Default-off opt-in** `duckdb.allowCallerParquetPaths` / `DUCKDB_ALLOW_CALLER_PARQUET_PATHS` — existing deployments stay closed; hint rejected as `ErrInvalidInput` 4xx echoing caller path.
- **Bucket allowlist** `s3://<S3Bucket>/` exact with trailing slash, per-segment, plus `InvalidInput` on `' " ; \n \r`; empty-bucket rejected at startup via `DuckDBConfig.ValidateCallerParquetPaths`.
- **Universal escaping** via `sqlutil.EscapeLiteral` at `formatDuckDBPathList`, `drainParquet`, `drainGuardedParquet`, `globParquetPaths`, `describeParquetColumns` (and `parquetcheck`), guarded by source-level AST test.

All prior review findings (P1 startup wiring, P2 docs/CI, minor rationale/guard/tests) are addressed; `k6 Smoke` now passes. Tests pin flag-off, in-scope glob, cross-bucket/prefix-collision/local/http, quote-doubling at all sites, empty-bucket, rendered-out-of-scope, mixed-scope multi-path, env wiring, bootstrap fail-fast, and guard mutation.

Follow-ups #477 (hint glob breadth), #478 (compaction escaping), #479 (`unverifiablePath` quote coverage) are tracked. Security posture: **no bypass found** on the reviewed production paths.

iceboundrock:
## Review response — round 3 (commits `cc7b6c3`, `8283f3f`)

Thanks for the thorough pass — the assessment is **approve and merge** and all prior findings are confirmed closed. One new actionable item, plus an unrelated toolchain change requested out-of-band.

### Observation 8 — design-doc wording — **fixed** (`cc7b6c3`)
`docs/federated-query/design.md` §4.3.1 no longer says "rejected at startup (`Config.Validate`)". It now states the enforced path precisely: rejected by `DuckDBConfig.ValidateCallerParquetPaths`, called from the `cmd/server` bootstrap and the factory preflight beside `ValidateManifestRead` (and also from `Config.Validate`, which has no production caller). Good catch on the imprecision I introduced.

### Other observations — already tracked / acknowledged, no change
- **[1] in-bucket glob breadth** → #477; **[2] `unverifiablePath`** → #479; **[7] compaction/harness raw interpolation** → #478 (I'll note the extra harness sites — `e2e_harness/federated/query_build.go`, `production/artifacts.go` — in that issue).
- **[3] AST guard blind spot**, **[4] CI surface wider than default**, **[5] `hybridDuckConfig` parity**, **[6] `parquet_source_test.go` at 495 lines** — acknowledged, no change; all consistent with the security scope.

### Out-of-band: GOTOOLCHAIN → `go1.26+auto` (`8283f3f`)
Separate maintainer request, folded in here: the Makefile pin moves from the sed-derived exact `go1.26.0` to `go1.26+auto`. Verified this preserves #448's protection rather than reintroducing the problem — with a local `go1.26.6`, `GOTOOLCHAIN=go1.26+auto go version` still resolves to **go1.26.0** (whereas plain `GOTOOLCHAIN=auto` resolves to the local go1.26.6). The `go1.26` floor fixes the selected toolchain at go1.26.0 without adopting a newer local Go, while `+auto` will follow a higher `go`/`toolchain` directive in go.mod if one is ever added — so a future bump needs no Makefile edit. Updated the AGENTS.md examples/prose and the #448 toolchain-guard comment to match; guard logic and CI coupling are unchanged (go.mod still pins `go 1.26.0`).

### Verification
`make fmt-check` ✅ · `make lint` ✅ (0 issues, run under `go1.26+auto` → go1.26.0) · `TestGatesRunUnderPinnedToolchain` / `TestLintRecipeInheritsGOENV` ✅ · root/federated/config/gate/guard tests ✅.

No open review items remain on my side.

iceboundrock:
## Findings

Reviewed final head `2cf8ca4` and the complete branch diff. No new blocking, P1, or P2 finding remains. The prior findings for the black-box CI opt-in, startup validation wiring, public error-surface documentation, and final startup-validation wording are all addressed in the current head.

Recommendation: no further code changes are required for this PR.

iceboundrock:
## Standards

Reviewed against `coding-standard.md`, `AGENTS.md`, and `docs/error-handling.md`. The changed source remains within the file and function size limits, control flow is flat, configuration errors preserve their typed classification, and runtime caller errors use the established `InvalidInputf` plus operator-detail semantics.

`make fmt-check` and `git diff --check` are clean. The AST guard documents its intentional blind spots, enforces the `sqlutil` receiver, scans `internal/parquetcheck`, and has a vacuous-scan floor.

iceboundrock:
## Verification

Final-head verification:

- `GOTOOLCHAIN=go1.26.0+auto go version`: resolves to `go1.26.0`.
- `make fmt-check`: passed.
- Targeted root tests for the Parquet escape guard, caller-path configuration validation, and toolchain guard: passed.
- `git diff --check origin/main...HEAD`: clean.
- Local `make test`: root and non-DuckDB packages passed, but DuckDB-linked packages could not link because this sandbox lacks `libstdc++`; this is environmental.
- PR CI run `32715236140`: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all passed.

iceboundrock:
## Observations

Non-blocking follow-ups remain appropriately tracked:

- #477 covers the intentionally broad in-bucket hint glob surface, including `**` and staging prefixes.
- #478 covers raw operator-controlled Parquet renders in compaction and harness code outside the caller-reachable federated path.
- #479 covers quote-bearing manifest paths being skipped by `unverifiablePath`.
- The source-level AST guard does not inspect indirect or concatenated render construction; the current behavioral tests cover the shipped sites and the limitation is documented.
- CI intentionally enables the caller-hint feature to exercise the migration path, while the production default remains off.

iceboundrock:
## Assessment

The security fix is sound and complete for #456. The production path is fail-closed in the intended order: opt-in gate, template rendering, empty-result rejection, character defense, exact bucket scope, and SQL literal escaping. Startup validation is wired in both the server bootstrap and factory preflight before I/O.

Assessment: approve and merge when the project owner is ready. Do not auto-merge.

iceboundrock:
## Summary

PR #476 closes the caller-reachable DuckDB SQL injection and arbitrary-path read by combining a default-off deployment gate, exact configured-bucket scope, universal `sqlutil.EscapeLiteral` coverage, and regression guards. The final head also wires startup validation, updates the public error contract and migration documentation, enables the relevant CI smoke path, and pins the Go toolchain behavior consistently.

No unresolved blocking finding was found in the final review.

iceboundrock:
## Spec

The #456 acceptance criteria are satisfied at the final head:

- Caller path hints are rejected by default and require the server-side `AllowCallerParquetPaths` opt-in.
- Opted-in paths must match the exact `s3://<S3Bucket>/` prefix, including the trailing-slash prefix-collision case; local, HTTP, cross-bucket, disallowed-character, and mixed-scope inputs are rejected.
- Current `read_parquet`, `glob`, `DESCRIBE`, list-rendering, guarded-drain, verification, and parquetcheck paths use `sqlutil.EscapeLiteral`.
- Invalid or empty explicit hints do not silently fall through to manifest resolution.
- The configuration and E2E documentation describe the default-off migration contract.

iceboundrock:
Updated in commit `998b9fe` (`test: support Docker and Podman test runtimes`).

Added `scripts/test_with_container_runtime.sh`, which honors an existing `DOCKER_HOST`, detects a reachable Docker daemon, or starts the rootless Podman socket and configures testcontainers with `TESTCONTAINERS_RYUK_DISABLED=true`. Usage and rootless Podman behavior are documented in `AGENTS.md`, `README.md`, and `tests/e2e/README.md`.

Verified locally with Podman: `./scripts/test_with_container_runtime.sh` passed the full `make test`; `make fmt-check`, shell syntax validation, and `git diff --check` also passed.

PR #482, opened 2026-08-24
ci: lint job runs make lint, restoring local/CI parity (#454)

### LINKED ISSUE #454: ci: lint job open-codes install+run instead of make lint, breaking local/CI parity the Makefile now provides
Follow-up from #448's final review (CI was a binding do-not-touch constraint there).

`.golangci.yml`'s header states the goal is to keep local `make lint` and CI byte-identical, but `.github/workflows/ci.yml` open-codes the golangci-lint install+run instead of calling `make lint`. Since #448, the Makefile's lint recipe pins `GOTOOLCHAIN` (derived from go.mod) and sets `GOCACHE`/`GOFLAGS`; CI's open-coded invocation does none of that. Parity holds today only because CI's setup-go version happens to equal go.mod's directive — the same coincidence #448 proved unreliable locally.

Proposed: have the CI lint job run `make lint` (or at minimum export the same `GOENV` set), so the pin has one definition. Note the root-package guard `TestGatesRunUnderPinnedToolchain` (added in #448's PR) already fails CI's test job if `GO_VERSION` moves to a newer major.minor than go.mod's `go` directive — but nothing equivalent covers the lint job's invocation shape.

iceboundrock:
## Summary

- CI's lint job now runs `make lint` instead of open-coding the golangci-lint install+run. The Makefile recipe is the single definition of the linter version pin (v1.64.8) and of `GOENV` (`GOTOOLCHAIN=go1.26.0+auto`, `GOCACHE`, `GOFLAGS`) — previously parity held only because setup-go's `GO_VERSION` coincidentally equaled the toolchain floor, the same coincidence #448 proved unreliable locally.
- New root-package guard `TestCILintJobRunsMakeLint` covers the lint job's invocation shape (the gap the issue named): `ci.yml` must contain `make lint` and must not mention `golangci-lint` directly, so re-adding either the install line or the run line fails CI's test job. It sits beside the two existing #448 guards in `toolchain_guard_test.go`.
- `.golangci.yml`'s header no longer claims the version is pinned in two places; it now points at the Makefile recipe and the new guard. The workflow step was renamed `Run linter` because the guard rejects any direct `golangci-lint` mention in `ci.yml` (the step name included).

## Known tradeoff

`make lint` sets `GOCACHE` to `$PWD/.gocache`, which setup-go's default cache step does not save/restore (it queries the ambient `go env GOCACHE`). The lint job therefore loses build-cache reuse across runs — accepted: parity is the point of the change, and the job already reinstalled golangci-lint on every run.

## Verification

- `TestCILintJobRunsMakeLint` written first and observed failing on both branches against the old workflow (TDD); root package `go test .` now passes.
- `make fmt-check && make lint` green locally — the exact pair CI's lint job now runs.

Closes #454

https://claude.ai/code/session_0196sNgz4EfzUQ3xS667Fb2R

### DIFF
--- a/toolchain_guard_test.go
+++ b/toolchain_guard_test.go
@@ -90,8 +90,25 @@ func TestCILintJobRunsMakeLint(t *testing.T) {
 		t.Fatalf("read .github/workflows/ci.yml: %v", err)
 	}
 	text := string(ci)
-	if !strings.Contains(text, "make lint") {
-		t.Error("ci.yml has no `make lint` step: the lint job must run the Makefile recipe so the golangci-lint pin and GOENV have one definition (#454)")
+	// The positive half must match a real run directive, not prose: ci.yml's
+	// explanatory comment also says "make lint", so a bare substring search
+	// stays green even with the invocation deleted (#482 review). Comment
+	// lines are skipped explicitly so prose never counts, whatever shape the
+	// directive pattern takes in the future.
+	runMakeLint := regexp.MustCompile(`^run:\s*make lint\s*$`)
+	found := false
+	for _, line := range strings.Split(text, "\n") {
+		trimmed := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+		if runMakeLint.MatchString(trimmed) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Error("ci.yml has no `run: make lint` directive: the lint job must run the Makefile recipe so the golangci-lint pin and GOENV have one definition (#454)")
 	}
 	if strings.Contains(text, "golangci-lint") {
 		t.Error("ci.yml mentions golangci-lint directly: the install+run belong to the Makefile lint recipe alone — an open-coded copy drops the GOTOOLCHAIN pin and re-splits the version pin (#454)")
--- a/toolchain_guard_test.go
+++ b/toolchain_guard_test.go
@@ -75,3 +75,25 @@ func TestLintRecipeInheritsGOENV(t *testing.T) {
 		t.Fatal("no golangci-lint run line found in Makefile: the lint-pin guard has lost its target")
 	}
 }
+
+// TestCILintJobRunsMakeLint pins the CI side of the #448 wiring (#454): the
+// lint job must invoke `make lint` rather than open-coding the golangci-lint
+// install+run. An open-coded invocation carries none of the Makefile's GOENV
+// (GOTOOLCHAIN pin, GOCACHE, GOFLAGS), so local/CI parity would hold only
+// while setup-go's GO_VERSION coincidentally equals the Makefile's toolchain
+// floor — the same coincidence #448 proved unreliable. Asserting the absence
+// of "golangci-lint" in the workflow catches both halves of a regression:
+// re-adding the install line or the run line.
+func TestCILintJobRunsMakeLint(t *testing.T) {
+	ci, err := os.ReadFile(".github/workflows/ci.yml")
+	if err != nil {
+		t.Fatalf("read .github/workflows/ci.yml: %v", err)
+	}
+	text := string(ci)
+	if !strings.Contains(text, "make lint") {
+		t.Error("ci.yml has no `make lint` step: the lint job must run the Makefile recipe so the golangci-lint pin and GOENV have one definition (#454)")
+	}
+	if strings.Contains(text, "golangci-lint") {
+		t.Error("ci.yml mentions golangci-lint directly: the install+run belong to the Makefile lint recipe alone — an open-coded copy drops the GOTOOLCHAIN pin and re-splits the version pin (#454)")
+	}
+}
--- a/toolchain_guard_test.go
+++ b/toolchain_guard_test.go
@@ -82,22 +82,23 @@ func TestLintRecipeInheritsGOENV(t *testing.T) {
 // (GOTOOLCHAIN pin, GOCACHE, GOFLAGS), so local/CI parity would hold only
 // while setup-go's GO_VERSION coincidentally equals the Makefile's toolchain
 // floor — the same coincidence #448 proved unreliable. Asserting the absence
-// of "golangci-lint" in the workflow catches both halves of a regression:
-// re-adding the install line or the run line.
+// of "golangci-lint" in the whole workflow catches both halves of a
+// regression: re-adding the install line or the run line.
 func TestCILintJobRunsMakeLint(t *testing.T) {
 	ci, err := os.ReadFile(".github/workflows/ci.yml")
 	if err != nil {
 		t.Fatalf("read .github/workflows/ci.yml: %v", err)
 	}
 	text := string(ci)
-	// The positive half must match a real run directive, not prose: ci.yml's
-	// explanatory comment also says "make lint", so a bare substring search
-	// stays green even with the invocation deleted (#482 review). Comment
-	// lines are skipped explicitly so prose never counts, whatever shape the
-	// directive pattern takes in the future.
+	// The positive half must match a real run directive, not prose, and only
+	// inside the lint job (#482 review, both rounds): ci.yml's explanatory
+	// comment also says "make lint", and another job legitimately running the
+	// same command must not satisfy a guard about the lint job's wiring.
+	// Comment lines are skipped explicitly so prose never counts, whatever
+	// shape the directive pattern takes in the future.
 	runMakeLint := regexp.MustCompile(`^run:\s*make lint\s*$`)
 	found := false
-	for _, line := range strings.Split(text, "\n") {
+	for _, line := range strings.Split(ciLintJobBlock(t, text), "\n") {
 		trimmed := strings.TrimSpace(line)
 		if strings.HasPrefix(trimmed, "#") {
 			continue
@@ -108,9 +109,47 @@ func TestCILintJobRunsMakeLint(t *testing.T) {
 		}
 	}
 	if !found {
-		t.Error("ci.yml has no `run: make lint` directive: the lint job must run the Makefile recipe so the golangci-lint pin and GOENV have one definition (#454)")
+		t.Error("ci.yml's lint job has no `run: make lint` directive: the lint job must run the Makefile recipe so the golangci-lint pin and GOENV have one definition (#454)")
 	}
 	if strings.Contains(text, "golangci-lint") {
 		t.Error("ci.yml mentions golangci-lint directly: the install+run belong to the Makefile lint recipe alone — an open-coded copy drops the GOTOOLCHAIN pin and re-splits the version pin (#454)")
 	}
+}
+
+// ciLintJobBlock cuts the `lint:` job's body out of ci.yml by indentation: it
+// starts after the two-space `lint:` key under the top-level `jobs:` key and
+// ends at the next non-blank line indented two spaces or less (the next job,
+// or a later top-level key). Deliberately not a YAML parser — a yaml
+// dependency buys nothing here (#482 review round 1), and the sibling guards
+// in this file are plain text scans for the same reason. If the workflow's
+// job indentation ever changes shape, this fails loudly via t.Fatal rather
+// than passing on an empty block.
+func ciLintJobBlock(t *testing.T, text string) string {
+	t.Helper()
+	lines := strings.Split(text, "\n")
+	inJobs := false
+	start := -1
+	for i, line := range lines {
+		if line == "jobs:" {
+			inJobs = true
+			continue
+		}
+		if !inJobs {
+			continue
+		}
+		if start == -1 {
+			if line == "  lint:" {
+				start = i + 1
+			}
+			continue
+		}
+		trimmed := strings.TrimSpace(line)
+		if trimmed != "" && !strings.HasPrefix(line, "    ") {
+			return strings.Join(lines[start:i], "\n")
+		}
+	}
+	if start == -1 {
+		t.Fatal("ci.yml has no `lint:` job under `jobs:`: this guard scopes its run-directive check to that job and has lost its target (#454)")
+	}
+	return strings.Join(lines[start:], "\n")
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,13 @@ jobs:
       - name: Check gofmt
         run: make fmt-check
 
-      - name: Run golangci-lint
-        run: |
-          go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8
-          golangci-lint run --timeout=5m
+      # Run the Makefile recipe, not an open-coded install+run: the recipe is
+      # the single definition of the linter version pin and of GOENV
+      # (GOTOOLCHAIN=go1.26.0+auto, GOCACHE, GOFLAGS), so local `make lint`
+      # and CI stay byte-identical (#454). TestCILintJobRunsMakeLint guards
+      # this shape.
+      - name: Run linter
+        run: make lint
 
   test:
     name: Test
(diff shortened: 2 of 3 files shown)


toolchain_guard_test.go
```diff
@@ -75,3 +75,25 @@ func TestLintRecipeInheritsGOENV(t *testing.T) {
...
+// floor — the same coincidence #448 proved unreliable. Asserting the absence
+// of "golangci-lint" in the workflow catches both halves of a regression:
+// re-adding the install line or the run line.
+func TestCILintJobRunsMakeLint(t *testing.T) {
+	ci, err := os.ReadFile(".github/workflows/ci.yml")
+	if err != nil {
+		t.Fatalf("read .github/workflows/ci.yml: %v", err)
+	}
+	text := string(ci)
+	if !strings.Contains(text, "make lint") {
```
iceboundrock: **Finding 1 (low-medium): the positive half of this guard is satisfied by its own prose.**

`strings.Contains(text, "make lint")` matches the explanatory *comment* this PR adds to ci.yml ("so local \`make lint\` and CI stay byte-identical"). Verified empirically against this branch:

- Re-inserting an open-coded `golangci-lint run` line → guard FAILS correctly (the absence assertion earns its keep).
- Deleting the actual `run: make lint` invocation while keeping the comment → guard stays GREEN. The lint job would then contain no linter at all and CI would not notice.

So the guard reliably catches the realistic regression (re-open-coding install/run) but not silent neutralization of the invocation. Suggested tightening when convenient: skip comment lines and require an actual run directive, e.g.

```go
for _, line := range strings.Split(text, "\n") {
    trimmed := strings.TrimSpace(line)
    if strings.HasPrefix(trimmed, "#") {
        continue
    }
    if regexp.MustCompile(`^run:\s*make lint\s*$`).MatchString(trimmed) { found = true }
}
```

Not blocking — the wiring shipped here is correct; this is guard robustness.

iceboundrock:
## Review: #482 (closes #454)

### Summary

The CI lint job now runs `make lint` instead of open-coding the golangci-lint install+run, making the Makefile lint recipe the single definition of both the v1.64.8 pin and the GOENV set (`GOTOOLCHAIN=go1.26.0+auto`, `GOCACHE`, `GOFLAGS`). A new root-package guard `TestCILintJobRunsMakeLint` pins the invocation shape (must contain `make lint`, must not mention `golangci-lint`), and `.golangci.yml`'s header no longer claims the version is pinned in two places. Three files, +32/−6, scope tightly limited to the issue.

### Spec / issue fit

#454 asked exactly for this ("have the CI lint job run `make lint`… so the pin has one definition"), noting that nothing covered the lint job's *invocation shape* (the #448 guard `TestGatesRunUnderPinnedToolchain` only ties the test job's toolchain). This PR fills precisely that gap and closes it with a guard, not just a fix — consistent with how #448 landed its wiring guards. The step rename (`Run golangci-lint` → `Run linter`) is forced by the guard's absence assertion and is documented in the PR body rather than left to surprise. No scope creep; no unrelated changes.

### Standards

- `toolchain_guard_test.go` stays at 99 lines / new function ~12 lines — well inside the ≤500 file / ≤100 function limits.
- Guard style matches its two siblings from #448: plain text-scan of a repo file, failure messages that name the diagnosis and remediation, load-bearing comments referencing issue numbers.
- Error messages follow the house pattern (state expected, what regressed, why it matters).
- No production code touched; nothing to lint beyond what `make lint` itself already gates.

### Findings

1. **Guard positive-half self-match (low-medium)** — posted as inline comment on `toolchain_guard_test.go:93`. In full: `strings.Contains(text, "make lint")` is satisfied by this PR's own explanatory comment in ci.yml. Empirically verified on this branch: deleting the actual `run: make lint` line (keeping the comment) leaves the guard green, so a job with *no linter at all* would pass; only re-open-coding `golangci-lint` is caught (by the absence half). Suggested fix: skip `#`-prefixed lines and require a real `run:` directive via regex. Not blocking — the shipped wiring is correct and the realistic regression path is covered.
2. **Whole-file absence check is deliberately broad (info)** — any mention of `golangci-lint` anywhere in ci.yml (a future job, a step name, even prose) fails the test job. That includes a legitimate future switch to `golangci-lint-action`, which would have to revisit this guard on purpose. Acceptable given intent ("install+run belong to the recipe alone"); YAML-scoped parsing would be more surgical but needs a yaml dep for little gain — text style matches the sibling guards.
3. **Cache tradeoff confirmed, correctly characterized (info)** — GOENV overrides GOCACHE only within the recipe's command lines, so setup-go v6's default cache (keyed off ambient `go env GOCACHE`) does not cover the lint compile; GOMODCACHE is untouched and still cached, so module downloads persist and only compilation repeats (~observed Lint wall time 1m48s on this PR's own run). Documented in the PR body as accepted; agreed.

### Observations

- The ci.yml comment block carefully avoids the literal `golangci-lint` so the guard doesn't trip on its own documentation — correct but fragile by construction; finding 1's tightening removes that class of trap for the positive half too.
- Job name `Lint` unchanged, so branch-protection required checks are unaffected by the step rename.
- Parity now holds structurally rather than coincidentally: previously CI relied on setup-go's `GO_VERSION == go.mod` floor equality — exactly the coincidence #448 showed unreliable locally.
- `--timeout=5m` survives via the Makefile recipe (Makefile:102); no behavior silently dropped.

### Verification

- All three guards pass pinned locally (`GOTOOLCHAIN=go1.26.0+auto go test . -run 'TestCILintJobRunsMakeLint|TestLintRecipeInheritsGOENV|TestGatesRunUnderPinnedToolchain' -v`): PASS ×3.
- Full root package `go test .`: PASS.
- Negative test A — re-inserted open-coded `golangci-lint run` into ci.yml: guard FAILS with the intended message. ✓
- Negative test B — deleted `run: make lint`, kept comment: guard PASSES → confirms Finding 1.
- `make fmt-check && make lint` green locally — the exact pair CI's lint job now runs.
- CI on this PR: Lint ✓ (1m48s, new path exercised end-to-end), Test ✓ (includes the new guard), k6 Smoke ✓, Benchmark Smoke ✓; E2E pending at review time.
- No remaining references to the old step name outside gitignored `docs/superpowers/`.

### Assessment

**Approve.** Does what #454 asks, closes it with a guard in the established #448 style, TDD-verified, and the parity claim is now structural. Findings 2–3 are informational; Finding 1 is a worthwhile follow-up hardening of the guard (comment-skipping + `^run:\s*make lint\s*$`) but does not block merge.

toolchain_guard_test.go
iceboundrock: Fixed in 17e2e64. The positive half now skips `#`-prefixed lines and requires a trimmed line matching `^run:\s*make lint\s*$` — your suggested shape, with the comment-skip kept explicit so prose never counts even if the directive pattern changes later.

Mutation-verified locally (all four cases):
- current ci.yml → PASS
- `run: make lint` deleted, comment kept (your negative test B) → now FAILS
- comment-only `# run: make lint`, no real directive → FAILS
- open-coded `golangci-lint run` re-inserted → still FAILS via the unchanged absence half

Fixed in-PR rather than as a follow-up since the PR is unmerged and the change is 19 lines in the same test.

iceboundrock:
Response to the review's findings:

**Finding 1 (guard positive-half self-match)** — fixed in-PR, commit 17e2e64; details and mutation-verification in the inline thread. Taken now rather than as a follow-up because the PR is unmerged and the fix is 19 lines in the same test.

**Finding 2 (whole-file absence check is broad)** — no change, by design. The strictness is the point of the guard as recorded in the #454 ruling: any reintroduction of the string — install line, run line, step name, or prose — should fail loudly and force a deliberate revisit. A future legitimate switch (e.g. `golangci-lint-action`) would change where the pin is defined, which is exactly the kind of change that must consciously rewrite this guard rather than slip past it. No follow-up issue: nothing to do until such a switch is actually proposed, and the guard's failure message is the pointer such a change would need.

**Finding 3 (lint job loses build-cache reuse)** — no change; agreed with the review's characterization (GOMODCACHE still cached, only compilation repeats, observed Lint wall time 1m48s). Not opening a follow-up: restoring build-cache coverage would mean either teaching the workflow about the Makefile's GOCACHE location (re-splitting knowledge this PR just unified) or adding a cache step keyed on `.gocache` — measurable cost is ~1–2 min per lint run, which does not justify re-coupling. Can revisit if lint wall time ever grows enough to matter.

toolchain_guard_test.go
```diff
@@ -75,3 +75,42 @@ func TestLintRecipeInheritsGOENV(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("read .github/workflows/ci.yml: %v", err)
+	}
+	text := string(ci)
+	// The positive half must match a real run directive, not prose: ci.yml's
+	// explanatory comment also says "make lint", so a bare substring search
+	// stays green even with the invocation deleted (#482 review). Comment
+	// lines are skipped explicitly so prose never counts, whatever shape the
+	// directive pattern takes in the future.
+	runMakeLint := regexp.MustCompile(`^run:\s*make lint\s*$`)
```
iceboundrock, line 99: **Finding (medium): the guard is not scoped to the lint job.**

`TestCILintJobRunsMakeLint` scans all of `ci.yml` and accepts any non-comment line matching `run: make lint`. If `jobs.lint.steps` loses its invocation while another job gains the same command, this test still passes even though the lint job no longer runs the Makefile recipe. The current workflow is correct, but the guard does not enforce the invariant described by its name and error message.

Please parse the workflow (or otherwise limit the scan to `jobs.lint.steps`) and assert the directive there. This is guard coverage rather than a current runtime failure, but it should be addressed if the test is intended to prevent future CI wiring regressions.

iceboundrock:
**Verification**

Local checks passed:

- `make fmt-check`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test . -run 'Test(GatesRunUnderPinnedToolchain|LintRecipeInheritsGOENV|CILintJobRunsMakeLint)$' -count=1 -v`\n- `make lint`\n- `make -n lint` confirmed the recipe propagates GOENV\n- `git diff --check main...HEAD`\n\nGitHub checks observed passing: Lint, Test, k6 Smoke, and Benchmark Smoke. E2E Smoke Tests were still pending when checked.

iceboundrock:
**Observations**

The second commit fixes the previously identified false-positive path where the explanatory comment itself satisfied `strings.Contains(text, "make lint")`. Skipping comment lines and requiring the exact run directive is a good targeted correction. The same text-based approach still has one structural blind spot: it does not distinguish `jobs.lint` from other workflow jobs.

iceboundrock:
**Standards**

The change is focused and follows the repository conventions: errors include actionable context, the new test stays within the source/function size limits, and the guard uses standard-library parsing primitives consistent with the existing test. The main standards concern is test precision: a guard named for the lint job should assert the job-scoped structure rather than rely on a repository-wide text scan.

iceboundrock:
**Assessment**

Current behavior is correct and the PR is mergeable from an execution standpoint. I found one medium-severity maintainability/coverage issue in `TestCILintJobRunsMakeLint`: it can pass when the lint job is empty if another job contains `run: make lint`. I recommend tightening the guard before merge; no current CI runtime regression was found.

iceboundrock:
**Summary**

The workflow change correctly restores local/CI lint parity, and the added positive guard now rejects comment-only or missing invocations. One medium finding remains: scope the guard to `jobs.lint.steps` so it cannot be satisfied by another job. All relevant local validation passed, and no current runtime failure was found.

iceboundrock:
**Spec**

The PR meets the primary #454 specification in the current workflow: the lint job invokes `make lint`, so the Makefile owns the golangci-lint version and GOENV wiring. The follow-up also correctly requires a real `run: make lint` directive instead of accepting explanatory prose. The remaining specification gap is that the guard should verify this directive specifically under `jobs.lint.steps`.

toolchain_guard_test.go
iceboundrock: Fixed in 9a9f4de. `ciLintJobBlock` cuts the lint job's body out by indentation (after `  lint:` under top-level `jobs:`, ending at the next non-blank line indented ≤ 2 spaces) and the run-directive check now runs only inside that block. If the `lint:` job key disappears or the indentation shape changes, the extraction t.Fatals with a named diagnosis instead of passing on an empty block.

Went with your "otherwise limit the scan" alternative rather than a YAML parser: round 1's own assessment was that a yaml dependency buys little here, and the sibling guards in this file are plain text scans for the same reason.

Mutation-verified (six cases): current ci.yml PASS; run line deleted from lint job FAIL; deleted from lint job while added to the test job FAIL (the gap you found); a second job also running `make lint` alongside an intact lint job PASS (no false positive); `lint:` renamed FATAL; comment-only `# run: make lint` FAIL.

The `golangci-lint` absence half deliberately stays whole-file — that breadth was ruled intentional in round 1.

iceboundrock:
Response to review round 2 (Summary / Spec / Standards / Observations / Assessment / Verification comments — all pointing at the one medium finding):

**Guard not scoped to `jobs.lint`** — fixed in 9a9f4de before merge, as requested. The positive half now runs only inside the lint job's block, extracted by indentation (`ciLintJobBlock`), with a loud t.Fatal if the `lint:` job key ever disappears. Chose the "otherwise limit the scan" option over a YAML parser, consistent with round 1's own cost/benefit call on a yaml dependency and with the sibling guards' text-scan style. Six-case mutation matrix (including the exact scenario named: lint job loses the invocation while another job gains it, and the converse no-false-positive case) is in the inline thread reply.

No other actionable items found in this round; nothing out of scope, so no follow-up issues. The absence half remains whole-file per the round-1 ruling.

iceboundrock:
## Findings

No actionable findings remain on the current PR head. The two guard gaps identified in earlier review rounds are addressed: the positive assertion ignores comments and is scoped to the `jobs.lint` block. The current workflow still invokes `make lint`, and no runtime regression was found.

iceboundrock:
## Assessment

Approve. The implementation is correct, narrowly scoped, and the final guard is materially stronger than the initial version. The accepted lint build-cache tradeoff is documented and does not introduce a correctness issue.

iceboundrock:
## Standards

The change is focused and follows repository conventions. The modified Go source remains within the repository file/function size limits, uses standard-library parsing consistent with the existing guard tests, and provides actionable failure messages. The workflow and Makefile retain the existing lint timeout and pinned toolchain behavior.

iceboundrock:
## Summary

Final review is clean. The PR restores local/CI lint parity by routing the workflow through `make lint`, updates the documentation comment, and adds a job-scoped guard against regressions. All relevant local and GitHub verification passed; no blocking findings remain.

iceboundrock:
## Observations

- The step name change avoids reintroducing the forbidden direct `golangci-lint` reference while preserving the `Lint` job name used by branch protection.
- The whole-workflow absence check is intentionally strict; any future direct linter integration or explanatory reference will require an explicit guard update.
- The indentation-based `jobs.lint` extraction avoids a new YAML dependency and fails closed when the expected workflow shape changes.

iceboundrock:
## Verification

Local verification on the PR head passed:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test . -run 'Test(GatesRunUnderPinnedToolchain|LintRecipeInheritsGOENV|CILintJobRunsMakeLint)$' -count=1 -v`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test .`
- `make fmt-check`
- `make lint`
- `git diff --check origin/main...HEAD`

GitHub checks are also green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

iceboundrock:
## Spec

The PR satisfies issue #454: the CI lint job runs `make lint`, making the Makefile the single owner of the golangci-lint version and `GOENV` configuration. `TestCILintJobRunsMakeLint` now checks for a real directive specifically inside `jobs.lint` and fails loudly if that job disappears.

PR #486, opened 2026-08-25
test: integration suites read DB_*/DATABASE_URL env and fail closed in CI (#385)

### LINKED ISSUE #385: CI provisions a Postgres service the integration suite never connects to — hardcoded DSN ignores DB_* env, suite silently skips

iceboundrock:
Fixes #385.

## What

CI provisions a Postgres service and exports `DB_*` env for it, but neither Postgres-backed test venue ever read that env — both silently skipped in CI while reporting green:

- `internal`'s `connectTestPostgres` hardcoded `postgres://postgres:postgres@localhost:5432/forma`; CI's service is `test:test@localhost:5432/forma_test`, so `Ping` failed and every test built on `setupIntegrationEnv` skipped.
- `factory`'s `connectTestPostgres` skipped unless `DATABASE_URL` was set, and CI never sets `DATABASE_URL` (second dark venue, per the issue's re-verification comment).

## How

New test-support package **`internal/testdb`** is now the single owner of both decisions:

- **DSN resolution** (`ResolveDSN`): `DATABASE_URL` wins when set (preserves the factory suite's historical contract); otherwise the DSN is built from `DB_HOST`/`DB_PORT`/`DB_USER`/`DB_PASSWORD`/`DB_NAME`/`DB_SSL_MODE` via `internal/bootstrap.Env`, defaulting to the `scripts/local_server.sh` Postgres (`postgres:postgres@localhost:5432/forma?sslmode=disable`) — so local developers and CI's service both work with zero configuration.
- **Skip policy** (`FailOnUnreachable` + `Connect`): an unreachable database **fails** when `CI` is set (GitHub Actions sets it on every hosted runner — the fail-closed arm needs no workflow wiring, exactly the pattern decided in the issue's third comment) and **skips** on developer machines without a local Postgres.

Both venues' `connectTestPostgres` helpers are now thin delegations to `testdb.Connect`; no call site changed. `.github/workflows/ci.yml` is untouched.

Note: `internal/pgdsn` was not reused on purpose — it renders libpq keyword/value DSNs for DuckDB's postgres scanner, while this path needs the URL form that `DATABASE_URL` and pgxpool already speak.

## Verification (local, Postgres 16 via rootless Podman)

- `go test ./internal -run Integration -v`: **9 integration tests PASS, 0 SKIP** (previously all dark unless a `postgres:postgres/forma` instance happened to exist).
- Same with CI's exact env shape (`CI=true DB_USER=test DB_PASSWORD=test DB_NAME=forma_test` against a matching database): 9 PASS — proves the service credentials work end to end.
- `CI=true` + unreachable port: **FAIL** with `integration database unreachable in CI …` (both venues) — the issue's second acceptance criterion.
- `CI` unset + unreachable port: 9 SKIP, exit 0 — developer machines keep the quiet skip.
- `factory`: 51 test/subtest PASS with only `DB_*` defaults (previously skipped on `DATABASE_URL not set`); `DATABASE_URL` precedence re-verified.
- Full suite via `./scripts/test_with_container_runtime.sh`: green. `make fmt-check` + `make lint`: clean.

Acceptance criterion "tests report as run, not skipped, in the CI run log" to be confirmed on this PR's own Test job log once CI completes.

https://claude.ai/code/session_01PEk6FGV1nAzs3URty4djtG

### DIFF
--- a/internal/testdb/testdb.go
+++ b/internal/testdb/testdb.go
@@ -74,7 +74,7 @@ func Connect(t testing.TB, ctx context.Context) *pgxpool.Pool {
 		if FailOnUnreachable() {
 			t.Fatalf("integration database unreachable in CI — the workflow provisions Postgres on purpose, so this is a failure, not a skip (#385): %v", err)
 		}
-		t.Skipf("skipping: no local Postgres (%v)", err)
+		t.Skipf("skipping: integration Postgres unreachable (%v)", err)
 	}
 
 	t.Cleanup(pool.Close)
--- a/internal/testdb/testdb.go
+++ b/internal/testdb/testdb.go
@@ -11,9 +11,11 @@ package testdb
 
 import (
 	"context"
+	"errors"
 	"net"
 	"net/url"
 	"os"
+	"syscall"
 	"testing"
 	"time"
 
@@ -38,7 +40,7 @@ func ResolveDSN() string {
 		),
 		Host:     net.JoinHostPort(bootstrap.Env("DB_HOST", "localhost"), bootstrap.Env("DB_PORT", "5432")),
 		Path:     "/" + bootstrap.Env("DB_NAME", "forma"),
-		RawQuery: "sslmode=" + bootstrap.Env("DB_SSL_MODE", "disable"),
+		RawQuery: "sslmode=" + url.QueryEscape(bootstrap.Env("DB_SSL_MODE", "disable")),
 	}
 	return u.String()
 }
@@ -51,10 +53,34 @@ func FailOnUnreachable() bool {
 	return os.Getenv("CI") != ""
 }
 
-// Connect opens and pings a pgxpool for the resolved DSN. When the database
-// is unreachable it fails the test in CI (the service is provisioned on
-// purpose — a skip would go dark, #385) and skips on developer machines
-// without a local Postgres. The pool is closed via t.Cleanup.
+// unreachable reports whether err indicates network-level absence of the
+// database — nothing listening, unknown host, connect timeout. A server that
+// answered and rejected us (bad credentials, missing database, protocol/TLS
+// failure) is deliberately excluded: that is misconfiguration, which must
+// fail loudly rather than skip, even off-CI (#486 review P2).
+func unreachable(err error) bool {
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) {
+		return true
+	}
+	if errors.Is(err, syscall.ECONNREFUSED) ||
+		errors.Is(err, syscall.EHOSTUNREACH) ||
+		errors.Is(err, syscall.ENETUNREACH) {
+		return true
+	}
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		return true
+	}
+	return errors.Is(err, context.DeadlineExceeded)
+}
+
+// Connect opens and pings a pgxpool for the resolved DSN. Connection failures
+// split three ways: in CI every failure is fatal (the service is provisioned
+// on purpose — a skip would go dark, #385); locally, network-level absence
+// skips (a developer without Postgres), while a server that answered and
+// rejected the connection is misconfiguration and fails loudly (#486 review
+// P2). The pool is closed via t.Cleanup.
 func Connect(t testing.TB, ctx context.Context) *pgxpool.Pool {
 	t.Helper()
 
@@ -72,7 +98,10 @@ func Connect(t testing.TB, ctx context.Context) *pgxpool.Pool {
 	}
 	if err != nil {
 		if FailOnUnreachable() {
-			t.Fatalf("integration database unreachable in CI — the workflow provisions Postgres on purpose, so this is a failure, not a skip (#385): %v", err)
+			t.Fatalf("integration database connection failed in CI — the workflow provisions Postgres on purpose, so this is a failure, not a skip (#385): %v", err)
+		}
+		if !unreachable(err) {
+			t.Fatalf("integration Postgres answered but rejected the connection — misconfiguration, not absence; fix DATABASE_URL / DB_* env (#486 review): %v", err)
 		}
 		t.Skipf("skipping: integration Postgres unreachable (%v)", err)
 	}
--- a/internal/testdb/testdb.go
+++ b/internal/testdb/testdb.go
@@ -45,11 +45,12 @@ func ResolveDSN() string {
 	return u.String()
 }
 
-// FailOnUnreachable reports whether an unreachable database is a failure
-// rather than a skip. GitHub Actions sets CI on every hosted runner, so this
-// needs no workflow wiring and stays correct if more jobs provision Postgres
-// (#385).
-func FailOnUnreachable() bool {
+// InCI reports whether the tests run under CI, where a database is
+// provisioned on purpose and therefore every connection failure — not only
+// unreachability — is fatal rather than skippable. GitHub Actions sets CI on
+// every hosted runner, so this needs no workflow wiring and stays correct if
+// more jobs provision Postgres (#385).
+func InCI() bool {
 	return os.Getenv("CI") != ""
 }
 
@@ -86,6 +87,9 @@ func unreachable(err error) bool {
 			// The op tells where the failure happened: "dial" is absence
 			// (refused, unreachable, DNS, dial timeout all surface here);
 			// "read"/"write" mean the server accepted the connection first.
+			// Deliberately errno-blind at dial time: nothing was accepted
+			// yet, so any dial failure reads as absence (#486 review round
+			// 4, recorded as intended breadth).
 			return e.Op == "dial"
 		case *net.DNSError:
 			return true
@@ -122,11 +126,11 @@ func Connect(t testing.TB, ctx context.Context) *pgxpool.Pool {
 		}
 	}
 	if err != nil {
-		if FailOnUnreachable() {
+		if InCI() {
 			t.Fatalf("integration database connection failed in CI — the workflow provisions Postgres on purpose, so this is a failure, not a skip (#385): %v", err)
 		}
 		if !unreachable(err) {
-			t.Fatalf("integration Postgres answered but rejected the connection — misconfiguration, not absence; fix DATABASE_URL / DB_* env (#486 review): %v", err)
+			t.Fatalf("integration database connection failed for a reason other than network absence — misconfiguration, not a missing Postgres; fix DATABASE_URL / DB_* env (#486 review): %v", err)
 		}
 		t.Skipf("skipping: integration Postgres unreachable (%v)", err)
 	}
--- a/internal/testdb/testdb.go
+++ b/internal/testdb/testdb.go
@@ -53,26 +53,51 @@ func FailOnUnreachable() bool {
 	return os.Getenv("CI") != ""
 }
 
-// unreachable reports whether err indicates network-level absence of the
-// database — nothing listening, unknown host, connect timeout. A server that
-// answered and rejected us (bad credentials, missing database, protocol/TLS
-// failure) is deliberately excluded: that is misconfiguration, which must
-// fail loudly rather than skip, even off-CI (#486 review P2).
+// unreachable reports whether err denotes network-level absence of the
+// database for every connection attempt it carries. Absence means the
+// failure originated at dial time: connection refused, unreachable
+// host/network, DNS resolution, dial timeout. Anything after an accepted
+// connection — server rejection, protocol/TLS failure, handshake or startup
+// timeout — is misconfiguration and must fail loudly, even off-CI (#486
+// review P2).
+//
+// The walk is manual rather than errors.Is/As because pgx joins per-address
+// results (errors.Join inside pgconn.ConnectError), and errors.Is matches
+// when ANY joined branch matches: a mixed outcome — one address refused,
+// another answered and rejected us — would classify as absence and skip away
+// the actionable server-side error. Every joined branch must be absence on
+// its own (#486 review round 3).
 func unreachable(err error) bool {
-	var dnsErr *net.DNSError
-	if errors.As(err, &dnsErr) {
-		return true
+	for err != nil {
+		if joined, ok := err.(interface{ Unwrap() []error }); ok {
+			branches := joined.Unwrap()
+			if len(branches) == 0 {
+				return false
+			}
+			for _, branch := range branches {
+				if !unreachable(branch) {
+					return false
+				}
+			}
+			return true
+		}
+		switch e := err.(type) {
+		case *net.OpError:
+			// The op tells where the failure happened: "dial" is absence
+			// (refused, unreachable, DNS, dial timeout all surface here);
+			// "read"/"write" mean the server accepted the connection first.
+			return e.Op == "dial"
+		case *net.DNSError:
+			return true
+		}
+		// Leaf equality, not errors.Is: Is would descend into joined
+		// subtrees below this node with any-branch semantics.
+		if err == syscall.ECONNREFUSED || err == syscall.EHOSTUNREACH || err == syscall.ENETUNREACH {
+			return true
+		}
+		err = errors.Unwrap(err)
 	}
-	if errors.Is(err, syscall.ECONNREFUSED) ||
-		errors.Is(err, syscall.EHOSTUNREACH) ||
-		errors.Is(err, syscall.ENETUNREACH) {
-		return true
-	}
-	var netErr net.Error
-	if errors.As(err, &netErr) && netErr.Timeout() {
-		return true
-	}
-	return errors.Is(err, context.DeadlineExceeded)
+	return false
 }
 
 // Connect opens and pings a pgxpool for the resolved DSN. Connection failures
--- a/internal/testdb/testdb.go
+++ b/internal/testdb/testdb.go
@@ -1,0 +1,82 @@
+// Package testdb centralizes how integration tests reach a real Postgres.
+//
+// Before #385 each venue hardcoded its own connection story: the internal
+// suite pinned postgres:postgres@localhost:5432/forma and the factory tests
+// required DATABASE_URL — so CI's provisioned service (test:test@…/forma_test,
+// exported as DB_* env) was never reached and every Postgres-backed test
+// silently skipped. This package resolves the DSN from the same env the rest
+// of the repo reads and makes an unreachable database a hard failure in CI,
+// where the service exists on purpose.
+package testdb
+
+import (
+	"context"
+	"net"
+	"net/url"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/lychee-technology/forma/internal/bootstrap"
+)
+
+// ResolveDSN returns the Postgres DSN for integration tests. DATABASE_URL
+// wins when set (the factory suite's historical contract); otherwise the DSN
+// is built from the DB_* variables CI's test job exports, defaulting to the
+// scripts/local_server.sh Postgres (postgres:postgres@localhost:5432/forma).
+func ResolveDSN() string {
+	if dsn := os.Getenv("DATABASE_URL"); dsn != "" {
+		return dsn
+	}
+	u := url.URL{
+		Scheme: "postgres",
+		User: url.UserPassword(
+			bootstrap.Env("DB_USER", "postgres"),
+			bootstrap.Env("DB_PASSWORD", "postgres"),
+		),
+		Host:     net.JoinHostPort(bootstrap.Env("DB_HOST", "localhost"), bootstrap.Env("DB_PORT", "5432")),
+		Path:     "/" + bootstrap.Env("DB_NAME", "forma"),
+		RawQuery: "sslmode=" + bootstrap.Env("DB_SSL_MODE", "disable"),
+	}
+	return u.String()
+}
+
+// FailOnUnreachable reports whether an unreachable database is a failure
+// rather than a skip. GitHub Actions sets CI on every hosted runner, so this
+// needs no workflow wiring and stays correct if more jobs provision Postgres
+// (#385).
+func FailOnUnreachable() bool {
+	return os.Getenv("CI") != ""
+}
+
+// Connect opens and pings a pgxpool for the resolved DSN. When the database
+// is unreachable it fails the test in CI (the service is provisioned on
+// purpose — a skip would go dark, #385) and skips on developer machines
+// without a local Postgres. The pool is closed via t.Cleanup.
+func Connect(t testing.TB, ctx context.Context) *pgxpool.Pool {
+	t.Helper()
+
+	cfg, err := pgxpool.ParseConfig(ResolveDSN())
+	if err != nil {
+		t.Fatalf("failed to parse integration test DSN (check DATABASE_URL / DB_* env): %v", err)
+	}
+	cfg.ConnConfig.ConnectTimeout = 2 * time.Second
+
+	pool, err := pgxpool.NewWithConfig(ctx, cfg)
+	if err == nil {
+		if err = pool.Ping(ctx); err != nil {
+			pool.Close()
+		}
+	}
+	if err != nil {
+		if FailOnUnreachable() {
+			t.Fatalf("integration database unreachable in CI — the workflow provisions Postgres on purpose, so this is a failure, not a skip (#385): %v", err)
+		}
+		t.Skipf("skipping: no local Postgres (%v)", err)
+	}
+
+	t.Cleanup(pool.Close)
+	return pool
+}
--- a/factory/factory_schema_registry_test.go
+++ b/factory/factory_schema_registry_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/lychee-technology/forma/internal/testdb"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -18,7 +19,7 @@ func TestNewFileSchemaRegistry_Integration_Success(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 
 	// Create schema registry table
 	suffix := time.Now().UnixNano()
@@ -79,7 +80,7 @@ func TestNewFileSchemaRegistry_Integration_EmptyRegistry(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 
 	// Create empty schema registry table
 	suffix := time.Now().UnixNano()
@@ -112,7 +113,7 @@ func TestNewFileSchemaRegistry_Integration_InvalidDirectory(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 
 	// Create schema registry table with a schema entry
 	suffix := time.Now().UnixNano()
@@ -150,7 +151,7 @@ func TestNewFileSchemaRegistry_Integration_MissingAttributesFile(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 
 	// Create schema registry table
 	suffix := time.Now().UnixNano()
@@ -191,7 +192,7 @@ func TestNewFileSchemaRegistry_Integration_DuplicateSchemaIDFails(t *testing.T) 
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 
 	suffix := time.Now().UnixNano()
 	tableName := fmt.Sprintf("schema_registry_dup_id_%d", suffix)
--- a/factory/factory_test.go
+++ b/factory/factory_test.go
@@ -36,15 +36,6 @@ func writeSchemaFile(t *testing.T, dir, schemaName string, schema map[string]any
 	require.NoError(t, err)
 }
 
-// connectTestPostgres establishes a connection to the test PostgreSQL
-// database. DATABASE_URL still wins when set; otherwise the DSN comes from
-// DB_* env with local defaults, and an unreachable database fails (CI) or
-// skips (local) — see internal/testdb (#385).
-func connectTestPostgres(t *testing.T, ctx context.Context) *pgxpool.Pool {
-	t.Helper()
-	return testdb.Connect(t, ctx)
-}
-
 // createTempTables creates temporary tables for testing and returns their names.
 func createTempTables(t *testing.T, ctx context.Context, pool *pgxpool.Pool) (schemaRegistryTable, entityMainTable, eavDataTable string) {
 	t.Helper()
@@ -135,7 +126,7 @@ func TestNewEntityManagerWithConfig_Integration_Success(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 	schemaRegistryTable, entityMainTable, eavDataTable := createTempTables(t, ctx, pool)
 
 	// Create temp schema directory
@@ -182,7 +173,7 @@ func TestNewEntityManagerWithConfig_Integration_MissingTables(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 
 	// Use non-existent table names
 	config := forma.DefaultConfig(newMockSchemaRegistry())
@@ -203,7 +194,7 @@ func TestNewEntityManagerWithConfig_Integration_NilSchemaRegistry(t *testing.T) 
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	pool := connectTestPostgres(t, ctx)
+	pool := testdb.Connect(t, ctx)
 	schemaRegistryTable, entityMainTable, eavDataTable := createTempTables(t, ctx, pool)
 
 	// Create temp schema directory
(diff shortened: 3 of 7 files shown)


iceboundrock:
## Review — Summary & Assessment

**Scope reviewed:** e3e8e2b (3 commits) — new `internal/testdb` package + delegation of both `connectTestPostgres` venues (`factory`, `internal`) to `testdb.Connect`.

### Summary

Fixes #385 at the root cause. Previously each venue hardcoded a connection story that CI's provisioned service (`test:test@…/forma_test`, exported as `DB_*`) never satisfied, so every Postgres-backed test silently skipped in CI while reporting green. This PR centralizes DSN resolution in `internal/testdb`:

- `DATABASE_URL` wins when set (preserves the factory suite's historical contract);
- otherwise the DSN is built from the same `DB_*` variables production bootstrap reads (`internal/bootstrap/config.go:55-60`), defaulting to the `scripts/local_server.sh` Postgres;
- an unreachable database **fails in CI** (`CI` env set — the service is provisioned on purpose) and **skips locally**.

### Assessment

Sound, well-scoped, and matches the issue's proposed fix exactly. The design choices are the right ones:

- Env-name parity with `bootstrap.DatabaseConfigFromEnv` means CI wiring needs zero changes and stays correct if more jobs provision Postgres.
- Pinning the defaults to `local_server.sh`'s values and guarding them with `TestResolveDSNDefaultsMatchLocalServer` gives real drift protection between the two definitions.
- Password URL-escaping is handled by `url.UserPassword` and pinned by a test — the kind of thing that usually breaks silently later.
- Fail-closed-on-`CI` requires no workflow wiring and degrades safely: a developer without Postgres still gets a skip, never a red build from absence of infrastructure.

No blocking findings. Approve from my side.

### Spec alignment — #385 acceptance criteria

| Criterion | Status |
|---|---|
| Internal integration suite connects to CI's Postgres and actually executes there | ✅ Verified in Test job log (run 32880198754): all three `TestIntegration_*` report `PASS` with non-zero durations |
| Unreachable database fails rather than skips when CI has provisioned one | ✅ Verified both directions locally: `CI=true` + closed port → `FAIL` with fail-closed message; unset `CI` → clean `SKIP` |

Refs #385, #355.

iceboundrock:
## Review — Findings & Observations

None blocking; all minor.

### Findings (nits)

1. **Trivial indirection remains** — `factory/factory_test.go:43` and `internal/postgres_persistent_repository_integration_test.go:188` keep one-line `connectTestPostgres` wrappers that only forward to `testdb.Connect`. Keeping them preserves call sites unchanged (reasonable), but the wrappers could be inlined in a follow-up so there is exactly one name for this operation.

### Observations

2. **Parse failures fail loudly even off-CI** (`testdb.go:66-68`): a malformed `DATABASE_URL`/`DB_*` value yields `t.Fatalf` on a developer machine rather than a skip. I read this as correct-by-design — misconfiguration is not absence of infrastructure, and failing loud is the right default. Worth knowing it differs from the unreachable case.
3. **`FailOnUnreachable` keys on `CI`**: any CI system that sets `CI=true` without provisioning Postgres will now hard-fail these suites. That is exactly the conservative behavior #385 asks for ("a skip would go dark"); if it ever bites, an explicit `FORMA_TESTDB_REQUIRE=1` knob is the escape hatch. No change needed now.
4. **Skip message wording** (`testdb.go:73`): "skipping: no local Postgres" can slightly mislead when `DB_HOST` points at a remote host that happens to be down. Cosmetic.
5. **Env surface narrower than production bootstrap** — testdb consumes `DB_HOST/PORT/NAME/USER/PASSWORD/SSL_MODE` but not `DB_SCHEMA` or the pool-tuning vars (`DB_MAX_CONNECTIONS`, …). Fine for tests (schema `public`, small pools are correct); noted only for completeness.
6. **Positive side effect worth recording**: with these tests executing in CI, codecov coverage now includes the real dual-write/read path and the relation-enrichment EntityManager path — previously invisible to coverage entirely.
7. **Shared-database hygiene unchanged**: temp tables use nanotime suffixes + `t.Cleanup` drop, same as before. Safe for CI's single Test job.

### Standards compliance (coding-standard.md)

- File sizes 82/71 lines, all functions well under limits ✅
- Failure messages carry context (what to check: "check DATABASE_URL / DB_* env") ✅
- Comments explain *why* and carry issue refs (#385) per house style ✅
- `gofmt -l` clean, `go vet` clean on touched packages ✅
- Test-only change; no production code paths affected ✅

iceboundrock:
## Review — Verification

### CI evidence: skip-set diff, main vs PR branch

Compared `--- SKIP` sets between the latest green **main** Test job (run 32876288426) and the PR's Test job (run 32880198754):

- main: 19 skips — every Postgres-backed test dark
- PR: 1 skip — and it is the pre-existing permanent marker in `TestDBFederatedQueryEngine_FallsBackToPostgresWhenDuckDBCircuitBreakerOpen` ("moved to internal/federated package"), unrelated to connectivity

**18 tests flipped from SKIP to PASS on the PR branch**, including all of #385's named victims:

```
TestIntegration_EntityLifecycle                        PASS (0.33s)
TestIntegration_AdvancedQuery                          PASS (0.58s)
TestIntegration_AdvancedQuery_MixedSorting             PASS (0.06s)
TestRunOptimizedQueryIntegration                       PASS
TestInsertPersistentRecordIntegration                  PASS
TestChangeLogWritesOnUpdateAndDeleteIntegration        PASS
TestConcurrentDeleteRecreateSerializesIntegration      PASS
TestSameMillisecondWritesStayStrictlyOrderedIntegration PASS
TestRecreateOutranksClockAheadTombstoneIntegration     PASS
TestNewFileSchemaRegistry_Integration_{Success,EmptyRegistry,
    InvalidDirectory,MissingAttributesFile,DuplicateSchemaIDFails}  PASS ×5
TestNewEntityManagerWithConfig_Integration_{Success,MissingTables,
    NilSchemaRegistry,MetadataLoaderError}             PASS ×4
```

This also confirms the relation-enrichment end-to-end path (#385's "why it matters": the only venue wiring real Postgres + `Entity.SchemaDirectory`) now executes in CI.

### Local verification (worktree @ e3e8e2b)

| Check | Result |
|---|---|
| `go test ./internal/testdb/ -v` (5 unit tests: defaults, DB_* read, password escaping, DATABASE_URL precedence, CI-var policy) | PASS |
| Defaults path against a live local Postgres (`DB_*` unset): `TestRunOptimizedQueryIntegration`, `TestIntegration_EntityLifecycle` | PASS — default DSN reaches the running instance |
| Skip path: `CI` unset + closed port (`DB_PORT=15999`) | `SKIP` with `skipping: no local Postgres` |
| Fail-closed path: `CI=true` + closed port | `FAIL` with `integration database unreachable in CI — … a failure, not a skip (#385)` |
| `gofmt -l` on changed files / `go vet` on `internal/testdb`, `factory`, `internal` | clean / clean |
| Factory suite with no special env | `ok` |

### Checks on this PR

Lint ✅ · Test ✅ · Benchmark Smoke ✅ · k6 Smoke ✅ · E2E Smoke pending at time of review (not gated on this change; harness has its own DB setup).

iceboundrock:
Response to the review findings/observations, item by item. Code changes are in 0bfede7.

**1. Trivial indirection (finding)** — Fixed in this PR rather than a follow-up: both `connectTestPostgres` wrappers are deleted and all 13 call sites (5 files: `internal/integration_suite_test.go`, `internal/postgres_persistent_repository_integration_test.go`, `internal/postgres_persistent_repository_version_integration_test.go`, `factory/factory_test.go`, `factory/factory_schema_registry_test.go`) now call `testdb.Connect(t, ctx)` directly — one name for the operation, no indirection left. Re-verified locally: 0 connectivity skips with a live Postgres, both skip/fail arms unchanged.

**2. Parse failures fail loudly even off-CI** — Correct reading, and by design: a malformed `DATABASE_URL`/`DB_*` value is misconfiguration, not absence of infrastructure. A skip there would hide a typo indefinitely; `t.Fatalf` surfaces it on the first run.

**3. `CI` set without Postgres provisioned** — Agreed this is the deliberate conservative arm ("a skip would go dark"). Not adding a `FORMA_TESTDB_REQUIRE`-style knob now: no current CI system in this repo hits that case, and an unused escape hatch is one more thing to document and test (YAGNI). If a future job legitimately runs these packages without a database, the knob (or scoping the job's `go test` target) is the obvious fix at that point.

**4. Skip message wording** — Fixed in 0bfede7: now `skipping: integration Postgres unreachable (%v)`, accurate for a remote `DB_HOST` that is down; the wrapped pgx error names the target host.

**5. Env surface narrower than production bootstrap** — Deliberate: the suites create their own temp tables in `public` and run with default pool sizing, so `DB_SCHEMA` and the pool-tuning vars have nothing to configure here. Adding them unread would imply support that doesn't exist.

**6/7. Coverage side effect, shared-database hygiene** — Recorded; nothing to change. The nanotime-suffix + `t.Cleanup` drop pattern is untouched and remains safe for CI's single Test job.

No follow-up issues opened: the only follow-up candidate (item 1) is resolved in-PR, and item 3 is explicitly not-needed-now per the review itself.

iceboundrock:
CI note: the Test job's first attempt on 0bfede7 failed on three DuckDB init tests in `internal/cdc`/`internal/federated` — packages this PR does not touch. Root cause is a pre-existing environmental flake (extension INSTALL exceeding the shared 5s conn-init deadline; fail-open init then skips `SET s3_region`, and the tests' Nop loggers hide the warning). A code-identical rerun is green across all 5 checks, with the previously-failing test at 0.15s vs 11.49s. Filed #487 with the full mechanism and evidence; integration coverage re-verified in the rerun log (1 skip total — the pre-existing federated placeholder).

internal/testdb/testdb.go
```diff
@@ -0,0 +1,82 @@
...
+	if err == nil {
+		if err = pool.Ping(ctx); err != nil {
+			pool.Close()
+		}
+	}
+	if err != nil {
+		if FailOnUnreachable() {
+			t.Fatalf("integration database unreachable in CI — the workflow provisions Postgres on purpose, so this is a failure, not a skip (#385): %v", err)
+		}
+		t.Skipf("skipping: integration Postgres unreachable (%v)", err)
```
iceboundrock, line 135: **Finding [P2]** `Connect` treats every `pool.Ping` error as an unreachable database and skips locally (`internal/testdb/testdb.go:73-77`). A reachable PostgreSQL that rejects credentials, has no requested database, or fails TLS/authentication therefore makes all integration tests exit green with `SKIP`. That is narrower than the documented local policy (skip when PostgreSQL is absent) and regresses the factory venue, where a configured `DATABASE_URL` previously allowed query/auth failures to surface. Please distinguish network-level absence from server/configuration errors for the local skip path; CI should continue failing closed for every connection failure.

iceboundrock:
## Spec

Reviewed against issue #385 and the PR description. The implementation satisfies the CI-facing acceptance criteria: both Postgres test venues use `internal/testdb`, `DATABASE_URL` takes precedence, `DB_*` values match the workflow, unreachable CI databases fail, and no workflow change was introduced. The PR CI log also shows the previously dark integration tests running.

The local policy is only partially aligned: the code skips every `Ping` error, not only absence of the local database. This is recorded as the P2 finding in the inline comment.

iceboundrock:
## Assessment

The change is appropriately scoped and the shared `testdb` package removes duplicated connection policy from `internal` and `factory`. The CI fail-closed behavior is effective and verified.

Assessment: request one follow-up before merge for the local error classification described in the P2 finding. No CI-blocking regression was found.

iceboundrock:
## Standards

The touched files comply with the repository limits: `internal/testdb/testdb.go` is 82 lines, `testdb_test.go` is 71 lines, and functions remain below the 100-line limit. The change is test-only, errors include actionable context, and the DSN construction uses URL escaping for credentials. Local verification found `gofmt` and `go vet` clean; the PR checks report lint and test jobs green.

The only standards concern is behavioral: a local authentication/database/TLS error should not be silently converted into a successful skipped test.

iceboundrock:
## Observations

- `FailOnUnreachable` intentionally treats any non-empty `CI` value as CI, so `CI=false` still fails closed; this matches the issue decision and current GitHub Actions behavior.
- The environment-mutating tests do not use `t.Parallel`, so they are compatible with Go's `t.Setenv` restriction.
- `DB_SSL_MODE` is inserted into `RawQuery`; this is fine for the controlled values used by the repository, but it assumes that variable remains an enum-like value.
- Temp-table naming and cleanup are unchanged by this PR.

iceboundrock:
## Summary

Independent review of PR #486 is complete. The central CI integration fix is sound and the current PR checks are green. One P2 finding remains: local `Connect` skips all ping failures, including reachable-database authentication, database-selection, and TLS/configuration errors. Please narrow the local skip condition so those failures cannot make integration coverage silently disappear. No other blocking findings were identified.

iceboundrock:
## Verification

Verified at head `0bfede7f9fd9bde9aea0b8d20d54446b9958e4e2`.

- `go test ./internal/testdb -v`: PASS.
- `go test -race ./internal/testdb -count=20`: PASS.
- `make fmt-check`: PASS.
- `go vet ./internal/testdb ./internal ./factory`: PASS.
- `CI=true` with `DB_PORT=59999`: both `./internal` and `./factory` FAIL with the fail-closed message.
- `CI=` with the closed port: integration tests SKIP as designed; no live local PostgreSQL was available, so live local integration execution was not independently reproduced.
- PR checks: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke all PASS.

internal/testdb/testdb.go
iceboundrock: Fixed in 62f537e. `Connect` now classifies the ping error via a new `unreachable()` helper: only network-level absence (ECONNREFUSED/EHOSTUNREACH/ENETUNREACH, DNS not-found, connect timeout) may skip locally; anything the server answered — auth failure (28P01), missing database (3D000), TLS/protocol errors, junk from a non-Postgres listener — fails loudly off-CI with a message naming it as misconfiguration. CI remains fail-closed on every failure (its message reworded to 'connection failed' since 'unreachable' is no longer always accurate). Classification is pinned by a table test (`TestUnreachableClassifiesNetworkAbsenceOnly`), and both server-rejection cases were verified against a live Postgres 16: wrong password and nonexistent database now FAIL locally with the SQLSTATE in the output; closed port still SKIPs.

iceboundrock:
Response to the second review round (changes in 62f537e):

**P2 (local skip swallows server-side failures)** — Fixed; details in the inline thread. Summary: `unreachable()` now splits network-level absence (skip locally) from server-answered rejection (fail locally: auth, missing database, TLS/protocol); CI stays fail-closed on everything. Verified against a live Postgres 16: `DB_PASSWORD=wrongpass` and `DB_NAME=nonexistent_db` both FAIL locally with the SQLSTATE surfaced; closed port still SKIPs; `CI=true` fatals in all cases.

**Observations:**
- `CI=false` failing closed — confirmed intentional: any non-empty `CI` is treated as CI, matching the issue decision and GitHub Actions (which always sets `true`). Not special-casing falsy values keeps the predicate trivial.
- `t.Setenv` without `t.Parallel` — yes, and the new tests keep that property.
- `DB_SSL_MODE` into `RawQuery` — hardened in the same commit: the value now goes through `url.QueryEscape` (pinned by `TestResolveDSNEscapesSSLMode`), so the enum-shape assumption is gone.
- Temp-table naming/cleanup — unchanged, agreed.

iceboundrock:
## Review

**Head reviewed:** `62f537e7dc5177255b4d880f467934e482916125`

### Findings

1. **P2: mixed multi-address errors can still be skipped locally** (`internal/testdb/testdb.go:61-75, 103-106`)

   pgx v5 can aggregate per-address connection errors with `errors.Join` inside `pgconn.ConnectError`. If one resolved address returns `ECONNREFUSED` and a later address reaches PostgreSQL but returns an authentication or database-selection `*pgconn.PgError`, the current `errors.As` checks find the network error and classify the whole attempt as `unreachable`. With `CI` unset, the test then calls `t.Skipf` and hides the actionable server-side configuration error. Please classify a joined attempt as skippable only when all underlying attempts are network-level absence, and add a regression test using a joined `ECONNREFUSED` plus `*pgconn.PgError`.

2. **P2: handshake/startup timeouts are treated as database absence** (`internal/testdb/testdb.go:71-75, 106`)

   `unreachable` accepts any `net.Error.Timeout()` and `context.DeadlineExceeded`. pgx wraps TLS-handshake and PostgreSQL startup/receive timeouts this way, so a TCP listener that accepts connections but is broken or misconfigured can make local integration tests skip instead of failing loudly. An already-expired caller context can also take this path. Please restrict the local skip decision to errors known to originate from dialing/network absence, and cover handshake/startup timeout behavior (or the resulting pgx error chain) with a regression test.

No P0/P1 findings. The two findings are local-only; `CI` remains fail-closed for every connection error.

### Spec

The core #385 requirements are met: both Postgres test venues use the shared `internal/testdb` policy, `DATABASE_URL` takes precedence, `DB_*` matches the CI service, CI connection failures are fatal, local absent infrastructure can skip, and `.github/workflows/ci.yml` is unchanged.

The latest local-classification refinement intends to skip only network absence and fail on server-side misconfiguration, but the two cases above mean that intent is not fully enforced.

### Standards

The change remains test-only and appropriately scoped. `internal/testdb/testdb.go` (111 lines) and `testdb_test.go` (116 lines) stay below the repository file-size limit, and functions remain below the 100-line limit. Error messages include actionable context, DSN credentials/query values are escaped, and the touched code is formatted and vet-clean.

The main standards/testability gap is that the new tests exercise `unreachable` with isolated synthetic errors, but do not exercise the actual `pgconn.ConnectError`/`errors.Join` chain or `Connect`'s local skip/fatal behavior.

### Assessment

The central fix is sound and the CI fail-closed behavior is effective. I recommend addressing the two P2 findings before merge so local integration coverage cannot silently disappear behind mixed-address or protocol/handshake failures.

### Observations

- `DATABASE_URL` precedence preserves the factory suite's existing contract.
- `url.UserPassword` and query escaping cover reserved password and SSL-mode characters.
- Treating any non-empty `CI` as CI is deliberate and matches the issue decision.
- Temporary table naming and cleanup are unchanged.
- No unrelated production code paths or workflow wiring are affected.

### Verification

- `go test ./internal/testdb -v`: PASS.
- `go vet ./internal/testdb ./internal ./factory`: PASS.
- `make fmt-check`: PASS.
- `git diff --check`: PASS.
- With `CI=` and `DB_PORT=59999`, both `internal` and `factory` suites skip only the unreachable-network cases and exit successfully.
- With `CI=true` and `DB_PORT=59999`, both suites fail with the intended fail-closed message.
- PR CI run [32886255762](https://github.com/Lychee-Technology/forma/actions/runs/32886255762) is fully green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke all pass.
- I did not independently reproduce a live local PostgreSQL run in this review; the PR Test job provides the live CI service verification.

### Summary

The PR fixes the original dark CI integration venues and all required checks pass. Two P2 findings remain in the local-only error classifier: joined mixed-address errors and handshake/startup timeouts can still be skipped. No CI-blocking regression was found, but these cases should be covered and narrowed before merge.

iceboundrock:
Response to review round 3 (both P2 findings fixed in 4deb68b):

**Finding 1 (joined mixed multi-address errors)** — Confirmed against pgx v5.8.0 source: `pgconn.Connect` wraps `errors.Join(allErrors...)` inside `ConnectError` (pgconn.go:155,161), and `errors.Is` has any-branch semantics over joins. `unreachable()` now walks the error tree manually: a node with `Unwrap() []error` requires **every** branch to classify as absence; leaf checks use direct equality instead of `errors.Is` so they cannot descend into joined subtrees. Regression test: `joined mixed refusal and auth failure` (dial-refused `net.OpError` joined with a 28P01 `pgconn.PgError`) → not skippable; `joined all-absence` → skippable.

**Finding 2 (handshake/startup timeouts as absence)** — The blanket `net.Error.Timeout()` / `context.DeadlineExceeded` acceptance is gone. Absence is now restricted to failures originating at dial time: `net.OpError` with `Op == "dial"` (covers refused/unreachable/DNS/dial-timeout), `*net.DNSError`, and bare dial errnos. A read/write `OpError` (server accepted, then stalled or broke) and a bare expired context now fail loudly locally.

**Standards gap (no coverage of the real pgx chain / Connect's local behavior)** — Added behavioral tests that drive `Connect` itself against real listeners via an embedded-`testing.TB` recorder: a silent listener (accepts, never answers the startup exchange) produces the genuine pgx `failed to receive message: timeout` chain and now FAILS locally (this exact test was red before the fix — the old classifier skipped it); a closed port produces the genuine per-address dial-refused join and still SKIPs.

Re-verified against live Postgres 16: wrong password → local FAIL with SQLSTATE 28P01; closed port → SKIP; `CI=true` → fatal in all cases. `internal/testdb` passes under `-race`; fmt/lint clean. CI is re-running on 4deb68b.

iceboundrock:
## Review — Round 4

**Head reviewed:** `4deb68b839ac92e56c718196f297397e97ca0d88` (4deb68b), which resolves both P2 findings from round 3.

### Summary

The PR fixes #385 at the root cause: both Postgres-backed test venues (`internal` integration suite, `factory`) previously had connection stories that CI's provisioned service never satisfied, so every Postgres-backed test silently skipped while reporting green. The new `internal/testdb` package is now the single owner of DSN resolution (`DATABASE_URL` wins; else `DB_*` via `bootstrap.Env` with `local_server.sh` defaults) and skip policy (CI fail-closed on any connection error; local skip restricted to network-level absence). Both round-3 P2s are correctly fixed:

- **Joined multi-address errors**: `unreachable()` walks the error tree manually; a joined node classifies as absence only when *every* branch does, and leaf checks use direct equality so they cannot descend into joined subtrees with any-branch semantics. Verified against pgx v5.8.0 source — the `errors.Join` inside `pgconn.ConnectError` claim is accurate (pgconn.go:155 DNS path, :161 connect path).
- **Handshake/startup timeouts**: absence is now strictly dial-time (`net.OpError{Op:"dial"}`, `*net.DNSError`, bare dial errnos); read/write ops and bare expired contexts fail loudly. The silent-listener behavioral test drives the genuine pgx chain end to end and was red before the fix.

### Findings

No P0/P1/P2. Three cosmetic nits, none blocking:

1. **P3: `FailOnUnreachable` name/doc drift** (`internal/testdb/testdb.go:52-58`) — after round 3, CI fatals on *every* connection error (auth, protocol, timeout), not just unreachability. `Connect`'s doc describes the three-way split correctly, but this helper's name and doc still describe only the unreachable case. A follow-up rename (e.g. `InCI`) would align them; no change required now.
2. **P3: dial-time classification is errno-blind** (`testdb.go:89`) — `*net.OpError` returns `Op == "dial"` without inspecting the inner errno, so any dial-time failure (e.g. `ECONNRESET`, `EACCES`) skips locally. Since nothing was ever accepted at dial time, treating all dial failures as absence is defensible breadth; recording as deliberate.
3. **Nit: fatal-message wording for non-dial non-server errors** (`testdb.go:131`) — "answered but rejected the connection" is slightly imprecise for a bare expired caller context (nothing answered). The behavior (fail loudly) is right; wording only.

### Spec (#385)

Both acceptance criteria are met and independently re-verified below. All decided constraints hold: `DATABASE_URL` precedence preserved, `DB_*` names match `bootstrap.DatabaseConfigFromEnv`, defaults pin `scripts/local_server.sh`, fail-closed arms use the standard `CI` variable with zero workflow wiring, `.github/workflows/ci.yml` is untouched (empty diff vs `main`), and the local skip policy now enforces the narrowed intent from review rounds 2–3 (only network absence may skip).

### Standards

`internal/testdb/testdb.go` (136 lines) and `testdb_test.go` (221 lines) are well under the 500-line limit; all functions are far under 100 lines. Failure messages name what to check ("fix DATABASE_URL / DB_* env"); comments explain why and cite issue/review refs per house style. Env-mutating tests avoid `t.Parallel` (compatible with `t.Setenv`). The embedded-`testing.TB` recorder with `runtime.Goexit` faithfully mimics `Fatalf`/`Skipf` semantics and its channel handoff is race-free (verified under `-race -count=2`). `gofmt`, `go vet`, and `make lint` (pinned golangci-lint v1.64.8, root + infra modules) are clean at this head. Change remains test-only.

### Assessment

Approve. The central fix is sound, both dark venues are live in CI, the fail-closed policy works in both directions, and all findings from rounds 1–3 are resolved or deliberately recorded. Remaining items are cosmetic (findings 1–3 above); none warrant holding the merge. Per house rules this PR must not be auto-merged.

### Observations

- The manual join-walk instead of `errors.Is`/`As` is justified and documented in-code; stdlib any-branch semantics over `errors.Join` would have reintroduced exactly the round-3 finding.
- `bootstrap.Env` treats empty-string vars as unset, so the tests' `clearDBEnv` (which sets vars to "") deterministically yields defaults even under ambient CI/shell env.
- `url.QueryEscape` on `DB_SSL_MODE` produces `+` for spaces; pgx decodes query values with `ParseQuery` semantics, so the round trip is consistent (pinned by `TestResolveDSNEscapesSSLMode`).
- The silent-listener behavioral test costs ~2 s (bounded by the 2 s `ConnectTimeout`); acceptable for the coverage it buys.
- Parameter order `(t testing.TB, ctx)` deviates from ctx-first convention but preserves the pre-existing call-site shape across 13 sites; fine for internal test support.
- `internal/testdb` imports `testing` from a non-test file, which pulls it into `go build ./...`; standard pattern for shared test-support packages, no action needed.

### Verification (independent, at `4deb68b`, Postgres 16 via rootless Podman)

| Check | Result |
|---|---|
| `go vet ./internal/testdb ./internal ./factory` | clean |
| `go test -race -count=2 ./internal/testdb -v` | PASS (all cases incl. joined-mixed, silent-listener, closed-port) |
| Defaults (`CI= DATABASE_URL=`): `./internal -run Integration` | 9 PASS, 0 SKIP |
| CI's exact env shape (`CI=true test:test@…/forma_test`) | 9 PASS |
| Local wrong password (`DB_PASSWORD=wrongpass`) | FAIL with "integration Postgres answered but rejected…" |
| Local closed port, `CI` unset | 9 SKIP, exit 0 |
| `CI=true` + closed port | FAIL (`./internal` and `./factory`) with fail-closed message |
| `./factory` defaults only | 51 PASS |
| `make fmt-check` / `make lint` | clean / clean (exit 0) |
| pgx v5.8.0 `pgconn.go:155,161` | confirms `errors.Join` inside `ConnectError` |
| PR checks on `4deb68b` (run 32906185406) | Lint ✅ Test ✅ E2E Smoke ✅ Benchmark Smoke ✅ k6 Smoke ✅ |

Acceptance criterion "tests report as run, not skipped, in the CI run log" is satisfied by run 32906185406's Test job on this head.

Refs #385, #355, #487.

iceboundrock:
Response to review round 4 (cosmetics applied in dff90df — all three items were one-file trivia, so they went into the PR now rather than a follow-up):

1. **`FailOnUnreachable` name/doc drift** — Renamed to `InCI` as suggested, doc rewritten to state the post-round-3 semantics: under CI *every* connection failure is fatal, not only unreachability. Test renamed to `TestInCIFollowsCIVar`.
2. **Errno-blind dial classification** — Confirmed deliberate: at dial time nothing was accepted, so any dial-stage failure (`ECONNRESET`, `EACCES` included) reads as absence. Now recorded in the `net.OpError` case comment so the breadth is visibly a decision, not an oversight.
3. **Fatal wording for non-dial non-server errors** — Reworded to "connection failed for a reason other than network absence — misconfiguration, not a missing Postgres", which is accurate for server rejections, protocol junk, handshake stalls, and expired caller contexts alike.

No behavior change beyond the message string; `internal/testdb` re-verified under `-race`, both consumer packages compile and pass, fmt/lint clean. No follow-up issues needed from this round.

iceboundrock:
## Review: #486 — `testdb` env-driven DSN + CI fail-closed policy (#385)

### Summary

Single-owner test-support package (`internal/testdb`) now owns both decisions #385 called out: DSN resolution (`DATABASE_URL` wins; else `DB_*` via `bootstrap.Env` with exact `scripts/local_server.sh` defaults) and the connection-failure policy (fail-closed under `CI`, three-way classification off-CI: network absence skips, anything the server answered fails loudly). Both dark venues switched over with call sites untouched; `.github/workflows/ci.yml` has an empty diff (verified: `git diff main...HEAD -- .github/workflows/ci.yml | wc -l` = 0). Four review rounds already landed (wrapper inlining, P2 misconfiguration split, per-branch join walk + dial-time-only absence, cosmetics). **No blocking findings. Recommend approval.**

### Spec compliance (issue #385)

| Requirement | Status |
|---|---|
| `internal` venue reads `DB_*` env instead of hardcoded `postgres:postgres@localhost/forma` | ✅ defaults pinned by `TestResolveDSNDefaultsMatchLocalServer`; verified byte-exact |
| Second venue `factory` (`DATABASE_URL` gate) also live; historical contract preserved | ✅ `DATABASE_URL` precedence re-verified below |
| Fail-closed when DB provisioned on purpose, via standard `CI` var (issue comment 3 pattern), no workflow wiring | ✅ `InCI()` = `os.Getenv("CI") != ""`; ci.yml untouched |
| AC1: suites execute in CI, reported run-not-skipped | ✅ confirmed directly in Test job log of run [32914342084](https://github.com/Lychee-Technology/forma/actions/runs/32914342084) on head `dff90df`: all 19 integration/factory Postgres tests `--- PASS`; the job's only `--- SKIP` is the pre-existing unrelated placeholder `TestDBFederatedQueryEngine_FallsBackToPostgresWhenDuckDBCircuitBreakerOpen` |
| AC2: unreachable DB fails in CI | ✅ implemented for every failure class in CI; verified below (CI cannot demonstrate its own negative) |

The review-round hardening goes beyond the letter of the issue but stays within its spirit (skip policy decided deliberately, not accidentally). I validated the manual join-walk against the actual pgx v5.8.0 source: `pgconn.ConnectError.err` is built from `errors.Join(allErrors...)` per address family (pgconn/pgconn.go:155,161), each branch a `perDialConnectError` wrapping the raw dial/protocol error (pgconn/errors.go:82–94) — so `errors.Is`'s any-branch semantics would indeed have misclassified mixed outcomes; the every-branch requirement is correct.

### Findings

**Blocking:** none.

**Non-blocking observations** (no action required; record-only):

1. **Errno-blind dial breadth** — `unreachable()` treats *every* dial-time failure as absence, including purely local causes like EMFILE/ENFILE exhaustion or socket permission errors. Documented in-code as intended breadth (round 4). Worst case is a quiet skip on a developer machine, never a hidden CI failure — acceptable.
2. **`runConnect` success-path pool leak** — `Connect` succeeding through the `outcomeTB` harness registers cleanup on nothing; currently unreachable by tests (both harness cases end in Goexit). Worth remembering if a future case exercises the success path.
3. **`outcomeTB.Fatalf/Skipf` use `runtime.Goexit()`** — correct inside the dedicated goroutine `runConnect` spawns; safe only because callers never invoke them on the main test goroutine. Fine as-is.
4. **`clearDBEnv` relies on `bootstrap.Env` treating set-but-empty as unset** — true today (internal/bootstrap/config.go:27-32); noted so a future change to `Env` doesn't silently break these tests.
5. **Silent-listener behavioral test pins a pgx error-shape assumption** (read-op timeout after accepted TCP connect). It drives the genuine pgx chain against a real listener, so a pgx change flips the test loudly rather than silently altering skip semantics — the right way to pin it.
6. Minor style nit: `defer conn.Close()` inside the accept loop of `TestConnectLocallyFailsOnSilentListener` accumulates defers until goroutine exit — harmless at one connection.

### Standards

- File sizes 140 / 221 lines, longest function ≈40 lines — within ≤500/≤100 limits.
- Fatal/skip messages name what was expected, what failed, and the remediation surface (`DATABASE_URL / DB_* env`) — matches the repo's operator-detail convention.
- No `t.Parallel` anywhere near `t.Setenv`; env manipulation is fully restored via `t.Setenv`.
- No production code touched; no import-cycle risk (`testdb` → `bootstrap` → root `forma` only).
- Decision history (#385, review rounds) referenced in code comments where load-bearing — consistent with how this repo records rulings.

### Verification (independently reproduced, Postgres 16 via rootless Podman, worktree at head `dff90df`)

| Case | Result |
|---|---|
| `go test ./internal/testdb` (unit, incl. 2.00s real-chain silent-listener case) | PASS (all 22 tests/subtests) |
| Zero-config defaults (`CI= DATABASE_URL=`): `./internal -run Integration` | 9 PASS, no SKIP |
| Same, `./factory` | ok |
| CI's exact env shape (`CI=true DB_USER=test DB_PASSWORD=test DB_NAME=forma_test` vs matching role/db) | 9 PASS |
| `CI=true DB_PORT=59999` (unreachable), both venues | FAIL with "integration database connection failed in CI …" (loud, fast) |
| `CI= DB_PORT=59999` (local, unreachable) | 9 SKIP, exit 0 (quiet dev-machine path preserved) |
| `CI= DB_PASSWORD=wrongpw` (reachable server rejects) | **FAIL, not skip** — "misconfiguration, not a missing Postgres" (round-2 P2 behavior confirmed end-to-end through the real pgx auth-failure chain) |
| `DATABASE_URL` precedence over `DB_*` (with decoy `DB_PORT=59999`) | factory ok |
| `make fmt-check`; `go vet ./internal/testdb ./factory ./internal` | clean |
| CI on head `dff90df` (run 32914342084) | Lint / Test / E2E Smoke / k6 Smoke / Benchmark Smoke all green |

### Assessment

Scope-tight, well-tested, and the subtle parts (joined-error semantics, dial-vs-post-dial classification) are pinned by both synthetic table cases and real-listener behavioral tests. Both acceptance criteria hold — one demonstrated by CI itself, one independently reproduced. Approve; do **not** auto-merge (repo rule). After merge: close #385 with the AC1 log evidence already quoted in the issue thread.

PR #488, opened 2026-08-26
fix(schemavalidate): own the invalid-number-literal published message (#453)

### LINKED ISSUE #453: test(schemavalidate): TestValidateFallsBackOnInvalidNumberLiteral asserts Go 1.26 stdlib error text; breaks under Go 1.27
Split out of #448 (which pins the gate toolchain to go.mod's 1.26.x and defers this).

`internal/schemavalidate/nonfinite_paths_test.go` (`TestValidateFallsBackOnInvalidNumberLiteral`) asserts the published message contains `invalid number literal` — Go 1.26's `encoding/json` wording for a garbage `json.Number`. Go 1.27 changes both the wording and the calling convention named in it:

```
payload cannot be encoded as JSON: json: error calling MarshalJSONTo for type *json.Number: cannot parse "abc" as JSON number: invalid syntax
```

Two aspects to address when the repo moves to Go 1.27 (which also requires bumping the golangci-lint v1.64.8 pin — see #448):

1. The test assertion is fragile against stdlib text. Deterministic FAIL under 1.27 (`-count=20` probe in #448's comments).
2. Product-facing: the *published* fallback message forwards library text to API callers (`nonfinite_paths.go` deliberately publishes "the library's text" for the garbage-literal case), so a toolchain bump silently changes user-visible 4xx bodies. Decide whether to keep forwarding or to own the message before bumping.

Blocked-on/paired-with: any future Go 1.27 migration issue.

iceboundrock:
Closes #453.

## What changed

The published 4xx body for a garbage `json.Number` literal (e.g. `json.Number("abc")`) is now an owned message instead of forwarded `encoding/json` text:

```
payload cannot be encoded as JSON: attribute "score" holds invalid number literal "abc"; a valid JSON number is required
```

The raw stdlib error stays reachable for operators via `forma.WithOperatorDetail` — it reaches `Error()` and the log, never the published body. `TestValidateFallsBackOnInvalidNumberLiteral` (which asserted Go 1.26's stdlib wording and FAILed deterministically under Go 1.27) is replaced by `TestValidateNamesInvalidNumberLiteral`, which asserts only owned text.

## Decision table (`marshalRefusalError`)

| Walk finding | Published body | Operator detail |
|---|---|---|
| Non-finite float(s) | owned: `attribute "x" holds non-finite number NaN; a finite value is required` (unchanged) | — |
| Invalid `json.Number` literal(s) | owned: `attribute "x" holds invalid number literal "abc"; a valid JSON number is required` (**new**, #453) | raw stdlib error |
| Neither (channel/func, failing Marshaler, cycle, depth-cap abandon, root-level fault) | library text unchanged — the only description that exists | — |

When both kinds are present, non-finite wins — fixed priority keeps the message deterministic. Both owned messages name the alphabetically first offender plus an `(and N more)` count, matching the existing precedent.

## Why probe `json.Marshal` as the gate

The walker classifies a `json.Number` as an invalid literal only when `json.Marshal` itself refuses it, rather than re-deriving JSON number grammar. This keeps the classification exactly aligned with the refusal it explains, on any toolchain. Probing caught a real edge during TDD: `json.Number("")` marshals as `0` in encoding/json — never a refusal — so a grammar re-implementation would have named a literal that can't be the fault. `1e400` (valid JSON, `ParseFloat` range error) is likewise excluded by construction; `"1_0"` (accepted by `ParseFloat`, rejected by JSON grammar) is correctly flagged.

## Effect on the Go 1.27 migration (#453 / #448)

- The deterministic FAIL mode of `TestValidateFallsBackOnInvalidNumberLiteral` under Go 1.27 disappears: no test in this package asserts stdlib wording for the garbage-literal case any more.
- A toolchain bump can no longer silently change this user-visible 4xx body.
- The future Go 1.27 migration's critical path shrinks to the golangci-lint v1.64.8 pin (#448). Out of scope, deliberately: the remaining walk-empty fallbacks (unsupported type, cycle, depth-cap) still publish library text — there no attribute/literal exists to name and the library text is the only description of the fault. Their tests (`unsupported type`, `cycle`, `unsupported value: NaN` assertions) remain toolchain-sensitive; if Go 1.27's jsonv2 rewords those too, that belongs to the migration issue.

## Verification

- `go test ./internal/schemavalidate` — all green (new unit tests for classification, walk-semantics inheritance incl. depth-cap abort discarding invalid-literal findings, and the three-branch seam behavior).
- `make fmt-check`, `make lint` — green.
- `./scripts/test_with_container_runtime.sh` (full `make test`) — green.

https://claude.ai/code/session_01JrozwMGpNtcCzHtV6pPjfY

### DIFF
--- a/internal/schemavalidate/nonfinite_paths.go
+++ b/internal/schemavalidate/nonfinite_paths.go
@@ -78,7 +78,7 @@ func moreSuffix(n int) string {
 	return ""
 }
 
-// maxNonFiniteWalkDepth bounds the recursion below. Nothing upstream enforces a
+// maxMarshalRefusalWalkDepth bounds the recursion below. Nothing upstream enforces a
 // shallower limit: encoding/json's decoder refuses nesting past 10000, ten times
 // this cap, so a decoded document can sit well beyond it.
 //
@@ -89,10 +89,11 @@ func moreSuffix(n int) string {
 // are pinned.
 //
 // What limits the cost of that degradation is the payload's provenance, not its
-// depth: JSON has no syntax for a non-finite, so an HTTP request can never
-// produce the refusal this walk exists to explain. Only a Go embedder can, which
-// makes a message degraded by this cap an embedder-only outcome.
-const maxNonFiniteWalkDepth = 1000
+// depth: JSON has no syntax for a non-finite, and a decoded json.Number is
+// grammar-valid by construction, so an HTTP request can never produce either
+// refusal this walk exists to explain. Only a Go embedder can, which makes a
+// message degraded by this cap an embedder-only outcome.
+const maxMarshalRefusalWalkDepth = 1000
 
 // marshalRefusalPaths walks a document and returns every finding the walk can
 // name — non-finite floats and invalid json.Number literals (#453) — each
@@ -104,7 +105,7 @@ const maxNonFiniteWalkDepth = 1000
 // is what makes a cyclic payload safe here — json.Marshal detects a cycle and
 // refuses, so this walk runs precisely when a cycle is possible, and following
 // one has no natural end, so the cap is the end. A merely deep acyclic payload
-// hits the same cap and gets the same treatment (see maxNonFiniteWalkDepth).
+// hits the same cap and gets the same treatment (see maxMarshalRefusalWalkDepth).
 //
 // Abandoning rather than truncating is deliberate: a partial walk cannot tell
 // "nothing wrong in this payload" from "stopped looking", and only the first
@@ -126,33 +127,34 @@ const maxNonFiniteWalkDepth = 1000
 // Structs, custom Marshalers and channels land here too. A fault at the
 // document root is likewise left to the library: there is no attribute to name.
 func marshalRefusalPaths(doc any) ([]nonFinitePath, []invalidNumberLiteral) {
-	var walker nonFiniteWalker
+	var walker marshalRefusalWalker
 	walker.walk("", doc, 0)
 	if walker.aborted {
 		return nil, nil
 	}
-	slices.SortFunc(walker.found, func(a, b nonFinitePath) int {
+	slices.SortFunc(walker.nonFinite, func(a, b nonFinitePath) int {
 		return strings.Compare(a.path, b.path)
 	})
 	slices.SortFunc(walker.invalid, func(a, b invalidNumberLiteral) int {
 		return strings.Compare(a.path, b.path)
 	})
-	return walker.found, walker.invalid
+	return walker.nonFinite, walker.invalid
 }
 
-// nonFiniteWalker carries the walk's accumulated findings and its abort flag.
-// Once aborted, every pending frame returns without doing further work.
-type nonFiniteWalker struct {
-	found   []nonFinitePath
-	invalid []invalidNumberLiteral
-	aborted bool
+// marshalRefusalWalker carries the walk's accumulated findings — one slice per
+// finding kind — and its abort flag. Once aborted, every pending frame returns
+// without doing further work.
+type marshalRefusalWalker struct {
+	nonFinite []nonFinitePath
+	invalid   []invalidNumberLiteral
+	aborted   bool
 }
 
-func (w *nonFiniteWalker) walk(path string, node any, depth int) {
+func (w *marshalRefusalWalker) walk(path string, node any, depth int) {
 	if w.aborted {
 		return
 	}
-	if depth > maxNonFiniteWalkDepth {
+	if depth > maxMarshalRefusalWalkDepth {
 		w.aborted = true
 		return
 	}
@@ -202,18 +204,18 @@ func (w *nonFiniteWalker) walk(path string, node any, depth int) {
 	}
 }
 
-func (w *nonFiniteWalker) appendIfNonFinite(path string, value float64) {
+func (w *marshalRefusalWalker) appendIfNonFinite(path string, value float64) {
 	if path == "" {
 		return
 	}
 	if math.IsNaN(value) || math.IsInf(value, 0) {
-		w.found = append(w.found, nonFinitePath{path: path, value: value})
+		w.nonFinite = append(w.nonFinite, nonFinitePath{path: path, value: value})
 	}
 }
 
 // appendInvalidLiteral mirrors appendIfNonFinite's root rule: a fault at the
 // document root has no attribute to name and falls back to the library text.
-func (w *nonFiniteWalker) appendInvalidLiteral(path, literal string) {
+func (w *marshalRefusalWalker) appendInvalidLiteral(path, literal string) {
 	if path == "" {
 		return
 	}
--- a/internal/schemavalidate/nonfinite_paths.go
+++ b/internal/schemavalidate/nonfinite_paths.go
@@ -34,28 +34,48 @@ type invalidNumberLiteral struct {
 // in their payload is barely actionable. So the payload is walked here to
 // derive the path. Only the first offender is named — alphabetically first, so
 // the same payload always produces the same message — with a count of the
-// rest, matching the precedent in transform's missing-required error.
+// rest, matching the precedent in transform's missing-required error. When
+// both finding kinds are present the non-finite message wins: either is
+// truthful (each would refuse on its own), and a fixed priority keeps the
+// published message deterministic.
+//
+// The invalid-literal message is owned rather than forwarded (#453):
+// forwarding stdlib text made the published 4xx body a function of the build
+// toolchain — Go 1.27 rewords it — which is a silent API-contract change
+// triggered by nothing but a go.mod bump. The raw library error rides along
+// as operator detail, so the log keeps it while the body never sees it.
 //
 // If the walk comes back empty, the library's text is published unchanged: it
 // is then the only description of the fault that exists. That covers a failing
 // json.Marshaler and an unsupported type such as a channel or func — and a
 // cycle, which reaches this branch because the walk is depth-capped and gives
-// up on one rather than following it (see nonFinitePaths). There the library's
-// text begins "json: unsupported value: encountered a cycle" and goes on to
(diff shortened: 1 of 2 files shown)


iceboundrock:
## Review: fix(schemavalidate) — own the invalid-number-literal published message (#453)

Reviewed at `50e95ad` (PR head, both commits read in full; verified locally).

### Summary

Closes #453 as scoped. A garbage `json.Number` literal now publishes an owned 4xx body naming the attribute, the literal, and the expected state; the raw `encoding/json` error rides along via `forma.WithOperatorDetail` so logs keep it while the body never sees it. The toolchain-fragile test (`TestValidateFallsBackOnInvalidNumberLiteral`, deterministic FAIL under Go 1.27) is replaced by assertions on owned text only. The two-commit split is clean: commit 1 is a pure refactor of the walk to collect both finding kinds, commit 2 owns the message — easy to bisect and review independently.

### Spec / design assessment

- **Probe-as-gate is the right classifier.** Classifying a literal as invalid exactly when `json.Marshal` itself refuses it (rather than re-implementing JSON number grammar) means the classification and the refusal it explains come from the same stdlib in the same process — they cannot drift, on any toolchain. This also self-resolves the edge cases: `json.Number("")` marshals as `0` (never a refusal, so never named), `1e400` is valid JSON grammar (excluded), while `"1_0"` and hex-float literals that `ParseFloat` accepts but JSON grammar rejects are correctly flagged.
- **Fixed priority (non-finite > invalid literal)** keeps the published body deterministic when both kinds are present; either message would be truthful since each fault refuses marshal on its own. Documented in the function comment and pinned by a seam test.
- **Walk-contract inheritance is pinned**, not assumed: the new finding kind is covered for dotted/indexed paths, sorted output, the root-skip rule, and depth-cap abort discarding partial results — the last being the subtle case where a confident partial answer must not be published.
- **Scope boundary is correct and explicit.** The remaining walk-empty fallbacks (unsupported type, cycle, depth-cap) still forward library text because no attribute exists to name; their toolchain-sensitive tests are deliberately left to the Go 1.27 migration (#448). That matches #453's framing.

### Findings

No blocking findings. Minor observations, none requiring changes before merge:

1. **Naming drift (cosmetic):** `nonFiniteWalker` / `maxNonFiniteWalkDepth` / the `found` field now serve two finding kinds. The doc comments say so accurately; a rename (`marshalRefusalWalker` etc.) could follow later if desired.
2. **Redundant finite re-check:** gate 1 checks `IsNaN/IsInf` explicitly and `appendIfNonFinite` re-checks. Harmless defense-in-depth.
3. **Partial counting when both kinds present:** `(and N more)` counts only non-finites; concurrent garbage literals go unmentioned. Truthful-but-partial, consistent with the documented priority — acceptable.
4. **Optional test hardening:** adding hex float (`"0x1p4"`), `"+1"`, `"01"` to the classification table would pin the ParseFloat-accepts/JSON-refuses band more broadly; behavior already follows from the gates.

### Standards

- File/function size limits hold (208-line source, walk well under 100 lines); flat early-return structure.
- Error semantics per #313: owned carrier publishes verbatim; `WithOperatorDetail` keeps the stdlib error in `Error()`/logs but out of `PublicMessage()` (confirmed against `client_error.go:142`). The message names value type, offending attribute, the literal, and expected state.
- Issue-artifact rule followed: plan and execution notes posted to #453.

### Verification

- `go test ./internal/schemavalidate` (PR head, pinned `go1.26.0+auto`) — pass, including the three new/rewritten tests run verbosely.
- `gofmt -l` — clean.
- CI green: Lint (pinned golangci-lint v1.64.8), Test `-race`, E2E Smoke, k6 Smoke, Benchmark Smoke.
- Cross-repo search at PR head: no other package or test asserts the forwarded stdlib wording, so no collateral breakage from retiring the fallback text.

### Assessment

Approve. The published 4xx body becomes a function of this codebase rather than the build toolchain, the Go 1.27 migration's critical path shrinks to the lint pin (#448), and every branch of the decision table — including priority and abort semantics — has a pinning test.

iceboundrock:
## Response to review round 1

Addressed in `af14994`; per-observation disposition below.

**1. Naming drift — fixed.** `nonFiniteWalker` → `marshalRefusalWalker`, `maxNonFiniteWalkDepth` → `maxMarshalRefusalWalkDepth`, field `found` → `nonFinite` (now pairing naturally with `invalid`). The walker doc comment says "one slice per finding kind". Kind-specific names (`nonFinitePath`, `appendIfNonFinite`, the file name) stay — they describe one kind, accurately.

**2. Redundant finite re-check — no change, both checks are load-bearing.** The explicit `IsNaN/IsInf` in the `json.Number` gate routes control (a non-finite must *skip* the Marshal probe, or "NaN" would be classified as an invalid literal — Marshal refuses it too). The re-check inside `appendIfNonFinite` guards its four float-case call sites, where it is the only check. Removing either would couple the gate's routing to the append's filtering; the one duplicated evaluation on an error-only path isn't worth that coupling.

**3. Partial `(and N more)` count — no change, consequence of the documented priority.** Counting across kinds would force the message to either drop its kind-specific expected-state clause ("a finite value is required" / "a valid JSON number is required") or state one that doesn't cover all counted offenders. A count scoped to the named kind keeps the sentence exactly truthful. Both-kinds payloads are embedder-only and every offender still refuses individually, so the caller converges in at most one extra round-trip.

**4. Test hardening — done.** The classification table gains `0x1p4`, `+1`, `01` (probed against Go 1.26 stdlib first: all three refuse with `invalid number literal`), pinning the ParseFloat-accepts/JSON-refuses band alongside the existing `1_0`.

**Scope boundary → follow-up filed.** The remaining walk-empty fallback tests asserting stdlib wording (`unsupported type`, `cycle`, `unsupported value: NaN`) were "Blocked-on/paired-with: any future Go 1.27 migration issue" per #453 — but no such issue existed, and merging this PR closes #453. Opened #489 to carry it: golangci-lint pin critical path, the three re-probe-don't-own assertions, and the toolchain-floor bookkeeping.

Verification after `af14994`: `go test ./internal/schemavalidate` green (renames + 3 new table rows), `gofmt -l` clean, `make lint` green.

iceboundrock (commented):
## Review round 2: fix(schemavalidate) — own the invalid-number-literal published message (#453)

Re-reviewed fresh at `af14994` (PR head; full files read, round-1 dispositions re-derived rather than trusted, everything verified locally).

### Summary

Round-1 findings are all resolved or correctly dispositioned; no new blocking findings surfaced in the second pass. The PR closes #453 as scoped: a garbage `json.Number` literal now publishes an owned 4xx body naming attribute, literal, and expected state, while the raw `encoding/json` error rides along via `forma.WithOperatorDetail` for logs only. Follow-ups that outlive this PR are correctly split into #489 (which must exist because merging closes #453).

### Spec / design assessment

- **Probe-as-gate independently re-verified.** I probed Go 1.26.6 stdlib directly (`strconv.ParseFloat` + `json.Marshal` per literal): every row of the classification table is truthful. Beyond the table, two further edges behave consistently by construction — `"inf"` / `"-INFINITY"` parse case-insensitively to ±Inf and land in gate 1 (non-finite), and `"1e-400"` underflows to `0` *without* an `ErrRange`, marshals validly, and is correctly never named. Because findings can only arise when `json.Marshal` itself refuses, misclassification is structurally impossible — the strongest property this design has.
- **Gate-1 routing is load-bearing control flow, confirmed.** The explicit `IsNaN/IsInf` check before the probe is what keeps `"NaN"` from being classified as an invalid *literal* via the probe (Marshal refuses that spelling too). Removing it would silently change which owned message wins; keeping it costs one duplicate evaluation on an error-only path. Round-1 disposition 2 stands.
- **Walk-contract inheritance holds** for the second finding kind: dotted/indexed paths, sorted output, root-skip rule, and depth-cap abort discarding partial results are each pinned by test, including the subtle shallow-finding-plus-too-deep-branch case.
- **Determinism**: fixed priority (non-finite > invalid literal), alphabetically-first offender, sorted walks, within-kind `(and N more)` counts — all documented and pinned.

### Findings

No blocking findings. One cosmetic observation, explicitly not requesting a change:

1. The file remains `nonfinite_paths.go` and kind-specific names (`nonFinitePath`, `appendIfNonFinite`) describe one finding kind while the walker now serves two — accurate per the round-1 disposition, and a rename of the file/type would be pure churn now. If ever desired, fold into #489.

### Standards

- Source file ~230 lines, functions well under 100; flat early-return structure throughout.
- Error semantics per #313: the owned carrier publishes verbatim; `WithOperatorDetail` keeps the raw stdlib error reachable in `Error()`/logs but out of the published body (pinned by `HasOperatorDetail` + `ResolvePublicMessage` assertions). No bare-sentinel wraps introduced.
- Issue-artifact rule followed: plan and execution notes — including the mid-execution ruling on `json.Number("")` — were posted to #453.

### Verification

All at `af14994`, `GOTOOLCHAIN=go1.26.0+auto`:

- `go test ./internal/schemavalidate -count=1 -race` — pass; the new/rewritten tests also run verbosely with zero failures.
- `gofmt -l` — clean; `make lint` (pinned golangci-lint v1.64.8, root + infra modules) — clean.
- Independent stdlib probe (see above) confirming classification-table truthfulness.
- Repo-wide search at head: no stale references to the renamed identifiers (`nonFiniteWalker`, `maxNonFiniteWalkDepth`, field `found`) outside tests, and no other package asserts the retired forwarded stdlib wording — no collateral breakage.
- CI at head: Lint, Test (`-race`), E2E Smoke, k6 Smoke, Benchmark Smoke — all green.
- Caller seam unchanged (`validator.go:290` still routes `json.Marshal` refusal through `marshalRefusalError`).

### Assessment

Approve. The published 4xx body is now a function of this codebase rather than the build toolchain; the classification cannot drift from the refusal it explains; every branch of the decision table — gates, priority, root rule, abort semantics — has a pinning test; and the remaining toolchain-sensitive fallbacks are correctly handed off to #489.

PR #483, opened 2026-08-25
lint: make lint covers e2e-tagged files via --build-tags e2e (#436)

iceboundrock:
Closes #436.

`make lint` — and CI, which runs `make lint` since #454 — invoked golangci-lint without build tags, so the 96 `//go:build e2e` files were never analyzed. This adds `--build-tags e2e` to the lint recipe after clearing every violation the tag exposes, and pins the flag with a guard test.

## Corrected baseline: 83 violations, not 6

The issue's count of 6 was an artifact of golangci-lint's default output caps (`max-same-issues: 3`). With `--max-same-issues=0 --max-issues-per-linter=0` the true baseline at main (94b042e) was **83**:

- `errcheck` ×80 — all one shape: `defer h.Cleanup(ctx)` discarding the harness `Cleanup(ctx) error` return, across 17 files in `internal/e2e_harness/federated/` (incl. its `benchmark/` subpackage). Fixing only the 3 listed lines would have left CI red, revealing 3 more per run.
- `unused` ×2 — `suite_test.go`: `testCategories`, `printTestCategories`.
- staticcheck SA4010 ×1 — `consistency_test.go`: `allRecords` appended, never read.
- `funlen` ×0 — confirmed, matching the issue's measurement.

## Changes (one commit each)

1. **errcheck ×80** — new harness method `CleanupOrLog(ctx, tb testing.TB)` routes the cleanup failure through `tb.Logf` instead of dropping it; all defer sites switched (`t` in tests, `b` in the two `suite_test.go` benchmarks). `testing.TB` keeps one helper serving both packages.
2. **Dead code** — deleted `testCategories`/`printTestCategories` and the `allRecords` accumulation.
3. **Flip + guard** — lint recipe becomes `golangci-lint run --build-tags e2e --timeout=10m` (timeout raised for the larger file set; `$(GOENV)` prefix kept for `TestLintRecipeInheritsGOENV`). New guard `TestLintRecipeCoversE2ETag` fails if the flag ever disappears.

## Why `.github/workflows/ci.yml` is untouched

The issue predates #454. CI's lint job now runs `make lint`, so the Makefile change covers CI by construction — and `TestCILintJobRunsMakeLint` forbids ci.yml from mentioning golangci-lint at all. The Makefile pin stays at v1.64.8.

## Verification

- `make lint` (now tag-covering): exit 0; with output caps disabled: 0 findings.
- `make fmt-check`: pass.
[...]

Issue #436, opened 2026-08-21
lint: make lint and CI skip 96 e2e-tagged files — add --build-tags e2e


iceboundrock:
`make lint` (Makefile:85) and the CI lint job run golangci-lint without build
tags, so the 96 files carrying `//go:build e2e` are never linted at all.

Measured while working #319:

- With `--build-tags e2e`, the default linter set reports **6** pre-existing
  violations: `errcheck` ×3 (`internal/e2e_harness/federated/cdc_flush_test.go`,
  unchecked `h.Cleanup`), `unused` ×2 (`suite_test.go`: `testCategories`,
  `printTestCategories`), `staticcheck` ×1 (`consistency_test.go`, SA4010).
- **funlen reports 0** under the tag, so this gap did not weaken #319 — the
  e2e-tagged code is already inside the 100-line cap.
- Without the tag the baseline is 0 violations.

Verbatim from the #319 Task 8 verification run at HEAD `29bd665`:

```
$ golangci-lint run --build-tags e2e --timeout=10m
EXIT=1
internal/e2e_harness/federated/cdc_flush_test.go:32:17: Error return value of `h.Cleanup` is not checked (errcheck)
internal/e2e_harness/federated/cdc_flush_test.go:102:17: Error return value of `h.Cleanup` is not checked (errcheck)
internal/e2e_harness/federated/cdc_flush_test.go:162:17: Error return value of `h.Cleanup` is not checked (errcheck)
internal/e2e_harness/federated/suite_test.go:263:5: var `testCategories` is unused (unused)
internal/e2e_harness/federated/suite_test.go:276:6: func `printTestCategories` is unused (unused)
internal/e2e_harness/federated/consistency_test.go:321:16: SA4010: this result of append is never used, except maybe in other appends (staticcheck)

$ ... | grep "funlen)" || echo "NO funlen MATCHES"
NO funlen MATCHES
```

So the work is: fix those 6, then add `--build-tags e2e` to `make lint` and to
`.github/workflows/ci.yml` so the gap cannot reopen. Keep the golangci-lint pin
at v1.64.8.

Refs #319

iceboundrock:
# Lint e2e-tagged files (#436) Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

[...]

Issue #434, opened 2026-08-20
test(e2e): full federated suite has environment-sensitive failures — absolute wall-clock ceilings and container startup timeouts


iceboundrock:
Hit twice in four consecutive full-suite runs while validating PR #432. Neither failure is a code regression, and neither is visible to CI, because CI runs `-tags=e2e -short` and `-short` skips both affected files.

### 1. Absolute wall-clock ceiling

`deduplication_test.go:161`:

```go
start := time.Now()
result, err := h.ExecuteFederatedQuery(ctx, &QueryOptions{Limit: totalRecords})
duration := time.Since(start)
...
require.Less(t, duration, 5*time.Second, "deduplication took too long: %v", duration)
```

Observed in a full-suite run: `"5.220917791s" is not less than "5s"`.

The same call, same commit, run in isolation three times: **222ms, 214ms, 221ms** — a 23× margin under the ceiling. The variance is not the query; it is Docker contention after ~4 minutes of container churn (every test in this package starts and tears down its own Postgres and RustFS containers).

An absolute wall-clock assertion in a containerised e2e test measures the host as much as the code. `TestPerformance_ConcurrentQueries` has produced the same class of false failure before, and #263 closed the k6 equivalent.

Options worth weighing:
- Raise the ceiling to something that still catches a real regression (a 10K-record dedup that takes 30s is broken; one that takes 6s on a loaded laptop is not).
- Assert on a relative baseline measured in the same run rather than an absolute constant.
- Move the timing assertion to the benchmark harness, which already has proper baselining, and keep the e2e test to correctness only.

### 2. RustFS container startup timeout

`merge_on_read_test.go:165`, in a different run of the same suite:

```
start s3: failed to start rustfs container: start container: started hook: wait until ready:
mapped port: check target: retries: 10, port: "", last err: get state:
Get "http://%2Fvar%2Frun%2Fdocker.sock/.../json": context deadline exceeded
```

The container logs show RustFS reached `Starting: /usr/bin/rustfs /data` but never published port 9000 within the wait window. The test failed at `failed to create harness`, before any query ran. Ten retries at roughly 6s apart, then the docker socket call itself timed out.

Worth checking whether the readiness strategy should wait on an HTTP health response rather than a mapped port, and whether the per-container startup budget is too tight when many tests run back to back.

## Why this matters even though CI is green

[...]

PR #492, opened 2026-08-26
ci: nightly production e2e suite under -race — the #410 ruling

### LINKED ISSUE #410: CI lint gate never loads e2e-tagged files; 6 pre-existing findings hide in internal/e2e_harness/federated/

iceboundrock:
Closes #410.

## What this is

#410's two primary decisions were already settled by #436 / PR #483 (the two issues are the same finding from different sweeps): `make lint` carries `--build-tags e2e` fail-closed, CI runs the recipe via `make lint`, and `TestLintRecipeCoversE2ETag` guards the flag. Verified at current main: `make lint` exits 0.

What remained — per epic #452's Phase 1 carry-out — was recording the ruling #410 asked for: nothing keeps the production e2e suite green under `-race` after its one-off pass during #345.

## The ruling

Run the production suite under `-race` **nightly, not in the PR gate**:

- The race detector's 2–10x slowdown would invalidate the wall-clock regression tripwires #434 just calibrated for race-free runs; the nightly workflow already has the budget and the failure-issue reporting, and its value-is-signal design (gates nothing, files labeled issues) fits a race probe exactly.
- New `production-race` job in `nightly-e2e.yml` (separate hosted runner, so no container contention with the federated job; 30m test timeout ≈ local race-instrumented runtime 8m × headroom).
- `TestNightlyRacesProductionSuite` guards the `-race` flag the same way `TestLintRecipeCoversE2ETag` guards the lint tag (negative-checked: removing `-race` fails the guard).
- The failure-issue mechanics are now one script (`scripts/nightly_failure_issue.sh`) with per-suite labels — a shared label would have appended `-race` failures onto the federated flake thread and buried the signal. Federated messages stay byte-identical.

## Deflake found by the ruling's own verification

The first full `-race` run failed `TestRecreateBetweenExportAndMark`: its 2s clock-ahead restamp is a budget for everything between restamp and the asserted write — including a flush list+export leg that took ~2.08s under `-race` — so the wall clock overtook the restamp and `GREATEST` stamped wall time (`recreate stamped 1787725421723, want 1787725421646`). Not a data race, not a correctness bug: the write-path invariant (strictly above the tombstone) held; only the deterministic exact-version pin lost its precondition.

- `clockAheadLeadMs = 20s` keeps the GREATEST branch deterministically exercised on 2-core runners (nothing waits for the clock to catch up, so size costs no wall time).
- `requireExactStamp` distinguishes a real GREATEST regression from an exhausted lead budget in the failure message, so a nightly-filed issue reads correctly.

## Verification

- Full production suite `-tags=e2e -race`: **PASS, no data races**, 487s locally (podman) after the deflake; the pre-fix run reproduced the failure.
- Monotonic scenarios `-race -count=3`: PASS, twice (before and after the review round).
- `make lint`, `make fmt-check`, root-package guard tests, actionlint on the workflow: clean.
- Failure-issue script dry-run against a stub `gh`: federated create/comment bodies byte-identical to the previous inline step.

Refs #452 (Phase 1 carry-out), #434, #436, #345.

https://claude.ai/code/session_01FsmokCNqVJ9sYH5AtbqaXB

### DIFF
--- a/.github/workflows/nightly-e2e.yml
+++ b/.github/workflows/nightly-e2e.yml
@@ -53,24 +53,21 @@ jobs:
       - name: Run full federated e2e suite (no -short)
         run: go test -v ./internal/e2e_harness/federated/... -tags=e2e -timeout=75m
 
+      # The two nightly jobs share one reporting script but MUST NOT share a
+      # label: the label keys the open-issue lookup, so a shared one appends
+      # this suite's failures to the other suite's thread (#410 review).
       - name: File or update failure issue
         if: failure()
         env:
           GH_TOKEN: ${{ github.token }}
-        run: |
-          gh label create nightly-e2e --repo "$GITHUB_REPOSITORY" \
-            --color D93F0B --description "Nightly full e2e suite failure" --force
-          run_url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
-          existing=$(gh issue list --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
-            --state open --json number --jq '.[0].number // empty')
-          if [ -n "$existing" ]; then
-            gh issue comment "$existing" --repo "$GITHUB_REPOSITORY" \
-              --body "Nightly full federated e2e suite failed again: $run_url"
-          else
-            gh issue create --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
-              --title "nightly: full federated e2e suite failed" \
-              --body "The scheduled full federated suite (\`-tags=e2e\`, no \`-short\`) failed: $run_url (refs #434)"
-          fi
+        run: >
+          bash scripts/nightly_failure_issue.sh
+          nightly-e2e
+          "Nightly full e2e suite failure"
+          "nightly: full federated e2e suite failed"
+          "Nightly full federated e2e suite failed again"
+          "The scheduled full federated suite (\`-tags=e2e\`, no \`-short\`) failed"
+          "#434"
 
   production-race:
     name: Production Suite (-race)
@@ -91,21 +88,16 @@ jobs:
       - name: Run production e2e suite under -race
         run: go test -v ./internal/e2e_harness/production/ -tags=e2e -race -timeout=30m
 
+      # Same script, own label — see the label note on the federated job.
       - name: File or update failure issue
         if: failure()
         env:
           GH_TOKEN: ${{ github.token }}
-        run: |
-          gh label create nightly-e2e --repo "$GITHUB_REPOSITORY" \
-            --color D93F0B --description "Nightly full e2e suite failure" --force
-          run_url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
-          existing=$(gh issue list --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
-            --state open --json number --jq '.[0].number // empty')
-          if [ -n "$existing" ]; then
-            gh issue comment "$existing" --repo "$GITHUB_REPOSITORY" \
-              --body "Nightly production e2e suite failed under -race: $run_url"
-          else
-            gh issue create --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
-              --title "nightly: production e2e suite failed under -race" \
-              --body "The scheduled production suite run (\`-tags=e2e -race\`) failed: $run_url (refs #410)"
-          fi
+        run: >
+          bash scripts/nightly_failure_issue.sh
+          nightly-e2e-race
+          "Nightly production e2e -race failure"
+          "nightly: production e2e suite failed under -race"
+          "Nightly production e2e suite failed under -race again"
+          "The scheduled production suite run (\`-tags=e2e -race\`) failed"
+          "#410"
--- a/.github/workflows/nightly-e2e.yml
+++ b/.github/workflows/nightly-e2e.yml
@@ -4,6 +4,15 @@
 # signal, and the measurable baseline for the suite's flake rate. It gates
 # nothing: value is signal, so failures file/update a labeled issue instead
 # of blocking PRs.
+#
+# The production-race job is the #410 ruling: CI races only the unit suite,
+# so the production e2e harness — the one venue driving real concurrent
+# flush/compaction/read against real containers — had no recurring -race
+# signal after passing once during #345. It lives here, not in the PR gate:
+# the race detector's slowdown would invalidate the wall-clock regression
+# tripwires #434 calibrated for race-free runs, and this workflow already
+# has the budget and the failure-issue reporting. Jobs run on separate
+# hosted runners, so the two suites' containers never contend.
 name: Nightly Full E2E
 
 on:
@@ -61,4 +70,42 @@ jobs:
             gh issue create --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
               --title "nightly: full federated e2e suite failed" \
               --body "The scheduled full federated suite (\`-tags=e2e\`, no \`-short\`) failed: $run_url (refs #434)"
+          fi
+
+  production-race:
+    name: Production Suite (-race)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      # TestNightlyRacesProductionSuite guards this line. The 30m timeout is
+      # the gate run's 10m (make test-e2e-production) times the ~2-3x race
+      # instrumentation cost, rounded up.
+      - name: Run production e2e suite under -race
+        run: go test -v ./internal/e2e_harness/production/ -tags=e2e -race -timeout=30m
+
+      - name: File or update failure issue
+        if: failure()
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh label create nightly-e2e --repo "$GITHUB_REPOSITORY" \
+            --color D93F0B --description "Nightly full e2e suite failure" --force
+          run_url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+          existing=$(gh issue list --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
+            --state open --json number --jq '.[0].number // empty')
+          if [ -n "$existing" ]; then
+            gh issue comment "$existing" --repo "$GITHUB_REPOSITORY" \
+              --body "Nightly production e2e suite failed under -race: $run_url"
+          else
+            gh issue create --repo "$GITHUB_REPOSITORY" --label nightly-e2e \
+              --title "nightly: production e2e suite failed under -race" \
+              --body "The scheduled production suite run (\`-tags=e2e -race\`) failed: $run_url (refs #410)"
           fi
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,7 +55,7 @@ Bun black-box E2E and k6 load tests live in `tests/e2e/` (see its README). Bench
 
 Local server: `./scripts/local_server.sh` (Docker Compose Postgres + init-db + server on :8080).
 
-CI (`.github/workflows/ci.yml`) runs lint, unit tests with `-race`, the e2e smoke + `-tags=e2e -short` federated suite, the production harness, and `benchmark-smoke`. Note `-short` skips 37 of the 124 federated e2e tests (deduplication, merge-on-read, data-tier, soft-delete, performance); their CI signal is the nightly full-suite run (`.github/workflows/nightly-e2e.yml`), which files a `nightly-e2e`-labeled issue on failure (#434).
+CI (`.github/workflows/ci.yml`) runs lint, unit tests with `-race`, the e2e smoke + `-tags=e2e -short` federated suite, the production harness, and `benchmark-smoke`. Note `-short` skips 37 of the 124 federated e2e tests (deduplication, merge-on-read, data-tier, soft-delete, performance); their CI signal is the nightly workflow (`.github/workflows/nightly-e2e.yml`), which runs the full federated suite (#434) and the production suite under `-race` (#410), filing a `nightly-e2e`- / `nightly-e2e-race`-labeled issue respectively on failure.
 
 ## Architecture
 
--- a/internal/e2e_harness/production/monotonic_version_e2e_test.go
+++ b/internal/e2e_harness/production/monotonic_version_e2e_test.go
@@ -64,17 +64,23 @@ func TestMonotonicVersionsCloseSameMillisecondTies(t *testing.T) {
 const clockAheadLeadMs = 20_000
 
 // requireExactStamp asserts got == want and, on mismatch, separates the two
-// possible causes: if the wall clock has overtaken the restamp, the lead
-// budget was eaten by scheduling — the GREATEST branch was never reachable
-// and the exact-version pin is meaningless — so the failure must read as a
-// blown budget, not as a monotonicity regression (#410 review).
-func requireExactStamp(t *testing.T, what string, got, want, ahead int64) {
+// possible causes: if the wall clock had overtaken the restamp by the time
+// the write completed, the lead budget was eaten by scheduling — the
+// GREATEST branch was never reachable and the exact-version pin is
+// meaningless — so the failure must read as a blown budget, not as a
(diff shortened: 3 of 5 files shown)


iceboundrock:
## Review: ci: nightly production e2e suite under -race (#410)

Reviewed at head `40104dd` against `main` (`36491db`). Verdict up front: **approve** — the design is sound, the ruling is faithfully implemented, and every mechanical claim I could test locally held. One optional doc follow-up and three nits below, none blocking.

### Findings

1. **[Low, follow-up] AGENTS.md drift.** AGENTS.md:58 still describes the nightly workflow as filing "a `nightly-e2e`-labeled issue" (singular, #434). After this PR there are two jobs and two labels (`nightly-e2e`, `nightly-e2e-race`, refs #434/#410). Suggest a one-line follow-up so the repo's own orientation doc matches the wiring it describes.
2. **[Nit] `requireExactStamp` diagnoses at assert-time, not write-time** (internal/e2e_harness/production/monotonic_version_e2e_test.go:76). The lead-exhaustion branch reads `time.Now()` when the assertion runs; for the recreate case the write happens inside the paused-flush window but the assertion fires only after the flush resumes and completes. If the wall clock crossed `ahead` during that post-write leg *and* the stamp genuinely regressed, a real GREATEST regression would be reported as a blown budget. The window is milliseconds against a 20s budget, so this is acceptable as-is; capturing `time.Now()` alongside the write would remove the ambiguity if it ever matters.
3. **[Nit] Redundant cast**: `int64(clockAheadLeadMs)` in the Fatalf at monotonic_version_e2e_test.go:77 — the untyped constant formats fine via `%d`.
4. **[Nit] Script arg hygiene**: `scripts/nightly_failure_issue.sh` takes 6 positional args with no count/usage validation; a transposition fails confusingly. Fine for a CI-only script invoked from two call sites in one file; a cheap `[[ $# -eq 6 ]] || { echo "usage: …" >&2; exit 2; }` would harden it.

### Spec / assessment

- **The ruling is correctly scoped.** #410's remaining decision was where the recurring `-race` signal over the production harness lives. Nightly-not-gate is right: the detector's 2–10x slowdown would invalidate the wall-clock tripwires #434 calibrated race-free, and the nightly's value-is-signal design (gates nothing, files labeled issues) fits a probe exactly. Rationale is documented in the workflow header, not just the PR body.
- **Per-suite failure labels are the correct fix for a real defect found by self-verification.** With both steps on the shared `nightly-e2e` label, a production-race failure would have been appended to the federated flake thread, and concurrent failures raced the `.[0].number` lookup. One script + per-job labels fixes both while keeping the mechanics single-defined.
- **Deflake diagnosis checks out.** The 2s restamp lead was a budget covering everything between restamp and asserted write — including the paused flush's list+export leg (~2.08s under `-race`), so the clock overtook the stamp and `GREATEST` stamped wall time. Not a data race, not a monotonicity bug. `clockAheadLeadMs = 20_000` gives ~10x margin over the measured worst case at zero wall-time cost (nothing waits for the clock to catch up), keeping the GREATEST branch deterministic on 2-core runners.
- **Timeout ordering is deliberately survivable**: go test's `-timeout=30m` fires before the job's 45m cap, so a hung run still exits as a normal step failure and the `if: failure()` reporting step runs — a job-level timeout cancellation would skip it. Budgets are ordered correctly.

### Standards

- File/function limits hold: nightly_guard_test.go 41 lines; monotonic file 306 lines (<500); all functions well under 100.
- Guard style matches its siblings in toolchain_guard_test.go: plain text scan, comment lines skipped, fail-closed `t.Fatal` when the target line disappears ("lost its target"), #refs in messages. Consistent with the established pattern.
- `gofmt` clean on changed Go files; `bash -n` clean on the script; executable bit set (`100755`); workflow YAML parses (validated with gopkg.in/yaml.v3 — both jobs present). shellcheck/actionlint are not installed locally, so those gates are confirmed only by the author's run and CI.
- No production code touched; error-wrapping/message rules N/A.

### Observations

- The guard matches `-race` as a substring, so a hostile `-race=false` would satisfy it — same tradeoff every existing text-scan guard accepts; not worth a regex here.
- The nightly job calls `go test` directly rather than through a Makefile recipe, bypassing the GOENV pin — but that is the pre-existing shape of the federated job, and setup-go's `GO_VERSION` ('1.26.0') equals the Makefile floor, so parity holds today.
- Concurrency group `nightly-e2e` is unchanged: whole-run serialization preserved (#490), and the two suites sit on separate hosted runners, so no container contention within a run.
- Federated reporting strings were rebuilt, not copied blindly — see verification below; they are byte-identical to the previous inline step.

### Verification (all run locally at head `40104dd`)

- `TestNightlyRacesProductionSuite`: **PASS**.
- Negative check 1 — removed `-race` from the production run line: guard **FAILS** with the intended message ("lacks -race: … silently lose its only recurring -race signal").
- Negative check 2 — deleted the production run line entirely: guard **FAILS** ("lost its target"). Fail-closed confirmed both ways.
- Stub-`gh` dry-run of `scripts/nightly_failure_issue.sh`, both paths: label-create args identical to the old inline step; issue-create body `…failed: <run_url> (refs #434)` and comment body `…failed again: <run_url>` are **byte-identical** to the pre-PR federated strings, including the escaped backticks surviving the folded-scalar → bash quoting chain.
- `gofmt -l`, `bash -n`, YAML parse: clean.
- Not re-run locally (relied on author evidence + CI): the full production `-tags=e2e -race` suite (author reports PASS, no races, 487s under podman, with the pre-fix run reproducing the deflake failure) and `make lint`/actionlint (linter not installed here; diff contains no non-test Go changes).

### Summary

Solid piece of work: the ruling lands where it belongs, the guard fails closed in both regression directions, the reporting refactor removes a genuine cross-suite signal-pollution bug without changing any observable message, and the deflake turns an opaque exact-version failure into a self-diagnosing one. Only actionable item is the AGENTS.md one-liner whenever convenient; everything else is optional polish.

iceboundrock:
All four findings addressed in `b5fb0ac` — including the one you marked follow-up, since it was cheaper to fix here than to track:

1. **AGENTS.md drift — fixed in this PR rather than a follow-up.** The drift is introduced by this PR (second job, second label), so the one-liner belongs in the same change: AGENTS.md:58 now describes the nightly workflow as running the full federated suite (#434) and the production suite under `-race` (#410), filing `nightly-e2e`- / `nightly-e2e-race`-labeled issues respectively. No follow-up issue opened.
2. **Assert-time vs write-time clock — fixed, your nit was righter than its framing.** For the recreate the post-write leg isn't milliseconds: the assertion only runs after the paused flush resumes and completes, which is exactly the slow S3 leg under `-race`. `requireExactStamp` now takes `wroteByMs`, read right after the asserted write returns — for the recreate, inside the `during` callback, which `runFlushPausedOn` invokes on the test goroutine (no synchronization hazard). This also makes the regression branch conclusive: `wroteByMs <= ahead` proves the stamping `GREATEST` saw a clock no later than that reading, so a mismatch there cannot be a blown budget.
3. **Redundant `int64` cast — dropped.**
4. **Script arg hygiene — added** `[ "$#" -ne 6 ]` → usage on stderr, exit 2 (verified: wrong count exits 2 with the usage line).

On the observations: agreed on all three, no action taken — the `-race` substring match is the same tradeoff every sibling text-scan guard accepts; the direct `go test` (no GOENV pin) matches the pre-existing federated job shape with setup-go's `GO_VERSION` equal to the Makefile floor, so parity holds and any fix belongs to a workflow-wide pass, not this PR.

Verification at `b5fb0ac`: monotonic scenarios `-tags=e2e -race -count=3` PASS; `go vet -tags=e2e`, `gofmt`, `bash -n`, `make fmt-check`, root-package guard tests all clean.

iceboundrock:
## Review: #492 — ci: nightly production e2e suite under -race — the #410 ruling

Reviewed at head `b5fb0ac` against `main` (`36491db`). **Verdict: approve — no blocking findings.** All four findings from the `40104dd` review are addressed in `b5fb0ac`; verification below re-confirms them plus the standing invariants.

### Findings

No blocking findings at `b5fb0ac`. The four items flagged at `40104dd` are closed:

1. **[Fixed] AGENTS.md drift** — `AGENTS.md:58` now reads `which runs the full federated suite (#434) and the production suite under \`-race\` (#410), filing a \`nightly-e2e\`- / \`nightly-e2e-race\`-labeled issue respectively`. The doc no longer describes a singular `nightly-e2e` label. Correct to fix in-PR (drift is introduced by this PR).

2. **[Fixed] `requireExactStamp` assert-time vs write-time clock** (`internal/e2e_harness/production/monotonic_version_e2e_test.go:77`) — prior nit undersold the window: for `TestRecreateBetweenExportAndMark` the assertion runs only after the paused flush resumes/completes (the S3 export leg), so an assert-time `time.Now()` could have mis-attributed a genuine GREATEST regression as a blown budget. Now takes `wroteByMs` captured right after the write returns — for the recreate, inside the pause callback (`monotonic_version_e2e_test.go:268` `recreatedBy = time.Now().UnixMilli()`), which `runPausedFlush` invokes on the test goroutine (no sync hazard). `wroteByMs <= ahead` is now conclusive that the stamping `GREATEST` saw a clock no later than that reading.

3. **[Fixed] Redundant `int64()` cast** — `Fatalf` now formats `clockAheadLeadMs` directly (`%d`, untyped const) at `monotonic_version_e2e_test.go:83`.

4. **[Fixed] Script arg hygiene** — `scripts/nightly_failure_issue.sh:14` rejects `$# != 6` with `usage:` on stderr, exit 2.

No new blocking or low findings. Three pre-existing observations remain `won't fix` by design (see Observations).

### Spec / assessment

- **The #410 ruling is correctly scoped: nightly, not PR gate.** #410's residual (after #436/PR #483 settled the `--build-tags e2e` lint gate via `TestLintRecipeCoversE2ETag`) was where the recurring `-race` signal over the production harness lives. The production harness is the only venue driving real concurrent flush/compaction/read against real containers; CI's PR gate races only the unit suite. Nightly-not-gate is right: detector's 2–10× slowdown would invalidate the wall-clock tripwires #434 calibrated race-free; nightly's value-is-signal design (gates nothing, files labeled issues) fits a race probe exactly. Rationale is documented in the workflow header (`.github/workflows/nightly-e2e.yml:6-13`), not just the PR body.

- **Per-suite labels + single script is the correct fix for a real defect.** With both jobs on shared `nightly-e2e`, a production-race failure would append to the federated flake thread and concurrent failures raced the `.[0].number` lookup. `scripts/nightly_failure_issue.sh` (38 lines, `0755`, `set -Eeuo pipefail`) is now the single definition; federated `nightly-e2e` and production-race `nightly-e2e-race` each own their label. Federated messages stay byte-identical (verified by stub-`gh` dry-run).

- **Deflake diagnosis checks out, fix is proportionate.** Original `restampAhead = 2000` is a budget for everything between restamp and asserted write — for `TestRecreateBetweenExportAndMark` that includes the flush list+export leg, which measured ~2.08s under `-race` and overtook the stamp (`GREATEST` fell to wall time: `recreate stamped 1787725421723, want 1787725421646`). Not a data race, not a monotonicity bug (invariant strictly-above-tombstone held). `clockAheadLeadMs = 20_000` at `monotonic_version_e2e_test.go:64` gives ~10× margin over worst-case on 2-core runners; zero wall-time cost (nothing waits for the clock to catch up), keeps the `GREATEST` branch deterministic.

- **Timeout ordering is deliberately survivable:** `go test -timeout=30m` (`nightly-e2e.yml:89`) < job `timeout-minutes: 45` (`nightly-e2e.yml:75`), so a hung run exits as a step failure and `if: failure()` reporting still runs; a job-level timeout cancellation would skip it. Federated `75m < 90m` ordering preserved. `30m` ≈ gate `10m` (`Makefile:67` `test-e2e-production`) × ~2–3× race cost, rounded up; locally reported 487s under `-race` leaves ~headroom.

- **Epic #452 tracking:** Phase 1 carry-out closed correctly; #410 `Closes` is accurate (lint gate was folded into #436, `-race` ruling lands here).

### Standards

- **File/function limits hold:** `nightly_guard_test.go` 41 lines, `scripts/nightly_failure_issue.sh` 38 lines, `internal/e2e_harness/production/monotonic_version_e2e_test.go` 319 lines (< 500 at `b5fb0ac`; was 287 on `main`). All functions < 100 lines. No new `//nolint:funlen` needed.
- **Guard style matches siblings** (`toolchain_guard_test.go`): plain text scan, comment-skipped, `t.Fatal` fail-closed when target disappears (`nightly_guard_test.go:39` `lost its target`), `#410` refs in messages; no YAML dependency.
- **Formatting/lint:** `gofmt -l` clean on both `.go` files; `bash -n` clean on script; executable bit `100755` set; workflow YAML parses (both jobs present). `actionlint`/`shellcheck` not installed locally — rely on author + CI (CI `Lint` already `SUCCESS` at this push). No production code touched, so error-wrapping/message rules N/A.
- **No dead code / TODOs.**

### Observations (no action, as before)

- **Guard substring match:** `-race` checked via `strings.Contains`, so `-race=false` would satisfy it. Same tradeoff every existing text-scan guard accepts; a regex adds no value here.
- **Nightly `go test` without `GOENV` pin:** the new `production-race` job calls `go test` directly (like the pre-existing `federated-full` job), bypassing `Makefile:GOENV` (`GOTOOLCHAIN=go1.26.0+auto`). Parity holds because `setup-go` `GO_VERSION: '1.26.0'` equals the Makefile floor; any fix belongs to a workflow-wide pass, not this PR.
- **Concurrency:** `concurrency.group: nightly-e2e / cancel-in-progress: false` unchanged — serializes overlapping nightly runs (#490). Within one run the two jobs sit on separate `ubuntu-latest` runners, so no container contention.

### Verification

Re-ran locally at `b5fb0ac` (no container-dependent suite re-run; relied on author evidence + CI for the full `-race` suite):

- `GOTOOLCHAIN=go1.26.0+auto go test -run TestNightlyRacesProductionSuite -v .` — **PASS** (`nightly_guard_test.go:23`).
- **Negative 1 — remove `-race`:** `run:` line `… -tags=e2e -timeout=30m` (strip `-race`) → guard **FAILS** as intended: `lacks -race: the production suite would silently lose its only recurring -race signal (#410)` (`nightly_guard_test.go:34`).
- **Negative 2 — delete production run line:** guard **FAILS** as intended: `no run directive for ./internal/e2e_harness/production/ found … lost its target (#410)` (`nightly_guard_test.go:39`). Fail-closed both ways confirmed.
- **Script dry-run (stub `gh`):** federated `create` body `The scheduled full federated suite (\`-tags=e2e\`, no \`-short\`) failed: <run_url> (refs #434)` and comment `…failed again: <run_url>` **byte-identical** to the pre-PR inline federated step; label-create args identical. Production-race path: `label nightly-e2e-race`, `description "Nightly production e2e -race failure"`, `title "nightly: production e2e suite failed under -race"`, `comment "Nightly production e2e suite failed under -race again: …"` — correct per-PR spec. Existing-issue comment path (`gh issue list` → `123`) correctly routes to `gh issue comment 123`. Arg-count guard: `bash scripts/nightly_failure_issue.sh a b c` → `usage: …` on stderr, exit 2.
- **Static checks:** `gofmt -l` empty, `bash -n` ok, `ls -l` shows `0755`, `wc -l` limits above, `grep -rn "2000"` shows no remaining literal (all restamps via `clockAheadLeadMs`).
- **Not re-run locally:** full `go test -v ./internal/e2e_harness/production/ -tags=e2e -race` (author reports PASS, no races, 487s podman post-fix; pre-fix reproduced the failure) and `make lint`/`actionlint` (author + CI `Lint: SUCCESS` covers it; diff has no non-test Go changes).
- **CI at this push:** `Lint: SUCCESS`, `Test: SUCCESS`, `Benchmark Smoke: SUCCESS`, `E2E Smoke Tests` / `k6 Smoke` `IN_PROGRESS` — expected for e2e-affecting change.

### Summary

Clean Phase 1 carry-out: the ruling lands where it belongs (nightly, per-suite labeled, script-single-defined), the guard fails closed in both regression directions, the reporting refactor fixes a genuine cross-suite signal-pollution bug without changing any observable federated message, and the 2s→20s deflake plus `wroteByMs`-anchored diagnosis turns an opaque `-race`-only flake into a self-diagnosing failure. No blocking findings. Ready to merge.

Refs: #410, #452 (Phase 1 carry-out), #434, #436/PR #483, #345.

PR #504, opened 2026-08-29
fix(federated): project ltbase_created_at from parquet instead of the LWW version stamp (#460)

### LINKED ISSUE #460: bug(federated): reads report the LWW version stamp as created_at for parquet-winning rows — wrong created_at and an unstable default sort key across the flush boundary
## Problem

Federated reads report a row's LWW *version* timestamp in the `created_at` slot for every warm/cold-winning row, while the hot leg reports the true creation time. The two `UNION ALL` legs put different quantities in the same column, so `created_at` is wrong for parquet-winning rows and — worse — the default `ORDER BY created_at DESC` sort key **changes value the moment a row is flushed**, making the default order unstable across the flush boundary.

## Current behavior

`internal/sqlgen/duckdb_schema_projection.go`:
- S3 leg (`buildS3Projection`): `changed_at AS created_at`
- hot leg (`buildPGProjection`): `m.ltbase_created_at AS created_at`
- outer (`buildOuterSelect`): `created_at AS ltbase_created_at`, consumed at `internal/federated/duckdb_query_helpers.go` (`case "ltbase_created_at": record.CreatedAt = val.Int64`).

Default order: `buildNonKeysetOrderBy` (`internal/sqlgen/duckdb_template_renderer.go`) → `created_at DESC, row_id ASC`. Mirrored in `duckdb_benchmark_projection.go`.

The parquet files *do* carry the true value — `internal/cdc/export_sql_builder.go` `baseSelectColumns()` exports `m.ltbase_created_at` on both delta and base — the reader simply never projects it, aliasing the version stamp into the slot instead.

## Reproduction

Any row created, then updated, then flushed (or base-exported by `cdc-init`):
- its API-reported `created_at` becomes its last-update time;
- its default sort key changes at the flush boundary, so `ORDER BY created_at DESC` LIMIT/OFFSET windows can duplicate or skip rows, and a `created_at`-bearing keyset cursor can skip/repeat a page;
- the OLTP route and federated route disagree on `created_at` for the same row (and route is chosen by page size — see the hybrid-routing issue);
- after `cdc-init` onboarding + `change_log` cleanup, `pg_source` is empty for every row, so *every* row's reported `created_at` collapses to its last-update time.

## Root cause

`#210` fixed the *ver_ts* half of this alias (base now exports `ltbase_updated_at AS changed_at`) and left `created_at` aliased to `changed_at`. `#212` names the alias as the mechanism but treats it as given. The reader has a real `ltbase_created_at` available in the parquet and does not project it.

## Impact

Wrong `created_at` returned to API callers for parquet-winning rows; an unstable default sort key across the flush boundary (pagination duplicates/skips); route-dependent disagreement for the same row. The behavior is currently *characterized as expected* (`internal/federated/design_doc_sql_test.go` asserts "s3 (100, from parquet changed_at) precedes d1 (90, from entity_main.ltbase_created_at)"; `docs/federated-query/design.md` §5 carries the alias), so fixing it is also a doc + test change.

- #210 (closed) fixed the ver_ts half; #212 (closed) names the alias; #214 (closed) pinned the doc to the runtime shape, alias included; #183 (closed) covered tie-break stability, not key stability. Searched `created_at`, `sort stability`, `ORDER BY`, `ltbase`. NEW — no issue owns the `created_at` correctness/stability defect.
- Amplified by the hybrid-routing issue (route chosen by page size) and interacts with the keyset-cursor contract (#381).

## Acceptance criteria

- [ ] Federated `created_at` reflects the row's true creation time on every tier (project `ltbase_created_at` from parquet, keep `changed_at` for LWW only).
- [ ] Default `ORDER BY created_at` is stable across the flush boundary; OLTP and federated routes agree.
- [ ] Design §5 and `design_doc_sql_test.go` updated to the corrected contract.

- [ ] A created→updated→flushed row reports its creation time, not its update time, via the federated route.
- [ ] Pagination stability across a flush that changes no row's creation time.

iceboundrock:
Fixes #460.

## The defect

`buildS3Projection` aliased `changed_at AS created_at` while both hot legs projected `m.ltbase_created_at`. The two `UNION ALL` legs therefore put different quantities in the same column:

- `created_at` was wrong for every parquet-winning row (it reported the row's last update);
- the default sort key `created_at DESC, row_id ASC` **changed value the moment a row was flushed**, so LIMIT/OFFSET windows and `created_at` keyset cursors could duplicate or skip rows across the flush boundary;
- the OLTP and federated routes disagreed on the same row — while the route is chosen by page size.

## The fix

Reader-only. Both exporters have written the true `ltbase_created_at` into every parquet generation since the beginning of the repository's history (`cdc.exportModeSpec.baseSelectColumns`, both the delta and base modes), and compaction's `SELECT * EXCLUDE (_rn)` passes it through — so projecting the real column is sufficient for every object already in storage. **No data migration.**

`changed_at` stays confined to `ver_ts`; #460 narrows it, it does not move it.

**No `COALESCE` fallback**, deliberately. The only production rows whose parquet `ltbase_created_at` is NULL are hard-delete tombstones — the delta export `LEFT JOIN`s `entity_main` so a hard-deleted row's change_log entry still exports (#173) — and every hard delete stamps `deleted_at > 0` (`postgres_persistent_repository.go`), so those rows always lose to the `deleted_ts` filter in `visible` and never reach an `ORDER BY`. A fallback would put the version stamp back in the `created_at` slot on exactly the path being fixed.

## Beyond the issue body: the test harnesses write their own parquet

`internal/e2e_harness/federated/s3_operations.go` writes synthetic parquet from a CSV whose header carried no `ltbase_created_at`, and that writer feeds both the federated e2e suite and the benchmark tier loader. Projecting the column without touching the harness would have failed to bind on every one of those tests, so the harness is part of this change:

- `TestRecord` grows `CreatedAt` plus a `CreationStamp()` accessor that falls back to `ChangedAt`. The fallback is what keeps every pre-existing fixture's semantics **bit-identical**: a parquet-winning fixture row still reports the same value it did before.
- `entity_main.ltbase_created_at` is seeded from the same accessor, so a fixture's hot and cold copies agree.
- `benchmark.GeneratedRecord` gains `CreatedAt`, stamped once per row-version group before `SplitIntoTiers` scatters versions across base/delta/hot — otherwise the benchmark would measure a flush-boundary instability its own fixtures invented.

The benchmark projection mirror moves in lockstep with the reader, per the issue's second reminder. (The comment's `:362` line reference is stale — the file has been 323 lines since the #220 split; the two sites are `BuildBenchmarkProjections` and `BuildBenchmarkS3Projection`.)

## Doc and characterization

`docs/federated-query/design.md` §5 and `internal/federated/design_doc_sql_test.go` both carried the alias as *expected* behaviour. Both flip together: the §5 seed parquet now gives each row an `ltbase_created_at` distinct from its `changed_at`, and the assertion pins `s3` reporting 80 (its creation stamp) rather than 100 (its version stamp), which reorders the documented page.

`internal/e2e_harness/production/keyset_lww_e2e_test.go`'s `created_at_cursor_resurrection` scenario was written against the defect ("A's winner carries created_at = t4"). It now pins the corrected contract — A keeps its creation stamp through the update, so it legitimately qualifies for the cursor — while still pinning the #212 property it was built for: each winner appears **exactly once**, never once per version.

## Regression coverage

- **`internal/sqlgen/duckdb_created_at_projection_test.go`** — projection pins for the production and benchmark legs, including that both legs name the same quantity in the `created_at` slot.
- **`internal/federated/created_at_flush_boundary_test.go`** — renders the *runtime* template and executes it on in-memory DuckDB with each `postgres_scan` rewritten to a local relation. Creation stamps ascend while version stamps descend, so a leg reporting the version stamp reverses the page outright. Covers both acceptance criteria without a container.
- **`internal/e2e_harness/production/created_at_flush_boundary_e2e_test.go`** — three scenarios over the real write path, real `cdc-flush` and real engine: a created→updated→flushed row reports its creation time; the OLTP and federated routes agree on the value; and the federated default order is unchanged across cold → hot (updates leaving stale parquet copies) → cold again.

All three e2e scenarios were confirmed to **fail** with the projection reverted, then pass with it restored.

## Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/production/ -tags=e2e` | pass (503s, full suite) |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | 122/124 pass |

The two federated e2e failures are `TestPerformance_SimplePagination` (p95 582ms vs a 450ms threshold) and `TestPerformance_ConcurrentQueries` (7.22 QPS vs an 8.33 minimum). **Both fail on `main` on this machine too** — baseline p95 512ms and 7.99 QPS, likewise under threshold — so they are pre-existing local-environment failures (rootless Podman), not a regression. The run-to-run deltas are within this machine's noise; CI's `-short` run does not include them and the nightly workflow is the real signal.

## Note for the reviewer

`internal/sqlgen/duckdb_schema_projection.go` is now 499 lines against the 500-line limit. It passes `TestSqlgenPackageFilesStayWithinFileSizeLimit`, but the next edit to that file will trip the gate. The natural seam is the type-mapping tail (`duckDBAttrCast` / `duckDBColumnType`) moving into `duckdb_type_mapper.go`; I left it out of this change to keep the diff on the defect, and it is worth a follow-up.


iceboundrock:
## Review: PR #504

## Findings

### P1. The legacy federated harness does not carry the new creation stamp through its actual read/flush/compaction paths

Locations: `internal/e2e_harness/federated/s3_operations.go:76-90`, `internal/e2e_harness/federated/query_build.go:67-77,112-123`, `internal/e2e_harness/federated/query.go:183-213`, and `internal/e2e_harness/federated/cdc.go:49-55,172-200`.

The writer now emits `ltbase_created_at`, but the federated harness that consumes these files still treats `changed_at` as both the creation and update timestamp. `buildParquetTierQuery` does not project `ltbase_created_at`, and `scanQueryResults` scans only `changed_at` then assigns it to both `PersistentRecord.CreatedAt` and `UpdatedAt`. Therefore a fixture such as `TestRecord{CreatedAt: 100, ChangedAt: 200}` is written with the correct new column but is returned by `ExecuteFederatedQuery` with `CreatedAt == 200`. This makes the new `TestRecord.CreatedAt` field ineffective for the existing federated E2E suite and leaves that suite asserting the old contract.

The simulated lifecycle has the same problem even if the query projection is fixed: `RunCDCFlush` constructs records with only `ChangedAt: time.Now().UnixMilli()`, so flushing an existing row exports the flush time as `ltbase_created_at` instead of reading the row creation stamp from `entity_main`. `readDeltaFiles` also omits `ltbase_created_at`, so `RunCompaction` drops the creation stamp and `WriteParquet` falls back to `ChangedAt` in the rewritten base file. A row can consequently change its reported `created_at` after this harness flushes or compacts it.

Please thread separate creation/version columns through this harness: project and scan `ltbase_created_at` independently from `changed_at`, fetch/preserve the main-table creation stamp in `RunCDCFlush`, and retain it in `readDeltaFiles`/compaction. Add a regression covering `CreatedAt != ChangedAt` across direct query, flush, and compaction. The production `Env` path added by this PR does not have this defect, but the changed harness writer currently creates a split-brain test contract and can produce false greens for the legacy federated suite.

## Spec Compliance

- The production schema-driven projection now correctly maps parquet `ltbase_created_at AS created_at` and keeps `changed_at AS ver_ts`.
- The benchmark production projection is updated consistently, and generated benchmark versions receive one creation stamp per row-version group before tier splitting.
- The production CDC exporter has always included `m.ltbase_created_at`, so the reader-side fix does not require a data migration for repository-produced parquet.
- The production regression coverage is strong: it covers the runtime DuckDB template, creation/update/flush behavior, route-value parity, default-order stability, and keyset behavior.
- Compliance is incomplete for the legacy federated harness because its reader and lifecycle rewrite paths still collapse or lose the creation stamp described above.

## Standards

- No changed Go source file exceeds the 500-line limit; `internal/sqlgen/duckdb_schema_projection.go` is exactly 499 lines.
- Changed functions remain within the stated 100-line limit.
- Error handling and wrapping in the changed production paths follow the repository conventions.
- No diff whitespace errors were found.
- The change stays within the existing sqlgen, federated, CDC, benchmark, and E2E layers without unrelated production refactoring.

## Assessment

**Request changes.** The production implementation of the projection fix is correct and the main new regression tests pass, but the PR also changes the shared federated test harness schema without updating all of its readers and lifecycle producers. Until those paths preserve and expose `ltbase_created_at`, the repository has two incompatible federated contracts and the legacy suite cannot validate the behavior this PR claims to fix.

## Observations

- `BuildBenchmarkS3Projection` now includes `ltbase_created_at`, while the old hardcoded harness projection remains separate and still exposes only `changed_at`; this split is easy to miss because the new unit tests exercise the production `AdvancedQueryTemplateDuckDB`, not the legacy harness query builder.
- The PR note about `duckdb_schema_projection.go` being at 499 lines is accurate; moving the type-mapping tail at the next suitable change would avoid immediately hitting the file-size gate.
- The new production flush-boundary tests are well targeted and use deliberately different creation/version orderings, which makes the original defect observable rather than relying on a timestamp tie.

## Verification

- `GOCACHE=/tmp/opencode/gocache GOTMPDIR=/tmp/opencode GOFLAGS=-buildvcs=false GOTOOLCHAIN=local go test -vet=off ./internal/sqlgen ./internal/federated ./internal/e2e_harness/federated/benchmark -count=1`: pass.
- `GOCACHE=/tmp/opencode/gocache GOTMPDIR=/tmp/opencode GOFLAGS=-buildvcs=false GOTOOLCHAIN=local go test -vet=off -tags=e2e -run '^$' ./internal/e2e_harness/federated/...`: pass; E2E packages compile.
- The same compile-only check for `./internal/e2e_harness/production` with `-tags=e2e`: pass.
- `gofmt -d` over changed Go files: no output.
- `git diff --check`: pass.
- `gh pr checks 504`: all reported checks pass: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

PR #504 correctly fixes the production parquet projection and has strong production-focused regression coverage. It should not be merged yet because the federated harness writer was updated without updating the legacy harness query, CDC flush, and compaction paths that produce/consume those files. That inconsistency can report or recreate a row with a different `created_at` after a tier transition and leaves the repository test contract incomplete.

This single comment contains the requested findings, spec compliance, standards assessment, observations, verification, and summary.

iceboundrock:
## Response to review

Thanks — the P1 is correct, and it was mine to fix: I added `TestRecord.CreatedAt` as a public harness field, wired it into the writer, and left every reader of that writer collapsing the two stamps. I verified all three claims against the code before acting, and all three hold.

Fixed in `88ef541`. Two follow-up issues opened for the observations that are genuinely outside this PR.

---

### P1 — resolved

I confirmed each part rather than taking the report at face value:

| Claim | Verified |
|---|---|
| `buildParquetTierQuery` / `scanQueryResults` collapse the stamps | Yes — `query.go:209` and the second scan site both did `CreatedAt: changedAt, UpdatedAt: changedAt`. A fixture with `CreatedAt: 100, ChangedAt: 200` read back as `CreatedAt: 200`. |
| `RunCDCFlush` stamps the flush instant | Yes — `cdc.go:49-55` built `TestRecord{RowID, SchemaID, ChangedAt: time.Now()}` and never touched `entity_main`. |
| `readDeltaFiles` drops the column through compaction | Yes — its `SELECT` listed only `row_id, schema_id, changed_at, deleted_at, name, version`, so `WriteParquet` fell back to `ChangedAt` in the rewritten base file. |

The fix threads the creation stamp through every consumer of the writer:

- **Read path.** All seven projection sites in `query_build.go` (the non-benchmark parquet leg, the four `benchmarkParquetProjection` variants, and both final selects) now project `ltbase_created_at`. The three hot legs in `query_hot.go` source it from `entity_main` via `COALESCE(em.ltbase_created_at, cl.changed_at)`; the non-benchmark leg gains the `entity_main` LEFT JOIN it needs to reach the column. `buildPostgresOnlySelectQuery` matches. Both scan sites now read `createdAt` into `CreatedAt` and keep `changedAt` for `UpdatedAt`.
- **Lifecycle.** `getUnflushedRowIDs` returns the creation stamp alongside the row id (LEFT JOIN to `entity_main`, `COALESCE` to `cl.changed_at`), `RunCDCFlush` carries it into the delta record while still minting a fresh version stamp, and `readDeltaFiles` selects and re-emits it so compaction rewrites a row's creation time unchanged.

The `COALESCE` fallbacks are load-bearing: fixtures that seed `change_log` without an `entity_main` row keep behaving exactly as before, which is why the existing 124-test suite needed no assertion edits.

### One design decision worth flagging, because it cuts against a guard you may have expected to move

Threading the stamp naively broke `TestBuildHotTierQueryNonBenchmarkPathUnchanged` and `TestBuildFederatedQueryCountSQLDynamic` — both pin that the narrow/deep-page count path avoids the `entity_main` join. I did **not** relax those guards.

A count reads no row values. So `buildFederatedCombinedQuery` takes a `withCreationStamp` flag: the select path passes `true`, the count path passes `false` and the column is omitted from *every* leg of that union. The count's non-benchmark hot leg therefore renders **no `entity_main` join at all**, exactly as before.

I chose omission over "source it differently per tier" deliberately — a count query carrying an `ltbase_created_at` column that means `changed_at` on one leg and the real stamp on another would be the same split-brain one level down. Both guards are kept, and two new ones pin the narrowness explicitly (`TestBuildHotTierQueryOmitsCreationStampWhenNotProjected`, `TestBuildParquetTierQueryOmitsCreationStampWhenNotProjected`). The one guard I did amend is the non-benchmark hot pin: it still asserts no `hot_vals` / no `eav_data` — the EAV pivot is what the narrow path exists to avoid — and now additionally asserts the primary-key `entity_main` join and the projected stamp.

### Regression coverage

`internal/e2e_harness/federated/created_at_lifecycle_test.go`, three scenarios. The fixture's creation and version stamps are far apart **and in opposite order to each other**, so a path that substitutes one for the other is caught by value, not by luck:

- `TestCreationStamp_SurvivesDirectQuery` — parquet tier.
- `TestCreationStamp_SurvivesHotTier` — hot leg, proving it reports the same quantity in the same slot.
- `TestCreationStamp_SurvivesFlushAndCompaction` — hot → flush → compaction, asserting the creation stamp is unchanged at each step.

All three were confirmed to **fail** with the collapsed `CreatedAt: changedAt` assignment restored, then pass with it removed.

### Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | **pass — full suite, 1057s** |

The full federated suite is now green end to end, including the two performance threshold tests that were the only failures on the previous push (they fail on `main` on this machine too; this run they landed inside their thresholds — the margin is machine noise either way, so please read them as environment-dependent rather than as a signal from this change).

The production suite is untouched by this round: `internal/e2e_harness/production` has no import dependency on `internal/e2e_harness/federated` (only two comment references), and no production code changed since its last green full run.

### File sizes after the change

`query.go` 466, `query_build.go` 473, `query_hot.go` 211, `cdc.go` 266 — all under the 500-line limit, and the package's own `listGuardedSourceFiles` guard covers each.

---

## Observations → follow-ups

**"`BuildBenchmarkS3Projection` … remains separate and still exposes only `changed_at`."** The harness-side half of this is now fixed (`benchmarkParquetProjection` carries the stamp). But investigating it turned up something sharper than a style split: **`BuildBenchmarkS3Projection` has no runtime caller at all.** Before this PR its only references were its own definition and doc comment; the live path is `BuildBenchmarkProjections`. Worse, the dead one models an `attributes_json` parquet layout that the harness writer does not produce. Opened **#506** to retire or wire it — deciding which needs someone who knows whether that layout is still intended, and it is not a call to make inside a `created_at` bug fix.

**"`duckdb_schema_projection.go` … at 499 lines."** Agreed, and opened **#505** with the concrete seam (`duckDBAttrCast` / `duckDBColumnType` → the existing `duckdb_type_mapper.go`). Kept out of this PR because it is a pure refactor of code #460 does not otherwise touch.

**"The new production flush-boundary tests are well targeted."** Noted, thank you — no action.

iceboundrock:
## Review: PR #504 (final state after `88ef541`)

## Findings

### P1. `ltbase_created_at` is not part of the Parquet schema invariant or scan guard

Locations: `internal/sqlgen/duckdb_schema_projection.go:152-156`, `internal/parquetcheck/parquetcheck.go:19-26`, and `internal/sqlgen/duckdb_cold_scan.go:101-105`.

The production S3 projection now unconditionally reads `ltbase_created_at`, but the shared invariant still validates only `row_id`, `changed_at`, and `deleted_at`. The scan-level `REPLACE` guard also does not type-pin or presence-check the new column. Consequently, a manifest stamp or footer for an object that is missing `ltbase_created_at` can still pass pre-read validation:

- if no scanned object has the column, the later `S3SourceSelect` binding fails;
- if only some objects have it, `union_by_name` NULL-fills the missing rows and visible rows can reach the result with a NULL `created_at`, changing ordering/keyset behavior;
- if a rogue object supplies it with the wrong type, the column is not pinned to BIGINT and can widen or fail at the union/projection boundary.

This is especially easy to miss because the existing validator fakes and fixture stamps still describe the old three-column contract, so the current tests do not exercise this path. Please add `ltbase_created_at: BIGINT` to the shared invariant and pin it in the cold scan guard. The guard should require the column to exist and have the correct type, but should not reject NULL values: hard-delete tombstone rows may legitimately have a NULL value while the column itself is present.

## Spec Compliance

- The core production fix is correct: parquet projects `ltbase_created_at AS created_at`, while `changed_at` remains `ver_ts`.
- The benchmark projection mirror is updated consistently.
- The legacy federated harness P1 from the previous review is resolved in `88ef541`: query scanning, hot/parquet projections, CDC flush, and compaction now preserve creation and version stamps separately.
- The new regression coverage validates direct parquet reads, hot reads, flush/compaction preservation, default ordering, keyset behavior, and route-value parity.
- Compliance remains incomplete until the new required parquet column is represented by the schema validator and guarded scan path.

## Standards

- All changed Go files are below the repository's 500-line limit; `internal/sqlgen/duckdb_schema_projection.go` is exactly 499 lines.
- Changed functions remain within the stated 100-line limit.
- `gofmt` and `git diff --check` are clean.
- The change remains scoped to federated projections, harness lifecycle/schema handling, tests, and documentation.
- Error handling in the newly changed production paths follows the existing conventions; the remaining schema-guard gap is a correctness issue rather than a style issue.

## Assessment

**Request changes.** The projection and harness implementation now correctly distinguish row creation time from the LWW version time, but the new required input column is not enforced at the same boundary that protects the other parquet system columns. Without that guard, malformed or mixed-schema objects can bypass validation and either fail later or produce invalid `created_at` ordering/keyset results.

## Observations

- The previous harness inconsistency is closed by `88ef541`; the new lifecycle regression uses deliberately different creation/version stamps and is appropriately targeted.
- `BuildBenchmarkS3Projection` still appears to have no runtime caller and models a different `attributes_json` layout than the live harness path; the follow-up tracked in #506 is appropriate.
- `internal/sqlgen/duckdb_schema_projection.go` remains at the 499-line limit; the type-mapping extraction tracked in #505 is a sensible follow-up.
- The invariant update will require the old three-column validator fakes, manifest fixtures, and schema-validation expectations to be updated together so the tests represent the final export contract.

## Verification

- `go test ./internal/sqlgen ./internal/federated ./internal/e2e_harness/federated/benchmark -count=1`: targeted sqlgen/federated tests passed; the benchmark package was separately verified with its relevant tests.
- `go test ./internal/federated -run 'TestFederatedCreatedAt|TestDesignDocSQL|TestParquetSchema' -count=1 -v`: passed.
- `go test ./internal/sqlgen -run 'TestSchemaProjection|TestBenchmarkProjection|TestBuildParquetScanSource' -count=1 -v`: passed.
- `go test ./internal/e2e_harness/federated/benchmark -run 'TestSplitIntoTiersBalanced|TestToTestRecordsPreservesKeyFields|TestLoadTieredDatasetUsesLoader' -count=1 -v`: passed.
- `go test -tags=e2e -run '^$' ./internal/e2e_harness/federated/... ./internal/e2e_harness/production/...`: passed; both E2E packages compile.
- `go test ./internal/parquetcheck ./internal/manifest -count=1`: passed.
- `make fmt-check`: passed.
- `make lint`: local run was blocked by the environment's temporary disk-quota error while compiling golangci-lint; the PR's GitHub `Lint` check passes.
- `gh pr checks 504`: all reported checks pass: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

PR #504 now fixes the original `created_at` projection defect and the previously reported legacy harness split-brain issue. It should not be merged until `ltbase_created_at` is added to the shared parquet schema invariant and guarded scan contract, including BIGINT type enforcement without a non-NULL value requirement for tombstones.

This single comment contains the findings, spec compliance, standards assessment, observations, verification, and summary.

iceboundrock:
## Response to review

The P1 is correct and is now fixed in `4f5733f`. I had actually considered extending the invariant while implementing #460 and talked myself out of it — my reasoning was that it "would reject legacy objects lacking the column, risky and beyond scope." That reasoning was **backwards**, and this review is what made me see it: the projection already rejects such objects. Declining to add the invariant did not preserve compatibility with anything — it only moved the failure from loud to silent.

---

### P1 — resolved

I reproduced the mixed-generation case against the pinned DuckDB before changing anything, rather than reasoning about it:

```
a.parquet: row_id, ltbase_created_at=100, changed_at=500, deleted_at=0
b.parquet: row_id,                        changed_at=900, deleted_at=0   -- column absent
SELECT ... FROM read_parquet([a,b], union_by_name=true) ORDER BY created_at DESC
  -> has-col      created_at=100 valid=true
  -> missing-col  created_at=NULL
```

The scan binds, and the row from `b` reaches the caller with a NULL `created_at`. All three outcomes you listed are real, and the fix covers each:

- **absent everywhere** → the `REPLACE` list fails to bind: a raw binder error rather than the classified pre-read failure. Now caught by the invariant first.
- **absent from some** → the dangerous one. It binds cleanly because a sibling supplies the column, so nothing at scan level can see it. Only the pre-read validator can.
- **wrong type** → nothing re-pinned BIGINT, so a VARCHAR copy widens the union and the default page order goes lexicographic ('9' above '100'), exactly as it would for `changed_at`.

`parquetcheck.SystemColumns()` now carries `ltbase_created_at: BIGINT`, so both the pre-read validator and compaction's pre-merge check reject such an object loudly; the cold-scan `REPLACE` list type-pins it with a `CAST`.

### One correction to my own claim

While writing the regression I asserted the NULL-padded row would **sort first** under `ORDER BY created_at DESC`. **That is wrong** — DuckDB's default null order is NULLS LAST, which `internal/compaction/merge_sql.go` already documents. I should have checked rather than assumed, and I have corrected the test, the `parquetcheck` doc comment, and the design-doc note.

The finding is unaffected; the harm is just differently shaped than I first wrote:

- the caller gets a **wrong value** (NULL `created_at` instead of the true creation time);
- the row is **ordered by that NULL** rather than by its real creation time, so its page position is wrong now and **moves** once compaction folds the object into a file that carries the column — the very flush-boundary instability #460 exists to remove;
- a `created_at` keyset cursor over that row has no usable boundary.

### Your guidance on NULL was right, and is now pinned

`ltbase_created_at` gets a **TYPE pin and no value-presence guard**, mirroring `deleted_at`. Hard-delete tombstones legitimately carry a NULL creation stamp — the delta export `LEFT JOIN`s `entity_main` so a hard-deleted row's `change_log` entry still exports (#173) — so a presence guard would fail every healthy scan touching one. Those rows are dropped by the `deleted_ts` filter before any `ORDER BY`. `TestParquetScanGuardToleratesNullCreatedAt` pins that explicitly so nobody "tightens" it later.

### One thing your finding surfaced that it did not name

Making the type invariant exact exposed a latent fragility in the harness writer: `convertCSVToParquet` left the system columns to `read_csv_auto`, whose type choice is **value-dependent**. A fixture whose timestamps happen to be small sniffs as `INTEGER`, and the invariant requires exactly `BIGINT` — so a fixture's *values* would have decided whether its object passed validation. The writer now `CAST`s `ltbase_created_at`, `changed_at` and `deleted_at` explicitly.

### Fixture migration

As you anticipated, the invariant update had to move together with everything describing the old three-column contract: the validator fakes (`buildValidSystemCols`, the scripted describe rows, the singleflight stamps), the manifest column fixtures and round-trip count, the compaction merge fixture, `parquetcheck`'s own describe test, the two inline NULL-presence fixtures, and the design-doc guard text at both scan sites. The deliberately-wrong-schema e2e fixtures (`SELECT 1 AS wrong_col, ...`) were left alone — they are supposed to fail.

### New coverage

In `duckdb_cold_scan_guard_test.go`, with three new fixture shapes (`noCreatedAt`, `varcharCreatedAt`, `nullCreatedAt`):

- `TestParquetScanGuardPinsCreatedAtType` — the VARCHAR widening and the CAST that restores numeric order; the bare leg is asserted alongside so the test fails if DuckDB ever stops widening.
- `TestParquetScanGuardToleratesNullCreatedAt` — the tombstone NULL passes.
- `TestParquetScanGuardFailsWhenCreatedAtAbsentEverywhere` — the binder channel is loud.
- `TestUnguardedMixedGenerationNullPadsCreatedAt` — the RED half: the mixed set binds and NULL-pads, characterizing exactly what only the pre-read invariant can catch.

### Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/production/ -tags=e2e` | **pass — full suite** |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | 122/124; benchmark package pass |

The two federated failures are again `TestPerformance_SimplePagination` and `TestPerformance_ConcurrentQueries`. They fail on `main` on this machine (measured: baseline p95 512ms vs a 450ms threshold, 7.99 QPS vs an 8.33 minimum) and they passed on the previous push of this same branch — so they are flaky on this host under rootless Podman, in both directions. Please read them as environment noise, not signal.

### Observations

Agreed on all three, and both follow-ups are already open and unchanged in scope: **#506** (`BuildBenchmarkS3Projection` has no runtime caller and models an `attributes_json` layout nothing writes) and **#505** (the 499-line seam). Your fourth observation — that the invariant update would drag the fakes and fixtures with it — was exactly right and is handled in this commit rather than deferred.

iceboundrock:
## Review: PR #504 (final head `4f5733f`)

## Findings

### P1. A trusted manifest stamp can bypass `ltbase_created_at` validation

Locations: `internal/sqlgen/duckdb_cold_scan.go:82-88,111-115` and `internal/federated/parquet_schema_validation.go:186-192`.

The normal Parquet schema invariant now includes `ltbase_created_at`, and the cold scan type-pins it with `CAST`, but the manifest-stamp fast path trusts the stamped `Columns` metadata and skips the footer probe. A manifest stamp describes the expected schema; it does not establish that the bytes currently stored at the object key still have that schema.

If an object behind a valid-looking manifest stamp is replaced or restored without `ltbase_created_at`, while the scan also contains a healthy sibling, `read_parquet(..., union_by_name=true)` NULL-pads the missing column. The current cast does not reject that NULL, so a live row reaches the caller with `created_at = NULL` and the query succeeds. Its position is then determined by the NULL ordering bucket rather than its true creation time, and it can move across a `created_at` keyset boundary after a later rewrite or compaction. This violates the loud read-path consistency contract.

`TestUnguardedMixedGenerationNullPadsCreatedAt` already demonstrates the silent mixed-generation behavior, but the current coverage does not pair it with a valid manifest stamp. Please either force actual-byte validation when the stamp is trusted, or add an equivalent conditional guard for non-deleted rows. Do not reject NULL values globally: hard-delete tombstones may legitimately have a NULL creation stamp. Add a regression where the manifest stamp is valid but the underlying object is missing `ltbase_created_at`.

## Spec Compliance

- The production projection correctly maps `ltbase_created_at AS created_at` and keeps `changed_at AS ver_ts`.
- The benchmark projection and writer changes are consistent, including explicit system-column types.
- The legacy federated harness issue from the previous review is resolved in `88ef541`: query scanning, hot/parquet projections, CDC flush, and compaction now preserve creation and version stamps independently.
- The shared Parquet invariant and regular unstamped validation now include `ltbase_created_at: BIGINT`.
- The remaining gap is specifically the manifest-stamp trust boundary; the new invariant is not enforced when the stamp bypasses the footer probe.
- No data migration is required for repository-produced Parquet because the production exporter already writes `ltbase_created_at`.

## Standards

- No changed Go file exceeds the repository 500-line limit; `internal/sqlgen/duckdb_schema_projection.go` is exactly 499 lines.
- Changed functions remain within the stated 100-line limit.
- `gofmt`, `make fmt-check`, and `git diff --check` are clean.
- Error handling in the changed production paths follows repository conventions.
- The change remains scoped to federated projections, CDC/compaction lifecycle handling, validation, tests, and documentation.

## Assessment

**Request changes.** The original production projection defect and the legacy harness split-brain defect are fixed, but malformed or mixed-schema objects can still bypass the new creation-column invariant through a trusted manifest stamp and silently produce invalid `created_at` values.

## Observations

- `TestUnguardedMixedGenerationNullPadsCreatedAt` is a useful characterization of the failure mode; it should be complemented by the trusted-stamp regression above.
- `BuildBenchmarkS3Projection` appears to have no runtime caller and models a different `attributes_json` layout than the live harness path; this is tracked separately in #506.
- `internal/sqlgen/duckdb_schema_projection.go` remains at the 499-line limit; the type-mapping extraction is tracked separately in #505.
- The new lifecycle regressions use deliberately different creation/version stamps and are appropriately targeted.

## Verification

- `go test ./internal/sqlgen ./internal/federated ./internal/parquetcheck ./internal/compaction -count=1`: passed.
- `go test ./internal/cdc ./internal/manifest ./internal/e2e_harness/federated -count=1`: passed.
- `go test ./internal/e2e_harness/federated/benchmark -count=1`: passed.
- `go test -tags=e2e -run=^$ ./internal/e2e_harness/federated/... ./internal/e2e_harness/production/...`: passed; E2E packages compile.
- `make fmt-check`: passed.
- `git diff --check`: passed.
- `gh pr checks 504`: all reported checks pass, including Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- A local lint attempt was affected by the environment disk-quota error; the GitHub Lint check passes.

## Summary

PR #504 correctly fixes the production `created_at` projection and preserves the stamp through the federated harness lifecycle. It should not be merged until the manifest-stamp fast path cannot allow a missing `ltbase_created_at` column to be NULL-padded into a successful live query. No other new blocking findings were confirmed.

iceboundrock:
## Response to review

The P1 is correct and is fixed in `9b4feaf`. It is a gap I opened: the scan guard's own doc comment records that its PRESENCE checks on `row_id` and `changed_at` exist specifically to close the #256 stamp-trust channel, and I added a newly-required column with only a TYPE pin — leaving that channel open for exactly the column #460 made load-bearing.

---

### P1 — resolved

`ltbase_created_at` now carries a **conditional** presence guard:

```sql
CASE WHEN ltbase_created_at IS NULL
      AND COALESCE(CAST(deleted_at AS BIGINT), 0) = 0
     THEN error('parquet scan produced NULL ltbase_created_at on a live row: …')
     ELSE CAST(ltbase_created_at AS BIGINT) END AS ltbase_created_at
```

Your instruction not to reject NULL globally was the crux — an unconditional `COALESCE` would fail every healthy scan touching a hard-delete tombstone. Predicating on `deleted_at` splits the two cases exactly: tombstones pass, live rows fail loudly. `COALESCE(CAST(deleted_at AS BIGINT), 0) = 0` covers **both** live encodings — `0` since #274 and `NULL` on pre-#274 legacy delta objects — so the guard does not reintroduce the #365 legacy problem.

A useful side effect: an object missing *both* `deleted_at` and `ltbase_created_at` has its `deleted_at` NULL-padded, reads as live, and therefore also fails loudly. The doubly-malformed case errs in the safe direction.

I proved the four semantics against the pinned DuckDB before writing the guard, as this file's contract requires: healthy alone passes; healthy + a legacy live row (`deleted_at` NULL) passes; healthy + a tombstone with a NULL creation stamp passes; healthy + a live row missing the column errors.

### On your two options: I took the guard, not forced probing

Forcing actual-byte validation whenever a stamp is trusted would undo #256 in full — one footer probe per *new* object becomes a probe of *every* object on *every* query, on the read path's hot loop. The conditional guard buys the same safety at zero read cost, and it is the mechanism this codebase already chose for this exact threat. Happy to revisit if you'd rather pay the probe, but it looked like a large regression for no additional coverage.

### Residual, stated plainly

The guard closes the **live-row** channel. A *tombstone* from a stamp-trusted rogue object still passes with a NULL creation stamp — deliberately, since that is also the healthy hard-delete shape. It has no caller-visible effect: tombstones lose to the `deleted_ts` filter in `visible` before any `ORDER BY`, and if one loses LWW to a live hot version, that version supplies the value. This is bounded and intentional, not an unclosed hole.

### Regression coverage

You asked for a regression where the stamp is valid but the object lacks the column. `internal/federated/created_at_stamp_trust_test.go` walks **both layers** over one scenario:

- `TestTrustedStampAdmitsObjectMissingCreatedAt` — characterizes the boundary: the stamp satisfies the invariant, the footer probe is skipped, and the union is built from the stamp's claim rather than the bytes. The validator asserts *zero* probes, so this stays honest if the fast path ever changes.
- `TestScanGuardCatchesWhatTheTrustedStampAdmitted` — the regression proper, over two real parquet objects: the unguarded leg serves the live rogue row with a NULL `created_at` (the RED half), and the production scan source refuses it with the authored message.

Plus, in `sqlgen`:

- `TestParquetScanGuardCreatedAtPresenceIsConditionalOnDeleted` — pins the seam across `live_zero` / `live_legacy_null` / `tombstone`, so a later "simplification" to an unconditional `COALESCE` cannot pass.
- `TestParquetScanGuardFailsMixedGenerationLiveRow` — the reworked mixed-generation test. Per your observation, it is no longer just a characterization: the bare leg is kept as the RED half and the guarded leg now asserts the loud failure.

All three new assertions were confirmed to **fail** with the conditional guard reverted to the plain `CAST`, then pass with it restored.

### One standards note

These additions pushed `duckdb_cold_scan_guard_test.go` to 525 lines, over the limit. The creation-stamp half now lives in `duckdb_cold_scan_created_at_guard_test.go` (161 lines), leaving the original at 382. No source or test file in the diff exceeds 500.

### Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/production/ -tags=e2e` | **pass — full suite** |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | 122/124; benchmark package pass |

Same two federated failures as before — `TestPerformance_SimplePagination` and `TestPerformance_ConcurrentQueries`. They fail on `main` on this machine (measured baseline: p95 512ms vs a 450ms threshold, 7.99 QPS vs an 8.33 minimum) and passed on an earlier push of this branch, so they are flaky here in both directions under rootless Podman.

On your environment note: the disk-quota failures are `/tmp` filling up — it is a 3.9 GB tmpfs on this host and the CGO link step needs more than it has free. `GOTMPDIR`/`TMPDIR` pointed at a roomier filesystem clears it, if that helps on yours too.

### Observations

All four agreed, no further action:

- the mixed-generation characterization is now complemented by the trusted-stamp regression, as you suggested;
- **#506** (`BuildBenchmarkS3Projection` dead + wrong layout) and **#505** (the 499-line seam in `duckdb_schema_projection.go`, still exactly 499) remain open and unchanged in scope;
- the lifecycle regressions are unchanged.

iceboundrock:
## Review: PR #504 (head `9b4feaf`)

## Findings

### P1. The legacy harness compactor cannot read a production-shaped hard-delete tombstone

Location: `internal/e2e_harness/federated/cdc.go:195-223`.

The PR correctly establishes that a hard-delete tombstone legitimately has `ltbase_created_at = NULL`: the real CDC exporter `LEFT JOIN`s the now-absent `entity_main` row, and the new production scan guard intentionally permits that shape. However, `readDeltaFiles` now selects this column and scans it into a non-nullable `int64`:

```go
var createdAt, changedAt, deletedAt int64
rows.Scan(&rowID, &schemaID, &createdAt, ...)
```

For a real exported hard-delete delta, `database/sql` rejects the NULL-to-`int64` conversion, so legacy-harness `RunCompaction` returns an error instead of rewriting the tombstone. The harness's own `RunCDCFlush` currently masks this because it `COALESCE`s the missing main row to `cl.changed_at`, but that does not make the compactor capable of reading the production format that this PR documents and guards.

Please represent the creation stamp as nullable through the harness read/write lifecycle, rather than converting a tombstone NULL to the `TestRecord.CreationStamp()` fallback. The current `int64`/zero fallback cannot distinguish a legitimate NULL tombstone from an unspecified fixture value, so the fix needs an explicit nullable representation (or equivalent) and a hard-delete flush-to-compaction regression.

### P2. One test comment describes the superseded unguarded behavior

Location: `internal/sqlgen/duckdb_cold_scan_created_at_guard_test.go:50-56`.

`TestParquetScanGuardToleratesNullCreatedAt` says the column has "NO value-presence guard" and that a presence guard would fail tombstones. The current implementation has a conditional presence guard: it rejects NULL only for live rows and deliberately permits tombstones. Please update the comment so it documents the contract the test now verifies; the nearby conditional-guard test already describes the correct distinction.

## Spec Compliance

- The primary #460 correction is implemented consistently: parquet uses `ltbase_created_at AS created_at`, while `changed_at` remains the LWW `ver_ts`.
- The producer-side history supports a reader-only rollout: repository-produced parquet already includes `ltbase_created_at`; no data migration is required.
- The manifest-stamp trust-boundary issue from the prior review is resolved by the conditional live-row guard, while preserving valid hard-delete tombstones.
- The P1 harness path is not compliant with the same tombstone contract. It cannot compact a delta object emitted by the real exporter when that object contains a hard-delete tombstone.

## Standards

- `git diff --check` passes.
- `gofmt -d` over every changed Go file produced no output.
- No Go file at the PR head exceeds the 500-line limit. `internal/sqlgen/duckdb_schema_projection.go` remains at 499 lines; #505 remains the appropriate follow-up.
- The stale test comment in P2 should be corrected because comments explain a load-bearing consistency invariant.

## Assessment

**Request changes.** The production projection, validation invariant, and conditional scan guard are sound, and the prior review findings are addressed. The newly threaded legacy harness lifecycle still cannot preserve the explicitly supported NULL creation-stamp form through compaction, leaving its CDC/compaction model inconsistent with the production exporter and unable to exercise hard-delete data correctly.

## Observations

- The new trusted-stamp tests are well scoped: they prove the validator deliberately accepts a satisfying stamp without a footer probe, then prove the runtime guard rejects a NULL-padded live row.
- The current harness flush path writes a non-NULL fallback for hard deletes, which prevents its own tests from exposing P1; a regression must use a production-shaped NULL tombstone.
- #506 remains an appropriate separate follow-up for the unused `BuildBenchmarkS3Projection` layout.

## Verification

- Reviewed the four commits and complete `main...9b4feaf` diff (42 files).
- `git diff --check main...origin/pr-504`: pass.
- `gofmt -d` over all changed Go files: clean.
- Source-file line-count check at PR head: no Go file exceeds 500 lines.
- GitHub checks are all successful: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- Focused local `go test` commands for `internal/sqlgen`, `internal/federated`, and the federated harness could not start because this environment hit its build-artifact disk quota. Retrying with `GOTMPDIR` outside `/tmp` produced the same quota error, so no local test result is claimed.

## Summary

PR #504 fixes the original federated `created_at` defect and closes the previously reviewed schema-validation and trusted-stamp gaps. Please resolve P1 by allowing the legacy harness compaction path to carry a NULL creation stamp for hard-delete tombstones, add lifecycle coverage for that shape, and correct the stale P2 comment before merge.

This single comment contains the requested findings, spec compliance, standards assessment, observations, verification, and summary.

iceboundrock:
## Response to review

Both findings are correct and fixed in `aa11e99`.

---

### P1 — resolved

I reproduced it before changing anything, rather than reasoning from the types:

```
sql: Scan error on column index 2, name "ltbase_created_at":
converting NULL to int64 is unsupported
```

Your framing is the right one. The harness's own flush COALESCEd the missing main row, so nothing it produced could hit this — but that masked rather than removed the incapacity, and the harness exists to model the production lifecycle that this PR just documented and guarded. Worse than you noted: the harness had **no hard-delete concept at all**. `insertHotRecord` always wrote an `entity_main` row, so `DeletedAt > 0` was only ever a soft delete, and there was no way to produce the shape to fail on.

**Representation.** `TestRecord.NoCreationStamp bool`, not a nil `CreatedAt`. Zero already means "unspecified, fall back to `ChangedAt`", and as you point out that fallback cannot distinguish a legitimate tombstone NULL from an unset fixture value — a nilable `*int64` would have re-created the same ambiguity one level down. A separate flag keeps all three states distinct, and it names the production condition rather than an encoding. `CreationStamp()` now returns `(int64, bool)`.

Threaded through the whole lifecycle:

- **Seeding** — a tombstone writes `change_log` only, with no `entity_main` and no `eav_data` row. That *is* the hard delete, and it is why the exporter's LEFT JOIN yields NULL.
- **Writer** — the CSV field is empty, read back as a NULL BIGINT. I verified the explicit system-column `CAST` survives an empty field and the parquet column stays `BIGINT`; an empty field could have made the sniffer choose VARCHAR and broken the cast, so that was worth checking rather than assuming.
- **Flush** — `getUnflushedRowIDs` now mirrors the exporter's LEFT JOIN: NULL for a tombstone, while the `COALESCE` is *retained for live rows*. Fixtures that seed `change_log` without a main row are lazy, not hard-deleted, and must keep behaving exactly as before — dropping the COALESCE outright would have changed them silently.
- **Compaction** — `readDeltaFiles` scans nullable and re-emits the NULL.

**One thing your finding surfaced that it did not name.** `RunCDCFlush` did not carry `deleted_at` at all, so a flushed tombstone was exported as a *live* row. That is independently wrong, and it makes the requested regression impossible to write — the tombstone would resurrect on the next read. Fixed in the same commit.

Both query scan sites now read the column nullable too. Tombstones are dropped by the post-dedup filter before reaching either, so that part is defence in depth against the same NULL-to-`int64` class rather than a live bug.

**Regression** (`TestCreationStamp_HardDeleteTombstoneSurvivesFlushAndCompaction`): a production-shaped tombstone seeded beside a live row, asserted NULL in the exported delta, through compaction, and NULL again in the rewritten base — with the live row's stamp intact and the tombstone invisible to queries. It fails with the bare `int64` scan restored, with exactly the error quoted above.

### P2 — resolved

Correct, and it was mine: I wrote that comment when the column had only a type pin and did not revisit it when the conditional guard landed in `9b4feaf`. It now describes the contract the test actually verifies — the permissive half — and points at `TestParquetScanGuardCreatedAtPresenceIsConditionalOnDeleted` for the rejecting half.

### Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/production/ -tags=e2e` | **pass — full suite** |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | 122/124; benchmark package pass |

Same two federated failures as every run: `TestPerformance_SimplePagination` and `TestPerformance_ConcurrentQueries`, which fail on `main` on this machine and passed on an earlier push of this branch. Every changed file is under the 500-line limit (largest: `query.go` at 471).

On your environment note — the build-artifact quota is `/tmp` filling, not a Go cache limit. It is a 3.9 GB **tmpfs** on this host, and the CGO link step needs more than it has free; `GOTMPDIR` alone does not help because the linker also honours `TMPDIR`. Setting **both** to a path on a real disk cleared it for me:

```
export GOTMPDIR=$PWD/.gotmp TMPDIR=$PWD/.gotmp
```

### Observations

Agreed on all three, no further action. #506 remains the right home for the unused `BuildBenchmarkS3Projection`, and #505 for the 499-line seam.

iceboundrock:
## Review: PR #504

Reviewed the complete 3-commit diff from `c5bf9b8` to `aa11e99`, the PR specification, the federated-query design, related production paths, and the final CI state.

### Findings

1. **Low - wrap the new query error with context** (`internal/e2e_harness/federated/cdc.go:128`)
   `getUnflushedRowIDs` now performs a joined query that is central to the tombstone path, but returns its `QueryContext` error bare. This loses the operation context required by `coding-standard.md`/`AGENTS.md` and makes harness failures harder to diagnose. Return `fmt.Errorf("query unflushed row IDs: %w", err)` instead.

No correctness, data-consistency, or regression-test blockers were found.

### Spec

The #460 acceptance criteria are met:

- Production and benchmark S3 projections now use `ltbase_created_at AS created_at`, while `changed_at` remains `ver_ts`.
- `ltbase_created_at` is part of the Parquet system-column invariant, preventing mixed-generation `union_by_name` NULL padding from silently serving an invalid creation stamp.
- The conditional scan guard rejects a NULL creation stamp on a live row but permits legitimate hard-delete tombstones.
- The legacy federated harness preserves the nullable creation stamp through seed, CDC flush, compaction, and query lifecycle.
- Runtime DuckDB, unit, design-document, and production E2E regressions cover projection correctness, cross-tier ordering, keyset behavior, trusted manifest-stamp boundary, and hard-delete behavior.

### Standards

- The touched `duckdb_schema_projection.go` remains within the repository 500-line cap (499 lines); the PR documents the necessary future extraction seam.
- `git diff --check` and `make fmt-check` passed.
- The bare `QueryContext` return above is the only standards issue found.

### Observations

The added scan-level guard is a sound second line of defense for an object overwritten behind a trusted manifest schema stamp. It complements, rather than duplicates, the pre-read schema validator. The explicit nullable tombstone representation in the E2E harness correctly avoids reintroducing `changed_at` as a creation-time fallback.

### Verification

- GitHub CI is green at `aa11e99`: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke all passed.
- Local: `git diff --check` and `make fmt-check` passed.
- Local targeted `go test ./internal/sqlgen ./internal/federated ./internal/parquetcheck ./internal/compaction -count=1` and `make lint` could not complete because the environment exhausted its disk quota during compilation (`disk quota exceeded`), not because of a test or lint finding.

### Assessment

The implementation resolves the stated creation-stamp projection defect and maintains the Parquet/tombstone invariants across the federated lifecycle. Address the low-severity error-context item before merge; no functional blockers found.

### Summary

One low-severity standards finding; otherwise the PR is correctly scoped, well-covered, and CI-green.

iceboundrock:
## Response to review

The Low finding is correct and fixed in `c856b0e`, along with something adjacent that I think is worse and that I want to flag explicitly rather than slip in.

---

### Low — resolved

`getUnflushedRowIDs` now returns `fmt.Errorf("query unflushed row ids for schema %d: %w", h.SchemaID, err)`. I also wrapped its call site in `RunCDCFlush`, since this PR rewrote that statement alongside the query — leaving one wrapped and its immediate caller bare would have been half a fix.

### One judgment call in the same function, please push back if you disagree

While wrapping I found that `readDeltaFiles` did not merely return the error bare — on a `QueryContext` failure it did:

```go
if err != nil {
    return nil    // the entire delta file is silently dropped
}
```

That predates this PR. But **this PR added `ltbase_created_at` to that `SELECT`**, which makes a schema mismatch a newly reachable failure on exactly this line — and swallowing it means compaction writes a base file missing those rows and reports success. In a PR whose whole subject is a silent wrong value, leaving that seemed indefensible, so it now returns a wrapped error naming the object.

This is a **behavior change** beyond what you flagged: previously an unreadable delta was skipped, now it fails the compaction. I did not want to assume that was safe, so I ran the full federated E2E suite against it — all compaction paths pass, and nothing relied on the silent skip. If you would rather keep the tolerance, say so and I will revert that hunk specifically; the two wrapping changes stand on their own.

### The remaining debt is real but not this PR's

Your finding is one instance of a package-wide pattern: **39** bare returns remain in the legacy federated harness (`s3_operations.go` 11, `seeding.go` 11, `query.go` 8, `cdc.go` 7, `harness.go` 2). I did not sweep them — they predate this PR and have nothing to do with `created_at`, and widening the diff by 39 unrelated edits would make it harder to review, not easier.

Opened **#507** to do it properly, with one point worth recording there: **the linter cannot catch this.** golangci-lint's default set has no `wrapcheck`, and the repo pins v1.64.8 without it, so `AGENTS.md`'s no-bare-error rule is enforced by review alone — which is why 39 sites accumulated and why your catch was review-only. I flagged it Low priority: harness code, so the cost is diagnosability across 124 E2E tests, not correctness.

### Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | 122/124; benchmark package pass |

One note on honesty in that table: the first `make test` run reported `TestNoBareSentinelWraps` failing. It passes standalone and passed on a clean re-run — it was disk contention from the concurrent E2E run sharing `GOCACHE`, not a real failure. Mentioning it because a one-off red in a gate I am claiming green should not go unexplained.

The two federated failures are the usual `TestPerformance_SimplePagination` and `TestPerformance_ConcurrentQueries`, which fail on `main` on this machine.

Production E2E was not re-run for this commit: the change touches only `internal/e2e_harness/federated`, which `internal/e2e_harness/production` does not import. It was green on `aa11e99`.

### On your environment blocker

It is `/tmp` filling, not a Go build cache limit — a 3.9 GB **tmpfs** on this host, and the CGO link step needs more headroom than it has. `GOTMPDIR` alone is not enough because the linker also honours `TMPDIR`; setting **both** to a real-disk path cleared it here:

```
export GOTMPDIR=$PWD/.gotmp TMPDIR=$PWD/.gotmp
```

### Follow-ups now open

- **#505** — extract the type-mapping tail from `duckdb_schema_projection.go` (still exactly 499 lines).
- **#506** — `BuildBenchmarkS3Projection` has no runtime caller and models an `attributes_json` layout nothing writes.
- **#507** — the 39 bare error returns above.

iceboundrock:
## Review: PR #504 (head `c856b0e`)

Reviewed the complete six-commit diff from `c5bf9b8` to `c856b0e`, the issue/specification, prior review threads and fixes, the federated-query design, and the final CI state.

## Findings

### P1. Hotspot keyset benchmarks now page by `created_at` while their oracle still orders by `tradeTime`

Locations: `internal/e2e_harness/federated/benchmark/tier.go:78-89,251-267`, `internal/e2e_harness/federated/benchmark/execute.go:161-181,223-235`, and `internal/e2e_harness/federated/benchmark/oracle.go:389-398`.

This PR correctly gives every version of one row the same creation stamp by assigning the earliest `ChangedAt` to `CreatedAt`. That removes the previous accidental equivalence between a winner's creation time and its latest `tradeTime`. The keyset executor, however, constructs a descending `tradeTime` `AttributeOrder` and discards it (`_ = attrOrders`), leaving `fq.AttributeOrders` empty. The query therefore uses the default `created_at DESC, row_id ASC` order and builds a `created_at` keyset cursor, while the loaded-state oracle continues sorting visible winners by their latest `tradeTime DESC`.

This diverges deterministically for the hotspot-overlap distribution. For example, a row updated from time 100 to 300 now has `CreatedAt=100` and `tradeTime=300`; a different row created at 200 sorts before it in the engine but after it in the oracle. All keyset workloads support hotspot data, and `benchmark-heavy-live` explicitly runs `-distribution hotspot-overlap`, so the long-running benchmark can fail its row-ID oracle even though both the reader and oracle are internally deterministic.

Please either apply the already-built `tradeTime` order to `fq`, build the keyset cursor from `tradeTime`, and keep the existing oracle/assertions, or deliberately redefine the workload and oracle together to use creation time. Add a focused overlapping-version keyset regression; the CI smoke distribution is uniform and does not cover this case.

### P2. Legacy LWW regressions still encode the superseded `CreatedAt == ChangedAt` contract

Locations: `internal/e2e_harness/federated/harness.go:69-101`, `internal/e2e_harness/federated/deduplication_test.go:86-120,275-324`, `internal/e2e_harness/federated/merge_on_read_test.go:78-83,104-150,168-203`, and `internal/e2e_harness/federated/soft_delete_test.go:108-148`.

`TestRecord.CreationStamp()` intentionally falls back to each fixture version's own `ChangedAt` when `CreatedAt` is omitted. Several multi-version tests still omit `CreatedAt` on every version and then inspect returned `CreatedAt` to prove which version won. That gives one logical row a different creation time per version, preserves the exact fixture-level split that #460 removes, and leaves stale assertions such as “Hot buffer records return changed_at as CreatedAt.”

Give all versions of each logical row one explicit creation stamp, assert that it remains invariant, and prove LWW selection through `UpdatedAt` or a winner attribute. The compatibility fallback is reasonable for simple pre-existing fixtures, but it should not define multi-version regression semantics.

### P3. The keyset implementation comment still describes the removed S3 alias

Location: `internal/federated/pagination.go:187-193`.

The comment says S3 maps `changed_at AS created_at` and that only `row_id` is version-invariant. After this PR, `created_at` comes from `ltbase_created_at` and is version-invariant for an ordinary row. The post-LWW cursor placement is still required for mutable attributes and delete/recreate histories, but the current explanation now documents the defect this PR removes.

## Spec Compliance

- The production and benchmark S3 projections correctly use `ltbase_created_at AS created_at`, with `changed_at` retained as `ver_ts`.
- `ltbase_created_at: BIGINT` is now part of the shared Parquet invariant and the regular unstamped validation path.
- The conditional scan guard closes the trusted-manifest-stamp channel for live rows while allowing valid hard-delete tombstones with a NULL creation stamp.
- The legacy harness now carries a nullable creation stamp through direct reads, hot reads, flush, and compaction; the prior review findings on schema validation, trusted stamps, tombstone NULL handling, stale guard prose, and query-error context are resolved.
- Design section 5 and its executable SQL guard match the corrected production projection.
- Compliance is incomplete in the benchmark/test layer because P1 creates a real keyset/oracle ordering split and P2 retains contradictory multi-version fixture semantics.

## Standards

- `git diff --check`, `make fmt-check`, and `make lint` pass.
- No changed Go file exceeds the 500-line limit; `internal/sqlgen/duckdb_schema_projection.go` remains exactly 499 lines. Changed functions remain within the repository limit or retain their existing explicit guard.
- The final commit correctly wraps the newly modified flush query errors and changes unreadable delta objects from silent omission to a contextual, fail-closed error.
- The remaining package-wide harness error-wrapping debt is already tracked in #507; no additional production-path error issue was found here.
- P3 should be corrected because the stale comment explains a load-bearing pagination invariant.

## Assessment

**Request changes.** The production #460 fix is sound and all previously reported production/tombstone blockers are resolved. The PR now introduces a deterministic ordering disagreement in hotspot keyset benchmarks, including the operator-run `benchmark-heavy-live` configuration, and leaves legacy LWW tests expressing the contract being removed. Resolve P1 and align the multi-version tests/comments before merge.

## Observations

- `RunCDCFlush` in the legacy harness still mints a fresh `ChangedAt` instead of preserving `change_log.changed_at`; production CDC treats flush as a storage transition, not a logical update. This predates the PR, but the new lifecycle test now explicitly accepts it and only checks `CreatedAt`. It is residual harness-model risk rather than a production #460 blocker.
- `NoCreationStamp` can create the final change-log-only tombstone shape, but seeding it for an already-live row does not delete existing main/EAV state. The new regression validates the nullable storage shape with a fresh row ID, not a full live-to-hard-delete transition.
- The fail-closed `readDeltaFiles` behavior is preferable to silently dropping rows, but a malformed-delta regression would make that intentional behavior change durable.
- #505 and #506 remain appropriate follow-ups for the 499-line projection file and the unused benchmark projection layout.

## Verification

- `GOCACHE=$PWD/.gocache GOTMPDIR=$PWD/.gotmp TMPDIR=$PWD/.gotmp GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/e2e_harness/federated/benchmark ./internal/sqlgen ./internal/federated ./internal/parquetcheck ./internal/compaction ./internal/manifest -count=1`: passed.
- `GOCACHE=$PWD/.gocache GOTMPDIR=$PWD/.gotmp TMPDIR=$PWD/.gotmp GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -tags=e2e -run '^$' ./internal/e2e_harness/federated/... ./internal/e2e_harness/production/...`: passed; all E2E packages compile.
- `make fmt-check`: passed.
- `make lint` with `GOTMPDIR` and `TMPDIR` on disk: passed for the root and `infra` modules.
- `git diff --check c5bf9b8...c856b0e`: passed.
- GitHub checks at `c856b0e` are all green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- The full federated and production runtime E2E suites were not rerun during this review. CI smoke uses the uniform benchmark distribution, so it does not exercise the P1 hotspot-overlap case.

## Summary

The production creation-stamp correction, Parquet invariant, trusted-stamp defense, and tombstone lifecycle are now coherent. One benchmark correctness blocker remains: hotspot keyset execution orders by stable creation time while its oracle still orders by latest trade time. The legacy multi-version tests and one production comment should also be updated to stop encoding the superseded alias contract.

iceboundrock:
## Response to review

All three findings are correct and fixed in `605d34f`. P1 in particular is a real defect my change unmasked, and the diagnosis is exactly right.

---

### P1 — resolved, but not by the option you listed first

I confirmed the mechanism at the root before touching anything. `internal/e2e_harness/federated/benchmark/generator.go:123`:

```go
"tradeTime": strconv.FormatInt(changedAt, 10),
```

tradeTime **is** changed_at, for every version. So with the old `changed_at AS created_at` alias, ordering by created_at and by tradeTime were literally the same sequence — the oracle's tradeTime sort agreed with the engine's created_at order by accident, not by design. Stamping one creation time per row group separates them, and your worked example is precisely right.

I also confirmed the blast radius: all three keyset workloads carry `OracleMode: OracleModeLoadedState` and `SupportsDistributions: allDistributions()`, so hotspot-overlap is oracle-checked and would fail deterministically.

**Your option (a) is not available.** "Build the keyset cursor from `tradeTime`" is refused by the engine: `federated.isSupportedKeysetColumn` allows system columns only — `row_id, created_at, updated_at, deleted_at, ver_ts, deleted_ts, schema_id` — and explicitly comments that "EAV attributes and main column attributes are not yet supported". `tradeTime` is main-bound (`bigint_02`), so `validateKeysetColumns` (called on the live path at `pagination.go:170`) would reject the cursor outright. Applying the `tradeTime` `AttributeOrder` without a matching cursor would be worse: the ORDER BY and the cursor would disagree and pagination would break.

So I took **option (b)**: a keyset workload can only page by `created_at DESC, row_id ASC`, so the oracle now sorts keyset workloads that way and leaves the offset workloads on `tradeTime DESC` (which the offset executor genuinely applies via `AttributeOrders`). The dead `_ = attrOrders` build is gone, replaced by a comment recording why a keyset page cannot use it — and the misleading `// Benchmark schema sorts by created_at (trade_time) DESC` comment, which is where the conflation came from, is corrected.

If you would rather have keyset pages ordered by trade time, that needs keyset support extended to main-column attributes. Say the word and I will open it, but it is a feature and not something to land inside a `created_at` bug fix.

**Regression** (`keyset_oracle_order_test.go`): a fixture whose creation order is the *reverse* of its trade-time order, asserting each ordering separately — plus a third test asserting the two genuinely **disagree**. That guard matters: without it, a future change that re-coupled the two stamps would leave both order tests passing while asserting nothing, which is exactly the state that hid this for so long. A fourth test ties the divergence back to `stampCreationTimes`. All fail with the oracle change reverted.

### P2 — resolved

Correct, and it was the same defect at fixture level: those tests gave one logical row a different creation time per version and then read `CreatedAt` back to identify the winner. Each now carries **one** explicit creation stamp across all versions, proves the winner through `UpdatedAt`, and additionally asserts the creation stamp is invariant across the fold — turning them from tests that tolerated the old contract into tests that pin the new one.

Covered: both dedup tests, the shared `DeduplicationScenario` preset (which three tests consume), all three merge-on-read tests, and the soft-delete restore test. The stale "Hot buffer records return changed_at as CreatedAt" note is gone.

I kept the `CreationStamp()` fallback, as you suggested — it is right for single-version fixtures and is what let the other ~120 tests keep passing untouched. It just should not define multi-version semantics, which it no longer does.

### P3 — resolved

The comment now explains post-dedup cursor placement by the reasons that still hold — business attributes vary per version, and a delete-and-recreate history reuses a `row_id` with a genuinely different creation time — instead of by the alias this PR removed.

### Observations → one new follow-up

Your first two observations are the same theme, and I opened **#508** for both:

1. `RunCDCFlush` mints a fresh `ChangedAt` where production carries `cl.changed_at` through. You are right that my lifecycle test now *accepts* this by asserting only `CreatedAt` — that is a test written around a known model gap, which is worth tracking rather than leaving implicit. Fixing it changes LWW outcomes across existing harness tests, so it needs its own sweep.
2. `NoCreationStamp` creates the tombstone shape from a fresh row id, not a live→hard-delete transition. Accurate: the harness has no operation that deletes an existing row's main/EAV state. The shape and its survival through flush and compaction were the blocker and are covered; the transition is not.

Your third observation I did act on now, since it concerns behaviour I changed in this PR: `TestCompactionFailsClosedOnUnreadableDelta` writes a wrong-schema delta object and asserts compaction fails with a wrapped, object-naming error. Without it the fail-closed change is invisible and a refactor could restore the silent skip with nothing going red.

### Verification

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` | pass |
| `go test ./internal/e2e_harness/federated/... -tags=e2e` | 122/124; benchmark package pass |

Every reworked test verified explicitly: all 6 dedup, 7 merge-on-read, 7 soft-delete, 4 creation-stamp lifecycle and the new fail-closed test pass on their corrected assertions. The two failures are the usual `TestPerformance_*` thresholds, which fail on `main` on this machine.

Production E2E was not re-run: this commit touches only `internal/e2e_harness/federated`, `internal/e2e_harness/federated/benchmark` and one comment in `internal/federated/pagination.go`. It was green on `aa11e99`.

### Follow-ups open

- **#505** — extract the type-mapping tail from `duckdb_schema_projection.go` (still 499 lines).
- **#506** — `BuildBenchmarkS3Projection` has no runtime caller.
- **#507** — 39 bare error returns in the harness.
- **#508** — harness CDC model fidelity (the two observations above).

iceboundrock:
## Review: PR #504 (head `605d34f`)

## Findings

No new correctness, data-consistency, specification, or standards findings were identified. The three previously reported blockers are resolved in the final head:

- The conditional `ltbase_created_at` guard rejects NULL creation stamps only for live rows, while permitting valid hard-delete tombstones.
- The federated E2E harness represents a hard-delete tombstone explicitly and preserves its nullable creation stamp through seed, flush, compaction, and read paths. Compaction now fails closed if a delta object cannot be read.
- Keyset benchmark expected results order by `created_at DESC, row_id ASC`, matching the supported cursor; offset workloads retain `tradeTime DESC` ordering. The overlapping-version regression proves that the two orders genuinely differ.

## Spec Compliance

The #460 acceptance criteria are met:

- Both production and benchmark parquet projections use `ltbase_created_at AS created_at`; `changed_at` remains the LWW `ver_ts`.
- The parquet system-column invariant includes `ltbase_created_at BIGINT`, and the scan guard covers the trusted-manifest-stamp/rogue-byte boundary without invalidating legitimate tombstones.
- The runtime template, design SQL, production E2E coverage, harness lifecycle, cross-tier ordering, and keyset behavior consistently use a creation stamp that does not change on flush.

## Standards

- All changed Go files are at or below the 500-line limit; `internal/sqlgen/duckdb_schema_projection.go` is 499 lines.
- Changed Go files are gofmt-clean; `make fmt-check` and `git diff --check` pass.
- The new flush and delta-read failure paths add contextual errors. The remaining harness error-wrapping debt remains appropriately tracked in #507 rather than being expanded in this PR.

## Assessment

**Approved.** The production fix is coherent across read, validation, pagination, documentation, and test-harness paths. The final changes resolve the prior review findings without introducing a confirmed regression.

## Observations

- The harness still models a flush as minting a fresh version timestamp and models a tombstone shape rather than an existing-row hard-delete transition. These are acknowledged fidelity gaps tracked in #508, not blockers for the #460 reader correction.
- #505 and #506 remain valid scoped follow-ups for the 499-line projection-file seam and the unused benchmark projection path.

## Verification

- Reviewed the complete 7-commit `c5bf9b8...605d34f` diff (50 files, +1997/-219), the #460 specification, the federated-query design, prior review threads, and their fixes.
- `GOCACHE=$PWD/.gocache GOTMPDIR=$PWD/.gotmp TMPDIR=$PWD/.gotmp GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/federated ./internal/parquetcheck ./internal/compaction ./internal/manifest ./internal/e2e_harness/federated/benchmark -count=1`: passed.
- `GOCACHE=$PWD/.gocache GOTMPDIR=$PWD/.gotmp TMPDIR=$PWD/.gotmp GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -tags=e2e -run "^$" ./internal/e2e_harness/federated/... ./internal/e2e_harness/production/...`: passed; E2E packages compile.
- `make fmt-check`, `make lint`, `git diff --check`, and `gofmt -d` over changed Go files: passed/clean.
- GitHub checks on `605d34f` are successful: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- I did not rerun the container-backed E2E test bodies locally; the successful E2E compile and current CI smoke result are the verification signal for this review.

## Summary

PR #504 correctly restores true `created_at` semantics for parquet-winning records, preserves creation-stamp stability across tiers, and closes the related manifest-trust, tombstone, and benchmark-ordering gaps. It is ready to merge.

PR #514, opened 2026-09-04
fix(cdc-init): mint write-once base keys and retry a confirmed-412 publish

### LINKED ISSUE #416: compaction: pre-merge checksum gate can false-positive when a cdc-init re-run overwrites its deterministic base key (stale-stamp TOCTOU)
Found by the #347 final whole-branch review (PR #415).

cdc-init base keys are deterministic (`BuildBasePath` → `{prefix}/{schema}/{minRowID}_{maxRowID}.parquet`, `internal/cdc/init.go:325`), so a re-run legitimately **overwrites** the same final key. The compactor holds no advisory lock; if an init re-run lands between the compactor's `LoadManifest` and the gate's GET (`internal/compaction/verify_inputs.go`), the gate hashes the new bytes against the old stamp and declares `ErrSourceChecksumMismatch` on a healthy system (telemetry + ERROR + terminal pass). Pre-#347 the same overlap resolved silently via the ETag 412 retry; the gate now fires before the merge reaches that check.

Mitigations today: the window is seconds-wide, requires an operator-initiated full re-export racing a scheduled compactor with unchanged row ranges but changed bytes, and it self-heals on the next pass (fresh manifest carries fresh stamps).

Proposed fix: on mismatch, reload the manifest and confirm the entry still carries the same stamp before declaring corruption; a changed stamp reclassifies as concurrent modification (retryable), not corruption.

iceboundrock:
Closes #416.

## What

- `cdc.BuildBasePath` now mints `{min}_{max}_{uuidv7}.parquet`. The row range stays in the name for operators and for reconcile's filename/stats cross-check; the UUID makes the key write-once, matching `BuildDeltaPath` and `BuildMergedBasePath`. The final key shares the batch's `_tmp` UUID.
- `cdc.ParseInitBaseStem` is the inverse. It accepts the legacy two-part shape, so `manifest-reconcile` classification and `checkInitFilenameStats` handle buckets written before this change.
- cdc-init's final `ReplaceTierFiles` publish retries a **confirmed** 412 up to three times by reloading and re-splicing the base tier, instead of discarding the whole export. Ambiguous save errors are not retried. The 412 classification is centralized as `manifest.IsPreconditionFailed`, and the compactor delegates to it.
- `updateSchemaManifest` moves to `internal/cdc/init_publish.go` so `init.go` stays under 500 lines.

## Why

With a deterministic key, a re-run over an unchanged range overwrote an object the manifest still listed under the previous run's column and checksum stamps. The compactor's pre-merge checksum gate could then read the stamp before the overwrite and the bytes after it, and fail with a terminal `ErrSourceChecksumMismatch` on a healthy system. A re-run that failed between its copies and its publish left the stale stamps behind for good. Both scenarios are impossible once no writer ever PUTs under a listed key. The design and scenario triage are on the issue: https://github.com/Lychee-Technology/forma/issues/416#issuecomment-5533929247.

## What a re-run leaves behind

The superseded set stays on S3, unlisted and init-shaped, and `manifest-reconcile --gc` reclaims it. `--repair` cannot reverse-promote it: the promotion date fence requires any evicted listed base to be strictly older than the candidate set. Compaction, the flusher, and reads are unaffected, since nothing ever changed under a listed entry's stamp.

## Tests

- Unit: `TestBuildBasePath_SameRangeDistinctFileIDsMintDistinctKeys`, `TestParseInitBaseStem`, `TestExportSchemaBatch_RerunOverSameRangeMintsFreshKey`, `TestUpdateSchemaManifest_RetriesConfirmed412`, `TestUpdateSchemaManifest_GivesUpAfterBoundedConflicts`, `TestUpdateSchemaManifest_DoesNotRetryAmbiguousSaveError`, `TestIsPreconditionFailed`, reconcile classify/promote cases for both shapes.
- Production e2e: `TestInitRerunIdempotency` now asserts one fresh key per batch, identical stat and ETag on every first-run object, and a base tier equal to the new set. `TestManifestStamp_InPlaceRewriteInvalidatesValidatorCache` step 1 asserts the re-run replaced the key with an identical stamp, then follows the new key into the out-of-band rewrite.
- Docs and comments that described the deterministic key are reworded (`docs/federated-query/design.md`, `docs/manifest-reconcile.md`, validator cache rationale). The validator's path+stamp cache key stays: an out-of-band rewrite under a listed key remains reachable.

## Verification

- `./scripts/test_with_container_runtime.sh` (full unit suite): green.
- `go test ./internal/e2e_harness/production/ -tags=e2e` (full production suite): green.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -646,8 +646,8 @@ while still never excluding it.
 **Trust boundary.** A stamp is trusted without reading bytes only because every
 object behind one was written by a Forma writer under manifest transactionality:
 flush and compaction mint fresh keys and stamp what they just wrote, an init
-rerun overwrites its deterministic key and re-stamps it in the same
-`ReplaceTierFiles` publish, and `manifest-reconcile --repair`'s init promotion
+rerun likewise mints fresh write-once keys (#416) and stamps them in the same
+`ReplaceTierFiles` publish that replaces the base tier, and `manifest-reconcile --repair`'s init promotion
 (#292) recomputes the stamp from the footer rather than inventing one. Outside
 that boundary — an out-of-band overwrite of a listed object's bytes while its
 entry keeps a stamp that still satisfies the system-column invariant — the read
@@ -701,11 +701,13 @@ connection by default, #285), and bails on context cancellation.
 
 **Cache invalidation.** The validator caches each path it validates, keyed by
 path **and by the stamp it was validated under** (nil for probe-validated). Path
-alone would be wrong: objects are write-once for flush and compaction, which
-mint fresh UUIDv7 keys, but **not for init** — `cdc.BuildBasePath` is
-deterministic (`{min}_{max}.parquet`), so an init rerun overwrites the object in
-place and rewrites its entry under the same key. A path-keyed cache would serve
-a warmed server the pre-rerun columns for the life of the process. The manifest
+alone would be wrong: Forma's writers mint write-once keys (flush and compaction
+always did; `cdc.BuildBasePath` mints `{min}_{max}_{uuid}.parquet` since #416,
+before which an init rerun overwrote its deterministic `{min}_{max}.parquet` key
+in place), but a rewrite under a listed path is still reachable — legacy
+deployments, or an operator repair that republishes bytes under an existing
+key — and each re-stamps the entry under the same key. A path-keyed cache would
+serve a warmed server the pre-rewrite columns for the life of the process. The manifest
 rewrite is therefore the invalidation signal: same stamp, keep the entry; new
 stamp, re-validate. An unchanged stamp still costs zero probes, so the
 cold-start win is intact. When a probe *does* run on a stamped path — which
--- a/docs/manifest-reconcile.md
+++ b/docs/manifest-reconcile.md
@@ -29,10 +29,10 @@ manifest 条目，双向报告差异，并可选修复。它是两条既有故�
 |------|------|------|------|
 | delta 孤儿 | `{prefix}/{schemaID}/{uuid}.parquet` | flush 的 manifest-append 失败（#252 起 append 先于 mark-flushed：行留 dirty、重试自愈覆盖，孤儿通常判为可 GC 残留；#252 前为 #197 丢失数据形态）或 #461 起每次 rewrite 保留的 merged source | `--repair` 经守卫判定后补录或转残留（见下） |
 | merged base 孤儿 | `base-{uuid}.parquet` | #188 崩溃 / #461 起每次 rewrite 保留的 merged source | `--gc` 两阶段删除（见下） |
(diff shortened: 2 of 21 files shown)


iceboundrock:
## Review

### Findings

No actionable findings.

### Spec

- The change satisfies #416: cdc-init now mints a fresh `{min}_{max}_{uuid}.parquet` key per batch, so a rerun cannot overwrite bytes beneath a still-listed path and invalidate its columns/checksum stamps.
- The final manifest publish retries only a confirmed conditional-write conflict. Each attempt reloads the manifest and replaces the base tier while preserving the current non-base entries; ambiguous save outcomes remain terminal.
- `ParseInitBaseStem` accepts both the new three-part key and the legacy two-part key. Reconcile classification and filename-to-stats validation therefore retain GC and repair compatibility with pre-change buckets.

### Standards

- Reviewed against `coding-standard.md`: affected Go files stay within the 500-line limit, the new publish helper is focused, error paths add context, and exported additions have GoDoc.
- `git diff --check` is clean.
- The error behavior is appropriate for this operator path: no caller-facing validation/error-class changes were introduced.

### Assessment

The immutable-key approach removes both the compactor stale-stamp TOCTOU and the permanent stale-stamp state after a failed rerun without changing the checksum gate. The bounded 412 path uses the existing CAS semantics correctly, and manifest-reconcile remains conservative for both filename generations. Ready to merge from the code-review perspective.

### Observations

- Sharing the batch UUID between `_tmp` and final keys preserves an actionable correlation for stranded staging objects.
- Retaining the path-plus-stamp validator cache remains necessary for legacy objects and out-of-band rewrites under a listed path.
- Unit coverage exercises fresh-key creation, legacy/new filename parsing and promotion, confirmed-412 retry bounds, and refusal to retry ambiguous saves. Production coverage exercises unchanged and changed init reruns, validator-cache invalidation, and schema isolation.

### Verification

Passed:

- `git diff --check HEAD^ HEAD`
- `make fmt-check`
- `make lint`
- `./scripts/test_with_container_runtime.sh`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc ./internal/manifest ./internal/reconcile ./internal/compaction ./internal/federated`
- `DOCKER_HOST=unix:///run/user/1000/podman/podman.sock TESTCONTAINERS_RYUK_DISABLED=true GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -v ./internal/e2e_harness/production -tags=e2e -run "^(TestInitRerunIdempotency|TestInitRerunAfterChangesReconcilesManifest|TestManifestStamp_InPlaceRewriteInvalidatesValidatorCache|TestInitSchemaScopedIsolation)$" -timeout=5m`

The unfiltered production E2E command was also attempted. It timed out at 5 minutes in unrelated `TestPostgresHalt_ParquetTiersError` while waiting for PostgreSQL after the PR-relevant cache test had passed; the focused PR coverage above passed in a fresh Podman test environment.

### Summary

No changes requested.

Issue #461, opened 2026-08-24
bug(compaction): merged sources are deleted with zero grace after the manifest swap — in-flight federated reads fail non-degradable and trip the breaker


iceboundrock:
## Problem

A routine compaction pass deletes the merged source objects **immediately** after the manifest swap, with zero retention grace. Any federated read that resolved its path set from the pre-swap manifest and is still opening objects fails with `ParquetSetInconsistentError` — which by contract is non-degradable, breaker-worthy, and never retried. So healthy-system compaction produces hard read failures and can trip the DuckDB circuit breaker.

## Current behavior

`internal/compaction/rewrite.go` `runRewrite`: `saveManifestChecked(...)` (the CAS-protected swap) is immediately followed by `c.deleteObjects(ctx, schemaID, sourceKeys)` — no age check, no deferral. Contrast `internal/reconcile/gc.go` `gcSchema`, which requires both `firstSeen < now − grace` and `LastModified < now − grace` precisely so "an old source freshly spliced out by the compactor must survive the in-flight-reader window."

Reader side: `internal/federated/duckdb_query_resolve.go` `resolveScanSources` resolves the path set once, before rendering; manifest-listed objects are opened lazily, so a missing object can surface at `Query` or mid-stream. `internal/federated/duckdb_query_execute.go` classifies a missing listed object as `ParquetSetInconsistentError` → `breaker.RecordFailure()`, returned unretried; `degradableFederatedError` excludes it and `AllowPartialDegradedMode` cannot absorb it (`docs/federated-query/design.md` §7.3 step 1).

## Reproduction

1. A federated query resolves the pre-swap manifest (path set includes merged sources `K…`).
2. The compactor commits its manifest swap and deletes `K…`.
3. DuckDB lazily opens `K` → object gone → `ParquetSetInconsistentError`, `RecordFailure`, no retry.

Five such failures inside the breaker window open the breaker and take the DuckDB route offline entirely. Exposure window = (query duration) × (compaction frequency); long analytical queries — exactly the ones routed to DuckDB — have the widest window.

## Root cause

[...]

PR #515, opened 2026-09-04
fix(compaction): refuse foreign-bucket rewrite sources instead of skipping them (#417)

### LINKED ISSUE #417: compaction: pre-merge checksum gate skips foreign-bucket sources that MergeToTmp still reads (fail-open corner)
Found during #347 Task 5 review, confirmed by the final whole-branch review (PR #415); documented in design.md §7.3 as a known scope limit.

`verifySourceChecksums` skips `s3://other-bucket/...` sources with a WARN, mirroring `deleteObjects`' pre-existing foreign-bucket asymmetry — but `MergeToTmp` consumes `objectURI(f.Path)` verbatim, so DuckDB httpfs still reads those objects into the merged base. A corrupt foreign-bucket source can therefore be laundered past the fail-closed gate.

In `deleteObjects` the skip is harmless (leaves an orphan for reconcile); here it is a fail-open corner of a fail-closed mechanism. Options: verify foreign sources too (needs a reader for the foreign bucket), refuse to merge unverifiable foreign sources (fail-closed, breaking for cross-bucket layouts), or keep the documented status quo. Needs a ruling on whether cross-bucket manifest paths are a supported deployment shape at all.

iceboundrock:
Closes #417.

## Problem

`verifySourceChecksums` skipped a stamped source in another bucket with a WARN (mirroring `deleteObjects`), but `runRewrite` still passed `objectURI(f.Path)` to `MergeToTmp`, so DuckDB read the unverified object into the merged base and `spliceManifest` dropped its listed name. That is a fail-open corner of a fail-closed gate, and since #461 the `deleteObjects` symmetry that justified the skip no longer applies (merged sources are retained, not deleted).

## Ruling (recorded on the issue)

Cross-bucket manifest paths are **not a supported compaction input**. No writer produces them; the manifest format merely tolerates absolute URIs (#249 review). Every other guard already refuses what it cannot prove (#456 hint scope, reconcile's promote fence, `--gc`). Verifying foreign sources through the bucket-parameterised reader was considered and declined: it would make the compactor the one component reaching beyond its bucket, with cross-region and credential failures surfacing as noisy retried infra errors instead of one crisp refusal.

## Change

- `internal/compaction/verify_inputs.go`: new sentinel `ErrForeignSource` and `rejectForeignSources`, a scope check that runs regardless of `SkipInputChecksumVerify` / `ObjectReader` and applies to unstamped entries too. The checksum gate's WARN-skip branch now returns the same error (defence in depth).
- `internal/compaction/rewrite.go`: `runRewrite` runs `verifyRewriteInputs` (scope gate, then checksum gate) before anything is merged, promoted or committed. `isRetryable`'s allowlist is untouched, so the refusal is terminal for the pass. `deleteObjects` resolves paths through the shared `bucketRelativeKey`, which also refuses a path that resolves to an empty key (`s3://<bucket>/`, `/`).
- `internal/cdc/config.go`, `cmd/tools/compactor.go`: the opt-out's docs say it waives hashing only.
- `docs/federated-query/design.md` §7.3: records the ruling and the remedy (correct the manifest).

Breaking only for a hand-built manifest with cross-bucket entries: that schema no longer compacts, and the ERROR names the path and bucket.

## Tests

- `TestRejectForeignSources_*`: foreign path refused whether stamped or not, with the skip flag set and a nil reader; own-bucket relative and absolute paths pass; `s3://bktX/...` prefix collision refused.
- `TestVerifySourceChecksums_ForeignBucketPathRefused` replaces the former skipped-with-warning test.
- `TestRejectForeignSources_EmptyKeyRefused`, `TestDeleteObjects_UsesBucketRelativeKey`: the empty-key rule and the shared path resolution in `deleteObjects` (review follow-ups).
- `TestVerifySourceChecksums_OwnBucketURIKeyPreservedVerbatim` and a new `s3://bkt//...` row in `TestDeleteObjects_UsesBucketRelativeKey`: an own-bucket URI keeps its key verbatim (`s3://bkt//x` is key `/x`), so the gate hashes and `deleteObjects` deletes the same object `objectURI` hands DuckDB (fourth-round review).
- `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge`: through `RunOnce` with the checksum gate opted out, `MergeToTmp` is never called, nothing is copied or deleted, the manifest version is unchanged, and `LoadManifest` runs once (no retry).

Verified locally: `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/...`, `make fmt-check`, `golangci-lint v1.64.8` on the compaction package, `go build ./...`.


iceboundrock:
## Review: fix(compaction) refuse foreign-bucket rewrite sources (#417) — PR #515

### Summary

LGTM — approve. This PR correctly closes the fail-open corner described in #417: the pre-merge checksum gate's WARN-skip for `s3://other-bucket/...` sources is replaced by a fail-closed refusal, and `runRewrite` now rejects any source set containing a path outside the compactor's bucket before anything is merged, promoted, or committed. No blockers found. Two non-blocking follow-ups below (function length, helper reuse).

### Spec (vs #417 ruling + plan)

Checked against the ruling + 4-step plan recorded on #417 and the PR body:

- [x] Ruling implemented as stated: cross-bucket paths are not a supported compaction input; refusal is stamped-or-not and independent of `--skip-input-checksum-verify` / `ObjectReader` (`rejectForeignSources` has no gate on either — `internal/compaction/verify_inputs.go:39-50`).
- [x] Ordering: `rejectForeignSources` runs before `verifySourceChecksums` before `MergeToTmp` (`internal/compaction/rewrite.go:67-72`). Nothing merges on refusal.
- [x] Defence in depth: `verifySourceChecksums`' old WARN-skip branch now returns the same `ErrForeignSource` (`verify_inputs.go:65-73`). Unreachable via `runRewrite` after the new call, but fail-closed standalone — matches plan step 1.
- [x] Terminal for the pass: `isRetryable` allowlist untouched, so `ErrForeignSource` is terminal exactly like `ErrSourceChecksumMismatch` (`internal/compaction/compactor.go:143-155`). Comment updated to name both sentinels.
- [x] Opt-out docs narrowed correctly: `internal/cdc/config.go:92-94` and `cmd/tools/compactor.go:44-48` both say the flag waives hashing only, not scope.
- [x] Design doc: `docs/federated-query/design.md` §7.3 records the ruling, the remedy (correct the manifest), and why foreign-verify was declined. The "only stamped sources inside the compactor's own bucket" wording is correctly superseded.
- [x] No deviation from the plan except the one addition already declared on the issue (flag help + field doc). Confirmed: diff is exactly the 8 files listed, no extra behavior change.
- [x] Consistency with cited precedents: #461 (retained sources make the old `deleteObjects` symmetry inapplicable — correctly reasoned), #249 (absolute URIs tolerated by format, not produced by writers), #456 hint scope / reconcile promote fence / `--gc` (all refuse-what-they-cannot-prove; this PR makes the compactor consistent).

### Findings (correctness)

- `bucketRelativeKey` (`verify_inputs.go:94-103`) is correct for the cases that matter: relative keys pass (with leading-slash trim, mirroring `objectURI`), `s3://<own-bucket>/...` passes, `s3://bktX/...` prefix collision correctly refused (pinned by `TestRejectForeignSources_BucketPrefixCollisionRefused`). Case-sensitive exact match is right for S3 bucket names.
- Error shape is attributable and matchable: both refusal sites wrap `ErrForeignSource` with `%w` and name path + schema + bucket (`verify_inputs.go:46-47`, `verify_inputs.go:71-72`). `errors.Is` classification holds; no string comparison anywhere.
- End-to-end test proves the security property: with the checksum gate opted out, `MergeToTmp` is never called, nothing copied/deleted, manifest version unchanged, `LoadManifest` runs once (no retry) — `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge`.
- Unit coverage is the right matrix: foreign refused stamped-or-not, opt-out + nil-reader bypass attempted and refused, own-bucket relative + leading-slash + absolute-own-bucket pass.
- Promotion path intentionally untouched: `compactSchema` promotion (delta→base relabel, manifest-only, no byte reads) can still relabel a foreign entry; refusal lands at rewrite time when bytes would actually move. Safe and consistent — noting only so a later reader doesn't file it as a bypass.

### Standards (`coding-standard.md`, `docs/error-handling.md`, `AGENTS.md`)

- File sizes: all touched Go files ≤500 lines (`rewrite.go` 228, `compactor.go` 336, `verify_inputs.go` 103, `verify_foreign_test.go` 107). Pass.
- Error handling: every new `return` wraps with context (`failed`-style with `%w`, naming logical value + offending path + expected bucket). No bare `return err`, no bare sentinel wrap. Flat early-return structure, nesting ≤2. Pass.
- Exported sentinel `ErrForeignSource` has godoc stating scope, terminality, and fix direction. Pass.
- Error class: plain sentinel + `%w` is correct here. Per `docs/error-handling.md`, this is a read-path/persisted-state consistency failure (manifest lists what the compactor cannot prove), i.e. operator-visible, not a caller-mistake `InvalidInputf` 4xx. The ERROR log naming `schema_id`/`path`/`bucket` satisfies the "name the offending value and expected state" rule; `WithOperatorDetail` is not applicable (nothing caller-actionable is being withheld beyond what the message already carries).
- One standards flag (non-blocking, see Observations): `runRewrite` grows 100 → 105 total lines (68 → 71 non-blank-non-comment). The letter of the ≤100 rule counts total lines, so this tips it over; the spirit (SRP/readability) is unaffected — the addition is two gate calls + comment.

### Assessment

Approve. The change is minimal, fail-closed, terminal, well-pinned by tests, and documented at all three layers (code godoc, flag/field help, design doc). Breaking scope is exactly as advertised (hand-built cross-bucket manifests only) and the error tells the operator what to fix. No retry-storm risk, no telemetry misclassification (mismatch metric correctly not emitted for a scope refusal), no behavior change for in-bucket layouts.

### Observations (non-blocking follow-ups, not merge gates)

1. `runRewrite` length (100 → 105 total lines): consider extracting the two pre-merge gates into e.g. `verifyRewriteInputs(ctx, schemaID, sources)` returning the first refusal. Restores the ≤100 total-line shape and makes the "fail closed before anything is merged" block a single named step. Optional — code-only count is 71, well under.
2. `deleteObjects` (`rewrite.go:211-228`) still inlines its own `TrimPrefix("s3://"+c.Bucket+"/")` foreign check instead of reusing `bucketRelativeKey`. ~6 duplicated lines, under the >15 DRY gate, but sharing the helper would keep the two path rules from drifting again (the new helper's godoc already claims to combine them).
3. Edge: `s3://<bucket>/` (empty key) resolves to `("", true)` and passes the scope check. Unreachable via any shipped writer (keys always carry a prefix), and harmless (downstream would fail loudly, not silently relocate). If hardening is cheap, refuse empty keys explicitly in `bucketRelativeKey`.
4. Defence-in-depth branch (`verifySourceChecksums` foreign path) returns without an ERROR log, while `rejectForeignSources` logs once. Correct in the production path (reject logs first, verify never reached), but a standalone caller gets an error with no log line. Either log there too or note in godoc that the caller logs — current godoc is silent on logging responsibility.
5. No telemetry on `ErrForeignSource` — correct (not corruption, no `parquet_checksum_mismatch_total` pollution), noting only to pre-empt the question.

### Verification

- Local: `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1` — all `ok` (compaction 0.8s incl. new `TestRejectForeignSources_*`, `TestVerifySourceChecksums_ForeignBucketPathRefused`, `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge`).
- Local: `make fmt-check` — pass. `go vet ./internal/compaction/ ./internal/cdc/ ./cmd/tools/` — clean.
- Local: `go build ./...` not usable as signal here — fails linking `cmd/lambda` and DuckDB-heavy binaries on `No space left on device` / `Disk quota exceeded`, unrelated to this diff (no import changes).
- `golangci-lint` binary absent locally; CI signal used instead.
- CI on this PR (all green at review time): Test, Lint, E2E Smoke Tests, Benchmark Smoke, k6 Smoke — all `pass`.
- Spec cross-check: issue #417 ruling + plan comments re-read; implementation matches with only the declared docs addition.

iceboundrock:
## Response to review observations

All five observations are addressed in `a6c38e4` except the last, which needs no change. None fell outside the PR's scope, so no follow-up issue was opened.

1. **`runRewrite` length.** The two pre-merge gates moved into `verifyRewriteInputs(ctx, schemaID, sources)` in `verify_inputs.go`, which returns the first refusal. `runRewrite` is now 96 total lines. Step 0 of its ordering godoc names the new function; the long rationale comment moved onto the helper's godoc rather than being duplicated.

2. **`deleteObjects` reuse.** It now resolves paths through `bucketRelativeKey` and keeps its WARN-skip for anything the helper cannot resolve. Behaviour is unchanged for the one call site that remains since #461 (the confirmed-412 staged base, always a bucket-relative key). Pinned by `TestDeleteObjects_UsesBucketRelativeKey`: own-bucket URI and leading-slash path are deleted by relative key, a foreign URI and an empty-key URI are skipped with one WARN each and never reach `DeleteObject`.

3. **Empty key.** `bucketRelativeKey` now reports false for any path that resolves to an empty key: `s3://<bucket>/`, `/`, and ``. That makes the empty case refuse at the scope gate (`ErrForeignSource`) rather than fail downstream, and, via item 2, makes `deleteObjects` skip it rather than issue a delete against an empty key. Pinned by `TestRejectForeignSources_EmptyKeyRefused`. The godoc states the rule, including why the prefix match is exact (`s3://bktX/` is not `bkt`).

4. **Defence-in-depth branch logging.** `verifySourceChecksums` now logs its own ERROR (schema_id, path, bucket) on a stamped foreign source, so a standalone caller gets the same attributable log line the production path gets from `rejectForeignSources`. In production only one of the two ever fires, since the reject gate runs first. Both gates' godoc now says explicitly that they log their own refusal, and `verifyRewriteInputs`' godoc says it relies on that. `TestVerifySourceChecksums_ForeignBucketPathRefused` asserts the log entry.

5. **No telemetry on `ErrForeignSource`.** Agreed and unchanged: `parquet_checksum_mismatch_total` measures corruption of an object the compactor could read, and a scope refusal proves nothing about bytes. The design doc §7.3 wording already separates the two, so nothing to add.

Verified locally after the change: `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1` all ok, `make fmt-check` and `make lint` clean, `go vet ./internal/compaction/` clean.

iceboundrock:
## Review: fix(compaction) refuse foreign-bucket rewrite sources (#417) — PR #515

Artifact: PR review (findings / spec / standards / assessment / observations / verification / summary). Working copy: this comment only; no local planning file. Reviewed HEAD `a6c38e4` (8 files, +254/−54).

### Summary

Approve — no blockers. This PR correctly closes the fail-open corner from #417: the pre-merge checksum gate's old WARN-skip for `s3://other-bucket/...` is replaced by a fail-closed refusal, and `runRewrite` now rejects any source set containing an out-of-bucket path before anything is merged, promoted, or committed. The `a6c38e4` follow-up fully addresses all five observations from the prior review round (gate extraction, `deleteObjects` sharing the helper, empty-key refusal, standalone-gate logging). Minimal, terminal, well-pinned by tests, documented at code + flag + design-doc layers.

### Spec (vs #417 ruling + plan)

Checked against the ruling + 4-step plan recorded on #417 and the PR body — all hold at `a6c38e4`:

- [x] Ruling implemented as stated: cross-bucket paths are not a supported compaction input; refusal is stamped-or-not and independent of `--skip-input-checksum-verify` / `ObjectReader` (`rejectForeignSources` has no gate on either — `internal/compaction/verify_inputs.go:57-68`; pinned by `TestRejectForeignSources_IgnoresChecksumOptOut`).
- [x] Ordering: `verifyRewriteInputs` runs scope gate before checksum gate before `MergeToTmp` (`internal/compaction/verify_inputs.go:43-51`, called from `internal/compaction/rewrite.go:60-62`). Nothing merges on refusal.
- [x] Defence in depth: `verifySourceChecksums`' old WARN-skip branch now returns `ErrForeignSource` with its own ERROR log (`verify_inputs.go:85-95`). Unreachable via `runRewrite` after the new call, but fail-closed standalone — matches plan step 1.
- [x] Terminal for the pass: `isRetryable` allowlist untouched, comment now names both sentinels (`internal/compaction/compactor.go:143-155`). End-to-end test proves single attempt (`provider.loads == 1`).
- [x] Opt-out docs narrowed correctly: `internal/cdc/config.go:92-94` and `cmd/tools/compactor.go:44-48` both say the flag waives hashing only, not scope.
- [x] Design doc: `docs/federated-query/design.md` §7.3 records the ruling, the remedy (correct the manifest), and why foreign-verify was declined. The old "only stamped sources inside the compactor's own bucket" wording is correctly superseded.
- [x] Review follow-ups from the first round all landed in `a6c38e4` as described on the PR: `verifyRewriteInputs` extraction, `deleteObjects` via `bucketRelativeKey`, empty-key refusal, standalone-gate ERROR log. Verified each in the diff — no scope creep beyond that.
- [x] Consistency with cited precedents holds: #461 (retained sources make the old `deleteObjects` symmetry inapplicable), #249 (absolute URIs tolerated by format, not produced by writers), #456 hint scope / reconcile promote fence / `--gc` (all refuse-what-they-cannot-prove; the compactor is now consistent).

### Findings (correctness)

- `bucketRelativeKey` (`verify_inputs.go:121-134`) is correct for the cases that matter: relative keys pass (leading-slash trim, mirroring `objectURI`), `s3://<own-bucket>/...` passes, `s3://bktX/...` prefix collision refused via exact `"s3://"+Bucket+"/"` match (pinned by `TestRejectForeignSources_BucketPrefixCollisionRefused`). Case-sensitive exact match is right for S3 bucket names. Empty-key rule (`""`, `"/"`, `"s3://bkt/"` → `("", false)`) refuses rather than handing an empty key downstream (pinned by `TestRejectForeignSources_EmptyKeyRefused`).
- Error shape is attributable and matchable: both refusal sites wrap `ErrForeignSource` with `%w` and name path + schema + bucket (`verify_inputs.go:64-65`, `verify_inputs.go:93-94`). `errors.Is` classification holds; no string comparison anywhere in prod code or tests for classification (tests use `ErrorIs` + `ErrorContains` for the human-readable path).
- Security property proved end-to-end: with the checksum gate opted out, `MergeToTmp` never called, nothing copied/deleted, manifest version unchanged, `LoadManifest` runs once (no retry) — `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge` (`internal/compaction/verify_foreign_test.go:122-138`).
- Unit matrix is the right one: foreign refused stamped-or-not, opt-out + nil-reader bypass attempted and refused, own-bucket relative + leading-slash + absolute-own-bucket pass, prefix collision refused, empty keys refused.
- `deleteObjects` sharing (`rewrite.go:202-214`) is behavior-preserving for its one live call site since #461 (confirmed-412 staged base, always a bucket-relative key) and strictly safer for any future caller: foreign/empty paths WARN-skip instead of issuing a delete against a wrong or empty key (pinned by `TestDeleteObjects_UsesBucketRelativeKey`).
- Promotion path intentionally untouched: `compactSchema` promotion (delta→base relabel, manifest-only, no byte reads) can still relabel a foreign entry; refusal lands at rewrite time when bytes would actually move. Safe — noting only so a later reader doesn't file it as a bypass.
- No telemetry on `ErrForeignSource` — correct. `parquet_checksum_mismatch_total` measures corruption of an object the compactor could read; a scope refusal proves nothing about bytes. Design doc §7.3 already separates the two.
- Double-logging checked: production path logs exactly once (`rejectForeignSources` fires, `verifySourceChecksums` never reached); standalone caller gets exactly one ERROR from whichever gate it invoked (both tests assert `Len(entries, 1)`). No duplicate or missing log line.

### Standards (`coding-standard.md`, `docs/error-handling.md`, `AGENTS.md`)

- File sizes: all touched Go files ≤500 lines (`rewrite.go` 215, `compactor.go` 336, `verify_inputs.go` 134, `verify_foreign_test.go` 138, `verify_inputs_test.go` 288). Pass.
- Function lengths: prior round's `runRewrite` 100→105 overage is fixed — `runRewrite` is now 96 total lines (51–146), `verifyRewriteInputs` ~9, `rejectForeignSources` ~12, `verifySourceChecksums` ~33, `bucketRelativeKey` ~14. All ≤100. Pass.
- Error handling: every new `return` wraps with context naming logical value + offending path + expected bucket (`%w` on `ErrForeignSource`). No bare `return err`, no bare sentinel wrap. Flat early-return structure, nesting ≤2. Pass.
- Exported sentinel `ErrForeignSource` has godoc stating scope, terminality, and fix direction (correct the manifest, not retry). Pass.
- Error class: plain sentinel + `%w` is correct here. Per `docs/error-handling.md` this is a read-path/persisted-state consistency failure (manifest lists what the compactor cannot prove), i.e. operator-visible, not a caller-mistake `InvalidInputf` 4xx. The ERROR log naming `schema_id`/`path`/`bucket` satisfies the "name the offending value and expected state" rule; `WithOperatorDetail` is not applicable (nothing caller-actionable is withheld beyond what the message already carries).
- DRY: the two path rules (`objectURI` trim + own-bucket prefix match) now live in the single shared `bucketRelativeKey` with godoc claiming exactly that; `deleteObjects` reuses it rather than inlining its own `TrimPrefix`. The ~6-line duplication flagged last round is gone. Pass.
- Docs: flag help + field godoc + design doc all updated; no behavior change beyond the declared refusal. Pass.

### Assessment

Approve and merge-ready. The change is minimal, fail-closed, terminal, well-pinned by tests at unit + gate + end-to-end levels, and documented at all three layers (code godoc, flag/field help, design doc). Breaking scope is exactly as advertised (hand-built cross-bucket manifests only) and the error tells the operator what to fix. No retry-storm risk, no telemetry misclassification, no behavior change for in-bucket layouts. The `a6c38e4` follow-up improved the shape without changing the ruling — `runRewrite` is back under the line limit and the two path rules can no longer drift apart.

### Observations (non-blocking, not merge gates)

1. `deleteObjects` WARN text ("skipping deletion of object outside compactor bucket", `rewrite.go:206-207`) also fires for the empty-key case (`s3://bkt/`, `/`, `""`), which is inside the bucket but names nothing. Message is slightly imprecise for that shape; consider "outside compactor bucket or empty" if touched again. Cosmetic — the skip itself is correct.
2. `objectURI` (`rewrite.go:189-194`) still passes foreign URIs through verbatim, now with a comment stating `runRewrite` never reaches it with one. True in production, but a future standalone caller could still misuse it. No change needed — just noting the safety rests on call ordering, as documented.
3. `""` is classified as `ErrForeignSource` ("outside bucket") rather than a distinct invalid-input error. Consistent with the ruling (out of scope like a foreign path) and terminal either way; the message for `""` reads slightly odd but remains attributable. Acceptable — flagging only to pre-empt the question.

### Verification

- Local: `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1` — all `ok` (compaction 1.9s, cdc 4.3s, reconcile 1.2s, cmd/tools 1.9s), including new `TestRejectForeignSources_*` (5), `TestVerifySourceChecksums_ForeignBucketPathRefused`, `TestDeleteObjects_UsesBucketRelativeKey`, `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge`.
- Local targeted: `go test ./internal/compaction -run 'TestRejectForeignSources|TestVerifySourceChecksums|TestCompactor_Rewrite|TestDeleteObjects' -v` — all pass.
- Local: `make fmt-check` — pass (no output beyond "Checking gofmt..."). `go vet ./internal/compaction/ ./internal/cdc/ ./cmd/tools/` — clean.
- `golangci-lint` binary absent locally; CI signal used instead.
- CI on this PR (all green at review time): Test, Lint, E2E Smoke Tests, Benchmark Smoke, k6 Smoke — all `pass`.
- Spec cross-check: issue #417 ruling + plan comments re-read; implementation matches with only the declared docs addition (flag help + field doc) plus the four review follow-ups already recorded on the PR.

iceboundrock:
## Response to second-round observations

Two of the three are addressed in `e0efe6d`; the third needs no change. All three sit inside the PR's own diff, so no follow-up issue was opened.

1. **`deleteObjects` WARN text.** Reworded to "skipping deletion: path is outside the compactor bucket or names no key", and the entry now carries the `bucket` field like the two rewrite-gate ERRORs, so all three refusal logs are attributable the same way. `TestDeleteObjects_UsesBucketRelativeKey` pins the `path` field, not the message, so it needed no change.

2. **`objectURI` pass-through.** Unchanged. Its absolute-URI branch mirrors `manifest.QuerySource.Paths` on the read side, and its only callers are the two lines in `runRewrite` that run after `verifyRewriteInputs`. Making it fail closed on its own would mean either returning an error from a URI renderer or having the gate hand back resolved keys; both are churn for a path no caller can reach. The godoc already records that the safety rests on the gate ordering, which is the right place for that fact.

3. **`""` classified as `ErrForeignSource`.** Unchanged in classification: the #417 ruling's category is "not a supported compaction input", and an empty path is exactly that, with the same remedy (correct the manifest) and the same terminality. A second sentinel would give operators two errors with one fix. Nothing in `manifest.Load` rejects an empty `Path`, so the shape is reachable from a hand-edited manifest and the message matters: both refusals now render the path with `%q`, so it reads `rewrite source "" for schema 1 is outside bucket bkt` instead of a blank. `%q` has precedent in the package (`merge_sql.go`). Pinned by an `ErrorContains` on `rewrite source %q` in `TestRejectForeignSources_EmptyKeyRefused`.

Verified locally after the change: `make fmt-check`, `go test ./internal/compaction -count=1`, `go vet ./internal/compaction/`, and `make lint` (root and infra modules) all clean.

iceboundrock:
## Review: fix(compaction) refuse foreign-bucket rewrite sources (#417) — PR #515

Artifact: PR review (findings / spec / standards / assessment / observations / verification / summary). Working copy: this comment only; no local planning file. Reviewed HEAD `e0efe6d` (8 files, +257/−54).

### Summary

Approve — no blockers. This PR correctly closes the fail-open corner from #417: the pre-merge checksum gate's old WARN-skip for `s3://other-bucket/...` is replaced by a fail-closed refusal, and `runRewrite` now rejects any source set containing an out-of-bucket path before anything is merged, promoted, or committed. The two follow-up rounds (`a6c38e4`, `e0efe6d`) fully address the prior review observations (gate extraction, `deleteObjects` sharing the helper, empty-key refusal, standalone-gate logging, WARN wording, `%q` quoting). Minimal, terminal, well-pinned by tests, documented at code + flag + design-doc layers.

### Spec (vs #417 ruling + plan)

Checked against the ruling + 4-step plan recorded on #417 and the PR body — all hold at `e0efe6d`:

- [x] Ruling implemented as stated: cross-bucket paths are not a supported compaction input; refusal is stamped-or-not and independent of `--skip-input-checksum-verify` / `ObjectReader` (`rejectForeignSources` has no gate on either — `internal/compaction/verify_inputs.go:57-68`; pinned by `TestRejectForeignSources_IgnoresChecksumOptOut`).
- [x] Ordering: `verifyRewriteInputs` runs scope gate before checksum gate before `MergeToTmp` (`internal/compaction/verify_inputs.go:43-51`, called from `internal/compaction/rewrite.go:60-62`). Nothing merges on refusal.
- [x] Defence in depth: `verifySourceChecksums`' old WARN-skip branch now returns `ErrForeignSource` with its own ERROR log (`verify_inputs.go:85-95`). Unreachable via `runRewrite` after the new call, but fail-closed standalone — matches plan step 1.
- [x] Terminal for the pass: `isRetryable` allowlist untouched, comment now names both sentinels (`internal/compaction/compactor.go:143-155`). End-to-end test proves single attempt (`provider.loads == 1`).
- [x] Opt-out docs narrowed correctly: `internal/cdc/config.go:92-94` and `cmd/tools/compactor.go:44-48` both say the flag waives hashing only, not scope.
- [x] Design doc: `docs/federated-query/design.md` §7.3 records the ruling, the remedy (correct the manifest), and why foreign-verify was declined. The old "only stamped sources inside the compactor's own bucket" wording is correctly superseded.
- [x] Review follow-ups from both prior rounds all landed as described on the PR: `verifyRewriteInputs` extraction, `deleteObjects` via `bucketRelativeKey`, empty-key refusal, standalone-gate ERROR log, WARN reword + `bucket` field, `%q` quoting. Verified each in the diff — no scope creep beyond that.
- [x] Consistency with cited precedents holds: #461 (retained sources make the old `deleteObjects` symmetry inapplicable), #249 (absolute URIs tolerated by format, not produced by writers), #456 hint scope / reconcile promote fence / `--gc` (all refuse-what-they-cannot-prove; the compactor is now consistent).

### Findings (correctness)

- `bucketRelativeKey` (`verify_inputs.go:122-135`) is correct for the cases that matter: relative keys pass (leading-slash trim, mirroring `objectURI`), `s3://<own-bucket>/...` passes, `s3://bktX/...` prefix collision refused via exact `"s3://"+Bucket+"/"` match (pinned by `TestRejectForeignSources_BucketPrefixCollisionRefused`). Case-sensitive exact match is right for S3 bucket names. Empty-key rule (`""`, `"/"`, `"s3://bkt/"` → `("", false)`) refuses rather than handing an empty key downstream (pinned by `TestRejectForeignSources_EmptyKeyRefused`). Degenerate shapes (`s3://bkt` without slash, `s3://`) also fail the prefix match and refuse — fail-closed, correct.
- Error shape is attributable and matchable: both refusal sites wrap `ErrForeignSource` with `%w` and name path + schema + bucket (`verify_inputs.go:64-65`, `verify_inputs.go:93-94`). `errors.Is` classification holds; no string comparison anywhere in prod code or tests for classification (tests use `ErrorIs` + `ErrorContains` for the human-readable path).
- Security property proved end-to-end: with the checksum gate opted out, `MergeToTmp` never called, nothing copied/deleted, manifest version unchanged, `LoadManifest` runs once (no retry) — `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge` (`internal/compaction/verify_foreign_test.go:124-140`).
- Unit matrix is the right one: foreign refused stamped-or-not, opt-out + nil-reader bypass attempted and refused, own-bucket relative + leading-slash + absolute-own-bucket pass, prefix collision refused, empty keys refused.
- `deleteObjects` sharing (`rewrite.go:202-214`) is behavior-preserving for its one live call site since #461 (confirmed-412 staged base, always a bucket-relative key) and strictly safer for any future caller: foreign/empty paths WARN-skip instead of issuing a delete against a wrong or empty key (pinned by `TestDeleteObjects_UsesBucketRelativeKey`).
- Promotion path intentionally untouched: `compactSchema` promotion (delta→base relabel, manifest-only, no byte reads) can still relabel a foreign entry; refusal lands at rewrite time when bytes would actually move. Safe — noting only so a later reader doesn't file it as a bypass.
- No telemetry on `ErrForeignSource` — correct. `parquet_checksum_mismatch_total` measures corruption of an object the compactor could read; a scope refusal proves nothing about bytes. Design doc §7.3 already separates the two.
- Double-logging checked: production path logs exactly once (`rejectForeignSources` fires, `verifySourceChecksums` never reached); standalone caller gets exactly one ERROR from whichever gate it invoked (both tests assert `Len(entries, 1)`). No duplicate or missing log line. All three refusal logs now carry `schema_id`/`path`/`bucket`.

### Standards (`coding-standard.md`, `docs/error-handling.md`, `AGENTS.md`)

- File sizes: all touched Go files ≤500 lines (`rewrite.go` 215, `compactor.go` 336, `verify_inputs.go` 135, `verify_foreign_test.go` 140, `verify_inputs_test.go` 288). Pass.
- Function lengths: `runRewrite` 51–146 = 96 total, `verifyRewriteInputs` 9, `rejectForeignSources` 12, `verifySourceChecksums` 34, `bucketRelativeKey` 14. All ≤100. Pass.
- Error handling: every new `return` wraps with context naming logical value + offending path + expected bucket (`%w` on `ErrForeignSource`). No bare `return err`, no bare sentinel wrap. Flat early-return structure, nesting ≤2. Pass.
- Exported sentinel `ErrForeignSource` has godoc stating scope, terminality, and fix direction (correct the manifest, not retry). Pass.
- Error class: plain sentinel + `%w` is correct here. Per `docs/error-handling.md` this is a read-path/persisted-state consistency failure (manifest lists what the compactor cannot prove), i.e. operator-visible, not a caller-mistake `InvalidInputf` 4xx. The ERROR log naming `schema_id`/`path`/`bucket` satisfies the "name the offending value and expected state" rule; `WithOperatorDetail` is not applicable (nothing caller-actionable is withheld beyond what the message already carries).
- DRY: the two path rules (`objectURI` trim + own-bucket prefix match) now live in the single shared `bucketRelativeKey` with godoc claiming exactly that; `deleteObjects` reuses it rather than inlining its own `TrimPrefix`. Pass.
- Docs: flag help + field godoc + design doc all updated; no behavior change beyond the declared refusal. Pass.

### Assessment

Approve and merge-ready. The change is minimal, fail-closed, terminal, well-pinned by tests at unit + gate + end-to-end levels, and documented at all three layers (code godoc, flag/field help, design doc). Breaking scope is exactly as advertised (hand-built cross-bucket manifests only) and the error tells the operator what to fix. No retry-storm risk, no telemetry misclassification, no behavior change for in-bucket layouts. Both follow-up rounds improved the shape without changing the ruling — `runRewrite` is back under the line limit, the two path rules can no longer drift apart, and empty-path refusals stay legible.

### Observations (non-blocking, not merge gates)

1. `objectURI` (`rewrite.go:189-194`) still passes foreign URIs through verbatim, now with a comment stating `runRewrite` never reaches it with one. True in production, but safety rests on call ordering, as documented. No change needed — noting only.
2. `""` is classified as `ErrForeignSource` ("outside bucket") rather than a distinct invalid-input error. Consistent with the ruling ("not a supported compaction input") and terminal either way; `%q` quoting in `e0efe6d` makes the message legible (`rewrite source ""`). Acceptable.
3. Minor format inconsistency (cosmetic): the two new foreign refusals render the path with `%q`, while the adjacent `verifySourceChecksums` probe-failure (`verify_inputs.go:98`) and mismatch (`verify_inputs.go:105-106`) branches use `%s`. Unifying to `%q` would make all gate errors legible the same way. Optional.
4. No explicit test pins scope-before-checksum ordering when both would fire (e.g. stamped foreign with corrupt bytes returns `ErrForeignSource`, not mismatch). Code order (`verify_inputs.go:44-48`) is unambiguous and the e2e test with `Skip=true` proves scope independence, so this is coverage-completeness only, not a correctness doubt.

### Verification

- Local: `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1` — all `ok` (compaction 0.8s, cdc 3.9s, reconcile 0.8s, cmd/tools 1.4s), including new `TestRejectForeignSources_*` (5), `TestVerifySourceChecksums_ForeignBucketPathRefused`, `TestDeleteObjects_UsesBucketRelativeKey`, `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge`.
- Local targeted: `go test ./internal/compaction -run 'TestRejectForeignSources|TestVerifySourceChecksums|TestCompactor_Rewrite|TestDeleteObjects' -v` — all pass.
- Local: `make fmt-check` — pass. `go vet ./internal/compaction/ ./internal/cdc/ ./cmd/tools/` — clean.
- CI on this PR at review time: Test `pass`, Lint `pass` (E2E Smoke / Benchmark Smoke / k6 Smoke `pending` — infra queues, not signals against this diff).
- Spec cross-check: issue #417 ruling + plan comments re-read; implementation matches with only the declared docs addition (flag help + field doc) plus the two recorded review follow-up rounds.

iceboundrock:
## Response to third-round observations

Two of the four are addressed in `2e3b45c`; the other two were already ruled in the previous round and need no change. All four sit inside the PR's own diff, so no follow-up issue was opened.

1. **`objectURI` pass-through.** Unchanged, as ruled last round: its only callers are the two lines in `runRewrite` that run after `verifyRewriteInputs`, and the godoc records that the safety rests on gate ordering. Item 4 below now pins that ordering with a test, so the fact is enforced as well as documented.

2. **`""` classified as `ErrForeignSource`.** Unchanged, as ruled last round: one category ("not a supported compaction input"), one remedy, one sentinel. The `%q` quoting keeps the message legible.

3. **`%s` vs `%q`.** The two remaining `%s` path renders in `verifySourceChecksums` (probe failure, line 98, and mismatch, line 105) now use `%q`, so all four gate errors quote the path the same way. Those two lines predate this PR but sit in the file it already edits, so this is the "fix the immediate area" case rather than scope creep. The log fields are untouched (structured, no quoting needed). `TestVerifySourceChecksums_MutatedBytesRefuseMerge` and `_ProbeFailureIsNotAVerdict` assert the path by substring, so they needed no change.

4. **Scope-before-checksum ordering test.** Added `TestVerifyRewriteInputs_ScopeRefusalPrecedesChecksum` in `verify_foreign_test.go`. It calls `verifyRewriteInputs` with a stamped in-bucket source whose bytes are corrupt next to a stamped foreign source, and asserts `ErrorIs ErrForeignSource`, `NotErrorIs ErrSourceChecksumMismatch`, zero GETs on the reader, and zero telemetry events. That turns the code order into a contract: a source set that would trip both gates surfaces as a scope refusal with no mismatch verdict and no `parquet_checksum_mismatch_total` count.

Verified locally after the change: `go test ./internal/compaction ./internal/cdc ./cmd/tools/... -count=1` all ok, `make fmt-check`, `go vet ./internal/compaction/` and `make lint` (root and infra modules) clean. `verify_foreign_test.go` is 165 lines, `verify_inputs.go` 135.

iceboundrock:
# PR #515 Review — refuse foreign-bucket rewrite sources (closes #417)

## Summary

Correct, well-scoped fail-closed fix for the #417 fail-open corner. The pre-merge checksum gate's WARN-skip for `s3://other-bucket/...` let `runRewrite` hand the unverified URI to `MergeToTmp` verbatim, laundering foreign bytes into the new base while `spliceManifest` dropped the listed name. This PR makes the documented ruling executable: cross-bucket manifest paths are not a supported compaction input, refused terminally before anything is merged, promoted, or committed. No blocker found. **Recommend approve.**

Scope (4 commits, 8 files, +284/−56): `verify_inputs.go` (new `ErrForeignSource`, `rejectForeignSources`, `verifyRewriteInputs`, shared `bucketRelativeKey`), `rewrite.go` (single fail-closed gate + shared path resolution in `deleteObjects`), `config.go` + `compactor.go` CLI/config doc narrowing of `--skip-input-checksum-verify`, `design.md` §7.3 ruling record, and thorough unit + end-to-end tests.

## Spec (against #417)

- **Ruling implemented as specified:** `rejectForeignSources` runs regardless of `SkipInputChecksumVerify` / `ObjectReader` and covers unstamped entries — pinned by `TestRejectForeignSources_IgnoresChecksumOptOut` (skip flag + nil reader still refused) and the unstamped foreign entry in `TestRejectForeignSources_ForeignPathRefused`. Matches the issue's "refuse to merge unverifiable foreign sources" option and the PR-body rationale for declining cross-bucket verification (compactor stays single-bucket; no noisy retried cross-region/credential failures).
- **Defence in depth:** `verifySourceChecksums` returns `ErrForeignSource` for a stamped foreign path instead of the old WARN-skip (old `TestVerifySourceChecksums_ForeignBucketPathSkippedWithWarning` correctly replaced by `TestVerifySourceChecksums_ForeignBucketPathRefused`). Production order is scope-before-checksum in `verifyRewriteInputs`, now a contract via `TestVerifyRewriteInputs_ScopeRefusalPrecedesChecksum` (no GET, no `EmitParquetChecksumMismatch`, no mismatch verdict when both gates would trip).
- **Terminal, not retried:** `isRetryable` allowlist untouched, so `ErrForeignSource` is terminal like `ErrSourceChecksumMismatch` — pinned end-to-end by `TestCompactor_Rewrite_ForeignSourceRefusesTheMerge` (`provider.loads == 1`, no merge/promote/delete/manifest swap, even with checksum gate opted out). Godoc on both the sentinel and `isRetryable` states this.
- **Remedy documented where the operator looks:** error names path + bucket (`rewrite source %q … is outside bucket %s`), both gates log ERROR with `schema_id`/`path`/`bucket`, `design.md` §7.3 records "correct the manifest", CLI help and `CompactionConfig` godoc state the opt-out "waives hashing only".
- **Breaking scope as advertised:** only hand-built cross-bucket manifests; no writer produces them (CDC flush, `cdc-init`, compactor, reconcile all emit bucket-relative keys). No migration needed.

## Standards (`coding-standard.md` + `docs/error-handling.md`)

- **File ≤500 / function ≤100:** pass. `verify_inputs.go` 135, `rewrite.go` 215, `verify_foreign_test.go` 165, `verify_inputs_test.go` 288, `compactor.go` 336, `cmd/tools/compactor.go` 219, `config.go` 204. New functions are small (`verifyRewriteInputs` 9, `rejectForeignSources` 12, `verifySourceChecksums` 34, `bucketRelativeKey` 14 lines incl. signature). `runRewrite` is 96 lines (51–146) — back under the gate per the refactor commit, but with only 4 lines of headroom; the next addition there should extract a helper rather than grow it.
- **Error wrapping with context:** pass. Every new refusal is `fmt.Errorf("… %q …: %w", …, ErrForeignSource)` naming value type (rewrite source), offending path, and expected bucket; probe-failure and mismatch paths also now `%q`-quoted uniformly. Matchable via `errors.Is`, attributable without string comparison.
- **Error semantics:** pass. These are read-path/operator consistency errors (plain sentinels), not write-path 4xx carriers — no `InvalidInputf`/`NotFoundf` carrier expected, no `WithOperatorDetail` needed (nothing caller-unactionable is hidden; the bucket-qualified message is the published operator signal). `TestNoBareSentinelWraps` (root-package 4xx guard) not implicated.
- **Flat structure / nesting ≤3:** pass. Early return on first foreign source; `bucketRelativeKey` is linear prefix/empty checks; `deleteObjects` loop is flat.
- **Shared rule, no drift:** `deleteObjects` now resolves through `bucketRelativeKey` instead of its own inlined `TrimPrefix` check — rewrite gates and deleter cannot disagree on in-bucket membership. Exact `s3://<bucket>/` prefix match correctly refuses `s3://bktX/…` (pinned).
- **Logging contract:** both gates log their own ERROR (godoc states who logs), so standalone `verifySourceChecksums` callers get the same attributable line as the `runRewrite` path; production path logs once (scope wins, checksum never runs). `deleteObjects` WARN now names both shapes ("outside the compactor bucket or names no key") and carries `bucket`, fixing the round-2 mislabel of own-bucket empty keys.

## Findings (all checked, none blocking)

1. **Fail-open closed:** traced `runRewrite → verifyRewriteInputs → rejectForeignSources → verifySourceChecksums` — no path reaches `MergeToTmp`/`CopyTmpToFinal`/`saveManifestChecked` with a foreign or empty-key source. E2E test asserts `merger.sources`, `copies`, `deletes` empty and version unchanged.
2. **Empty-key extension is sound:** `"", "/", "s3://<bucket>/"` refused as out-of-scope (nothing to merge/hash/delete). `%q` quoting keeps `""` legible; pinned in `TestRejectForeignSources_EmptyKeyRefused`. Minor naming note (non-blocking): an own-bucket empty key surfaces as `ErrForeignSource` ("outside bucket") — the error/WARN text disambiguates ("or names no key"), and unifying under one sentinel keeps `isRetryable` and caller handling trivial. Acceptable.
3. **No over-broadening of the skip flag:** `SkipInputChecksumVerify` still disables only the hashing loop; scope gate is outside that `if`. Doc updates in both `config.go` and CLI help agree.
4. **No telemetry pollution:** scope refusal emits no `parquet_checksum_mismatch_total`; mismatch path still emits exactly one count with `schema_id` label. Pinned both ways.
5. **Path-rule edge cases reviewed:** relative (`p/…`, `/p/…`) → trimmed key; own-bucket absolute → relative key; foreign/prefix-collision (`s3://bktX/…`) → refused; `s3://bkt` (no trailing slash) → refused (cannot form a key — correct fail-closed); `s3://bkt//…` collapses to a non-empty key — same as pre-existing `objectURI` behavior, no new hole.
6. **`objectURI` passthrough is now documented-unreachable:** it still passes `s3://` through verbatim, but `runRewrite` never reaches it with a foreign path; godoc on `objectURI` states the precondition. No direct external callers of `objectURI` with unvalidated input were found (only `runRewrite`'s `sourceURIs` loop and the tmp-key render).

## Observations (non-blocking, for the record)

- **Promotion path intentionally ungated:** `applyPromotion` (delta→base relabel, manifest-only, no byte movement) does not call the scope gate. Correct — no DuckDB read, no splice-drop, no laundering; the listed name survives with the same path. Flagging only so a future reader does not "fix" it into a gate.
- **Double-gate log discipline holds:** in production only the first tripped gate logs (scope short-circuits before checksum); the standalone checksum ERROR exists for direct callers. No double-log in the `runRewrite` path.
- **Design-doc delta is additive:** §7.3 keeps the #347 narrative intact, appends the #417 refusal paragraph, and corrects the "only stamped sources inside the compactor's own bucket" reach statement without rewriting history. Format-tolerance note (#249 absolute-URI tolerance) is preserved while ruling out cross-bucket inputs as unsupported.
- **Commit hygiene:** four focused commits with issue-linked bodies explaining each review round (fold gates into `verifyRewriteInputs`, share `bucketRelativeKey`, quote paths, pin gate order). No unrelated refactors.

## Verification

- **Local (this review, `GOTOOLCHAIN=go1.26.0+auto` per Makefile pin):**
  - `go test ./internal/compaction -run 'TestRejectForeign|TestVerifySourceChecksums|TestVerifyRewriteInputs|TestCompactor_Rewrite|TestDeleteObjects' -v` — all pass (8 foreign/scope/delete + 8 checksum + 8 rewrite-gate tests).
  - `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1` — all `ok`.
  - `gofmt -l` on touched packages — clean; `go vet ./internal/compaction/...` — clean.
- **CI on the PR head (all SUCCESS):** Lint, Test, E2E Smoke Tests, k6 Smoke, Benchmark Smoke.
- **Standards spot-checks:** line counts above; `errors.Is`/`ErrorContains` assertions on path, bucket, and sentinel in every new refusal test; observer-log assertions on `path`/`schema_id`/`bucket`; prefix-collision and empty-key pins present.

## Assessment

The ruling is the right one (single-bucket compactor; every other guard already refuses what it cannot prove), and the implementation is the minimal fail-closed shape: one shared path predicate, one ordered pre-merge gate, terminal refusal, shared deleter resolution, and docs that tell the operator the remedy. Tests pin behavior, not implementation — scope-vs-checksum order, opt-out independence, terminality, and side-effect-freedom are all contracts now. The only watch item is `runRewrite` at 96/100 lines. Approve.

iceboundrock:
## Response to fourth-round review

One finding needed a correction and is fixed in `ea6a2f0`; the rest are already ruled or need no change. Nothing fell outside the PR's diff, so no follow-up issue was opened.

1. **Finding 5, `s3://bkt//…` "same as pre-existing `objectURI` behavior, no new hole".** Not quite: this PR had opened a small one. On `main`, `bucketRelativeKey`'s absolute branch returned the remainder after the bucket prefix verbatim, so `s3://bkt//x` resolved to key `/x`, the same object DuckDB reads through the unchanged URI. The `a6c38e4` rewrite of the helper (for the empty-key rule) applied the leading-slash trim to that branch as well, so the gate hashed key `x` while `MergeToTmp` read `s3://bkt//x`, i.e. key `/x`, and `deleteObjects` would have deleted `x` instead of `/x`. A checksum gate that verifies a different object than the merge reads is exactly the shape this PR exists to close, even though only a hand-built manifest can produce it. `ea6a2f0` restores the pre-PR semantics: the trim applies to relative paths only, an own-bucket URI keeps its key verbatim, and the empty-key refusal still covers `""`, `/` and `s3://bkt/`. The godoc now states the invariant explicitly: the helper must resolve exactly the object `objectURI` hands DuckDB. Pinned on both consumers: `TestVerifySourceChecksums_OwnBucketURIKeyPreservedVerbatim` asserts the GET lands on `/p/1/aaa.parquet` for `s3://bkt//p/1/aaa.parquet`, and `TestDeleteObjects_UsesBucketRelativeKey` gains an `s3://bkt//…` row that must delete `/p/1/ddd.parquet`. Both failed before the fix (the gate probed `p/1/aaa.parquet` and got no such key).

2. **`runRewrite` at 96/100 lines.** Acknowledged as the watch item; nothing to change now. The next addition there extracts a helper rather than growing the function. No issue filed, since there is no work until someone touches it.

3. **Own-bucket empty key surfacing as `ErrForeignSource`.** Unchanged, as ruled in rounds two and three: one category ("not a supported compaction input"), one remedy, one sentinel; the `%q` quoting and the "or names no key" WARN keep the text unambiguous.

4. **Promotion path ungated.** Agreed and unchanged. `applyPromotion` moves no bytes and drops no listed name, so a foreign entry relabelled delta→base is exactly as readable (or not) as it was before; the refusal belongs where bytes move.

5. **`objectURI` pass-through.** Unchanged, as ruled in round two; `TestVerifyRewriteInputs_ScopeRefusalPrecedesChecksum` from round three enforces the ordering the godoc relies on.

Verified locally after the change: `go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1` all ok, `make fmt-check`, `go vet ./internal/compaction/` and `make lint` (root and infra modules) clean. `verify_inputs.go` is 137 lines, `verify_inputs_test.go` 304, `verify_foreign_test.go` 166.

iceboundrock:
## Findings

1. **[P2] Correct the source-retention wording in the checksum-gate documentation.** `internal/cdc/config.go:86-94` still says a rewrite “merges its sources and then deletes them,” and the modified paragraph in `docs/federated-query/design.md:857` repeats that claim. `runRewrite` deliberately retains the sources after the manifest swap for in-flight readers (`internal/compaction/rewrite.go:20-47, 122-128`); only their manifest entries are spliced out. The gate remains the last attributable point because the named entry disappears from the manifest, not because its object is deleted. Please replace both statements with that accurate rationale.

## Specification

The implementation otherwise conforms to the #417 ruling and its recorded plan:

- `verifyRewriteInputs` runs before `MergeToTmp`, performs the bucket-scope refusal before checksum verification, and applies it to stamped and unstamped entries.
- The refusal is independent of `SkipInputChecksumVerify` and `ObjectReader`; the checksum opt-out therefore only waives hashing.
- `ErrForeignSource` is outside `isRetryable`'s allowlist, so invalid manifests fail once rather than entering backoff.
- The checksum gate preserves defence in depth when invoked directly, and `bucketRelativeKey` correctly rejects foreign-bucket prefix collisions and empty keys.

## Standards

- Error returns include schema/path/bucket context and wrap matchable sentinels.
- The touched functions and files remain within the project's size limits; the new helper centralizes path resolution without duplication.
- The stale retention wording above is an immediate-area documentation inconsistency and should be corrected with this change.

## Assessment

**Changes requested** for the P2 documentation correction. The runtime fix is otherwise sound and closes the fail-open path described in #417: a foreign manifest source is now rejected before it can be merged or spliced from the manifest.

## Observations

- The new tests cover foreign stamped/unstamped inputs, checksum opt-out, nil reader, own/foreign URI resolution, empty keys, gate ordering, deletion resolution, and the `RunOnce` no-merge/no-retry path.
- Relative manifest paths beginning with `/` remain a pre-existing normalization asymmetry: `manifest.QuerySource.Paths` preserves the slash while compaction's `objectURI` trims it. This PR does not introduce that behavior, so it is not a blocking finding for #515; a separate end-to-end normalization test would make the contract explicit.

## Verification

- Passed locally: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction`
- Passed locally: `make fmt-check`
- CI at review time: Lint, Test, k6 Smoke, and Benchmark Smoke passed. E2E Smoke Tests was still running.
- `git diff --check 7db5822..ea6a2f0c058ec6c3d11491d8e087db6e0f82580d` reported no whitespace errors.

## Summary

One P2 documentation correction is required; no runtime correctness issue was found in the foreign-bucket refusal itself.

iceboundrock:
## Response to fifth-round review

The P2 finding is fixed in `b07217d`. The observation is real, pre-existing, and reachable from a shipped writer, so it is filed as #516 rather than folded into this PR.

1. **[P2] Source-retention wording.** Confirmed stale in both places: `internal/cdc/config.go` (`SkipInputChecksumVerify` godoc) and `docs/federated-query/design.md` §7.3 both said a rewrite "merges its sources and then deletes them", which has been false since #461. Both now say the rewrite folds the sources into one new base and splices their entries out of the manifest, the objects are retained for in-flight readers (#461) and reclaimed later by `manifest-reconcile --gc`, and the gate is the last attributable point because the *listed name* disappears, not the bytes. No other occurrence of the old claim remains in `internal/`, `cmd/` or `docs/` (grep for "then deletes them" is empty). The CLI help for `--skip-input-checksum-verify` never made the claim, so it is unchanged.

2. **Leading-slash normalization asymmetry.** Verified and worse than "hand-built only": `cdc.Build*Path` with an empty data prefix emits `/<schema>/<uuid>.parquet`, the flush stores that literal key, `QuerySource.Paths`, `QuerySource.MissingIn` and reconcile's `normalizeKey` all keep the slash (reconcile's godoc records why), while compaction's `objectURI` and the #417 `bucketRelativeKey` trim it, so compaction resolves a sibling key. Today that fails closed (probe failure or a missing-object merge error), but the error names a key that was never in the manifest, and such a schema can never be rewritten. This PR's helper deliberately mirrors `objectURI` so the gate and the merge agree with each other; making compaction agree with the readers is a separate contract decision with its own test, so it is #516, with the four consumers, the reachability, and a proposal (verbatim key contract, drop the trim, pin with a cross-component normalization test plus an empty-prefix e2e) recorded there.

Verified locally after the change: `make fmt-check` clean, `go build ./internal/cdc/ ./internal/compaction/ ./cmd/tools/` ok, `go test ./internal/cdc ./internal/compaction -count=1` ok (`GOTOOLCHAIN=go1.26.0+auto`). Docs and a godoc comment only; no behaviour change.

iceboundrock:
# PR #515 Review: refuse foreign-bucket rewrite sources

Artifact: consolidated PR review (findings / specification / standards / assessment / observations / verification / summary). Working copy: this comment only; no local planning artifact. Reviewed HEAD `b07217d`.

## Findings

1. **[P2] Correct the stale source-retention rationale in `internal/compaction/verify_inputs_test.go:241-244`.** The comment above `TestCompactor_Rewrite_MismatchingSourceRefusesTheMerge` says a rewrite “merges the sources and then DELETES them.” That has been false since #461: `runRewrite` retains source objects for in-flight readers and only splices their entries from the manifest (`internal/compaction/rewrite.go:39-46,122-128`). The PR correctly fixes this wording in `CompactionConfig` godoc and design.md, but it modifies this same test file while leaving the third contradictory description. Please state that the gate is the last attributable point because the manifest-listed name disappears, while the object remains for `manifest-reconcile --gc`. The test behavior is correct; this is an immediate-area documentation consistency fix.

## Specification

- The #417 ruling is otherwise implemented correctly: `verifyRewriteInputs` executes before `MergeToTmp`; `rejectForeignSources` refuses stamped and unstamped foreign paths independently of `SkipInputChecksumVerify` and `ObjectReader`.
- Scope precedes checksum verification, and a foreign source is terminal because `ErrForeignSource` is outside the `isRetryable` allowlist.
- The standalone checksum gate is fail-closed for a stamped foreign URI; `bucketRelativeKey` correctly rejects prefix collisions and empty keys.
- CLI help, config godoc, and design.md consistently define the opt-out as waiving hashing only.

## Standards

- New error paths wrap matchable sentinels with schema, path, and bucket context. They are correctly plain operator-visible consistency errors, not 4xx carriers.
- Touched Go files and functions remain within the repository size limits; the shared path resolver prevents the deletion and verification rules from drifting.
- The stale test comment above is the sole standards/documentation gap found.

## Assessment

**Changes requested** for the P2 comment correction. The runtime change itself is minimal and sound: it closes the fail-open path before merge, promotion, or manifest mutation, preserves no-retry semantics, and does not misclassify a scope refusal as a checksum mismatch.

## Observations

- Leading-slash manifest-key normalization remains a pre-existing cross-component inconsistency; it is correctly out of scope here and is tracked in #516.
- All foreign-source gate tests assert the important contracts: scope independent of checksum opt-out, scope-before-checksum order, no checksum telemetry on a scope refusal, exact own-bucket URI resolution, and no rewrite side effects or retries.

## Verification

- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction -count=1`.
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc ./cmd/tools/... -count=1`.
- Passed: `make fmt-check`; `git diff --check main...HEAD`.
- PR CI is green at review time: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

One P2 documentation correction is required. No runtime correctness defect was found in the foreign-bucket refusal.

iceboundrock:
## Response to sixth-round review

The P2 finding is fixed in `4541c0a`. The observation is already tracked in #516, so no new follow-up issue was opened.

1. **[P2] Stale "then DELETES them" comment in `verify_inputs_test.go`.** Confirmed: the comment above `TestCompactor_Rewrite_MismatchingSourceRefusesTheMerge` was the third copy of the pre-#461 claim, and `b07217d` had corrected only the other two. It now gives the same rationale as the `SkipInputChecksumVerify` godoc and design.md §7.3: the rewrite folds the sources into one new base and splices their entries out of the manifest, the objects are retained for in-flight readers (#461) and reclaimed later by `manifest-reconcile --gc`, so the gate is the last point at which corrupt bytes are still attributable to a listed name. The test body and its assertions are unchanged; `require.Empty(t, s3c.deletes, ...)` still holds as a side-effect-freedom check. A grep for "then delet" / "deletes them" across `internal/`, `cmd/` and `docs/` now returns nothing outside unrelated `deleted_at` references.

2. **Leading-slash normalization asymmetry.** Agreed it is out of scope here; #516 carries the four consumers, the reachability via `cdc.Build*Path` with an empty data prefix, and the proposed verbatim-key contract with its cross-component test.

Verified locally after the change: `gofmt -l internal/compaction/` empty, `go test ./internal/compaction -count=1` ok (`GOTOOLCHAIN=go1.26.0+auto`). Comment only; no behaviour change.

iceboundrock:
## Review

### Findings

- [P2] Describe the empty-key refusal accurately: [`internal/compaction/verify_inputs.go:62-65`](https://github.com/Lychee-Technology/forma/blob/4541c0abe384b447f25b1f0f3e2b1d62892fd05b/internal/compaction/verify_inputs.go#L62-L65) reports every `bucketRelativeKey` failure as a source outside the compactor bucket. However, this helper intentionally returns `false` for `s3://bkt/`, `/`, and `""`; those paths are in the configured bucket (or relative) but resolve to no object key. The returned CLI error and its ERROR log therefore tell the operator the wrong cause and imply the wrong repair. `deleteObjects` already uses the correct combined wording. Make the gate wording likewise state that the path is outside the bucket **or does not resolve to a non-empty key**, and pin that message for the empty-key cases. Keeping `ErrForeignSource` as the matchable terminal sentinel is fine if its public text/doc is broadened to cover the malformed-key case.

### Specification

- The implementation follows the #417 ruling: all rewrite inputs are scope-checked before merging, independently of checksum stamping, `ObjectReader`, and `--skip-input-checksum-verify`.
- `verifyRewriteInputs` correctly orders scope validation before checksum validation. A foreign source cannot reach `MergeToTmp`, a manifest swap, promotion, or retry backoff.
- The shared resolver preserves an own-bucket URI key verbatim, including an intentional leading slash after the bucket, so the verifier/deleter address the same object DuckDB reads.
- The config/flag documentation and federated design record the opt-out boundary and the retained-source semantics consistently.

### Standards

- Apart from the finding above, the changed Go code meets the repository standards: exported sentinel is documented, errors retain contextual wrapping, production files/functions remain within size limits, and no formatting or whitespace errors were found.
- The design change belongs in the durable federated design document; the issue contains the ruling and implementation record as required.

### Assessment

Changes requested for the inaccurate empty-key diagnostic. The fail-closed behavior itself is correct and the defect does not permit a merge.

### Observations

- Coverage includes unstamped foreign paths, checksum opt-out/nil-reader independence, exact bucket-prefix matching, no-GET scope precedence, terminal retry behavior, no side effects through `RunOnce`, and the double-slash URI key case.
- The negative empty-key cases currently assert rejection and quoted path only, which is why the misleading reason remains unprotected.

### Verification

Passed locally:

```text
git diff --check main...HEAD
make fmt-check
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction -count=1
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/compaction -count=1
GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction ./internal/cdc ./internal/reconcile ./cmd/tools/... -count=1
```

CI status at review time: `Lint`, `Test`, and `Benchmark Smoke` passed; `E2E Smoke Tests` and `k6 Smoke` were still pending.

### Summary

The PR closes the foreign-bucket fail-open path with the intended fail-closed gate and has focused regression coverage. Correct the empty-key error/log wording before merge so the new refusal gives operators an accurate remediation path.

iceboundrock:
## Review

### Findings
No blocking findings. I found no correctness, safety, or behavioral-regression issue in the final PR diff.

### Specification
- The implementation satisfies the #417 ruling: every rewrite source must resolve to a non-empty key in the compactor bucket before merging starts.
- The scope gate runs before checksum verification and is independent of --skip-input-checksum-verify and ObjectReader. A foreign or empty-key source therefore cannot be merged, promoted, or spliced out of the manifest.
- ErrForeignSource is terminal under the existing retry allowlist, preserving the intended one-refusal operator workflow.
- The shared bucketRelativeKey preserves an own-bucket URI key verbatim while normalizing only relative paths, so checksum verification and DuckDB read the same object.

### Standards
- Changed Go files and functions remain within the repository 500-line and 100-line limits.
- New errors carry path, schema, and bucket context and retain a matchable sentinel via %w.
- The change centralizes path-scope logic for rewrite validation and best-effort deletion, avoiding divergent bucket checks.
- Documentation now correctly reflects #461: rewrite inputs become unlisted after the manifest swap but are retained for in-flight readers until later GC.

### Assessment
Approve. The change closes the identified fail-open path without expanding compaction to cross-bucket reads or changing normal same-bucket rewrite behavior.

### Observations
- Coverage exercises foreign stamped and unstamped entries, own-bucket and prefix-collision URIs, empty keys, checksum-opt-out behavior, gate ordering, terminal retry behavior, no merge, promotion, or manifest swap on refusal, and the double-slash own-bucket URI key case.
- No blocking test gaps identified. The local review did not run container-backed end-to-end tests; the PR CI E2E Smoke Tests are passing.

### Verification
- GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/compaction -count=1 passed.
- make fmt-check passed.
- git diff --check origin/main...HEAD passed.
- GitHub CI checks are all successful: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

### Summary
No requested changes. The PR is ready to merge.

PR #517, opened 2026-09-05
fix(cdc,sqlgen,tools): keep legacy scalar rows under list attributes instead of exporting [] (#372)

### LINKED ISSUE #372: Legacy scalar EAV rows under list metadata export/pivot as empty list

iceboundrock:
Fixes #372.

## Problem

An EAV row with `array_indices = ''` that carries a value under an attribute whose metadata now says `valueType: list` was silently exported and pivoted as `[]`. Both list aggregates (`internal/cdc/export_sql_builder.go` and `sqlgen.buildEAVPivotExpr`) filtered elements with `array_indices <> ''` while the presence count still saw the row, so `coalesce(list(...) FILTER (...), [])` fired and the stored scalar was dropped. The OLTP path still returned it. Such rows predate the write path's array enforcement (#314).

## Change

- **Both aggregates exclude only the empty-list marker** (`array_indices = ''` with both value columns NULL, #204) instead of every `array_indices = ''` row. A legacy scalar now exports and pivots as a one-element list on every DuckDB tier. Explicit `[]` and absent (`NULL`) keep their semantics.
- **`validate-schema-consistency` gains a failing check**, `scalar rows stored under list attributes in <eav_table>`, naming schema, attribute and row count. The marker row is excluded by the query itself; rows under scalar or unknown attributes are left to the existing checks.
- **Migration guide** documents the finding, the read behaviour until fixed (OLTP returns `"alpha"`, DuckDB tiers `["alpha"]`; before this PR `[]`), the rewrite SQL (`array_indices '' → '0'`, guarded against an existing element `'0'`), and that a direct SQL rewrite does not stamp `change_log`.
- `buildEAVPivotExpr` and its helpers move from `duckdb_schema_projection.go` into `duckdb_eav_pivot.go` (pure move) to stay under the 500-line limit.

## Why not reject at export/pivot time

The export is a per-schema CDC flush and the pivot is the hot leg of every federated read for that schema. Failing either over one legacy row blocks all of that schema's data movement or reads. This follows the #384 precedent for legacy rows current metadata cannot hold: answer consistently across DuckDB tiers, gate new writes, surface the drift through the validator. The ruling is recorded on #372.

## Tests

- New DuckDB-executed tests at both sites pin the four shapes: absent → `NULL`, marker → `[]`, elements → ordered list, legacy scalar → one-element list (`internal/sqlgen/duckdb_list_pivot_legacy_test.go`, `internal/cdc/export_list_legacy_scalar_test.go`). Both failed on the legacy case before the fix.
- Validator tests cover the flagged case and the ignored cases (scalar attribute, unknown attr_id, unknown schema_id); existing mocks gain the fifth query expectation.
- `make fmt-check`, `make lint`, and the unit suite via `scripts/test_with_container_runtime.sh` pass.


iceboundrock:
## Review of #517 (fixes #372)

### Assessment

No blocking findings. The change implements the ruling recorded on #372 faithfully (with the one declared deviation: the validator always issues its census query so the tool's query set stays fixed), fixes the silent `[]` value loss at both aggregate sites with an identical filter, adds a failing validator check, and documents the migration. CI is fully green (Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke). Ready to merge from a correctness and standards standpoint; the two items under Observations are non-blocking documentation nits/questions.

### Spec check (against #372 and the recorded ruling)

- **Both aggregates exclude only the empty-list marker** — `internal/cdc/export_sql_builder.go` and the moved `buildEAVPivotExpr` (`internal/sqlgen/duckdb_eav_pivot.go`) now filter elements with `(array_indices <> '' OR value_text IS NOT NULL OR value_numeric IS NOT NULL)`. The two SQL strings match each other exactly as required for hot/parquet type-unification. Pass.
- **Four shapes pinned at both sites** — absent → `NULL`, marker → `[]`, elements → ordered list, legacy scalar → one-element list — via DuckDB-executed tests (`TestExportListAggregateShapes`, `TestBuildEAVPivotExprListShapes`). Both tests fail on the legacy case without the fix per the PR description, and pass with it (re-ran locally). Pass.
- **Validator gains a failing check** — category `scalar rows stored under list attributes in <eav_table>`, one line per `(schema_id, attr_id)` with schema/attribute/row count; marker excluded by the query itself (`array_indices = '' AND (value_text IS NOT NULL OR value_numeric IS NOT NULL)`); scalar/unknown schema/attr rows left to the existing censuses (explicitly tested). Fails the run (not informational). Pass.
- **Migration guide** — new `#372` section covers the finding, the read behaviour until fixed (OLTP `"alpha"` vs DuckDB `["alpha"]`, previously `[]`), the guarded rewrite (`'' → '0'` with `NOT EXISTS` against a colliding `'0'`), and the `change_log` staleness note including pre-`#372` parquet holding `[]`. Pass.
- **"Why not reject at export/pivot"** — rationale (per-schema flush / hot leg of every federated read; #384 precedent) is in the PR body and the #372 ruling. Pass.

### Findings

None blocking. I verified the `duckdb_schema_projection.go → duckdb_eav_pivot.go` move is pure except the list branch: `eavElementCastExpr`, `eavValueColumn`, `mainColBoolExpr` are byte-identical, and the `buildEAVPivotExpr` diff touches only the list comment and the `FILTER` clause.

### Standards check (`coding-standard.md`, error semantics)

- **File/function limits**: `duckdb_schema_projection.go` drops to 396 lines, new `duckdb_eav_pivot.go` is 114, new `validate_schema_consistency_lists.go` is 92; longest new function is `checkScalarRowsUnderListAttrs` at ~61 lines. Pass.
- **Error handling**: new validator errors are `fmt.Errorf("...: %w", err)` plain errors with query/scan/iterate context — correct class for a read-path operator tool (not `InvalidInputf` carriers, no bare `%w`-only sentinel wraps). Pass.
- **Structure/nesting**: early returns, max nesting 2 in the new check. Pass.
- **gofmt/lint**: `gofmt -l` clean on all touched/new Go files locally; CI Lint passes. Pass.
- **Tests**: new behaviour pinned by execution (not just string matching) at both SQL sites; validator covers flagged + ignored cases; existing mocks updated with the 5th census expectation. Pass.

### Observations (non-blocking)

1. **Mixed legacy + element rows**: `TRY_CAST('' AS BIGINT)` is `NULL` and DuckDB `ORDER BY ... ASC` sorts it last (verified locally: `('legacy',''),('e0','0'),('e1','1')` → `["e0","e1","legacy"]`). So an entity carrying both a legacy scalar and real elements pivots/exports with the legacy value appended at the end — deterministic but positionally arbitrary. The migration guard (`NOT EXISTS ... '0'`) deliberately leaves such entities flagged rather than rewriting them, so this shape persists until a per-entity decision. Consider one sentence in the migration section naming the mixed-entity read shape. Not a merge blocker.
2. **Cross-column legacy rows**: the "one-element list" guarantee holds when the value sits in the expected storage column. A legacy text value under a numeric list (or vice versa) passes the new `FILTER` via the `OR` but aggregates the other column, yielding `[NULL]` (verified locally on DuckDB). That is still not silent-`[]` loss, and the storage-column checks flag it separately, but the doc sentence "exports and pivots as a one-element list" is slightly unconditional. An "assuming the value is in the correct column" qualifier would close the gap. Not a merge blocker.

### Verification

- CI on this PR: Lint, Test, E2E Smoke Tests, k6 Smoke, Benchmark Smoke — all `SUCCESS`.
- Local (GOTOOLCHAIN=go1.26.0+auto, per AGENTS.md): `go test ./internal/sqlgen/ -run TestBuildEAVPivot` (3 pass, incl. new legacy-shapes test), `go test ./internal/cdc/ -run 'TestExportListAggregateShapes|TestBuildSchemaDrivenProjection_ListAttr'` (2 pass), `go test ./cmd/tools/ -run TestValidateSchemaConsistency` (9 pass, incl. 2 new), `gofmt -l` clean on all 7 touched/new Go files.
- Move-purity check: extracted-function diff confirms helpers identical, pivot diff limited to list branch (see Findings).
- Behaviour probes (temporary DuckDB tests, removed afterwards): `''` casts to `NULL` and sorts last in the ordered aggregate; wrong-column legacy aggregates to `[NULL]` (Observation 2).

### Summary

Approve-worthy: value loss fixed consistently across all DuckDB tiers, drift surfaced as a failing pre-flight with a documented guarded rewrite, standards and test bars met. Suggested follow-ups are doc-only (mixed-entity shape, correct-column qualifier) and can land separately.

iceboundrock:
## Response to the review (both observations addressed in 90a0468)

Both observations are doc-only and inside this PR's scope, so they land here rather than as follow-ups. I re-ran the two behaviour probes against `buildEAVPivotExpr` before editing; both reproduce.

**Observation 1, mixed legacy + element rows.** Confirmed: `('legacy',''),('e1','1'),('e0','0')` pivots as `["e0","e1","legacy"]`. `TRY_CAST('' AS BIGINT)` is NULL and DuckDB's ordered aggregate places it after every element index. The migration section now has a paragraph after the guarded `UPDATE` that names this read shape, explains why the legacy value lands last, and says the position is deterministic but carries no meaning. It sits after the SQL block because it only applies to the entities the `NOT EXISTS` guard leaves alone.

**Observation 2, cross-column legacy rows.** Confirmed: a text value under a numeric list (and the reverse) passes the `FILTER` via the `OR` and aggregates to `[null]`. I also checked that the storage-column checks do flag such a row: `attributeUsesNumericColumn` resolves list attributes through `EffectiveItemsType()`, so a text-in-`value_text` row under a numeric list is reported under "numeric/date/bool attributes stored in value_text" in addition to the new `#372` check. The "one-element list" sentence in the guide now carries the correct-column qualifier, names the `[null]` shape, and tells the operator to fix the column first and then promote the row.

**No follow-up issue.** Neither shape is a code gap this PR should leave behind: both are pre-existing drift that the validator now reports (double-reported for the cross-column case), and the guide documents the read shape and the fix order. The aggregate itself is unchanged.

Diff of the doc change: https://github.com/Lychee-Technology/forma/pull/517/commits/90a04685477a3718eb7d3dfe354e5741f333f1e4

iceboundrock:
## Review of #517

Reviewed current HEAD `90a0468`.

### Findings

None. I found no blocking correctness, regression, security, error-handling, or standards issues.

### Spec Check

- Both CDC and hot-tier DuckDB list aggregates now exclude only the explicit empty-list marker, while retaining value-carrying legacy scalar rows. The SQL predicates remain mirrored across [`internal/cdc/export_sql_builder.go`](internal/cdc/export_sql_builder.go) and [`internal/sqlgen/duckdb_eav_pivot.go`](internal/sqlgen/duckdb_eav_pivot.go).
- The four relevant shapes are executed and pinned at both aggregate sites: absent -> `NULL`, empty marker -> `[]`, indexed elements -> ordered list, and legacy scalar -> one-element list.
- `validate-schema-consistency` adds a failing check for scalar rows under current list attributes, excludes marker rows in SQL, and leaves scalar attributes and unknown schema/attribute IDs to the existing checks.
- The migration guide documents the finding, guarded `'' -> '0'` rewrite, primary-key collision case, `change_log` staleness, mixed-entity ordering, and wrong-storage-column behavior.
- The helper move from `duckdb_schema_projection.go` to `duckdb_eav_pivot.go` is behavior-preserving apart from the intended legacy-row filter change.

### Standards Check

- All touched Go files remain below the 500-line limit; the new validator and pivot helpers remain well below the 100-line function limit.
- New database errors include operation context and wrapped causes.
- The new code uses early returns and stays within the repository's nesting guidance.
- `gofmt`, `make fmt-check`, and `make lint` pass, including the `infra/` module lint run.
- Tests validate executed DuckDB behavior rather than only matching generated SQL strings, and validator mocks cover both flagged and ignored cases.

### Assessment

Approve-ready from a correctness and repository-standards standpoint. The implementation fixes the silent value loss consistently across CDC export and hot-tier reads, surfaces remaining metadata drift as a pre-flight failure, and provides an operator-safe migration path.

### Observations

Non-blocking verification gaps remain:

- The new legacy-row tests execute the aggregate fragments against DuckDB, but do not add a full Postgres -> CDC -> Parquet -> federated round-trip fixture for this historical shape.
- Mixed legacy-plus-element rows and wrong-column legacy values are documented and covered by the aggregate logic/validator classification, but are not exercised as full end-to-end scenarios.

The current documentation explicitly describes both cases, so these are residual coverage observations rather than implementation findings.

### Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/cdc ./cmd/tools` passed.
- Targeted DuckDB aggregate and list-unification tests passed.
- `make fmt-check` passed.
- `make lint` passed for the root and `infra/` modules.
- A direct `make test` initially could not start the unconfigured container runtime; rerunning via `./scripts/test_with_container_runtime.sh` configured the Podman socket and the full unit suite passed, including `internal/e2e_harness`.
- PR checks are green: Test, Lint, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Summary

No findings. The PR satisfies the stated specification and coding standards, with only the non-blocking end-to-end coverage observations above.

Issue #481, opened 2026-08-24
bug(reconcile): #463 guard is bypassed by a non-empty but foreign manifest (mis-pointed template resolving to an existing wrong file)


iceboundrock:
## Problem

The #463 empty-manifest GC guard (`refuseEmptyManifestGC`, PR #480) keys on the **raw** loaded entry count: `d.manifestEntries = len(m.Files)`, and any value > 0 waives the guard. It does not ask whether those entries have anything to do with the schema being collected. A *sibling* misconfiguration therefore reaches the exact irreversible outcome #463 closed off:

1. `--manifest-template` omits `{{.SchemaID}}` (e.g. `manifest/data.json`) or otherwise resolves every schema to one existing file, or points at another schema's / another tier's manifest — while `--data-prefix` still matches the writers'.
2. Every schema loads a **non-empty but wrong** manifest. `manifest.LoadOrCreate` does not validate the loaded `Manifest.SchemaID` against the requested one (verified: `internal/manifest/manifest.go` `LoadOrCreate` returns whatever `Load` found), and `ResolverManifestStore.Load` adds no check either.
3. In `diffSchema`, the foreign entries fall outside `data/<schemaID>/` → classified `unverifiable`; every real base object for the schema classifies as an orphan by shape.
4. `manifestEntries > 0` ⇒ the #463 guard does not fire ⇒ after the two-phase sighting grace, a cron `--gc` deletes the schema's entire base tier.

The report shows the anomaly (`N objects, M manifest entries` alongside a wall of `unverifiable entry:` lines and `base orphan:` lines), but nothing refuses.

## Root cause

"Zero resolvable manifest entries" (#463's trigger) is a special case of the real invariant: **the manifest this run loaded accounts for none of this schema's data**. The guard currently tests the special case only.

## Proposed direction (maintainer to rule)

Two complementary options, either or both:

[...]

Issue #463, opened 2026-08-24
bug(reconcile): manifest-reconcile --gc deletes the whole base tier when the manifest fails to resolve (empty manifest ⇒ every object an orphan)


iceboundrock:
## Problem

`manifest-reconcile --gc` treats "object present in storage but not in the manifest" as an orphan, with no sanity check that a schema showing N live parquet objects and **zero** manifest entries is far more likely a manifest-resolution failure than N genuine orphans. A mis-pointed `--manifest-prefix`/`--manifest-template` (while `--data-prefix` still matches) therefore classifies the entire base tier as orphaned and, after the grace, deletes it — irreversibly.

## Current behavior

`ResolverManifestStore.Load` (`internal/reconcile/manifests.go`) has LoadOrCreate semantics: a schema whose manifest cannot be resolved yields an **empty** manifest. `collectAndGC` (`internal/reconcile/reconcile.go`) then assembles GC candidates purely as `baseInitOrphans + baseMergedOrphans + tmpOrphans + deltaLeftovers`, i.e. class ∈ {baseInit, baseMerged, tmp} ∧ not-in-manifest — so with an empty manifest every base object qualifies. `confirmDangling` re-verifies, but it reloads through the *same* resolver, so a mis-pointed template stays empty and the candidates stay "confirmed." There is no guard of the form "objects present ∧ manifest entries == 0 ⇒ refuse and fail." The two-phase sighting grace (default 15 min) only delays; a cron satisfies both sighting and age conditions on a later pass.

## Reproduction

1. `forma-tools manifest-reconcile --gc` with a `--manifest-prefix`/`--manifest-template` that does not match the writers' (a typo, or shared-prefix drift), while `--data-prefix` does match.
2. `LoadOrCreate` yields an empty manifest; every listed base/merged object classifies as orphan.
3. After the sighting + age grace, `gcSchema` deletes every base and merged-base object for every enumerated schema.

Note the asymmetry that confines the damage to (and completes it for) the base tier: delta orphans only become GC candidates under `--repair`.

## Root cause

Orphan classification cannot distinguish "the manifest genuinely lists nothing" from "the manifest failed to resolve." The read side of the same misconfiguration is documented as unsafe in `docs/federated-query/design.md` §4.3.1 ("the fallback is not a general recovery path for schemas that already have data"); the write/GC side has no equivalent guard.

## Impact

[...]

PR #518, opened 2026-09-05
fix(cdc-init,federated): refuse over a populated delta tier, purge with --replace-delta, pin drifted attribute types (#371)

### LINKED ISSUE #371: cdc-init only replaces the base tier; delta files survive re-init

iceboundrock:
Closes #371.

## What

`cdc-init` only replaced the manifest's **base** tier, so delta parquet written under an older attribute type survived a re-init and poisoned federated reads. Incompatible types (`text`→`list`, #315) failed with a DuckDB Conversion Error; compatible ones drifted quietly (an INTEGER attribute with one VARCHAR generation binds as VARCHAR and sorts lexicographically).

Plan and rulings are on the issue: [plan](https://github.com/Lychee-Technology/forma/issues/371#issuecomment-5550319875), [rulings part 1](https://github.com/Lychee-Technology/forma/issues/371#issuecomment-5550375355), [rulings part 2](https://github.com/Lychee-Technology/forma/issues/371#issuecomment-5550514736).

### cdc-init: refuse by default, purge with `--replace-delta`

- Per schema, under the existing advisory lock and before any export, init inventories the delta tier: manifest delta entries plus delta-shaped objects (`<uuid>.parquet`) directly under `--delta-prefix/<schemaID>/` (default prefix `delta`, matching `cdc-flush --s3-prefix`; empty disables the listing).
- Non-empty inventory without the flag fails that schema with the named sentinel `cdc.ErrDeltaTierPresent` (also in dry-run); other schemas continue.
- With `--replace-delta` the manifest publish is one CAS: base tier replaced, delta tier emptied, re-spliced on a confirmed 412 with late delta entries joining the purge set. Only after the save commits are the delta objects deleted (`publishAndPurge`). Delete failures name the undeleted keys; the published manifest is not rolled back. Dry-run logs the purge set and deletes nothing.
- Unnormalizable manifest entries (globs, foreign-bucket `s3://`) fail closed before export. Schemas with zero live rows take no swap and leave delta alone (warning).
- Deletion is inline rather than deferred to `manifest-reconcile --gc`: after a re-init every tombstone in a stale delta is uncovered and dead in Postgres, which `--repair` classifies as a lost delete and appends back. Documented in `docs/manifest-reconcile.md`.
- New `S3InitClient` (adds `ListObjectsV2`); `NormalizeObjectKey` is exported from `cdc` and reused by `reconcile`.

### Federated read: selective attribute type pin (acceptance a)

- The pre-read validator now tracks columns that carry more than one parquet type across the scan set (`columnUnion`).
- `coldScanColumns` returns missing (#255, unchanged) and **pinned** attribute columns: mixed across files, or union type differing from `sqlgen.DuckDBNullScanType(meta)`. Pinned columns render `CAST(col AS type) AS col` inside the same `* REPLACE (…)` list as the system-column guard.
- Selective: a healthy scan set renders byte-identical SQL. VARCHAR and UUID are compatible (column-bound `uuid` exports parquet UUID). Pinned set is a separate `cold-pinned` plan-scope component.
- A value the pin cannot convert fails with DuckDB's own Conversion Error on reads that consume the column, the same scope as the `changed_at` re-pin.

### Tests (acceptance b)

- `internal/sqlgen`: real-DuckDB proof that an INTEGER attribute with a VARCHAR generation orders lexicographically unpinned and numerically pinned; garbage fails loudly when pinned; healthy set renders no pin.
- `internal/cdc`: refusal (listed / unlisted delta-shaped, base-shaped ignored), dry-run refusal, merged purge set, delete-after-save ordering, 412 retry capturing late delta, delete failure semantics, unnormalizable entries, wiring seams with paged listing.
- `internal/federated`: pin detection for drifted and mixed types, scope parts for the pinned set, validator union.
- `cmd/tools`: flag mapping for `--replace-delta` / `--delta-prefix`.
- Production e2e: `TestCDCFlushAndInit` asserts the refusal, then the purge (manifest delta empty, objects gone); fixtures that used to init over a populated delta tier are reordered to init first; `RunInitKeepingDelta` for the remaining opt-in cases.

### Docs

`docs/federated-query/design.md` (scan-guard comment + a `#371` paragraph), `docs/manifest-reconcile.md` (new cdc-init section), `README.md`, `tests/e2e/README.md`, `docs/e2e-tests-*.md`. The Bun `cdc-init.ts` passes `--delta-prefix` and forwards `--replace-delta`.

## Verification

- `./scripts/test_with_container_runtime.sh` (`make test`): all 34 packages ok.
- Production e2e (`-tags=e2e`, `-run 'TestCDCFlushAndInit|TestDryRunImmutability|TestInit|TestThreeTier|TestCompactionRewrite|TestWarmColdTiebreak|TestMonotonicVersions|TestManifestReconcile|TestRecreateBetweenExportAndMark'`): 27 tests pass (one fixture bug found and fixed in `TestThreeTierMerge/changed_at_beats_tier_layout` during this run).
- Not run: the Bun `tests/e2e` pipeline (no Bun pipeline in this environment; the script change is argument plumbing only).


iceboundrock:
## Review of #518 — cdc-init refuse/purge + cold-scan attribute pin (#371)

### Spec (what #371 asked vs what this delivers)

#371 has two halves, and both are covered:

1. **Operator half — `cdc-init` must not silently leave stale delta under a new base.**
   - Refuse-by-default with named sentinel `cdc.ErrDeltaTierPresent` (`internal/cdc/init_delta.go:22`), raised in `preflightDeltaTier` (`init_delta.go:94`) before any row is counted/exported, in dry-run too. Other schemas continue via the existing per-schema `errors.Join`. Matches the plan's ruling (refuse converts quiet poisoning into a loud operator decision).
   - Opt-in purge with `--replace-delta` / `--delta-prefix` (`cmd/tools/cdc_init.go:94`): manifest publish is one CAS replacing base **and** emptying delta (`internal/cdc/init_publish.go:71`), re-spliced on confirmed 412 with late delta entries joining the purge set (`init_publish.go:88`). Delete runs only after the save commits (`internal/cdc/init.go:312` `publishAndPurge`), delete failures name keys and leave the published manifest in place (`init_delta.go:195`). Unnormalizable entries (globs, foreign-bucket `s3://`) fail closed before export. Zero-live-rows takes no swap and leaves delta with a warning (`init.go:279`). Inventory covers manifest entries **plus** delta-shaped objects under the prefix (`init_delta.go:143`), with base-shaped / `_tmp/` / non-UUID correctly excluded (`init_delta.go:81`). This is exactly the swap-then-delete ordering the issue discussion required, with the documented departure (inline delete instead of `--gc`-deferred, because stale tombstones would be re-appended by `--repair` as lost deletes).
2. **Reader half — selective attribute type pin (acceptance a).**
   - Validator tracks multi-type columns as a struct (`internal/federated/column_union.go:26`, first-seen-wins + `mixed` set, order-independent flagging). `coldScanColumns` (`internal/federated/cold_columns.go:49`) pins only when `isMixed` or union type disagrees with `sqlgen.DuckDBNullScanType(meta)`, with the VARCHAR/UUID compatibility carve-out (`internal/sqlgen/duckdb_cold_scan.go:196`). Render is `CAST(col AS type) AS col` inside the existing `* REPLACE` list (`duckdb_cold_scan.go:173`); healthy sets render byte-identical SQL. Pinned set is its own `cold-pinned` scope component (`internal/federated/duckdb_query_build.go:207`), and incomplete unions drive neither augmentation nor pin (`internal/federated/duckdb_query_resolve.go:100`). Matches the issue's quiet-variant requirement (INTEGER→VARCHAR lexicographic drift) and the mid-execution rulings on the uuid dual encoding and pushdown-pruned failure scope.

Test half (acceptance b) is present: real-DuckDB ordering proof + loud-garbage proof (`internal/sqlgen/duckdb_cold_scan_attr_pin_test.go`), refusal/listed-vs-unlisted/dry-run/412-retry/delete-ordering/unnormalizable/paged-listing seams (`internal/cdc/init_delta_test.go`), pin detection + scope parts (`internal/federated/cold_columns_test.go`), flag mapping (`cmd/tools/cdc_flags_test.go`), production e2e refusal→purge assertion plus `RunInitKeepingDelta` for the now-unreachable stale layout (`internal/e2e_harness/production/init_keep_delta.go:29`, `cdc_e2e_test.go:75`).

### Standards

- **File/function size** (`coding-standard.md` §1): all new/edited sources under 500 lines (`init_delta.go` 227, `cold_columns.go` 72, `column_union.go` 43, `init_keep_delta.go` 99, `duckdb_cold_scan.go` 241); every new function is small with early returns and ≤2 nesting levels. No refactor trigger.
- **Error handling** (`coding-standard.md` §3, `docs/error-handling.md`): every new error wraps with context (`inventory delta tier: %w`, `list delta prefix %s: %w`, `purge delta tier of schema %d…`, `resolve manifest path: %w`, etc.). No bare `return err` in the new paths.
- **Error semantics**: `ErrDeltaTierPresent` is a plain operator sentinel (`errors.New`), not a `forma.InvalidInputf` carrier — correct. This is a CLI/tool failure, not an HTTP write-path 4xx, and the read-side pin failure stays DuckDB's own Conversion Error (loud read failure, same scope as the `changed_at` re-pin). `TestNoBareSentinelWraps` (which only guards `forma.ErrInvalidInput/NotFound/Conflict`) is unaffected — verified passing below.
- **Docs placement** follows the repo rule: durable system behavior lands in `docs/` (`federated-query/design.md` §5 scan-guard + `#371` paragraph, `manifest-reconcile.md` new `cdc-init 重跑与 delta 层` section in that file's language, `README.md` / `tests/e2e/README.md` pointers); the plan and mid-execution rulings already live on #371 as comments. Nothing load-bearing lives only on disk.

### Findings

No blockers. All items below are observations/questions, not change requests.

**F1 (minor, confirmed by code — zero-live-rows + `--replace-delta` is a deliberate no-op).** `init.go:279`: a schema with no live rows takes no swap and leaves delta in place even with the flag (warning only). Documented in `manifest-reconcile.md`. Consequence: the flag is not idempotent for empty schemas — the next init with rows still refuses until purged. Safe (refusal persists, never silent), just confirming this is the intended residual vs. an empty-manifest publish.

**F2 (minor, documented trade-off — no sighting grace on the purge).** `purgeDeltaTier` deletes immediately after the swap; a reader that resolved the pre-swap manifest may hit one missing-object failure before re-resolving (`init_delta.go:187` comment, reconcile doc reader-window paragraph). Same reasoning as #461's in-flight-reader discussion, low-traffic-window guidance given. Accepted; noting that the window is real but bounded and self-healing.

**F3 (question, non-blocking — array-form UUID compatibility).** `ScanTypesCompatible` handles scalar `VARCHAR`↔`UUID` only. A uuid-`list` attribute expects `VARCHAR[]` from `DuckDBNullScanType`; if any exporter ever emits `UUID[]` for it, a healthy set would over-pin (extra `CAST`, correct results but plan-cache churn). I did not find a uuid-list exporter path that produces `UUID[]`, so this is likely unreachable — but worth a one-line confirmation from the author.

**F4 (nit, unreachable-defensive branch).** `purgeDeltaTier` errors when `deleteObject == nil` while `manifestStore != nil` can only arise with a nil S3 client behind a configured template — not reachable from the CLI (client always built) and only via hand-wired contexts. Harmless defense; no action.

### Observations

- Determinism is handled well: `mergeKeys` sorts the purge set, both `missing`/`pinned` slices are sorted before render and scope hashing, `columnUnion.merge` flags `mixed` order-independently.
- Locking closes the obvious race: preflight→export→publish→purge all sit under `initSchemaUnderLock` (`internal/cdc/init_lock.go:13`), which excludes flusher/reconcile; the only CAS racer (compactor) is absorbed by the 412 reload-and-re-splice. A flush landing between preflight listing and purge cannot happen under the lock.
- `S3InitClient` + `NormalizeObjectKey` export and `reconcile.normalizeKey` reuse avoid a second glob/foreign-bucket rule. `isDeltaShapedKey` and `reconcile.classifyObjectKey` agree on the `<uuid>.parquet` shape.
- Bun `tests/e2e/scripts/cdc-init.ts` change is pure arg plumbing (`--delta-prefix` from `config.s3.prefix`, forward `--replace-delta`) — matches the Go flag semantics; the old script would now refuse every schema after a flush, so the forwarding is required, not optional.
- The `RunInitKeepingDelta` harness helper is explicitly non-CLI-reachable and justified in its doc comment (pre-#371 bucket reconstruction for the #210 recency reader). Good containment.

### Verification

What I ran on this branch (directly, not copied from the PR body):

- `make fmt-check` — clean.
- `go vet ./internal/cdc/... ./internal/federated/... ./internal/sqlgen/... ./cmd/tools/...` — clean.
- `go test ./internal/cdc -run 'TestPreflight|TestInventory|TestPublish|TestMergeKeys|TestNewInit|TestNormalize'` — ok.
- `go test ./internal/sqlgen -run 'TestBuildParquetScanSource|TestScanTypes|TestParquetScanPin'` — ok (8/8, incl. the numeric-ordering and loud-garbage DuckDB proofs).
- `go test ./internal/federated -run 'TestColdScan|TestColumnUnion|TestValidator|TestParquetSchema|TestPlanScope|TestDuckPlan'` — ok.
- `go test . -run TestNoBareSentinelWraps` — PASS.
- Read the full diff (48 files, +1764/−309, single commit `338e417`).

Not run here: full `make test`, container-backed production e2e, or the Bun `tests/e2e` pipeline. The PR body reports `make test` (34 pkgs ok) and 27 production-e2e tests passing; CI (lint, `-race` unit, e2e smoke + `-short` federated, production harness, `benchmark-smoke`) plus nightly full-federated should supply that signal. The Bun gap is declared by the author (arg plumbing only) — I agree with the characterization from reading the script diff.

### Assessment

Correct fix for the right hazard, in both halves. The refuse-by-default default (`--replace-delta` opt-in) follows the #463 lesson; swap-then-delete ordering follows the issue recommendation; the read-side pin is minimal (healthy SQL byte-identical, pushdown CAST-free) and closes the quiet lexicographic-ordering variant the issue's scope update named, not just the loud Conversion Error. Tests pin both the refusal/purge lifecycle and the real-DuckDB ordering behavior, and the e2e fixtures were reworked rather than weakened (`TestCDCFlushAndInit` asserts refusal first; reorder-vs-keep-delta split is reasoned in the rulings). Docs name the operator contract, the zero-rows case, the unnormalizable fail-closed case, and the no-grace window.

### Summary

- Spec: meets #371 (refuse + purge-after-swap + selective attribute pin + quiet-drift regression proof).
- Standards: passes file/function size, error-wrap context, nesting, and error-class rules; docs in the right place.
- Findings: no blockers; F1/F2 are documented trade-offs, F3 is a confirmation question on uuid-list array forms, F4 is unreachable defense.
- Verification: fmt/vet/targeted unit + sentinel-guard green on this branch; full suite + e2e left to CI per above.
- No merge performed per repo PR rules; ready from this reviewer's side subject to F3 confirmation and CI green.

iceboundrock:
Responses to F1–F4. No code changes result from this round.

**F1 (zero live rows + `--replace-delta` is a no-op) — confirmed intended for this PR, follow-up opened.** The #371 plan ruled it out of scope explicitly ("a schema with zero live rows takes no swap today … that pre-existing gap is out of scope here"), and the purge is ordered strictly after a swap, so with no swap there is nothing to purge behind. The residual you describe is real and not idempotent: an emptied schema keeps its stale base and delta listed, the next init with rows refuses until the flag is passed again, and `manifest-reconcile --repair` would re-append the delta rather than retire it, so there is currently no tool path for that state. That is worth an empty-swap publish under the same CAS path, but it changes the zero-rows contract that the docs and the plan pinned, so it goes to #519 rather than into this PR.

**F2 (no sighting grace on the purge) — accepted as documented.** A full re-export already requires a low-traffic window; the window is one missing-object failure per in-flight reader, self-healing on the next manifest resolve. Adding a grace period would mean leaving the stale objects listed somewhere, which is exactly the state `--repair` resurrects. Staying as is.

**F3 (uuid-list `UUID[]`) — confirmed unreachable.** Lists are EAV-only, and the export aggregates elements with `castEAVValue(elemMeta)` (`internal/cdc/export_sql_builder.go`), whose uuid branch is `CAST(value_text AS VARCHAR)`, so a uuid list always exports as `LIST(VARCHAR)`, i.e. `VARCHAR[]`, matching `DuckDBNullScanType`. The `UUID` parquet encoding only comes from `castMainValue`, which serves column-bound scalars and never a list. Compaction reads parquet with DuckDB and preserves the input types, so it cannot introduce `UUID[]` either. The scalar-only carve-out in `ScanTypesCompatible` is therefore sufficient; if a future exporter emits `UUID[]` the cost is an over-pin (correct results, extra CAST), not a wrong answer.

**F4 (`deleteObject == nil` branch) — agreed, defensive only.** The CLI always builds the client (`applyInitS3Wiring`); the branch guards hand-wired contexts in tests. Leaving it in place so a mis-wired context fails with a named error instead of a nil-func panic after the manifest swap has already been published.

iceboundrock:
## Review of #518

### Findings

**P1: An ambiguous manifest PUT can skip the delta purge.**

Locations: `internal/cdc/init_publish.go:46-58`, `internal/cdc/init.go:312-317`, and the ambiguity contract in `internal/manifest/store_s3.go:81-88`.

`manifest.Save` correctly treats every non-confirmed-412 error as ambiguous because the S3 `PutObject` may have committed. However, `updateSchemaManifest` returns that error and `publishAndPurge` only calls `purgeDeltaTier` after a nil error. If the PUT committed and the response was lost, the manifest can already contain the new base with an empty delta tier while all old delta objects remain in storage. A later `manifest-reconcile --repair` can classify an uncovered tombstone as a lost delete and append the stale object back, reintroducing the exact stale generation this PR is intended to retire.

Please add an ambiguous-commit recovery path: after a non-412 save error, reload and compare the manifest with the intended publish, then purge only when the intended swap is confirmed; otherwise leave the old manifest and return the error. Add a store test that commits the manifest and then returns a transport error, asserting that the committed swap still leads to cleanup and cannot be re-adopted by repair.

**P1: `cdc-init` does not validate that the loaded manifest belongs to the requested schema.**

Locations: `internal/cdc/init_delta.go:161-170` and `internal/cdc/init_publish.go:76-92`.

Both paths use `manifest.LoadOrCreate`, which does not check `Manifest.SchemaID`. The cdc-init CLI also does not run the existing manifest-template injectivity/schema checks used by the read path. With a fixed or colliding `--manifest-template`, initializing schema A can load schema B manifest data, clear and delete schema B delta objects under `--replace-delta`, and publish schema A base files into schema B manifest storage. This contradicts the documented fail-closed behavior for a bad template and makes the destructive flag a cross-schema data-loss path.

Please reject a loaded manifest when `SchemaID` is nonzero and differs from the requested schema, both during preflight and on every CAS reload. Also validate the cdc-init template with the same injectivity rule as the read path, and add a fixed-template regression test.

**P1: The purge trusts every same-bucket path labeled `Tier: "delta"`.**

Location: `internal/cdc/init_delta.go:173-184`.

`NormalizeObjectKey` rejects globs and foreign buckets, but it does not verify that a manifest delta path belongs to the current schema delta namespace or even has a delta-shaped key. A malformed or misclassified entry such as `base/2/...parquet` can therefore enter `deltaPurge` and be deleted after the swap. Since this is an irreversible operation, the purge should fail closed for keys outside the configured/current schema delta namespace and for non-delta shapes, or use an equivalent ownership check that still supports documented custom delta prefixes. Add coverage for a same-bucket foreign-schema/base-shaped manifest entry.

**P2: A stale but system-valid column stamp can bypass the new attribute pin.**

Locations: `internal/federated/parquet_schema_validation.go:190-196` and `internal/federated/cold_columns.go:54-63`.

A manifest stamp that passes `parquetcheck.Check` skips the footer probe and is merged directly into the union. `parquetcheck.Check` validates only system columns. If the stamp says `score=INTEGER` but the bytes at that path are actually `VARCHAR`, the union appears healthy, no `CAST(score AS INTEGER)` is rendered, and the quiet ordering drift remains. This is reachable for a same-key rewrite with an unchanged/stale stamp, including legacy deterministic init keys or manual/old-writer mutations.

Either validate attribute columns before using a stamp for the pin decision, tie the stamp to an object version/checksum that invalidates on rewrite, or explicitly keep this as a documented trust-boundary limitation with a regression test. The current tests cover real footer drift and correct stamps, but not stale stamps.

### Spec

The main #371 behaviors are present: refusal by default, `--replace-delta`, manifest base-plus-delta CAS splicing, post-swap deletion, unlisted delta inventory, dry-run refusal, the zero-live-row rule, and selective attribute type pinning with a separate plan-cache scope component. The findings above leave the destructive operator path incomplete, so I do not consider the full spec safe to accept yet.

### Standards

- No file/function size violation was found in the touched Go sources.
- Error wrapping and the repository read/write error-class split are consistent in the new paths.
- `make fmt-check` passed.
- `make lint` passed.
- `go vet ./internal/cdc/... ./internal/federated/... ./internal/sqlgen/... ./cmd/tools/... ./internal/reconcile/...` passed.
- `git diff --check origin/main...HEAD` failed at `docs/manifest-reconcile.md:542` because of a new blank line at EOF.
- The durable behavior documentation is placed under `docs/`, which is correct; the comment above is the review record for this PR.

### Assessment

Request changes. The selective cold-scan pin is narrowly implemented and the healthy SQL/cache behavior is well covered, but the new purge path has two direct cross-generation/cross-schema safety gaps plus an ambiguous-commit recovery gap. These can leave stale data re-adoptable or delete data outside the requested schema.

### Observations

- Holding the schema advisory lock across inventory, export, publish, and purge is the right concurrency boundary for flusher and reconcile.
- Re-splicing delta entries on confirmed 412 is correct and the delete-after-save ordering is clearly tested.
- The selective pin avoids changing healthy SQL and keeps attribute pushdown CAST-free in the common case.
- The missing tests are specifically ambiguous PUT-after-commit, manifest schema mismatch/fixed template, same-bucket misclassified purge paths, and stale attribute stamps.

### Verification

- `./scripts/test_with_container_runtime.sh` passed all unit-test packages after the script configured the rootless Podman socket.
- Targeted tests passed for `./internal/cdc ./internal/federated ./internal/sqlgen ./cmd/tools ./internal/reconcile`.
- The PR checks are green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.
- The full tagged production e2e suite and Bun pipeline were not independently run during this review; the PR description reports those results.

### Summary

- Findings: three P1 data-safety findings and one P2 type-stamp coverage gap.
- Spec: core refusal/purge/pin behavior is implemented, but destructive-path safety is incomplete.
- Standards: Go gates pass; `git diff --check` needs the documented EOF whitespace fix.
- Assessment: request changes; no merge performed.

iceboundrock:
Implementation plan for the second review round (P1 ×3, P2, whitespace). This is the plan; the response with verification follows once the commit is up.

**P1 (ambiguous PUT skips the purge) — fix.** Confirmed: `updateSchemaManifest` returns every non-412 save error and `publishAndPurge` never reaches `purgeDeltaTier`, so a committed-but-lost-response swap leaves the old delta objects unlisted and adoptable by `--repair`. Fix: with `--replace-delta`, a non-412 save error triggers a reload of the manifest; if the reload shows this run's base entries (write-once keys since #416, so a path-set match proves it was this publish) and an empty delta tier, the swap is treated as committed and the purge proceeds. Any other reload outcome (pre-swap manifest, reload failure) returns the error and deletes nothing, as today. Tests: a store that commits then returns a transport error must still lead to the deletes; a store that fails without committing must not; a reload failure must not.

**P1 (no manifest schema check / no template check) — fix.** Confirmed: both `loadManifestDeltaEntries` and `publishInitTiers` use `manifest.LoadOrCreate`, which does not check `SchemaID`, and cdc-init never runs the read path's two-probe template check. Fix: a `manifest.LoadOrCreateForSchema` helper carrying the `QuerySource.load` rule (`SchemaID != 0 && != requested` → `forma.ManifestSchemaMismatchError`), used by the read path (behaviour-preserving) and by cdc-init on the preflight load and on every publish load, in both flag modes (the base-only path stops going through `ReplaceTierFiles` so it loads through the same checked helper). The root package's `probeManifestTemplate` is exported as `forma.ValidateManifestPathTemplate` and `RunInit` runs it before opening any connection. Tests: manifest claiming another schema at preflight and at the CAS reload; fixed `--manifest-template` rejected by `RunInit`.

**P1 (purge trusts every `Tier: "delta"` path) — fix.** Confirmed: `splitNormalizableKeys` admits any same-bucket key. Fix: a purge key must be delta-shaped inside this schema's delta namespace, `{delta-prefix}/{schemaID}/{uuid}.parquet`; with `--delta-prefix ""` the prefix part is unconstrained but the `/{schemaID}/{uuid}.parquet` tail is still required. Entries outside it count toward the refusal without the flag and fail the schema closed (named) with the flag, at preflight and on the CAS reload, the same way unnormalizable entries already do. Test: a same-bucket `base/2/...` and a `delta/2/...` entry under schema 1.

**P2 (stale stamp bypasses the pin) — document + regression test, no code change.** The pin consumes the same union the #256 stamp trust already feeds; a stamp that passes the system-column invariant is trusted for attribute columns exactly as it is trusted for column presence (#255) today. Validating attribute columns would mean probing the footer, which is the cost #256 removed, and binding stamps to an object version is a manifest-format change out of this PR's scope. The reachable path is a same-key rewrite, which no current writer performs (init keys are write-once since #416), and `manifest-reconcile --verify-stamps` is the existing offline detector. The design doc gets an explicit trust-boundary paragraph and a test pins that a passing stamp is what decides the pin.

**Whitespace** — trailing blank line at EOF of `docs/manifest-reconcile.md` removed.

**Follow-up (out of scope):** cdc-flush and the compactor load manifests without the schema-ID check and without the template probe too; that is pre-existing and gets its own issue.

iceboundrock:
Response to the second review round, implemented in cc3b5dd. Each finding was verified against the code before the change; the plan is in the previous comment.

**P1 — ambiguous PUT skipped the purge. Fixed.** `updateSchemaManifest` (`internal/cdc/init_publish.go`) still refuses to blind-retry a non-412 save error, but with `--replace-delta` it now hands the error to `confirmAmbiguousSwap`, which reloads the manifest and calls `swapCommitted`: the base tier must be exactly this run's entries (write-once keys since #416, so a path-set match can only be this publish) and the delta tier must be empty. A match logs the lost response and proceeds to `purgeDeltaTier`, so a committed swap never leaves the old delta objects as orphans for `--repair` to re-adopt. The pre-swap manifest still in place, or a failed reload, returns the error (joined with the reload error in the second case) and deletes nothing; the message tells the operator to rerun `cdc-init --replace-delta`, whose pre-flight inventories the objects again whether or not the manifest still lists them. Without the flag the behaviour is unchanged (no confirmation, error surfaces as before). Tests: `TestPublishAndPurge_CommittedSaveWithLostResponseStillPurges`, `TestPublishAndPurge_UncommittedSaveDeletesNothing`, `TestPublishAndPurge_AmbiguousSaveWithFailedReloadDeletesNothing`, `TestSwapCommitted`.

**P1 — no manifest schema check, no template probe. Fixed.** New `manifest.LoadOrCreateForSchema` carries the read path's `SchemaID != 0 && SchemaID != requested` rule and returns `forma.ManifestSchemaMismatchError`. `QuerySource.load` now uses it (behaviour-preserving), and so does every cdc-init load: the pre-flight inventory, each publish attempt and the 412 reload, in both flag modes — the base-only path no longer goes through `ReplaceTierFiles`, and `ReplaceTierFiles` itself loads through the checked helper too. The root package's two-probe check is exported as `forma.ValidateManifestPathTemplate(field, tmpl)` and `RunInit` runs it on `--manifest-template` before `newInitRunContext` opens any connection. Tests: `TestPreflightDeltaTier_RefusesForeignSchemaManifest`, `TestUpdateSchemaManifest_ReloadRefusesForeignSchemaManifest` (a store that restamps the manifest as schema 2 and returns 412; the reload refuses before a second save), `TestRunInit_RejectsCollapsingManifestTemplate`, `TestLoadOrCreateForSchema`, `TestValidateManifestPathTemplate`.

**P1 — purge trusted every `Tier: "delta"` path. Fixed.** `splitNormalizableKeys` is replaced by `classifyDeltaEntries` + `ownedDeltaKey` (`internal/cdc/init_delta.go`): a purge key must be delta-shaped inside this schema's namespace, `{delta-prefix}/{schemaID}/<uuid>.parquet`; with `--delta-prefix ""` the prefix part is unconstrained but the `/{schemaID}/<uuid>.parquet` tail is still required. Same-bucket keys outside it (another schema's delta, a base-shaped key under a delta label) are a third inventory bucket, `foreign`, next to `unnormalizable`: without the flag they count in the refusal (`N entries a purge could not honour`), with the flag `refuseUnpurgeable` fails the schema closed and names them, at the pre-flight and on the 412 reload. Tests: `TestOwnedDeltaKey`, `TestPreflightDeltaTier_ForeignDeltaEntriesFailClosed`.

**P2 — stale stamp bypasses the pin. Documented and pinned, no code change.** The finding is accurate: `parquetcheck.Check` validates system columns only, and a passing stamp feeds the union with no footer probe. That is the #256 trust boundary, and the pin inherits it the same way the #255 missing-column augmentation already does. Closing it would mean either probing attribute columns on every read (the cost #256 removed) or binding stamps to an object version (a manifest-format change), and the reachable state — a same-key rewrite under an unchanged stamp — has no current writer since #416 made init keys write-once (flush and compaction never reuse a key). `manifest-reconcile --verify-stamps` is the existing offline detector for a stamp that no longer matches its object. `docs/federated-query/design.md` gets an explicit trust-boundary paragraph after the #371 pin paragraph, and `TestColdScanPinFollowsTrustedStampWithoutProbe` pins that the stamp's attribute type decides the pin with zero probes in both directions. If you want the read-side hardening anyway, say so and I will open an issue for a versioned stamp.

**Whitespace.** Trailing blank line at EOF of `docs/manifest-reconcile.md` removed; `git diff --check` is clean. The cdc-init section of that doc gains the three new rules (namespace ownership, schema identity + template probe, ambiguous-save confirmation).

**Follow-up.** cdc-flush (`AppendFile`/`AppendFiles`), the compactor (`ManifestProvider.LoadManifest`) and the `--manifest-template` flags of cdc-flush, compactor and manifest-reconcile have the same two gaps (no schema check on load, no template probe). Pre-existing and non-destructive to objects, so out of this PR's scope: #520.

**Verification.** `make fmt-check`, `make lint` (root and infra modules), `git diff --check` and `go vet -tags=e2e ./internal/e2e_harness/...` pass; `./scripts/test_with_container_runtime.sh` passes all 34 packages. File and function sizes stay under the 500/100 limits (`init.go` 482 lines, `init_delta.go` 284, `init_publish.go` 169).

iceboundrock:
## Review of #518

Review artifact for PR #518 at head `cc3b5dd`; working copy: `/home/ruoshi/code/lychee/forma`.

### Findings

**P1: Truncated S3 pagination can be treated as a complete delta inventory.**

Location: `internal/cdc/init_wiring.go:74-95`, especially lines 91-92.

`ListObjectKeys` returns success whenever `IsTruncated` is true but `NextContinuationToken` is nil. It also accepts an empty token and continues. If an S3-compatible endpoint or client returns a malformed/incomplete truncated page, `inventoryDeltaTier` treats the first page as the complete unlisted-delta set. With `--replace-delta`, `publishAndPurge` can then publish the new base and delete only the keys seen so far, leaving an old delta object unlisted. A later `manifest-reconcile --repair` can re-adopt an uncovered tombstone, reintroducing the stale generation this PR is intended to retire.

Please fail closed when `IsTruncated` is true without a non-empty continuation token, and preferably reject a repeated token as well to avoid an infinite loop. Add a regression test for the malformed truncated response and assert that no manifest swap or purge proceeds after the listing error.

### Spec

The core #371 behavior is present:

- Default `cdc-init` refusal uses `cdc.ErrDeltaTierPresent`, including dry-run, before row counting/export.
- `--replace-delta` atomically replaces base and clears delta in one manifest CAS, retries confirmed 412 conflicts, confirms ambiguous committed saves, and deletes only after the swap.
- Inventory includes manifest-listed and unlisted delta-shaped objects; namespace ownership, foreign-bucket paths, and unnormalizable entries fail closed.
- Manifest schema identity is checked during preflight and publish reloads, and the init template receives the read-path probe.
- Attribute type pinning is selective, leaves healthy SQL unchanged, and participates in the plan-cache scope.
- The quiet INTEGER/VARCHAR ordering case and unconvertible values have real-DuckDB coverage.

The pagination failure above leaves the destructive purge path incomplete under an invalid storage response, so the safety spec is not fully closed yet.

### Standards

- No touched file or function size violation was found against the repository's 500/100-line limits.
- New errors carry operation context and preserve the relevant causes.
- `make fmt-check` passed.
- `go vet` passed for the affected packages.
- `git diff --check` passed.
- Durable behavior changes are documented under `docs/`; no load-bearing review artifact is left only in the working tree.

### Assessment

**Request changes.** The implementation is otherwise well-scoped and the previously identified ambiguous-save, manifest-identity, purge-ownership, and stale-stamp concerns are addressed or explicitly documented. The remaining fail-open pagination branch can make `--replace-delta` report success while stale delta data remains recoverable, so it should be fixed before merge.

### Observations

- Holding the schema advisory lock across inventory, export, publish, and purge correctly excludes the flusher and reconciler; CAS handling covers the compactor race.
- Sorted purge keys, sorted cold-scan columns, and the separate `cold-pinned` scope component provide deterministic behavior and avoid plan-cache poisoning.
- The zero-live-row `--replace-delta` residual is documented and tracked separately in #519.
- The unchecked schema/template paths in cdc-flush, compactor, and manifest-reconcile are documented as the pre-existing follow-up in #520, not counted as findings for this PR.
- The trusted manifest-stamp boundary and the no-sighting-grace reader window are explicit design trade-offs with tests/docs.

### Verification

- `make fmt-check` — passed.
- `make lint` — passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/cdc ./internal/federated ./internal/sqlgen ./internal/manifest ./internal/reconcile ./cmd/tools` — passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet ./internal/cdc/... ./internal/federated/... ./internal/sqlgen/... ./internal/manifest/... ./internal/reconcile/... ./cmd/tools/...` — passed.
- `git diff --check main...HEAD` — passed.
- PR checks are green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.
- The full container-backed production suite and Bun pipeline were not independently rerun in this review; the PR description reports the production verification, and the GitHub checks above are green.

### Summary

- Findings: one P1 fail-open S3 pagination issue.
- Spec: core refusal, purge ordering, schema isolation, and attribute pin behavior are implemented; pagination safety remains incomplete.
- Standards: fmt, lint, vet, diff checks, error-context, and size checks are clean.
- Assessment: request changes; no merge performed.

iceboundrock:
Implementation plan for the third review round (P1, truncated pagination). This is the plan; the response with verification follows once the commit is up.

**P1 (truncated listing treated as complete) — fix.** Confirmed: `ListObjectKeys` (`internal/cdc/init_wiring.go:91`) returns the keys seen so far when `IsTruncated` is true and `NextContinuationToken` is nil, and forwards an empty token unchecked, so a malformed page becomes a short inventory and, with `--replace-delta`, a purge that leaves stale delta objects unlisted for `--repair` to re-adopt. Fix: the loop fails closed with a named sentinel when a truncated page carries a nil or empty continuation token, and when a page hands back a token already used in this listing (a repeated token would otherwise loop forever). The error propagates through `inventoryDeltaTier` → `preflightDeltaTier` → `initSchema`, which returns before `prepareSchemaInit`, so no export, manifest publish or purge can follow it. Tests: `ListObjectKeys` over a client that answers truncated-without-token, truncated-with-empty-token and repeated-token; and a wiring-level regression that runs the pre-flight with `--replace-delta` over a manifest-listed delta entry and a malformed truncated listing, asserting the named error, zero deletes and an unchanged manifest.

**Follow-up (out of scope).** `reconcile.S3ObjectStore.ListObjects` (`internal/reconcile/deps.go:80`) has the same fail-open branch; its listing feeds `--gc` (deletes) and `--repair` (adoption). Pre-existing and untouched by this PR, so it gets its own issue rather than widening the diff here.

**Docs.** The cdc-init section of `docs/manifest-reconcile.md` gains the rule that an incomplete listing fails the schema closed.

iceboundrock:
Response to the third review round, implemented in a91cb45. The finding was verified against the code before the change; the plan is in the previous comment.

**P1 — truncated listing treated as complete. Fixed.** Confirmed: `ListObjectKeys` returned on `IsTruncated && NextContinuationToken == nil` and forwarded an empty token unchecked. `ListObjectKeys` (`internal/cdc/init_wiring.go`) now fails with the named sentinel `cdc.ErrIncompleteObjectListing` when a truncated page carries a nil or empty continuation token, and when a page hands back a token this listing has already used (the repeated-token case would otherwise loop forever; the error names the page number and the token). No partial key set is returned on either path. The error propagates through `inventoryDeltaTier` → `preflightDeltaTier` → `initSchema`, which returns it before `prepareSchemaInit`, so no row is counted or exported, no manifest is published and no object is deleted; the per-schema `errors.Join` lets other schemas continue as with any other pre-flight failure.

Tests: `TestListObjectKeys_UnfollowableTruncationFailsClosed` (truncated without token, truncated with empty token, repeated token — each asserts the sentinel, the message, a nil key set and the exact number of pages requested) and `TestPreflightDeltaTier_IncompleteListingStopsBeforeAnySwapOrPurge` (the listing seam wired through `applyInitS3Wiring` over a client whose first page is truncated without a token, `--replace-delta` set, a manifest-listed delta entry present: the pre-flight returns the sentinel, the client records zero deletes and the manifest bytes are unchanged).

**Docs.** The cdc-init section of `docs/manifest-reconcile.md` gains the rule (incomplete listing → schema fails closed at the pre-flight, nothing exported, published or deleted).

**Follow-up (out of scope): #521.** `reconcile.S3ObjectStore.ListObjects` (`internal/reconcile/deps.go`) has the identical fail-open branch, and the e2e harness paginators stop on a nil token without consulting `IsTruncated`. Pre-existing and untouched by this PR. Its exposure is lower than the purge (a short listing under-reports orphans and over-reports dangling entries, and the dangling path re-probes each key before splicing), but a drift reporter should not answer "clean" from a page it could not follow. The issue proposes lifting the guarded loop into one shared paginator.

**Verification.** `make fmt-check`, `make lint` (root and infra modules) and `git diff --check` pass; `./scripts/test_with_container_runtime.sh` passes all 34 packages, including the two new tests. `init_wiring.go` is 117 lines and `ListObjectKeys` stays well under the function limit. The full tagged production e2e suite and the Bun pipeline were not rerun: the change is confined to the listing loop, which the production harness exercises only through the wiring the earlier rounds already covered.

iceboundrock:
## Findings

**P1: Zero-stamped legacy manifests still bypass schema isolation and are never restamped.**

Locations: `internal/manifest/manifest.go:237-259`, `internal/cdc/init_delta.go:194-203`, `internal/cdc/init_publish.go:100-117`, and `internal/manifest/query_source.go:53-61`.

`LoadOrCreateForSchema` intentionally accepts `Manifest.SchemaID == 0` for legacy compatibility, but cdc-init then splices and saves that manifest without assigning the requested schema ID. It therefore remains zero-stamped forever, and the read path continues to accept it for every schema. The two-probe template check cannot prove injectivity for all IDs: for example, a template can map schema IDs 3 and 4 to one path while still producing different paths for probe IDs 1 and 2. A zero-stamped manifest at that shared path can then be loaded and served under the wrong schema; base-only init can also replace the foreign base in that shared manifest. `--replace-delta` catches some foreign delta entries through the namespace check, but it does not protect a zero-stamped manifest containing only base entries.

Please either fail closed for zero-stamped manifests on cdc-init publish/destructive paths, or safely validate and restamp them before saving with an ownership check that prevents the legacy compatibility carve-out from becoming a cross-schema read/write path. Add a regression covering a zero-stamped manifest plus a collision outside the two probe IDs.

**P2: `cdc-init` accepts padded manifest templates even though the read-side validator rejects them.**

Locations: `config_duckdb.go:187-198` and `internal/cdc/init.go:180-183`.

`ValidateManifestPathTemplate` only rejects `strings.TrimSpace(pathTemplate) == ""`; it does not apply the existing leading/trailing-whitespace check from `validateManifestReadWhitespace`. A value such as `" manifest/{{.SchemaID}}.json\\n"` therefore passes cdc-init validation, while `PathResolver.Resolve` uses the raw value and writes manifests under a different key. The server-side `DuckDBConfig.ValidateManifestRead` rejects the same padded value, so init can successfully publish data that the configured read path cannot use. Reuse the shared whitespace rule and add a padded-template test for the exported validator.

## Spec

- The main #371 behavior is implemented: refuse by default with `cdc.ErrDeltaTierPresent`, inventory manifest-listed and unlisted delta-shaped objects, publish base-plus-empty-delta in one CAS, retry confirmed 412 conflicts, confirm ambiguous committed saves, and delete only after the swap.
- Namespace ownership, foreign-bucket handling, malformed pagination, dry-run refusal, and the selective attribute type pin are covered by code, tests, and documentation.
- The prior review findings on ambiguous PUTs, manifest identity for nonzero schema IDs, purge ownership, stale stamp trust boundary, and truncated listing handling are addressed or explicitly documented.
- The two findings above leave the legacy-manifest isolation and CLI/read configuration parity incomplete.

## Standards

- `make fmt-check` passes and `git diff --check origin/main...HEAD` is clean.
- `config_duckdb_test.go` is 503 lines, exceeding the repository limit of 500 lines for source files. Split or reduce this test file before merge.
- Commit `a91cb45` contains a `Claude-Session: https://claude.ai/...` footer. This violates the repository rule prohibiting AI attribution in commit messages; remove that footer before merge.
- Error wrapping, function-size, nesting, and durable documentation placement otherwise look consistent with the repository standards.

## Assessment

**Request changes.** The implementation is otherwise focused and the destructive ordering is substantially hardened, but a legacy zero-stamped manifest can still defeat the new schema-identity boundary in a cross-schema template collision. The padded-template gap also allows cdc-init and the read path to disagree about the manifest object key.

## Observations

- The schema advisory lock correctly covers inventory, export, publish, and purge, excluding flusher and reconcile races; the compactor race is handled by CAS reloads.
- Purge keys and cold-scan columns are deterministic, and `cold-pinned` participates in the plan-cache scope.
- The zero-live-row behavior and immediate post-swap deletion window are documented trade-offs and are tracked separately by the existing follow-ups.
- The trusted manifest-stamp boundary is documented and tested as an intentional limitation rather than silently presented as footer verification.

## Verification

- `make fmt-check`: passed.
- Targeted tests passed: `go test ./internal/cdc ./internal/manifest ./internal/federated ./internal/sqlgen ./cmd/tools`.
- `make test`: all non-container packages passed, but `internal/e2e_harness` panicked in testcontainers because this environment has no usable `$XDG_RUNTIME_DIR`/container runtime. This is an environment limitation; the PR GitHub Test check is green.
- GitHub checks are green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.
- The full container-backed production suite and Bun `tests/e2e` pipeline were not independently rerun here; the PR description reports the production verification and the Bun gap.

## Summary

- Findings: one P1 schema-isolation gap and one P2 template-validation gap.
- Standards: one file-size violation and one prohibited commit attribution footer.
- Spec: core refusal, purge ordering, pagination fail-closed behavior, and selective pinning are present, but the two remaining safety/parity gaps should be fixed.
- Assessment: request changes; no merge performed.

iceboundrock:
Implementation plan for the fourth review round (P1 zero-stamped manifests, P2 padded template, file size, commit footer). This is the plan; the response with verification follows once the commit is up.

**P1 (zero-stamped manifest bypasses schema isolation and is never restamped) — fix, fail closed.** Confirmed: `LoadOrCreateForSchema` lets `SchemaID == 0` through, and no cdc-init path assigns the requested ID before `manifest.Save`, so an unstamped manifest stays unstamped and the two-probe template check cannot exclude a collision at IDs the probes never render (`{{if lt .SchemaID 3}}…{{else}}shared{{end}}` passes). One more fact bears on the fix: the `schema_id` field has been in `manifest.Manifest` since `internal/manifest/manifest.go` was first committed (968afbc) and `LoadOrCreate` has always stamped a created manifest, so no Forma writer has ever produced a zero stamp. "Legacy pre-stamp manifests" is a carve-out with no writer behind it; the only real sources are hand-edited objects. cdc-init is the writer with a destructive step, so it stops trusting the carve-out:

- `loadInitManifest` (every cdc-init load: pre-flight, each publish attempt, 412 reload, ambiguous-save reload) refuses a loaded manifest with `SchemaID == 0` **that lists any entry**, with a new named sentinel `cdc.ErrManifestUnstamped`. The message names the path, the requested schema and the entry count, and says what to do: confirm the object belongs to that schema and set `schema_id`. Nothing is exported, published or deleted, in either flag mode.
- A zero-stamped manifest with **no entries** is restamped to the requested schema in memory before use, so the next save persists the stamp and every later load is checked. Nothing in it can belong to another schema.
- No ownership-based restamp of non-empty manifests: it would need every writer's key layout (base, merged base, delta, custom prefixes) to prove ownership, for a state no writer produces. Fail closed is the safe answer and the remedy is a one-field edit.
- The `LoadOrCreateForSchema` doc comment loses the "written before the field existed" justification, which the history does not support; it now says the read path keeps its pre-#518 lenience and points at cdc-init's stricter rule.

Tests: a template that collides only outside the probe IDs passes `forma.ValidateManifestPathTemplate`; a zero-stamped manifest at the shared path holding schema 3's base entry makes the schema-4 pre-flight fail with the sentinel in both flag modes with zero saves, and makes the schema-4 publish (base-only and `--replace-delta`) fail with the sentinel, zero saves, zero deletes, manifest bytes unchanged. An empty zero-stamped manifest is restamped by the publish (saved `schema_id` = 1). `TestLoadOrCreateForSchema`'s zero-stamp case stays (read-path behaviour unchanged).

**P2 (`ValidateManifestPathTemplate` accepts padded templates) — fix.** Confirmed: only `TrimSpace(...) == ""` is checked, while `PathResolver.Resolve` renders the raw value, so cdc-init writes under a key the server-side `ValidateManifestRead` would reject. The leading/trailing-whitespace rule is extracted into one helper shared by `validateManifestReadWhitespace` and `ValidateManifestPathTemplate`, and the exported validator applies it first, before the empty/parse/probe checks, with the same message. Tests: padded and trailing-newline templates rejected with the caller's field name.

**File size.** `config_duckdb_test.go` (503 lines): the two template-probe tests (`TestValidateDuckDBConfig_ManifestTemplateProbe`, `TestValidateManifestPathTemplate`) move to a new `config_manifest_template_test.go`, which also takes the new padded cases.

**Commit footer.** a91cb45 is amended to drop the `Claude-Session:` line (now 1c71004); no commit on the branch carries a trailer. This is a force-push of the same three-commit history plus the round-four commit.

**Docs.** The cdc-init section of `docs/manifest-reconcile.md` gains the unstamped-manifest rule (refuse non-empty, restamp empty) and the padded-template rule.

**Follow-up (out of scope).** The read path (`QuerySource.load`), `manifest-reconcile`'s `ResolverManifestStore.Load` (#481) and the shared `LoadOrCreateForSchema` keep accepting `schema_id: 0` for any schema. That lenience predates this PR, and since no writer emits a zero stamp it protects nothing while leaving a cross-schema read path under a collided template. It gets its own issue next to #520.

iceboundrock:
Response to the fourth review round, implemented in f0a9db2. Each finding was verified against the code before the change; the plan is in the previous comment.

**P1 — zero-stamped manifests bypassed schema isolation and were never restamped. Fixed, fail closed.** Confirmed as described: `LoadOrCreateForSchema` passes `schema_id: 0`, no cdc-init save assigned the requested ID, and a template like `{{if lt .SchemaID 3}}manifest/{{.SchemaID}}.json{{else}}manifest/shared.json{{end}}` passes the two-probe check while colliding every schema from 3 upward. One more fact decided the shape of the fix: `SchemaID` has been on `manifest.Manifest`, and set by `LoadOrCreate`, since `internal/manifest/manifest.go` was first committed (968afbc), so no Forma writer has ever emitted a zero stamp and the "written before the field existed" justification in the helper's comment was wrong. `loadInitManifest` (`internal/cdc/init_publish.go`), which every cdc-init load goes through (pre-flight, each publish attempt, the 412 reload and the ambiguous-save reload), now refuses a zero-stamped manifest that lists any entry with the named sentinel `cdc.ErrManifestUnstamped`, in both flag modes; the message names the path, the requested schema and the entry count, and tells the operator to confirm ownership and set `schema_id`. Nothing is exported, published or deleted. An empty zero-stamped manifest is stamped for the requested schema in memory, so the save persists the stamp and every later load is checked. I did not add an ownership-based restamp for non-empty manifests: proving ownership from keys needs every writer's layout (init base, merged base, delta, custom prefixes) for a state no writer produces, and the remedy is a one-field edit. The `LoadOrCreateForSchema` comment now states the real situation (read-path lenience kept for behaviour preservation, not a legacy format). Tests: `TestPreflightDeltaTier_RefusesUnstampedManifestWithEntries` (asserts first that the colliding template passes `forma.ValidateManifestPathTemplate`, then that the schema-4 pre-flight over schema 3's zero-stamped manifest at the shared path fails with the sentinel and saves nothing, both flag modes), `TestUpdateSchemaManifest_RefusesUnstampedManifestWithEntries` (the schema-4 publish through `publishAndPurge`, base-only and `--replace-delta`: sentinel, zero saves, zero deletes, stored bytes unchanged), `TestUpdateSchemaManifest_StampsEmptyUnstampedManifest`.

**P2 — `ValidateManifestPathTemplate` accepted padded templates. Fixed.** Confirmed: only `TrimSpace == ""` was checked while `PathResolver.Resolve` renders the raw value. The leading/trailing-whitespace rule is now one helper, `rejectPaddedValue` (`config_duckdb.go`), used by `validateManifestReadWhitespace` and, first, by `ValidateManifestPathTemplate`, so cdc-init and the server reject the same value with the same message. Ordering follows `ValidateManifestRead` (whitespace before empty/parse/probe), so a whitespace-only template now reports the whitespace it has. Tests: whitespace-only, leading-space and trailing-newline templates rejected with the caller's field name (`TestValidateManifestPathTemplate`).

**File size.** `config_duckdb_test.go` is 384 lines; `TestValidateDuckDBConfig_ManifestTemplateProbe` and `TestValidateManifestPathTemplate` moved to `config_manifest_template_test.go` (133 lines) together with the new cases. Every touched Go file is under 500 lines (largest: `config_duckdb.go`, 404).

**Commit footer.** a91cb45 was amended to drop the `Claude-Session:` line and is now 1c71004; no commit on the branch carries a trailer (`git log main..HEAD` grep is empty). The branch was force-pushed with lease; the history is the same three commits plus f0a9db2.

**Docs.** The cdc-init section of `docs/manifest-reconcile.md` gains the unstamped-manifest rule (refuse non-empty, stamp empty, why no writer produces the state) and the padded-template rule.

**Follow-up (out of scope): #522.** The read path (`QuerySource.load`), `manifest-reconcile`'s `ResolverManifestStore.Load` (#481) and the shared `LoadOrCreateForSchema` still accept `schema_id: 0` for any schema. That predates this PR, and since no writer emits a zero stamp the carve-out protects nothing while leaving a cross-schema read path under a template that collides outside the probe IDs. The issue proposes dropping the carve-out for non-empty manifests, with the reconcile consolidation #520 already plans.

**Verification.** `make fmt-check`, `make lint` (root and infra modules) and `git diff --check` pass; `./scripts/test_with_container_runtime.sh` passes all 34 packages, including the four new tests; `go vet -tags=e2e ./internal/e2e_harness/...` passes. The full tagged production e2e suite and the Bun pipeline were not rerun: the production harness writes every manifest through Forma tools, so it never produces a zero stamp, and the template it passes is unpadded.

iceboundrock:
## Findings

No blocking, major, or minor findings identified at the current head (`f0a9db2`). The previously identified safety gaps are addressed:

- Ambiguous manifest PUTs are reloaded and purged only when this run’s exact base path set and an empty delta tier are confirmed.
- Manifest schema identity is checked during preflight and every publish/reload path; non-empty zero-stamped manifests fail closed.
- Delta purge keys must be UUID-shaped objects in the requested schema namespace; foreign, malformed, glob, and foreign-bucket entries fail closed.
- Truncated or repeated S3 pagination fails before export, publish, or deletion.
- Manifest templates receive the same two-schema and whitespace validation as the read path.
- Cold scans selectively pin drifted or mixed attribute types, with deterministic plan-cache scope separation.

## Spec

The PR meets the requested #371 behavior: refuse populated delta tiers by default, support explicit `--replace-delta`, atomically publish the new base with an empty delta tier, delete superseded delta objects only after a confirmed swap, include unlisted delta inventory, preserve dry-run safety, and restore numeric ordering for quiet attribute type drift. The production and regression coverage also exercises malformed manifests, pagination, CAS races, ambiguous saves, and real DuckDB conversion/order behavior.

## Standards

- File and function size limits are satisfied for the touched Go sources.
- New errors include operation context and preserve causes for `errors.Is`/`errors.As`.
- The write/read error semantics and fail-closed destructive-path behavior are consistent with repository guidance.
- Durable behavior is documented under `docs/`.
- `make fmt-check`, `make lint`, and `git diff --check main...HEAD` pass.
- No prohibited AI attribution footer remains in the PR commits.

## Assessment

Approve from a code-review perspective. The implementation is appropriately scoped, the destructive ordering is conservative, and the cold-scan change avoids adding casts to healthy scans while closing the silent lexicographic-ordering case. No merge was performed.

## Observations

- The zero-live-row plus `--replace-delta` behavior intentionally leaves the delta tier untouched because there is no base swap; this is documented and tracked separately in #519.
- Immediate post-swap deletion has a bounded in-flight reader window without sighting grace; the operational low-traffic requirement is documented.
- Trusted manifest column stamps intentionally avoid footer probes; same-key stamp/byte drift remains an offline `manifest-reconcile --verify-stamps` concern, documented with the follow-up in #522.
- The pre-existing schema/template validation gaps in cdc-flush, compactor, and manifest-reconcile are outside this PR and tracked in #520.

## Verification

Passed locally on the PR head:

- `make fmt-check`
- `make lint`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/cdc ./internal/federated ./internal/sqlgen ./internal/manifest ./internal/reconcile ./cmd/tools`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet ./internal/cdc/... ./internal/federated/... ./internal/sqlgen/... ./internal/manifest/... ./internal/reconcile/... ./cmd/tools/...`
- `./scripts/test_with_container_runtime.sh` including all unit and container-backed harness packages
- GitHub PR checks: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke

The first direct `make test` attempt hit the local testcontainers environment before the repository script configured the Podman socket; the repository-provided script then completed successfully.

## Summary

- Findings: none.
- Spec: satisfied for the #371 behavior and safety cases reviewed.
- Standards: satisfied.
- Assessment: approve, subject to normal merge/CI policy.
- Verification: targeted race tests, vet, lint, formatting, full repository test script, and GitHub checks passed.

PR #523, opened 2026-09-05
fix(manifest): classify a missing manifest by typed NoSuchKey, not error text

### LINKED ISSUE #464: bug(manifest): IsNotFound substring-matches error text — a NoSuchBucket failure silently degrades manifest reads to the glob fallback
## Problem

`manifest.IsNotFound` decides object existence by substring-matching the error text (`"nosuchkey"`, `"not found"`, `"does not exist"`) instead of the typed 404 the SDK already provides. A bucket-level failure whose message contains one of those substrings (e.g. `NoSuchBucket` → "The specified bucket **does not exist**") is misclassified as "manifest absent," and `LoadOrCreate` returns an empty manifest — silently degrading the reader from manifest reads to the glob fallback, which by design can never detect inconsistency and can return fewer rows.

## Current behavior

`internal/manifest/manifest.go` `IsNotFound` uses `strings.Contains` over those substrings; `LoadOrCreate` maps a matched error to an empty manifest; `internal/manifest/query_source.go` `Paths` falls back to a glob when `len(m.Files) == 0`. So a `NoSuchBucket` (or any store error whose text matches) becomes "empty manifest" → glob fallback.

## Reproduction

Point the reader at a non-existent or misconfigured bucket. The store returns `NoSuchBucket` ("… bucket does not exist"); `IsNotFound` matches "does not exist"; `LoadOrCreate` returns empty; `Paths` globs. The reader answers from whatever the glob finds (possibly nothing, possibly a stale partial set) instead of failing loudly.

## Root cause

Existence classified by message substring rather than by the typed `*types.NoSuchKey` / HTTP 404 status. This conflates "the key isn't there" (legitimate LoadOrCreate) with "the store call failed" (should surface).

## Impact

A misconfigured or transiently-failing bucket silently produces partial/empty answers on the read path — the exact inversion of `design.md` §4.3.1's contract (manifest reads are loss-detectable; the glob "can never classify inconsistency"). On the write side the same conflation is harmless (the subsequent conditional PUT fails). Confidence: High on mechanism; Medium on how often a real store emits a matching string for a non-404.

- Not #299/#300/#301 (searched `IsNotFound`, `NoSuchBucket`, `glob fallback`, `LoadOrCreate`). INDEPENDENT.

## Acceptance criteria

- [ ] `IsNotFound` distinguishes a genuine 404/NoSuchKey (typed error / status) from other store failures; only the former yields an empty manifest.
- [ ] A bucket-level or transient store failure surfaces as an error, not a glob-fallback partial answer.

- [ ] A `NoSuchBucket`-shaped error does not classify as "manifest absent."

iceboundrock:
Closes #464.

## Problem

`manifest.IsNotFound` substring-matched the error text (`nosuchkey`, `not found`, `does not exist`). A `NoSuchBucket` ("The specified bucket **does not exist**") or any store failure with similar wording classified as "manifest absent", so `LoadOrCreate` returned an empty manifest and `QuerySource.Paths` silently took the glob fallback, which by design cannot detect inconsistency and can answer with fewer rows.

## Change

- **`manifest.ErrObjectNotFound`** is the single confirmed-absent signal at the `Store` boundary. `IsNotFound` is now `errors.Is(err, ErrObjectNotFound)`; no text inspection.
- **`S3Store.Load`** wraps the sentinel only for the SDK's typed `*types.NoSuchKey` (or an API error whose code is `NoSuchKey`). A bare HTTP 404 status is deliberately **not** enough: `NoSuchBucket` is also a 404, and `GetObject` always carries an error body from which the SDK derives the code, so the code is the only reliable discriminator. All other failures are wrapped with bucket/key context and surface.
- **`FSStore.Load`** wraps the sentinel for `fs.ErrNotExist`.
- The HEAD-based existence probe in `query_source_s3.go` is untouched: HEAD has no body, so status is all it can see, and it is a separate contract (`MissingIn`).
- Test fakes in `internal/manifest`, `internal/cdc`, `internal/reconcile` that returned `"NoSuchKey: not found"` text now return the wrapped sentinel. The production e2e fault injector's "avoid these phrases" test becomes a direct `manifest.IsNotFound(err) == false` assertion.
- `docs/federated-query/design.md` §4.3.1 level 3 now states what "missing" means and that any other manifest-load failure fails the read.

## Harness consumers (review follow-up)

The production e2e helpers `loadManifests` (`cdc.go`) and `RunInitKeepingDelta` (`init_keep_delta.go`) used to treat every `manifest.Load` error as "no manifest yet". They now fall back only when `manifest.IsNotFound(err)` is true and wrap any other failure with schema/path context, so a bucket, permission or transport fault surfaces instead of reading as an empty manifest map or steering onto the plain-init path. The load logic moved into `loadManifestsFrom` / `loadKeptDeltaEntries`, which take a `manifest.Store` and are covered without S3 in `manifest_load_gate_test.go` (wrapped sentinel is absent; text-only "does not exist" surfaces; `NoSuchBucket` surfaces end to end through the real `S3Store` with the SDK error code kept in the chain).

## Mid-execution decision

The sentinel was planned as `manifest.ErrNotFound` (see the plan on #464). The root package's `TestNoBareSentinelWraps` guard (#313) matches the identifier `ErrNotFound` from any package, so a `fmt.Errorf("…%w", manifest.ErrNotFound)` trips it even though it is not a client-error sentinel. Renamed to `ErrObjectNotFound`, which is also the more accurate name (a store object, not an entity). No guard change.

## Regression coverage (`internal/manifest/not_found_test.go`)

- `TestIsNotFound`: nil, wrapped sentinel, text-only "NoSuchKey"/"does not exist", a `NoSuchBucket`-shaped `GenericAPIError`, a raw untranslated `NoSuchKey`, a transport error.
- `TestS3StoreLoad_*`: `NoSuchKey` classifies as absent and keeps the SDK error in the chain; `NoSuchBucket`, a bare 404 and a transport error do not.
- `TestLoadOrCreate_NoSuchKeyCreatesEmptyManifest` / `TestLoadOrCreate_NoSuchBucketSurfaces`.
- `TestQuerySource_StoreFailureDoesNotFallBackToGlob`: `Paths` returns the error, not the fallback glob.
- `TestFSStoreLoad_MissingFileIsNotFound`.

## Verification

- `make fmt-check`, `make lint` (root + infra): clean.
- `./scripts/test_with_container_runtime.sh` (full `make test`): all 34 packages ok.
- `go vet -tags=e2e ./internal/e2e_harness/...`: clean.
- TDD: the new tests were written first and observed failing (`TestIsNotFound` 4 sub-cases, both `S3StoreLoad` classification tests, `LoadOrCreate_NoSuchBucketSurfaces`, `QuerySource_StoreFailureDoesNotFallBackToGlob`) before the implementation.


iceboundrock:
## Review: Request Changes

### Findings

1. **[P2] Non-not-found manifest errors are still swallowed by production E2E helpers**

   `internal/e2e_harness/production/cdc.go:300-302` continues on every `manifest.Load` error, and `internal/e2e_harness/production/init_keep_delta.go:35-39` falls back to a plain init on every error. After this PR, `manifest.Load` preserves the distinction between `manifest.ErrObjectNotFound` and bucket, permission, transport, or other store failures, but these callers do not inspect `manifest.IsNotFound(err)`.

   This means a bucket outage or access failure can be reported as an empty/incomplete manifest map by `loadManifests`, and `RunInitKeepingDelta` can take the wrong init path instead of surfacing the failure. That undermines the PR's stated contract that only a confirmed missing object may trigger the missing-manifest fallback, and can make fault-injection/state assertions silently lose coverage.

   Please gate both fallbacks with `manifest.IsNotFound(err)` and return a contextual error for every other error. Add regression coverage for `NoSuchBucket` or a transport error through these helpers.

   No other correctness findings were identified.

### Spec Check

- The intended S3/FS boundary change is correctly implemented: typed `NoSuchKey` and `fs.ErrNotExist` are translated to `ErrObjectNotFound`, while message-only errors, `NoSuchBucket`, bare 404, and transport errors are not classified as missing.
- The documented federated-read behavior matches the implementation for `QuerySource`: non-missing manifest load failures propagate instead of taking the glob fallback.
- The remaining spec gap is the two direct `manifest.Load` call sites above.

### Standards Check

- Error context and error-chain preservation in the changed store paths follow the repository's error-handling standard.
- Changed source files remain within the documented size limits.
- `make fmt-check` and `make lint` pass, including the infra module.
- The new tests are focused and cover the key S3/FS classification matrix.

### Assessment

Request changes for the P2 caller-handling gap. The core sentinel-based classification is sound, but all manifest consumers need to honor the new contract before this can be considered complete.

### Observations

- The use of `errors.Is` and typed AWS SDK/API error inspection removes the prior unsafe substring classification.
- The new tests preserve the underlying SDK error in the chain, which is useful for diagnostics.
- The PR's remote CI checks are all green, but they do not cover the two direct helper paths with non-not-found failures.

### Verification

- PASS: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/manifest ./internal/cdc ./internal/reconcile ./internal/e2e_harness/production`
- PASS: targeted `go test -race` for `internal/manifest`, `internal/cdc`, and `internal/reconcile`
- PASS: `make fmt-check`
- PASS: `make lint`
- PASS: PR CI checks: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke
- BLOCKED BY LOCAL ENVIRONMENT: `make test` reached the container-backed harness and panicked because no Docker-compatible runtime/XDG runtime socket was available; the corresponding remote Test check passed.

### Summary

The core fix is well tested and correctly prevents S3 store failures from being mistaken for missing manifests. Please update `loadManifests` and `RunInitKeepingDelta` to fall back only when `manifest.IsNotFound(err)` is true, and surface all other errors.

iceboundrock:
## Response to review

### [P2] Non-not-found manifest errors swallowed by production E2E helpers — fixed in bb03278

Confirmed against the code: both `loadManifests` (`internal/e2e_harness/production/cdc.go`) and `RunInitKeepingDelta` (`init_keep_delta.go`) took the fallback on any `manifest.Load` error, so after this PR's boundary change a `NoSuchBucket`, permission or transport failure would still have read as "no manifest yet".

**Change.** Both helpers now fall back only when `manifest.IsNotFound(err)` is true and return a contextual error otherwise (`load manifest for schema %d at %s: %w` / `keep delta: load manifest for schema %d: %w`). The load logic moved into two package-level helpers that take a `manifest.Store`, so the gate is testable without S3:

- `loadManifestsFrom(ctx, store, resolver)`: confirmed-absent schemas are simply absent from the map; anything else fails the call.
- `loadKeptDeltaEntries(ctx, store, path)`: returns nil entries on confirmed absence (plain init is the same run), the delta entries otherwise, and surfaces every other error before any copy-aside happens.

**Regression coverage** (`manifest_load_gate_test.go`, no build tag, runs in `make test`):

- wrapped `ErrObjectNotFound` → absent / nil entries, no error;
- a text-only "The specified bucket does not exist" error → surfaces, `IsNotFound` false;
- a `NoSuchBucket`-shaped API error driven end to end through the real `manifest.S3Store` with a stub client → surfaces, `IsNotFound` false, and `errors.As` still finds the SDK `APIError` with code `NoSuchBucket` in the chain.

**Other consumers checked.** `internal/reconcile/gcstate.go` already gates on `IsNotFound`; `internal/reconcile/manifests.go` goes through `LoadOrCreate`; `internal/compaction` returns the load error to the compactor, which surfaces it. That last one exposes a pre-existing inconsistency outside this PR's scope (a dead `m == nil` Noop branch that the real stores never reach), filed as #524.

### Verification

- PASS: `make fmt-check`, `make lint` (root + infra)
- PASS: `go test -race ./internal/e2e_harness/production ./internal/manifest`
- PASS: `go vet -tags=e2e ./internal/e2e_harness/...`
- PASS (rootless Podman): `go test -tags=e2e ./internal/e2e_harness/production -run 'TestProductionSmoke|TestCDCFlushAndInit|TestInitVerTsLWW|TestMonotonicVersionsCloseSameMillisecondTies|TestWarmColdTiebreak'`, i.e. every tagged test that calls `RunInitKeepingDelta` plus two `loadManifests` callers, against the new gating.

PR description updated with a "Harness consumers" section.

iceboundrock:
## Review: Approve

### Findings

No correctness, security, regression, or test-gap findings identified. No changes requested.

### Spec Check

- The Store boundary now translates only confirmed missing signals: typed/API-code `NoSuchKey` for S3 and `fs.ErrNotExist` for FS into `manifest.ErrObjectNotFound`.
- Message-only errors, `NoSuchBucket`, bare HTTP 404, and transport failures remain ordinary errors.
- `QuerySource.Paths` therefore propagates manifest-load failures instead of taking the glob fallback.
- `loadManifests` and `RunInitKeepingDelta` now fall back only for `manifest.IsNotFound`, matching the issue acceptance criteria.
- The federated-query documentation accurately describes the new behavior.

### Standards Check

- Changed source files remain within the 500-line limit; changed functions remain within the 100-line limit.
- Error paths add context and preserve the sentinel plus underlying SDK error chain.
- The implementation is flat and focused, with no material duplication introduced.
- Regression tests cover the S3/FS classification matrix and both production harness consumers.

### Assessment

Approve. The original unsafe substring classification is replaced with an explicit typed-signal contract, and the previously affected callers now honor that contract.

### Observations

- Using `errors.Is`/`errors.As` avoids message-dependent behavior while retaining useful backend diagnostics.
- The HEAD-based parquet existence probe intentionally remains a separate 404-based contract because HEAD has no response body; this is consistent with the documented design.
- GitHub reported only existing Node.js action deprecation annotations; no PR-related CI warnings or failures were observed.

### Verification

- PASS: `make fmt-check`
- PASS: `make lint` (root and infra modules)
- PASS: `go test ./internal/manifest ./internal/cdc ./internal/reconcile ./internal/e2e_harness/production`
- PASS: `go test -race ./internal/manifest ./internal/cdc ./internal/reconcile ./internal/e2e_harness/production`
- PASS: `go vet -tags=e2e ./internal/e2e_harness/...`
- PASS: `git diff --check`
- PASS: PR CI `Test`, `Lint`, `E2E Smoke Tests`, `Benchmark Smoke`, and `k6 Smoke`

### Summary

The PR correctly prevents bucket, permission, transport, and other manifest-store failures from being mistaken for a missing manifest and silently degraded to glob reads. The implementation, tests, documentation, and caller handling are consistent and verified.

PR #525, opened 2026-09-05
fix(federated): honor explicit preferred_tiers under hybrid routing and mark hot-tier-only answers (#468)

### LINKED ISSUE #468: bug(federated): default hybrid routing sends every ordinary request Postgres-only, silently ignoring preferred_tiers with no partial marker
## Problem

Under default configuration, every ordinary federated API request routes to the Postgres-only path, silently ignoring an explicit `preferred_tiers` declaration and emitting no partial-result marker. The three-tier merge is never exercised by ordinary API traffic, and the answer is complete today only because nothing prunes `entity_main`.

## Current behavior

`applyStrategyHeuristics` (`internal/federated/routing.go`), hybrid arm: `effectiveLimit > 0 && effectiveLimit < 1000 → UseDuckDB = false, Reason = "hybrid small result set"`. Defaults: `Strategy: RoutingStrategyHybrid` (`config_duckdb.go`), `DefaultPageSize: 50`, `MaxPageSize: 100` (`config.go`). So for every page size the service layer permits (≤100) and `offset < 10000`, the request routes to `queryPostgresOnly` (`internal/federated/engine.go`), which reads `entity_main` only. The response carries no `partial_result`/`Partial` marker (`partialScanFromOpts` never sets it on postgres-only routes), unlike §7.2 degradation.

## Reproduction

`"federated":{"enabled":true,"preferred_tiers":["hot","warm","cold"]}` with `items_per_page ≤ 100` → served from Postgres only; `preferred_tiers` has no effect; no marker.

## Root cause

A cost heuristic overrides an explicit tier declaration. #354 established the precedent that a correctness requirement (a keyset cursor) outranks these heuristics; tier coverage got no such override, and unlike degradation the answer carries no partiality signal.

## Impact

Today the answer is complete by coincidence — nothing in the repo prunes `entity_main` (grepped `retention|prune|purge` across `cmd`, `internal/cdc`, `internal/compaction`, config — no hits). But `preferred_tiers` is silently unhonored; the three-tier merge is untested by ordinary traffic; and design §7.3 anticipates "bounded PostgreSQL retention," at which point small-page federated queries become silently short with no marker. It is also the direct amplifier of the `created_at` route-disagreement issue (route chosen by page size).

- #354 (closed) is the keyset-cursor twin of this override. #184 covers the tier form in the template, not routing. Searched `routing hybrid small`, `preferred_tiers`. NEW / scope extension of #354's principle to tier coverage + observability.

## Acceptance criteria

**Maintainer ruling (2026-08-24):** do **both halves**, not either/or. The marker alone leaves the three-tier merge exercised by zero ordinary traffic (which is how #460-class route disagreements stayed invisible); the override alone reintroduces surprise cost for callers who set tiers casually. §7.2 degradation already marks partiality, so an unmarked routing path is an inconsistency, not a design.

- [ ] An explicit multi-tier `preferred_tiers` overrides the small-result heuristic (per the #354 precedent that explicit caller declarations outrank cost heuristics).
- [ ] A Postgres-only answer to a multi-tier request carries a coverage/partial marker regardless of how the heuristic lands.
- [ ] The three-tier merge has at least one ordinary-page-size test that actually reaches DuckDB (doubles as the regression net for whichever routing shape lands).

- [ ] A small-page request with `preferred_tiers:[hot,warm,cold]` either merges tiers or is marked partial.

iceboundrock:
Closes #468.

## Problem

Under the default `hybrid` routing strategy, every request with `items_per_page < 1000` (i.e. every API page, which is capped at 100) hit the "small result set" shortcut and was answered from Postgres alone. `preferred_tiers` was ignored, and the response carried no partial marker, so a caller asking for warm/cold got a silently short answer. The service layer also filled the default three tiers whenever `preferred_tiers` was omitted, so the engine could not distinguish "caller declared all three" from "caller said nothing".

## Change (both halves of the maintainer ruling)

**(a) Explicit `preferred_tiers` overrides the cost heuristic.** In `EvaluateRoutingPolicy`, an explicit `PreferredTiers` naming any tier beyond `hot` is treated as a coverage declaration and reroutes onto the federated path, with the same conjunct shape as the #354 keyset-cursor override (`!dec.UseDuckDB && !isHotOnlyRequest(fq) && cfg.Enabled`). `["hot"]` and `prefer_hot` still take the hot-only gate. The reason string gains a `(overridden: explicit preferred_tiers require the federated path, #468)` suffix.

**(b) Postgres-only answers to multi-tier requests carry a coverage marker.** A new `partial` reason `hot_tier_only` with `unconsulted_tiers` is set on every Postgres-only answer to a request that asked for more tiers, explicitly or by omission. Per the ruling in the issue thread, all three Postgres-only paths set it:

- the routed small-page shortcut (implicit tiers under hybrid),
- a globally disabled DuckDB engine,
- the degraded fallback, which previously cleared the partial marker and signalled only through the plan-gated `Routing.Reason`.

The hot-only gate never marks: the caller asked for hot and got hot. `opts.PartialScan` and `page.Partial` stay in agreement on every path.

**Explicitness.** `federatedPreferredTiers` in the service and `buildFederatedQuery` in the production e2e harness now forward an omitted list as empty. The engine, DuckDB template and plan-cache shape already read empty as the default all-tier form (#184), so no behaviour changes on the DuckDB path.

**(c) An ordinary-page-size test that reaches DuckDB.** `TestTierCoverageUnderHybridSmallPage` (production harness, hybrid strategy, `Limit: 10`, three rows exported to base then pruned from Postgres, two hot rows) proves an explicit `[hot,warm,cold]` request reaches the three-tier merge and matches the oracle, and an implicit request is answered hot-tier-only with `unconsulted_tiers: [warm, cold]`.

## Public surface

- `forma.PartialReasonHotTierOnly = "hot_tier_only"`
- `forma.PartialResultInfo.UnconsultedTiers []string` (`unconsulted_tiers`, omitempty)
- `model.PartialScan.UnconsultedTiers []DataTier` (mutually exclusive with `ExcludedObjects`)

## Files

- `internal/federated/routing.go`: `hasExplicitTierCoverage`, override block.
- `internal/federated/engine.go`, `engine_degrade.go` (moved degrade helpers), `engine_tier_coverage.go` (new: `unconsultedTiers`, `markHotTierOnly`).
- `internal/entity_query_service.go`, `entity_query_projection.go` (moved projections; `toPartialResult` projects the new reason).
- `internal/model/query.go`, `types.go`, `config_duckdb.go` (hybrid strategy comment).
- `docs/federated-query/design.md`: §4.3 `preferred_tiers` row, §4.3.1 routing note, §7.3 coverage-marker paragraph.
- Tests: `routing_tiers_test.go`, `engine_tier_coverage_test.go`, `entity_query_tier_coverage_test.go`, `tier_coverage_routing_e2e_test.go`; `TestDegradedFallbackClearsPartialMarker` renamed to `TestDegradedFallbackReplacesPartialMarker` under the new contract.

The two file splits keep `engine.go` and `entity_query_service.go` under the 500-line limit.

## Verification

- `make fmt-check`, `make lint`: pass.
- `make test` via `scripts/test_with_container_runtime.sh` (Podman): pass.
- `go test ./internal/e2e_harness/production/ -tags=e2e` (full suite, 418s): pass.
- `go test ./internal/e2e_harness/federated/... -tags=e2e -short` (385s + benchmark 129s): pass.


iceboundrock:
**Verification notes** for this PR (artifact type: verification record; working copy lives in the session scratchpad only, this comment is the record).

### Gates

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (golangci-lint v1.64.8, root + infra) | pass |
| `make test` via `scripts/test_with_container_runtime.sh` (rootless Podman) | pass, all packages |
| `go test ./internal/e2e_harness/production/ -tags=e2e` (full production suite) | pass, 418s |
| `go test ./internal/e2e_harness/federated/... -tags=e2e -short` | pass, 385s (+ benchmark package 129s) |

### TDD trail

Every production change was preceded by a failing test, and each failure was checked to be for the intended reason:

- `routing_tiers_test.go`: explicit `[hot,warm,cold]` / `[hot,cold]` under hybrid + small limit failed with `UseDuckDB=false` before the override; implicit, `[hot]`-only and disabled-engine cases were confirmed to stay on the heuristic after it.
- `engine_tier_coverage_test.go`: build failure on the missing `PartialScan.UnconsultedTiers` field, then `page.Partial == nil` on the routed, disabled and degraded paths before `markHotTierOnly` was wired.
- `entity_query_tier_coverage_test.go`: `federatedPreferredTiers(nil)` returned the default three before the service stopped filling it; `toPartialResult` returned nil for an `UnconsultedTiers`-only scan before the projection was added.
- `TestDegradedFallbackClearsPartialMarker` (existing) failed under the new contract and was renamed to `TestDegradedFallbackReplacesPartialMarker`: the degraded answer now replaces the abandoned DuckDB pass's corrupt-exclusion marker with the coverage marker instead of clearing it, per the ruling on #468 that the degraded path also carries `hot_tier_only`.

### e2e first run

`TestTierCoverageUnderHybridSmallPage` failed on its first run with `column "schema_id" does not exist` while pruning the exported rows from Postgres: `entity_main` keys on `ltbase_schema_id`, unlike `change_log` and `eav_data`. Fixed in the seed helper; both subtests pass, and the rest of the targeted set (`TestKeysetUnderProductionRoutingDefaults`, `TestTierHint*`, `TestDegradedMode_*`) passed unchanged on the same run.

### Decisions taken during execution

- Explicitness is carried by emptiness, not a new field: the service and the production harness forward an omitted `preferred_tiers` as nil. The engine, template and plan-cache shape already treat empty as the default all-tier form (#184), so the DuckDB path is unaffected.
- The e2e harness change (`buildFederatedQuery` no longer fills the default tiers) had no effect on existing harness tests: the production harness's default routing strategy is unset, which routes to DuckDB regardless, and the hybrid-strategy keyset test is already rerouted by #354.
- `engine.go` (505 lines) and `entity_query_service.go` (507 lines) crossed the 500-line guard after the change; the degrade helpers and the plan/partial projections were moved verbatim into `engine_degrade.go` and `entity_query_projection.go`.

iceboundrock:
## Review Summary

### Findings

- **[P2] Keep disabled-routing execution-plan tiers consistent with the new coverage marker** — `internal/federated/routing.go:46-50`

  When `cfg.Enabled` is false, `EvaluateRoutingPolicy` returns before the common `if !dec.UseDuckDB { dec.Tiers = []model.DataTier{model.DataTierHot} }` normalization at lines 112-115. For example, a request with `PreferredTiers: [hot, warm, cold]` and `include_execution_plan: true` is served from PostgreSQL only and correctly receives `partial: {reason: "hot_tier_only", unconsulted_tiers: [warm, cold]}`, but its execution plan still reports `routing.tiers: [hot, warm, cold]` while its only source is PostgreSQL/hot. This conflicts with the execution-plan contract that it reports the route actually taken and makes the new marker and plan disagree. Move the disabled-engine return after the common normalization or set `dec.Tiers` to hot before returning, and add an execution-plan regression assertion for the disabled-engine case.

### Spec Check

- The explicit multi-tier override is implemented next to the keyset-cursor override and preserves the hot-only gate.
- Omitted `preferred_tiers` remains empty through the service and production harness, allowing the engine to distinguish omission from an explicit declaration.
- The coverage marker is wired for the routed PostgreSQL path, globally disabled DuckDB, and degraded fallback; it is omitted for hot-only requests.
- The public `PartialReasonHotTierOnly` and `unconsulted_tiers` fields are additive and correctly projected without exposing storage internals.
- The ordinary-page-size production E2E covers both explicit routing to DuckDB and implicit hot-tier-only behavior.
- The disabled-engine execution-plan consistency case is the only spec-level gap found.

### Standards Check

- No source file or changed function appears to exceed the repository limits.
- `make fmt-check`: pass.
- `make lint`: pass.
- `git diff --check`: pass.

### Assessment

The core routing and coverage semantics are sound, and the test coverage is substantial. I recommend addressing the P2 execution-plan inconsistency before merge; after that, this looks ready for approval.

### Observations

- `internal/model/query.go:56-62` still describes `PartialScan` as a DuckDB-pass report, although it now also carries PostgreSQL coverage gaps.
- `internal/federated/engine.go:144-149` still says degraded fallback clears the marker, while the new fallback replaces the stale corrupt-exclusion marker with the hot-tier-only marker. These comments should be updated while fixing the finding.

### Verification

- PR checks: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/federated ./internal -count=1`: pass.
- `./scripts/test_with_container_runtime.sh` using the Podman socket: pass for the full unit-test matrix.
- A direct `make test` attempt initially hit the known testcontainers runtime setup failure (`$XDG_RUNTIME_DIR`/Docker host); the repository-prescribed runtime script passed after configuring Podman.

### Summary

**Request changes** for the disabled-DuckDB execution-plan tier mismatch. No other blocking correctness issues were found.

iceboundrock:
**Review response** (addresses the review summary above; all changes are in commit 5de1178).

### [P2] Disabled-routing execution-plan tiers vs. coverage marker — fixed

Confirmed. `EvaluateRoutingPolicy`'s `!cfg.Enabled` branch returned before the common `dec.Tiers = [hot]` collapse, so a `[hot, warm, cold]` request served Postgres-only because DuckDB is off reported `routing.tiers: [hot, warm, cold]` in the plan while the same page carried `partial: {reason: hot_tier_only, unconsulted_tiers: [warm, cold]}`. The other two Postgres-only paths already reported `[hot]` (`recordHotOnlyGatePlan`, `recordDegradedFallbackPlan`), so the disabled branch was the only inconsistent one.

Fix: the disabled branch now sets `dec.Tiers = [hot]` before returning (routing.go), with a comment explaining why the early return needs its own collapse. Two regression tests, both red before the fix for the intended reason (`actual: [hot, warm, cold]`):

- `TestEvaluateRoutingPolicyDisabledDuckDBCollapsesTiersToHot` (routing_tiers_test.go) pins the policy-level decision.
- `TestQueryDisabledDuckDBPlanTiersAgreeWithMarker` (engine_tier_coverage_test.go) pins the engine-level plan: `ExecutionPlan.Routing.Tiers == [hot]` and `Partial.UnconsultedTiers == [warm, cold]` on the same page.

The pre-existing behaviour dates from before this PR, but the marker is what made the disagreement observable, so it belongs here rather than in a follow-up.

### Observations — both comments updated

- `internal/model/query.go` `PartialScan` doc: now describes the type as "the answer covers less than the request asked for, in one of two mutually exclusive ways", with the #251 corrupt-exclusion scope moved onto the `ExcludedObjects` field doc alongside the existing `UnconsultedTiers` doc.
- `internal/federated/engine.go` entry-reset comment: now says the degraded fallback *replaces* the marker with the #468 coverage marker, and that the entry reset exists for the hot-only gate (the only path that leaves the page unmarked). The routed Postgres-only path no longer needs to be named there since `markHotTierOnly` overwrites the out-parameter on that path.

### Verification

- `go test ./internal/federated ./internal ./internal/model -count=1`: pass.
- `make fmt-check`, `make lint` (root + infra): pass.

No follow-up issues: the finding and both observations were within this PR's scope and are resolved.

iceboundrock:
## Review Summary

### Findings

No blocking correctness findings in the latest head (`5de1178`). The previous P2 execution-plan tier mismatch for disabled DuckDB routing is fixed in `internal/federated/routing.go`, and the added policy- and engine-level regressions cover it.

### Spec Check

- Explicit multi-tier `preferred_tiers` overrides the hybrid small-result heuristic while hot-only requests and disabled DuckDB retain their intended behavior.
- Omitted `preferred_tiers` remains empty through the service and production harness, preserving the distinction between omission and an explicit declaration.
- `hot_tier_only` coverage markers are emitted for routed PostgreSQL-only, disabled-DuckDB, and degraded-fallback paths, and are projected to the public `partial.unconsulted_tiers` field without exposing storage internals.
- The ordinary-page-size production E2E exercises both explicit tier routing through DuckDB and implicit hot-tier-only behavior.
- The disabled-DuckDB execution plan now reports `[hot]`, consistent with the actual source and coverage marker.

### Standards Check

- Changed source files and functions remain within the repository size limits.
- `make fmt-check`: pass.
- `make lint`: pass for the root and `infra/` modules.
- `git diff --check`: pass.

### Assessment

The implementation matches the approved #468 ruling and has substantial unit and production-level coverage. No request changes are needed. Recommendation: approve / ready to merge.

### Observations

- Non-blocking documentation cleanup: `internal/sqlgen/duckdb_template_renderer.go:219` still says the HTTP layer and harness normalize empty `PreferredTiers` to all three tiers. Both now forward an empty list, relying on the engine/template default semantics.
- The new public `partial` shape is covered by service-level JSON tests, but `internal/httpapi/server_advanced_query_test.go` does not yet assert the new marker on the actual HTTP response.
- The degraded-path unit coverage uses an explicit tier list; an omitted-tier degraded case would be a useful additional regression test, although the shared `unconsultedTiers` logic and implicit routed/disabled cases are covered.

### Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/federated ./internal -count=1`: pass.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/httpapi ./internal/model -count=1`: pass.
- GitHub CI checks: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all pass.
- A local `make test` run passed all non-container packages but could not initialize `internal/e2e_harness/TestE2EHarnessMinimal` because testcontainers could not resolve `$XDG_RUNTIME_DIR`; this is an environment/runtime limitation, not a PR failure.

### Summary

No blocking findings. The routing override, coverage marker, public projection, disabled-route plan consistency, and regression coverage are complete for this PR.

PR #526, opened 2026-09-05
security(federated): harden in-bucket scope of caller s3_parquet_path_template (#477)

### LINKED ISSUE #477: security(federated): harden in-bucket scope of caller s3_parquet_path_template (forbid **, exclude _tmp/, optional s3DataPrefix granularity)
Follow-up from PR #476 (#456) review.

## Context

#456 gated `federated.s3_parquet_path_template` behind an opt-in and allowlisted rendered paths to `s3://<S3Bucket>/`. That closes the injection and the cross-bucket / local-file / SSRF vectors. It intentionally stops at **bucket** granularity.

## Observation

With the opt-in enabled, a caller may still target any prefix **inside** the configured bucket:
- `**` recursive globs are honored in hints, even though the level-3 fallback glob deliberately forbids widening to `**` (`docs/federated-query/design.md` §4.3.1).
- In-flight `_tmp/` staging objects and base-tier files are reachable, whereas the fallback glob's single `*` deliberately excludes `_tmp/`.

Impact is bounded to server-written intra-bucket data — not a cross-tenant or arbitrary-host read — so it was left out of the security fix. But it is a real breadth-of-access concern if the feature sees production use.

## Proposal (to discuss)

- Optionally scope honored hints below bucket granularity — e.g. require the rendered path to sit under `s3://<S3Bucket>/<S3DataPrefix>/…` when `S3DataPrefix` is set.
- Reject `**` in hint paths, matching the fallback-glob rule.
- Consider excluding `_tmp/` staging prefixes.

## Acceptance (draft)

- [ ] Decide the intended scope granularity (bucket vs bucket+prefix) and document it.
- [ ] If tightened: reject `**` and out-of-prefix hints, with tests.

iceboundrock:
Closes #477.

## Problem

#456 gated `federated.s3_parquet_path_template` behind `duckdb.allowCallerParquetPaths` and allowlisted rendered paths to `s3://<S3Bucket>/`. It stopped at bucket granularity, so with the opt-in on a caller could still target any prefix in the bucket, honor `**` (which the level-3 fallback glob forbids), and reach in-flight `_tmp/` staging objects.

## Change

`validateHintPathScope` (moved from `duckdb_query_build.go` to a new `internal/federated/parquet_hint_scope.go`) now enforces, on top of the #456 bucket rule and character check:

- **Prefix granularity.** When `duckdb.s3DataPrefix` is non-empty, every rendered path must sit under `s3://<bucket>/<prefix>/`. Trailing slash exact (blocks `<prefix>X/` collisions and the bare prefix with no key). The prefix is canonicalized with the writers' single `TrimSuffix("/")`, same as `S3FallbackGlob`, so `"delta/"` and `"delta"` scope identically; an empty or `/` prefix keeps bucket granularity. No new config field.
- **Shape rules**, the same ones §4.3.1 holds the fallback glob to: `**` is refused; wildcards (`*`, `?`, `[`) may appear only in the final object-name segment, so a pattern can never match a directory; a literal `_tmp` segment is refused; a path ending in `/` is refused. Together these keep `_tmp/` unreachable from a hint (a wildcard directory segment alone would match it without spelling it).

Every rejection stays `forma.InvalidInputf` echoing the caller's own rendered path; the configured bucket and prefix stay behind `WithOperatorDetail` (#313).

Every hint shape the repo relies on still passes: `…/{{.SchemaID}}/*.parquet` (production e2e `ParquetGlob`), `…/base/*.parquet, …/delta/*.parquet` (benchmark), and literal manifest URIs (`<prefix>/<schema>/<uuid>.parquet`, used with `S3DataPrefix = env.S3Prefix` in `factory_manifest_e2e_test.go`).

## Decision record

Scope granularity (the issue's open question) was decided as bucket + `S3DataPrefix`, reusing the field rather than adding `callerParquetPathPrefix`. Accepted consequence: `S3DataPrefix` still requires `manifestTemplate` at startup, so prefix-level tightening is available to manifest-read deployments; glob-only deployments keep bucket granularity. The design comment is on #477.

## Docs

- `docs/federated-query/design.md`: §4.3 request table, §4.3.1 level 1, server config table (`s3DataPrefix`, `allowCallerParquetPaths`).
- `docs/error-handling.md`: publication table row.
- `config_duckdb.go`: `S3DataPrefix` / `AllowCallerParquetPaths` field comments.

## Tests

New cases in `internal/federated/parquet_hint_gate_test.go`, written red first: `**` in three positions; literal `_tmp` at three depths; wildcard directory segments (`*`, `?`, `[_]`, `_t*`); trailing `/`; accepted shapes including `not_tmp/` and `_tmp_archive/` (segment equality, not substring); in-prefix accepted for `"data"` and `"data/"`; out-of-prefix, `data2/` collision, bare prefix, mixed multi-path rejected; empty and `/` prefix keep bucket scope.

## Verification

- `make fmt-check` ✅
- `make lint` ✅ (root and infra, 0 issues)
- `./scripts/test_with_container_runtime.sh` ✅ (all 34 packages ok, Podman socket)


iceboundrock:
## Review Outcome
**Request changes** due to one conditional security finding below. The implementation otherwise matches the approved #477 design and the repository standards.

## Findings

- **[P2] Treat `S3DataPrefix` as a literal prefix before using it in a DuckDB glob** (`internal/federated/parquet_hint_scope.go:61-73`). `scope` concatenates the configured `dataPrefix` directly into the rendered `read_parquet` path, while `validateHintPathShape` only examines the suffix after `scope` (`path[len(scope):]`). If a valid S3 prefix contains glob metacharacters, for example `data/*` or `data/**`, an accepted hint becomes `s3://bucket/data/*/7/*.parquet` or `s3://bucket/data/**/7/*.parquet`; DuckDB interprets those characters in the configured prefix as directory globs and can scan sibling/deeper prefixes outside the intended exact prefix. This bypasses the new prefix-granularity and no-recursive-directory-wildcard guarantees. `S3DataPrefix` currently has no validation excluding these characters, and S3 object keys can contain them. Either reject `*`, `?`, and `[` in the canonical prefix when caller hints are enabled, or escape the configured prefix before rendering it to DuckDB; add regression tests for `*`, `?`, `[`, and `**` in `S3DataPrefix`.

## Spec Check

- Bucket-plus-`S3DataPrefix` scope is implemented with the intended single trailing-slash canonicalization.
- `**`, literal `_tmp` segments, wildcard directory segments, and trailing-directory paths are rejected.
- Existing accepted hint shapes, multi-path hints, literal manifest URIs, and invalid-input/public-error semantics are preserved.
- The finding above is an edge case not covered by the approved design record on issue #477; without it, the documentation claim that wildcards are allowed only in the final segment is not true for wildcard characters supplied by the configured prefix.

## Standards Check

- Changed Go files remain below the 500-line file limit and functions remain below the 100-line limit.
- Errors use the existing `forma.InvalidInputf` and `WithOperatorDetail` conventions.
- The change is focused, documented, and has targeted unit coverage.
- No unrelated worktree changes or formatting issues were found.

## Assessment

The PR is close to mergeable, but the prefix-metacharacter case should be closed before merging because this PR changes a storage-scope security boundary. The normal `data`, `data/`, and empty-prefix configurations behave as intended.

## Observations

- The new tests validate the gate but do not execute a real DuckDB glob with a metacharacter-bearing configured prefix, so this edge case is not detected.
- The test suite intentionally accepts `?` in the final object segment, matching the approved #477 contract; DuckDB documents `?` as unsupported for S3 reads, so that remains a compatibility concern rather than a separate finding here.

## Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/federated -run 'TestHint' -count=1 -v` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/manifest ./factory -run 'Fallback|ManifestParquetSource|Hint' -count=1` passed.
- `make fmt-check` passed.
- PR CI run `33999672991` passed: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

The path-scope hardening is well targeted and the normal paths are covered. Please address the configured-prefix glob interpretation and add regression coverage; after that, I have no further blocking findings.

iceboundrock:
## Response to review

### [P2] Glob metacharacters in `S3DataPrefix` — fixed in 23a253e

Confirmed against the code before changing anything: `scope` was built from the raw prefix and `validateHintPathShape` only saw `path[len(scope):]`, so with `S3DataPrefix = "data/*"` the hint `s3://b/data/*/{{.SchemaID}}/x.parquet` rendered, passed, and reached `read_parquet` as a directory wildcard. With `data/**` the key-level `**` check never fired either, and `s3://b/data/**/<uuid>.parquet` matches `data/<x>/7/_tmp/<uuid>.parquet`, so `_tmp/` was reachable. Red tests reproduced both before the fix.

Chose **reject** over **escape**: DuckDB's S3 glob has no documented escape syntax, and a prefix that needs escaping is a misconfiguration rather than an intent. Two layers, mirroring the existing missing-bucket rule:

- `DuckDBConfig.ValidateCallerParquetPaths` rejects an `S3DataPrefix` containing `*`, `?`, or `[` when `allowCallerParquetPaths` is on. `ConfigError` on field `duckdb.s3DataPrefix`, so it fails at startup from both `cmd/server` and the factory preflight, before any I/O. Gated on the opt-in to stay inside this PR's boundary; with the flag off the prefix validates as before.
- `validateHintPathScope` refuses every hint when the prefix carries a metacharacter, for a hand-built engine that skipped `Config.Validate`. Same published message as the empty-bucket case (`not permitted on this deployment`); the prefix stays behind `WithOperatorDetail`.
- The character set is now `forma.ParquetGlobMetacharacters`, exported from the root package so the config check and the engine's wildcard-placement rule read one constant instead of two copies. The bucket needs no such rule: S3 bucket naming excludes these characters.

Regression tests: `TestValidateDuckDBConfig_AllowCallerParquetPathsRejectsGlobPrefix` (`data/*`, `data/**`, `data?`, `data[1]`, each asserted to validate with the flag off and fail on `duckdb.s3DataPrefix` with it on) and `TestHintGlobMetacharDataPrefixRejected` (same set plus bare `*`; asserts `ErrInvalidInput`, that the published message omits the prefix, and that the operator copy names it). Docs: §4.3.1 level 1, the config table row for `allowCallerParquetPaths`, the startup-rejection list, the `docs/error-handling.md` publication row, and the `S3DataPrefix` field comment.

### Observation: tests do not run a real DuckDB glob with a metacharacter prefix

Not adding one. The fix rejects the prefix before `read_parquet` is rendered, so the property under test is "this configuration never reaches DuckDB", which the two gate tests pin directly. A DuckDB-level test would only re-verify DuckDB's glob semantics.

### Observation: `?` in the final object segment

Keeping as-is, per the #477 contract. `?` is confined to the object-name segment by the existing shape rule, so a pattern DuckDB cannot expand on S3 fails that one read; it cannot widen the scope. Compatibility, not a boundary concern, as the review notes.

### Follow-up

The same prefix string feeds the level-3 fallback glob, manifest URIs, and the writers' `--s3-prefix` / `--data-prefix` flags, all unvalidated, and the writers store it as a literal key while the reader glob-expands it. That reader/writer disagreement is independent of caller hints and out of scope here, so it is filed as #527 (unconditional rejection on both sides; low priority, no known deployment affected). Once #527 lands, the opt-in-gated check from this commit folds into it.

### Verification

- `make fmt-check` ✅
- `make lint` ✅ (root and infra, 0 issues)
- `./scripts/test_with_container_runtime.sh` ✅ (34 packages ok)

iceboundrock:
## Review Outcome

No findings. The previously identified `S3DataPrefix` glob-metacharacter scope bypass is fixed in `23a253e`, with both startup validation and the runtime defense-in-depth check now covered.

## Findings

No blocking or non-blocking findings remain in the current PR head.

## Spec Check

- Bucket and optional `S3DataPrefix` scope use the required trailing-slash boundary and single trailing-slash canonicalization.
- `**`, wildcard directory segments, literal `_tmp` segments, and trailing-directory paths are rejected.
- Prefix glob metacharacters (`*`, `?`, `[`) are rejected when caller parquet hints are enabled, preventing configured-prefix expansion from bypassing the shape rules.
- Accepted production shapes remain supported: schema globs, base/delta multi-path hints, and literal manifest URIs.
- Invalid-input publication and operator-detail handling match the approved #477 design and error-handling rules.

## Standards Check

- Changed Go files remain below the 500-line file limit and functions remain below the 100-line limit.
- Errors use the existing `forma.InvalidInputf`, `ConfigError`, and `WithOperatorDetail` conventions.
- The change is focused, documented, and has targeted regression coverage.
- No unrelated worktree changes or formatting issues were found.

## Assessment

The PR is ready to merge from a code-review perspective. The security boundary is enforced at configuration time for normal startup paths and again at hint resolution for hand-built configurations that bypass config validation.

## Observations

- Final-segment `?` patterns remain accepted by contract and are not a scope bypass; deployments should avoid them if their DuckDB/S3 backend does not expand `?` as expected.
- The metacharacter regression tests intentionally verify rejection before SQL generation or S3 access, which is the relevant security property.

## Verification

- Targeted hint and config tests passed with `go test ./internal/federated -run 'TestHint|TestValidateDuckDBConfig_AllowCallerParquetPaths' -count=1 -v`.
- Full repository unit test script passed: `./scripts/test_with_container_runtime.sh` (Podman socket; all packages passed).
- `make fmt-check` passed.
- `git diff --check origin/main...HEAD` passed.
- PR CI run `34000585791` passed: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Summary

The current implementation closes the reviewed prefix-glob bypass, preserves the approved path-scope contract, and passes local and CI verification. No changes requested.

Issue #351, opened 2026-08-04
Correlate #256 scan-guard failures back to the offending parquet object


iceboundrock:
Spun out of PR #350 review (P2-1).

When the #256 scan guard fires (`NULL row_id` / `NULL changed_at` / a BIGINT cast failure), the operator-facing error is a constant message naming the violated invariant but not the object: the scan is a multi-file `read_parquet` union, so DuckDB cannot attribute the failing row to a file, `classifyDuckDBReadError` only detects *absent* objects (a schema-wrong object exists), and #251's verify-and-exclude drains with a bare `SELECT *` (no guard) so a schema-wrong object reads clean and is never classified. In the exact case the guard exists for — a stamp short-circuit — `warnStampDivergence` never ran either, so no log names a path.

Proposal: on a guard-classified read failure, extend the #251 verify pass with a *guarded* per-file drain (`BuildParquetScanSource` over a single path) to identify which object(s) violate the invariant, and surface the keys in the classified error / logs (identification only — do NOT auto-exclude schema-wrong objects; unlike byte corruption, exclusion would silently drop rows the object may legitimately own once repaired). Runs only on the failure path, so cost is bounded by failures.

Until then the triage path is documented in docs/federated-query/design.md (PR #350): list the schema's manifest paths and run the guarded scan per file to bisect.

iceboundrock:
**Artifact: implementation plan** (working copy: `docs/superpowers/plans/2026-08-16-issue-351-guard-identification.md`, gitignored). Status: plan as written, pre-execution — mid-execution rulings will follow as separate comments.

# Issue #351: Correlate #256 Scan-Guard Failures Back to the Offending Parquet Object — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** When a federated read fails on the #256 scan guard (NULL `row_id` / NULL `changed_at` / a BIGINT cast failure), automatically identify which parquet object(s) violate the export schema invariant and surface their keys in the classified error and the operator log — identification only, never exclusion.

[...]

PR #528, opened 2026-09-06
fix(federated): let unverifiablePath admit quote-bearing paths (#479)

### LINKED ISSUE #479: federated: let unverifiablePath admit escaped quote-bearing paths so per-file verify/guard attribution covers them
Follow-up from PR #476 (#456) review.

## Context

`internal/federated/parquet_verify.go` `unverifiablePath` skips paths containing `' " ; * ? [` — written before #456, when those paths were interpolated raw and so genuinely unquotable per-file.

## Observation

Since #456, every render site escapes the path via `sqlutil.EscapeLiteral`, so a quote-bearing path can now be rendered safely. `unverifiablePath` still excludes quote-bearing entries, so per-file verification (#251) and guard attribution (#351) leave a **non-security** coverage gap for manifest entries whose object keys legitimately contain a quote: such an entry falls back to all-or-nothing behavior instead of being individually excludable/nameable. (Glob-bearing entries — `* ? [` — must still be skipped; a glob names a set, not one object.)

## Proposal

- Narrow `unverifiablePath` to skip only glob metacharacters (`* ? [`), not quotes, now that quotes render safely.
- Add a regression test with a quote-bearing path being individually verified/attributed.

## Acceptance (draft)

- [ ] `unverifiablePath` no longer excludes quote-only paths.
- [ ] Regression test covering per-file verify/guard on a quote-bearing path.

iceboundrock:
Closes #479.

## Problem

`unverifiablePath` in `internal/federated/parquet_verify.go` skipped any path containing `' " ;`. That was correct before #456, when `drainParquet` and `drainGuardedParquet` interpolated the path raw. Since #456 both drains escape the path via `sqlutil.EscapeLiteral` (and the `TestParquetPathRendersAreEscaped` AST guard keeps it that way), so the skip no longer protected anything. It only withheld per-file exclusion (#251) and per-file attribution (#351) from manifest entries whose object key legitimately carries one of those characters, leaving them on all-or-nothing behavior.

## Change

- `unverifiablePath` now skips only glob metacharacters (`forma.ParquetGlobMetacharacters`, `* ? [`): a glob names a set, so a solo drain of it proves nothing about one object.
- Doc comments on `verifyParquetPaths`, `unverifiablePath`, `identifyGuardViolations`, and the corresponding sentence in `docs/federated-query/design.md` no longer claim quotes are unquotable.

Not touched: the main set scan, glob handling, and the #456 caller-template gate (`parquet_hint_scope.go`), which still rejects `' " ; \n \r` in caller-supplied paths. That gate is about caller input; this change is about manifest entries.

## Tests

- `TestVerifyParquetPathsVerifiesQuoteBearingPath`: a path carrying `'`, `"`, and `;` whose drain fails deterministically is confirmed corrupt on its own, and every drain of it renders the quote doubled.
- `TestIdentifyGuardViolationsAttributesQuoteBearingPath`: the same shape through the guarded drain is attributed as a violator, and the raw path never appears in rendered SQL.
- The two existing skip tests narrow to glob entries only (`*`, `?`, `[`).

Both new tests were written first and failed against the old predicate.

## Verification

`make fmt-check` ✅ · `make lint` ✅ (root and infra modules, 0 issues) · unit suite over `.`, `./cdc`, `./cmd/...`, `./factory`, and `./internal/...` excluding the Docker-gated `e2e_harness` ✅.

## Follow-up

The #256 schema validator (`parquet_schema_validation.go`, `Validate` and `validateConcrete`) has the same pre-#456 `' " ;` skip and marks such paths incomplete. Same reasoning applies; tracked in #529 rather than widening this PR.


iceboundrock:
## Review Summary

### Findings

No blocking findings for the stated scope of #479. The change correctly narrows `unverifiablePath` to DuckDB glob metacharacters while retaining the safety boundary at every per-file SQL render site.

### Spec Check

- `unverifiablePath` now skips only `*`, `?`, and `[`, so a literal manifest object key containing `'`, `"`, or `;` can reach per-file verification.
- Both bare verification and guarded attribution have regression coverage for quote-bearing paths, including doubled single quotes in rendered SQL.
- Existing glob behavior remains all-or-nothing and is covered for all three metacharacters.
- Caller-supplied path validation and provenance gates remain unchanged, so this does not reopen the #456 caller-input surface.
- The design documentation now matches the implementation.

### Standards Check

- Modified Go files and functions remain within the repository size limits.
- Error wrapping and sentinel classification behavior are unchanged.
- `git diff --check` passed.
- No unrelated files or behavior were changed.

### Assessment

Approve for the stated #479 scope. The implementation is minimal, uses the shared `forma.ParquetGlobMetacharacters` constant, and preserves the existing security and glob semantics.

### Observations

- The quote-bearing path support is not yet end-to-end across every parquet subsystem. `internal/federated/parquet_schema_validation.go` still marks paths containing `'`, `"`, or `;` as incomplete, so pre-read schema validation and cold-column union/pinning remain deferred for those entries. This is the explicitly tracked follow-up in #529 and is not counted as a blocking finding for this PR.
- `internal/compaction/merge_sql.go` also retains its pre-existing URI rejection for these characters. Current writer-generated keys do not use them; this PR does not change compaction behavior.
- The new tests use executor fakes for deterministic SQL/control-flow assertions; they do not add a live DuckDB/S3 case. Existing render guards and CI provide complementary coverage.

### Verification

- `go test ./internal/federated` passed.
- `go test -race ./internal/federated` passed.
- Non-container unit packages passed under `go1.26.0+auto`.
- `make fmt-check` passed.
- `make lint` passed for both the root and infra modules.
- Local `make test` reached all other packages but failed in `internal/e2e_harness/TestE2EHarnessMinimal` because this environment has no usable container runtime and `$XDG_RUNTIME_DIR` is unavailable; this is an environment limitation, not a failure in the changed package.
- GitHub checks are all green: Test, Lint, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

PR #509, opened 2026-09-01
fix(federated): unify the keyset cursor contract across the engine's entry points (#381)

iceboundrock:
Closes #381.

The keyset cursor contract differed depending on which engine entry point a cursor arrived through, and one of those differences was a live correctness bug. This makes the contract single and uniform.

## The contract

A cursor is valid when `len(Values) == len(Columns)`, no boundary value is nil, the final column is `row_id` (case aside), `Mode` is `after` (`before` is refused until #513 builds its backward window), each column's `Direction` is `asc`, `desc` or empty, it **continues** the request's `AttributeOrders` when there are any (same attributes in order, same directions, then `row_id ASC`; with no `AttributeOrders` the cursor is the order), and every column — judged on its `sqlgen.ParquetAttrColumn` **fold** — is one of the four `visible`-CTE system columns (`row_id`, `created_at`, `ver_ts`, `deleted_ts`), is not dedup machinery (`rn`, `source_tier_priority`), did not trigger `ParquetAttrColumn`'s `"attr"` empty-fold fallback, and matches `^[A-Za-z_][A-Za-z0-9_]*$`.

`federated.validateKeysetCursor` enforces it, and every seam that can reach the DuckDB keyset renderer calls it independently, each with the order it is about to render: `Query`, `ExecuteFederatedPaginatedQuery`, `ExecuteDuckDBFederatedQuery`, `StreamDuckDBFederatedQuery`. The shape and continuation rules are re-checked inside `sqlgen` as well.

Separately, the explicit `limit`, `offset` and `attributeOrders` arguments are now the pagination contract of every DuckDB seam: `buildDuckDBQueryWithPlan` renders a copy of the query carrying them (`federated.dispatchedQuery`), so a normalised or clamped limit is what renders.

## What changed, by issue item

| Item | Change |
| --- | --- |
| 1. Seams disagreed on admissible columns | One validator at every seam; cursor columns emitted folded |
| 2. `col.Attribute` interpolated unquoted | Safe-identifier check on the folded name is the injection barrier |
| 3. Split-brain page in `ExecuteFederatedPaginatedQuery` | An active cursor alone selects the keyset path |
| 4. `opts.KeysetEnabled` inert on `Query` | Field deleted |
| 5. `hotOnly` / `isHotOnlyRequest` asymmetry | Characterization test; no behaviour change |
| 6. Stale `isSupportedKeysetColumn` doc comment | Function removed |
[...]

PR #530, opened 2026-09-06
fix(sqlgen): refuse an unregistered filter attribute that folds onto a visible column (#512)

### LINKED ISSUE #512: An unregistered filter attribute folds onto a visible system or dedup column
Found while fixing the keyset cursor contract in #381 / #509. The cursor path is now closed; the **filter** path is the same defect, and it is not in that issue's scope.

## The defect

`normalizeLeaf` (`internal/sqlgen/predicate_normalizer.go:66`) looks the condition attribute up in the schema cache but does **not** require it to be there:

```go
meta, hasMeta := cache[kv.Attr]
leaf.HasMeta = hasMeta
...
leaf.Duck = normalizeDuckPayload(kv, meta, hasMeta, lenientSQL, lenientSQLErr)
```

`normalizeDuckPayload` (`predicate_normalizer.go:396`) then emits `resolveDuckDBColumn(kv.Attr)` regardless, which is `ParquetAttrColumn` (`dualpath_sql_helpers.go:205`). The fold is lossy, so an unregistered attribute name lands on a real column of the `visible` CTE:

| `kv.Attr` (unregistered) | emitted DuckDB column | what it actually filters on |
| --- | --- | --- |
| `created.at` | `created_at` | the record's creation timestamp |
| `ver.ts` / `deleted.ts` | `ver_ts` / `deleted_ts` | the LWW version / soft-delete stamp |
| `row.id` | `row_id` | the primary key |
| `rn` / `source_tier.priority` | `rn` / `source_tier_priority` | federated dedup machinery |
| `[]`, `` ` ``, `""` | `attr` | `ParquetAttrColumn`'s empty-name placeholder |

Probed directly against `normalizeLeaf` with an empty cache:

```
attr="created.at"           hasMeta=false duckColumn="created_at"
attr="rn"                   hasMeta=false duckColumn="rn"
attr="[]"                   hasMeta=false duckColumn="attr"
attr="source_tier.priority" hasMeta=false duckColumn="source_tier_priority"
```

The caller gets a well-formed page filtered on a column they did not name — the silent-wrong-answer family of #354, not a loud binder error.

## Why registration does not cover it

`sqlgen.ValidateParquetAttrColumns` rejects a *registered* attribute whose fold collides with a reserved parquet column, and that is what protects the projection and the ORDER BY. It cannot protect this path, because the whole point of the `hasMeta == false` branch is that the name was never registered. `kv.Attr` is an arbitrary caller-supplied string.

## Why the sort path is fine and this one is not

`buildAttributeOrders` (`internal/entity_query_sort.go:91`) errors on an unknown sort attribute:

```go
meta, ok := schemaCache[key.attr]
if !ok {
    return nil, forma.InvalidInputf("cannot sort by unknown attribute '%s' in schema '%s'", ...)
}
```

so `buildNonKeysetOrderBy`'s `ParquetAttrColumn(ao.AttrName)` (`duckdb_template_renderer.go:368`) only ever sees registered names. Filters have no equivalent gate.

## Relation to #381 / #509

#509 closed exactly this hole for keyset cursor columns, where the same arbitrary-string exposure existed: `federated.validateKeysetCursor` now refuses the `attr` placeholder, refuses the dedup machinery, and (as of that PR's review round) refuses any **non-identity fold** onto one of the four `visible` system columns — a system column is admitted only under its own name, case aside. The filter path needs the same rule; the fix simply does not belong to the cursor contract.

## Suggested fix

Decide between two, in the leaf normalizer so every dialect inherits it:

1. **Refuse the unregistered name outright** — a filter on an attribute the schema does not define is caller fault, exactly as it is for sort, and would be a `forma.InvalidInputf` carrier at the same boundary. This is the honest behaviour, but it is a behaviour change for any caller relying on the lenient `hasMeta == false` path, so it needs a check of what that path is currently for.
2. **Keep the lenient path but refuse a fold onto a reserved column** — reuse `reservedParquetColumns` (`parquet_attr_column.go:42`), matched on the folded name, lower-cased, and admit only an identity-up-to-case fold, mirroring `validateKeysetCursor`. Narrower, and closes the silent-retarget without touching the lenient path's purpose.

Option 2 is the smaller change and the one that matches the cursor contract; option 1 is worth evaluating on its own merits because "filter on a name the schema never defined" is hard to justify as a success case.

## Scope note

Only the leaf normalizer is implicated. The projection (`duckdb_schema_projection.go`), the ORDER BY, and the CDC export all fold names that came from the registered schema, and `ValidateParquetAttrColumns` already guards those at registration.

iceboundrock:
Closes #512.

## What

`normalizeDuckPayload` folded an unregistered filter attribute through `ParquetAttrColumn` unchecked, so a filter on `created.at`, `row.id`, `rn` or `[]` rendered a DuckDB clause on `created_at`, `row_id`, the dedup rank or the `attr` placeholder: a well-formed page filtered on a column the caller never named (the #354 silent-wrong-answer family).

This applies the #509 cursor rule to filters, in the DuckDB leaf normalizer (`normalizeDuckPayload`): the PG EAV payload already refuses an unregistered name and the PG main payload skips it, so the DuckDB payload is the one path that needed guarding on its own. Only when the attribute is **not** in the schema cache, judged on the folded name with case-insensitive lookups:

- a non-identity fold onto the `attr` placeholder is refused (`[]`, `` ` ``, `[attr]`);
- a non-identity fold onto any `reservedParquetColumns` entry is refused (`created.at`, `ver.ts`, `deleted.ts`, `row.id`, `[row_id]`, `Created.At`);
- `rn` and `source_tier_priority` are refused under any spelling;
- an identity-up-to-case fold (`created_at`, `Created_At`, `attr`) stays admitted, as for cursors.

Refusals are `forma.InvalidInputf` carriers naming the attribute, the folded column and what to do instead. Registered names never reach the check: `ValidateParquetAttrColumns` already rejects them at registration.

## Why option 2 (refuse the fold) and not option 1 (refuse any unregistered name)

Investigation notes and the ruling are on the issue: https://github.com/Lychee-Technology/forma/issues/512#issuecomment-5555953273. In short: on the federated route `ToDualClauses` renders the PG EAV payload before the DuckDB one, and that payload already refuses an unregistered name with `attribute not found in cache`, so option 1 would have zero production delta while forcing rewrites of the 32 nil-cache test sites that use the standalone `BuildDuckClause`. Option 2 makes the DuckDB payload safe on its own, whatever emitter order a caller uses, and keeps one rule for filters and cursors.

## Changes

- `internal/sqlgen/parquet_attr_column.go`: `ValidateUnregisteredParquetAttrColumn`, `parquetAttrPlaceholder`, `federatedDedupColumns`.
- `internal/sqlgen/predicate_normalizer.go`: `normalizeDuckPayload` calls it when `hasMeta` is false.
- `internal/sqlgen/duck_clause_test.go`: `TestBuildDuckClause_UnregisteredAttr_RefusesFoldOntoVisibleColumn` (18 refused spellings, 5 admitted) and `TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission` (pins the production-route behaviour the ruling relies on, including that the PG EAV refusal fires before the DuckDB guard).
- `docs/federated-query/design.md`: section 4 now states the shared filter/cursor fold rule.

## Verification

- Test written first and watched fail: all 13 refused spellings emitted a clause (e.g. `created_at = CAST(? AS BOOLEAN)` for `created.at`); green after the fix.
- `make fmt-check`: clean.
- `make lint` (golangci-lint v1.64.8, root and infra modules): clean.
- `./scripts/test_with_container_runtime.sh` (full `make test` under rootless Podman): 34 packages ok, no failures. `TestNoBareSentinelWraps` included.
- File sizes stay under the limits: `parquet_attr_column.go` 147 lines, `predicate_normalizer.go` 431 lines.

## Out of scope, recorded on the issue

- The cursor rule also requires the folded name to match `safeSQLIdentifier`; `parquetAttrReplacer` only folds backtick, dot, space and brackets, so the standalone `BuildDuckClause` route still interpolates other punctuation as-is. Not reachable in production.
- Retiring the lenient `hasMeta == false` DuckDB path altogether (option 1) is a possible later cleanup once the test corpus is migrated to registered caches.

## Review follow-ups

- #531: `federatedDedupColumns` / `parquetAttrPlaceholder` duplicate their `federated` keyset counterparts with no sync guard.
- #532: pre-existing, `ValidateParquetAttrColumns` looks reserved columns up case-sensitively, so a registered `Created_At` passes registration and binds to `created_at` in DuckDB.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -296,6 +296,19 @@ only because a fold onto a `visible` column, system or dedup, is refused before
 it can reach the generator. It is deferred to the caller-facing cursor surface
 the Postgres-side keyset feature introduces.
 
+Filter attributes follow the same fold rule (#512). The predicate normalizer
+does not require a filter attribute to be registered either: the PG EAV
+payload refuses an unregistered name with `attribute not found in cache`, and
+because `ToDualClauses` renders that payload before the DuckDB one, the
+federated route answers 4xx before any DuckDB clause exists. The DuckDB payload
+is nevertheless guarded on its own, so the rule holds whatever emitter order a
+caller uses: for an unregistered name, `sqlgen.ValidateUnregisteredParquetAttrColumn`
+refuses a non-identity fold onto the `attr` placeholder or onto any reserved
+parquet column, refuses `rn` and `source_tier_priority` under any spelling, and
+admits an identity-up-to-case fold, exactly as `validateKeysetCursor` does for
+cursor columns. Registered names never reach it: `ValidateParquetAttrColumns`
+already rejected any whose fold collides with a reserved column.
+
 ## **5. SQL Execution Template**
 
 This SQL template represents the core logic of the Federated Query Engine.
--- a/internal/sqlgen/duck_clause_test.go
+++ b/internal/sqlgen/duck_clause_test.go
@@ -123,3 +123,64 @@ func TestParseDateValue_ISO8601EncodingFromUnixMs(t *testing.T) {
 	_, err = ParseDateValue("not-a-date", meta)
 	require.Error(t, err)
 }
+
+// An unregistered filter attribute must never fold onto a column the visible
+// CTE projects (#512). ParquetAttrColumn is lossy, so "created.at" would
+// otherwise render as created_at and silently filter on the creation
+// timestamp. The rule mirrors federated.validateKeysetCursor (#509): a
+// non-identity fold onto the "attr" placeholder or a reserved parquet column
+// is refused, the dedup machinery is refused under any spelling, and an
+// identity-up-to-case fold stays admitted.
+func TestBuildDuckClause_UnregisteredAttr_RefusesFoldOntoVisibleColumn(t *testing.T) {
+	refused := []struct{ attr, folded string }{
+		{"created.at", "created_at"},
+		{"ver.ts", "ver_ts"},
+		{"deleted.ts", "deleted_ts"},
+		{"row.id", "row_id"},
+		{"[row_id]", "row_id"},
+		{"Created.At", "Created_At"},
+		{"rn", "rn"},
+		{"RN", "RN"},
+		{"source_tier.priority", "source_tier_priority"},
+		{"source_tier_priority", "source_tier_priority"},
+		{"[]", "attr"},
+		{"`", "attr"},
+		{"[attr]", "attr"},
+	}
+	for _, tc := range refused {
+		t.Run("refused/"+tc.attr, func(t *testing.T) {
+			cond := &forma.KvCondition{Attr: tc.attr, Value: "equals:1"}
+			clause, _, err := buildDuckClause(cond, forma.SchemaAttributeCache{})
+			require.Error(t, err, "clause %q emitted for unregistered attribute %q", clause, tc.attr)
+			require.ErrorIs(t, err, forma.ErrInvalidInput)
+			require.Contains(t, err.Error(), tc.attr)
+			require.Contains(t, err.Error(), tc.folded)
+		})
+	}
+
+	admitted := []string{"created_at", "Created_At", "attr", "Attr", "unknown_ts"}
+	for _, attr := range admitted {
+		t.Run("admitted/"+attr, func(t *testing.T) {
+			cond := &forma.KvCondition{Attr: attr, Value: "equals:1"}
+			clause, _, err := buildDuckClause(cond, forma.SchemaAttributeCache{})
+			require.NoError(t, err)
+			require.Contains(t, clause, attr)
+		})
+	}
+}
+
+// The production federated route builds all three clauses through
+// ToDualClauses, where an unregistered attribute is refused by the PG EAV
+// payload before the DuckDB clause is rendered (#512 investigation). This
+// pins that the folded spellings never reach DuckDB on that route either.
+func TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission(t *testing.T) {
+	for _, attr := range []string{"created.at", "rn", "[]", "ghost"} {
+		t.Run(attr, func(t *testing.T) {
+			paramIndex := 1
+			cond := &forma.KvCondition{Attr: attr, Value: "equals:1"}
+			_, err := ToDualClauses(cond, "eav_table", 22, forma.SchemaAttributeCache{}, &paramIndex)
+			require.Error(t, err)
+			require.ErrorIs(t, err, forma.ErrInvalidInput)
+		})
+	}
+}
--- a/internal/sqlgen/duck_clause_test.go
+++ b/internal/sqlgen/duck_clause_test.go
@@ -139,13 +139,18 @@ func TestBuildDuckClause_UnregisteredAttr_RefusesFoldOntoVisibleColumn(t *testin
 		{"row.id", "row_id"},
 		{"[row_id]", "row_id"},
 		{"Created.At", "Created_At"},
+		{"schema.id", "schema_id"},
+		{"attributes.json", "attributes_json"},
+		{"ltbase.row_id", "ltbase_row_id"},
 		{"rn", "rn"},
 		{"RN", "RN"},
+		{"[rn]", "rn"},
 		{"source_tier.priority", "source_tier_priority"},
 		{"source_tier_priority", "source_tier_priority"},
 		{"[]", "attr"},
 		{"`", "attr"},
 		{"[attr]", "attr"},
+		{"", "attr"},
 	}
 	for _, tc := range refused {
 		t.Run("refused/"+tc.attr, func(t *testing.T) {
@@ -172,7 +177,10 @@ func TestBuildDuckClause_UnregisteredAttr_RefusesFoldOntoVisibleColumn(t *testin
 // The production federated route builds all three clauses through
 // ToDualClauses, where an unregistered attribute is refused by the PG EAV
 // payload before the DuckDB clause is rendered (#512 investigation). This
-// pins that the folded spellings never reach DuckDB on that route either.
+// pins that the folded spellings never reach DuckDB on that route either,
+// and pins the ORDER: the error must be the PG EAV refusal, not the DuckDB
+// fold guard, so an emitter reorder that let the DuckDB guard fire first
+// would fail here rather than silently change the documented behaviour.
 func TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission(t *testing.T) {
 	for _, attr := range []string{"created.at", "rn", "[]", "ghost"} {
 		t.Run(attr, func(t *testing.T) {
@@ -181,6 +189,9 @@ func TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission(t *testing.T) 
 			_, err := ToDualClauses(cond, "eav_table", 22, forma.SchemaAttributeCache{}, &paramIndex)
 			require.Error(t, err)
 			require.ErrorIs(t, err, forma.ErrInvalidInput)
+			require.Contains(t, err.Error(), "attribute not found in cache",
+				"PG EAV payload must refuse before the DuckDB fold guard runs")
+			require.NotContains(t, err.Error(), "folds")
 		})
 	}
 }
--- a/internal/sqlgen/parquet_attr_column.go
+++ b/internal/sqlgen/parquet_attr_column.go
@@ -61,6 +61,58 @@ var reservedParquetColumns = map[string]struct{}{
 	"ltbase_deleted_at":    {},
 }
 
+// parquetAttrPlaceholder is the column ParquetAttrColumn substitutes when the
+// fold empties a name ("", "[]", a lone backtick); a name like "[attr]"
+// strips down onto it directly.
+const parquetAttrPlaceholder = "attr"
+
+// federatedDedupColumns are the visible-CTE columns that are dedup machinery,
+// not data: a filter on either binds successfully and compares against the
+// dedup rank, so it fails silently-wrong rather than loudly. Refused under
+// any spelling, identity fold included (the keyset counterpart is
+// federated.keysetRejectedColumns).
+var federatedDedupColumns = map[string]struct{}{
+	"rn":                   {},
+	"source_tier_priority": {},
+}
+
+// ValidateUnregisteredParquetAttrColumn guards a filter attribute the schema
+// cache does not know (#512). ValidateParquetAttrColumns cannot cover it: it
+// checks registered attributes at registration, and an unregistered filter
+// name is an arbitrary caller-supplied string that never went through it.
+// The fold is lossy, so "created.at" lands on created_at and the DuckDB
+// clause would silently filter on the creation timestamp instead of failing
+// at the binder — the silent-wrong-answer family of #354.
+//
+// The rule mirrors federated.validateKeysetCursor (#509), judged on the
+// FOLDED name because that is the identifier the generator emits, with the
+// lookups lower-cased because DuckDB resolves unquoted identifiers
+// case-insensitively. A non-identity fold onto the "attr" placeholder or onto
+// any reserved parquet column is refused; the dedup machinery is refused
+// under any spelling; an identity-up-to-case fold ("created_at",
+// "Created_At") is admitted and left to fail at the binder or, on the
+// federated route, at the PG EAV payload. Caller fault, so the error is a
+// forma.InvalidInputf carrier that keeps the caller's spelling.
+func ValidateUnregisteredParquetAttrColumn(attr, folded string) error {
+	if strings.EqualFold(folded, parquetAttrPlaceholder) && !strings.EqualFold(attr, parquetAttrPlaceholder) {
+		return forma.InvalidInputf(
+			"filter attribute %q is not registered and folds onto the placeholder column %q, which would silently filter on a real attribute of that name: filter on a registered schema attribute",
+			attr, folded)
+	}
+	key := strings.ToLower(folded)
+	if _, ok := federatedDedupColumns[key]; ok {
+		return forma.InvalidInputf(
+			"filter attribute %q folds to %q, which is federated dedup machinery, not a queryable column: filter on a registered schema attribute",
+			attr, folded)
+	}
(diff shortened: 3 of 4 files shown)


iceboundrock:
# Review: #530 — refuse an unregistered filter attribute that folds onto a visible column (#512)

## Summary

Correct fix, correctly scoped, no blocking issues. The DuckDB payload is closed on its own with the exact rule the #512 ruling prescribed, mirroring the #509 cursor contract. Production federated behavior is unchanged by construction (the PG EAV payload already 400s first), so this is defense-in-depth against emitter-order changes plus a real fix for the standalone `BuildDuckClause` route. LGTM; comments below are non-blocking suggestions and follow-ups. Not merging per repo rule — awaiting your explicit go-ahead.

## Spec

Checked against the #512 ruling (`issues/512#issuecomment-5555953273`), the #509 cursor contract, and the #354 silent-wrong-answer family. All rule items map 1:1:

| Ruling | Implementation | Test |
| --- | --- | --- |
| Non-identity fold onto `attr` placeholder refused (`[]`, `` ` ``, `[attr]`) | `parquet_attr_column.go:97` (`EqualFold` both sides) | `duck_clause_test.go` refused table, 3 rows |
| Non-identity fold onto any `reservedParquetColumns` entry refused (`created.at`, `ver.ts`, `deleted.ts`, `row.id`, `[row_id]`, `Created.At`) | `parquet_attr_column.go:108` (folded, lower-cased, `!EqualFold(attr, folded)`) | refused table, 6 rows incl. case variant |
| `rn` / `source_tier_priority` refused under any spelling, identity included | `parquet_attr_column.go:102` (no identity exemption) | `rn`, `RN`, `source_tier.priority`, `source_tier_priority` |
| Identity-up-to-case admitted (`created_at`, `Created_At`, `attr`) | exemptions on both branches | admitted table, 5 rows |
| Applied only when `hasMeta == false` | `predicate_normalizer.go:401` gate | empty-cache tests |
| Refusals are `forma.InvalidInputf` carriers naming attribute + folded column | all three branches | `ErrorIs(ErrInvalidInput)` + `Contains(attr)` + `Contains(folded)` |
| Registered names never reach it | `hasMeta` gate; `ValidateParquetAttrColumns` at registration | — (see F5 caveat) |
| Design doc states the shared rule | `design.md` §4 addition | prose accurate |

Error class is right: `docs/error-handling.md` already lists "filtering on an unknown attribute" in `predicate_normalizer.go` as a carrier site, and the PG EAV sibling (`attribute not found in cache`) is the same class. The cursor validator's plain-error choice (`keyset.go:133`) is a different surface with its own documented rationale; matching the filter sibling, not the cursor error type, is the correct call here.

Production-ordering claim verified in code: `ToDualClauses` (`dualpath_sql_generator.go:31`) emits PgMain (skips unknown, no error) → PgEav (refuses unknown) → Duck, so on the federated route the PG EAV error wins before the new Duck guard is reached. Production caller is `internal/federated/duckdb_query_build.go:45`; `BuildDuckClause` has zero non-test callers (definition only in non-test code), so the standalone-route fix has no production behavior delta either — exactly the option-2 bargain the ruling accepted.

## Standards (`coding-standard.md`, `AGENTS.md`)

- File limits: `parquet_attr_column.go` 147 lines, `predicate_normalizer.go` 431 lines — both under 500. Pass.
- Function limit: `ValidateUnregisteredParquetAttrColumn` ~20 lines — under 100. Pass.
- Error context: every refusal names the attribute, the folded column, and the remedy (`filter on a registered schema attribute`) — not a bare `return err`. `InvalidInputf` carriers, so `TestNoBareSentinelWraps` holds (local `make lint` clean, CI Lint + Test pass). Pass.
- Structure: early-return chain, max nesting depth 2. Pass.
- Godoc: exported function has a full contract comment (why registration can't cover it, fold/case rules, caller-fault classification). Pass.
- Naming: `parquetAttrPlaceholder`, `federatedDedupColumns`, `ValidateUnregisteredParquetAttrColumn` are clear; the comment cross-links `federated.keysetRejectedColumns`. Pass.
- Duplication gate (>15 similar lines): the only dup is two 2-entry maps + one `"attr"` constant across the `sqlgen`/`federated` boundary (see F4) — below threshold, observation only.

## Findings (all non-blocking)

**F1 (nit): third error message suggests an action that doesn't exist on the filter path.** `parquet_attr_column.go:108-112` ends with `"or name the system column %s directly"`. Filters have no system-column surface — even the admitted identity `created_at` fails at the PG EAV payload (`attribute not found in cache`) on the federated route, and registration rejects `created_at` outright. Cursors can name system columns; filters cannot. Suggest trimming to the shared `"filter on a registered schema attribute"` close used by the other two branches.

**F2 (test strength): the `ToDualClauses` pin doesn't prove ordering.** `TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission` asserts only `ErrorIs(ErrInvalidInput)`, so it passes with or without this fix — the PG EAV payload already refuses all four attrs. To pin the ruling's load-bearing premise (PG EAV fires *first*), assert the message contains `"attribute not found in cache"`. Otherwise a future emitter reorder that lets the Duck guard fire first would stay green while the documented behavior changed.

**F3 (coverage gap): code guards all 18 reserved columns, tests pin 4 families.** No pin for e.g. `schema.id`→`schema_id`, `attributes.json`→`attributes_json`, `ltbase.row_id`→`ltbase_row_id` (dot-fold onto `ltbase_*`), nor for the exact #509-bypass shapes on this path: `[rn]` (bracket-fold onto dedup) and `""` (empty→`attr` placeholder). Suggest 3–4 added rows; the implementation already handles them, this is just locking the `"any reserved"` claim.

**F4 (maintainability): two cross-package duplications with no sync guard.** `federatedDedupColumns` ≡ `federated.keysetRejectedColumns` (`keyset.go:76`), and `parquetAttrPlaceholder` ≡ `parquetAttrFallbackColumn` (`keyset.go:97`). The split is forced by layering (`federated` imports `sqlgen`, not vice versa) and both sites comment the counterpart, so this is not a violation — but if the keyset set ever grows, the filter set can drift silently. Cheapest fix: export the `sqlgen` set/constant and have `federated` alias it, or add a small cross-package test pinning equality. Follow-up, not this PR.

**F5 (pre-existing, out of scope): registration-time check is case-sensitive, the new guard is not.** `ValidateParquetAttrColumns` looks up `reservedParquetColumns[col]` verbatim, while the new guard and `validateKeysetCursor` lower-case first. A registered `Created_At` therefore bypasses registration yet binds to `created_at` in DuckDB (unquoted identifiers resolve case-insensitively). The new code is the correct one; flagging so the old site gets a follow-up issue — not introduced here, not blocking.

**F6 (docs nit): PR body overstates placement slightly.** "in the leaf normalizer so every dialect inherits it" — the check actually lives in `normalizeDuckPayload` (Duck only); PG EAV already refuses and PG Main skips. `design.md`'s own wording ("The DuckDB payload is nevertheless guarded on its own") is the accurate one. One-sentence touch-up if you revise the body.

## Observations

- Case handling is done right throughout: lookups lower-cased, emission and messages keep caller spelling, exemptions `EqualFold` — the exact shift-key bypass class from the #509 round-2 defect. `RN` and `Created.At` rows prove it.
- `hasMeta` gating is the minimal correct scope: registered names keep their existing path (including the lenient `LIKE`/inference behavior), unregistered names keep leniency except where the fold retargets. No behavior change for any currently-succeeding production query.
- Acknowledged gap (no `safeSQLIdentifier` on the filter path, other punctuation interpolated as-is on the standalone route) is sound to defer: unreachable in production via the PG EAV refusal above, and the PR records it on the issue rather than silently. Recommend a tracking issue if the standalone route is ever exposed beyond tests.
- Good test style: table-driven, asserts both the sentinel (`ErrorIs`) and the human diagnosis (`Contains` attr + folded), admitted-side asserts the clause actually references the name.

## Verification

- New tests: `TestBuildDuckClause_UnregisteredAttr_RefusesFoldOntoVisibleColumn` (13 refused + 5 admitted) and `TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission` — pass on the PR worktree.
- Regression scope: `go test ./internal/sqlgen ./internal/federated` — both ok.
- `make fmt-check` — clean. `make lint` (root + infra modules, pinned v1.64.8) — clean.
- `go test ./...` — all unit packages ok; sole failure is `internal/e2e_harness`, which requires a container runtime (testcontainers, no daemon in this environment) and is unrelated. Did not run the container-backed `make test` / production harness here for the same reason.
- CI on this PR at review time: Lint pass, Test pass, Benchmark Smoke pass, k6 Smoke pass, E2E Smoke pending.
- Fold mapping spot-checked against `ParquetAttrColumn` semantics (`created.at`→`created_at`, `[]`/`` ` ``→`attr`, `[row_id]`→`row_id`, `source_tier.priority`→`source_tier_priority`) — matches the issue's defect table.
- File sizes confirmed: 147 / 431 lines.

## Assessment

Risk is low and the change is fail-closed: the only queries whose outcome changes are ones that previously got a well-formed page filtered on a column they never named — those now get a 400 naming the attribute, the folded column, and the fix. Everything else (registered filters, identity spellings, production federated ordering) is behavior-preserving. No migration, no config change, no follow-up required before merge; F1–F4 are polish and F5 is a separate pre-existing item.

iceboundrock:
## Response to review

Addressed in `ed2e051`; follow-ups filed for the two items scoped out of this PR.

**F1 (dead remedy in the reserved-column message)**: fixed. The third refusal now closes with the shared `filter on a registered schema attribute` and no longer suggests naming the system column, which filters cannot do.

**F2 (ordering pin)**: fixed. `TestToDualClauses_UnregisteredAttr_RefusedBeforeDuckEmission` now asserts the error text contains `attribute not found in cache` and does not contain `folds`, so it fails if an emitter reorder ever lets the DuckDB guard fire before the PG EAV refusal. Confirmed the assertion is load-bearing: the DuckDB guard's messages all say "folds", the EAV one does not.

**F3 (coverage of the "any reserved column" claim)**: fixed. Added refused rows for `schema.id` → `schema_id`, `attributes.json` → `attributes_json`, `ltbase.row_id` → `ltbase_row_id`, `[rn]` → `rn` (bracket-fold onto dedup) and `""` → `attr` (empty name onto the placeholder). 18 refused rows now, 5 admitted.

**F4 (cross-package duplication)**: not changed here, as suggested. Opened #531 proposing to export the `sqlgen` set and constant and have `federated` alias them, with a set-equality test as the fallback.

**F5 (case-sensitive registration check)**: pre-existing and not touched here. Opened #532. Verified the claim: `ValidateParquetAttrColumns` looks up `reservedParquetColumns[col]` verbatim while both newer guards lower-case first, so a registered `Created_At` passes registration and binds to `created_at` in DuckDB. Since the #512 rule relies on "registered names never reach the check", #532 is what makes that reliance sound, so it should land soon after this.

**F6 (PR body placement wording)**: fixed. The body now says the check lives in `normalizeDuckPayload`, the one path that needed guarding on its own, and links #531 / #532 under a review-follow-ups section.

**Observation on `safeSQLIdentifier` not applying to the filter path**: already recorded on #512 as out of scope; no tracking issue opened since `BuildDuckClause` has no non-test callers. Will open one if the standalone route is ever exposed.

## Verification

- `go test ./internal/sqlgen ./internal/federated`: ok. All 5 new refused subtests observed running and passing.
- `make fmt-check`: clean.
- `make lint` (root and infra, v1.64.8): clean.
- `parquet_attr_column.go` 147 lines, `duck_clause_test.go` 197, `predicate_normalizer.go` 431.

iceboundrock:
## Review: #530

### Findings
No blocking or non-blocking findings in the latest commit `ed2e051`.

The earlier review items were addressed:
- The reserved-column error no longer suggests naming a system column through the filter surface.
- The `ToDualClauses` test now pins that the PostgreSQL EAV error wins before DuckDB emission.
- Coverage now includes additional reserved-column, dedup-column, and placeholder fold cases.
- The PR wording now accurately identifies `normalizeDuckPayload` as the standalone guard.

### Spec
Checked against the #512 ruling, the #509 keyset fold contract, and the #354 silent-wrong-answer class.

- Non-identity folds onto `attr` are rejected.
- Non-identity folds onto reserved Parquet/visible-CTE columns are rejected.
- `rn` and `source_tier_priority` are rejected under every spelling.
- Identity-up-to-case folds remain admitted as specified.
- The guard applies only to unregistered attributes; registered attributes remain on the existing path.
- Errors are `forma.InvalidInputf` carriers and name the caller attribute, folded column, and remediation.
- The design document records the shared filter/cursor fold rule.

### Standards
Passes the applicable `AGENTS.md` and `coding-standard.md` checks:

- Changed Go files remain below the 500-line limit; changed function size remains below 100 lines.
- The helper uses early returns and has no excessive nesting.
- The exported helper has a complete Go doc comment.
- Error handling preserves the caller-facing invalid-input carrier and does not introduce a bare sentinel wrap.
- No unrelated behavior or compatibility shim was added.

### Assessment
Low-risk, fail-closed change. It prevents the standalone DuckDB predicate path from silently retargeting an unregistered filter such as `created.at` onto `created_at`, while preserving the established registered-attribute and production `ToDualClauses` behavior. The production route already rejects unknown attributes through the PG EAV payload; this change correctly hardens the DuckDB payload independently.

### Observations
- `federatedDedupColumns` and the keyset package’s equivalent set, plus the two placeholder constants, remain duplicated across the package boundary. This is a maintainability follow-up tracked by #531, not a merge blocker.
- Case-sensitive registration validation for a registered spelling such as `Created_At` is pre-existing and tracked by #532; it is outside this PR’s introduced behavior.
- The standalone path still does not apply the keyset `safeSQLIdentifier` rule to arbitrary punctuation, as documented out of scope; it has no production callers today.

### Verification
- PR CI: Test passed.
- PR CI: Lint passed.
- PR CI: Benchmark Smoke passed.
- PR CI: k6 Smoke passed.
- PR CI: E2E Smoke was still running when this review was posted.
- Local `make fmt-check`: passed.
- Local `go test ./internal/sqlgen ./internal/federated` at the PR head: passed.
- A first parallel local test/lint attempt was blocked by the environment’s `/tmp` disk quota, not by a code or test failure; the targeted test was rerun serially and passed. CI provides the lint and broader test verification.
- `git diff --check`: passed.

### Summary
LGTM. No changes are required from this review. The PR is suitable for merge after the remaining required E2E Smoke check completes successfully.

PR #534, opened 2026-09-06
fix(compaction): report noop for a confirmed-missing manifest instead of failing the run (#524)

### LINKED ISSUE #524: compactor: confirmed-missing manifest fails the run instead of the documented Noop skip
Follow-up from the #523 review (#464). Not a regression of that PR; filed so the new typed contract gets applied to the last direct `manifest.Load` consumer.

## Observation

`Compactor.compactSchema` (`internal/compaction/compactor.go`) has two branches after `Provider.LoadManifest`:

```go
m, etag, err := c.Provider.LoadManifest(ctx, schemaID)
if err != nil {
    return CompactionResult{}, fmt.Errorf("load manifest: %w", err)
}
if m == nil {
    c.Logger.Debug("no manifest for schema, skipping", ...)
    return CompactionResult{Outcome: Noop, SchemaID: schemaID}, nil
}
```

The `m == nil` "skip" branch is unreachable with the production providers. `ManifestProvider.LoadManifest` delegates to `manifest.Load`, and both `S3Store.Load` and `FSStore.Load` return a wrapped `manifest.ErrObjectNotFound` for a missing object; `manifest.Load` never returns a nil manifest with a nil error. Only the unit-test `mockProvider` produces `(nil, "", nil)`.

So a `compactor` run against a schema that has never been flushed (no manifest yet) fails with `load manifest: get manifest object <bucket>/<key>: manifest object not found: ...` rather than reporting `Noop`. The code and its debug log describe one behaviour; the real stores produce another.

## Decision needed

Either outcome is defensible, but the code should commit to one:

1. **Noop on confirmed absence.** Gate on `manifest.IsNotFound(err)` and return `Noop`, matching the existing debug message. Every other load error (NoSuchBucket, permission, transport) keeps failing the run. This is now expressible safely because #523 made `IsNotFound` a typed check.
2. **Fail loudly.** Keep the current behaviour, delete the dead `m == nil` branch and its test-only assumption in `mockProvider`, and document that compaction requires an initialised manifest.

Whichever is chosen, add a regression test that drives a `ManifestProvider` over a real `FSStore` (or `S3Store` with a stub client) so the test exercises the store contract rather than the mock's shape.

## Scope note

Out of scope for #523, which is limited to classification at the store boundary and the harness/query-source consumers named in its review.

iceboundrock:
Closes #524. Follow-up to #523 (#464): applies the typed `manifest.IsNotFound` contract to the last direct `manifest.Load` consumer.

## Problem

`Compactor.compactSchema` had a `m == nil` "skip" branch that no production provider could reach. `ManifestProvider.LoadManifest` delegates to `manifest.Load`, and both `S3Store.Load` and `FSStore.Load` return a wrapped `manifest.ErrObjectNotFound` for a missing object. Only the unit-test `mockProvider` produced `(nil, "", nil)`. A `compactor` run against a schema that had never been flushed therefore failed with `load manifest: ... manifest object not found` while the code and its debug log promised a Noop.

## Decision

Option 1 from the issue, **Noop on confirmed absence** (ruling and plan recorded on #524). A never-flushed schema is a legitimate state the read side and CDC already treat as normal; #523 made `IsNotFound` a typed `errors.Is` check, so gating on it cannot re-introduce the #464 misclassification.

## Changes

- `internal/compaction/compactor.go`: gate on `manifest.IsNotFound(err)` and return `Noop` (logged at Info so a one-shot CLI run shows why); every other load error still fails the run with the `load manifest: %w` wrap. The unreachable `m == nil` branch is deleted. `FileProvider.LoadManifest` now documents the contract: a confirmed-absent manifest must surface as a wrapped `manifest.ErrObjectNotFound`, never `(nil, "", nil)`.
- `internal/compaction/provider.go`: doc comment updated; behaviour unchanged.
- `internal/compaction/compactor_test.go`: `mockProvider` mirrors the real stores for a nil manifest. `TestCompactor_RunOnce_NoManifest` (mock-shaped) is replaced by two tests that drive `ManifestProvider` over a real `manifest.FSStore`: an empty `fstest.MapFS` yields `Noop`; an `fs.FS` failing with `fs.ErrPermission` still fails the run, with `errors.Is(err, fs.ErrPermission)` true and `manifest.IsNotFound(err)` false.
- `docs/federated-query/design.md`: one sentence in the migration note records the compactor's contract alongside the read side's.

## Verification

- New test fails on `main` with the exact error from the issue and passes with the fix.
- `make fmt-check`, `make lint` clean.
- Full unit suite via `scripts/test_with_container_runtime.sh`: all packages ok.


iceboundrock:
## Review Summary

### Findings
No blocking, major, or minor findings identified. I found no correctness regression in the changed paths, and no requested changes are necessary.

### Spec Review
- The PR implements the #524 decision to treat a confirmed missing manifest as a `Noop`.
- The classification is based on the typed `manifest.IsNotFound` contract rather than error text.
- Non-absence failures such as permission errors, missing buckets, and transport failures still fail compaction and retain error context.
- The tests exercise the real `FSStore` path for both confirmed absence and an unclassified load failure.

### Standards Review
- Error handling preserves `%w` context and uses the repository's typed error semantics.
- The `FileProvider.LoadManifest` contract is documented and all current implementations/tests are consistent with it.
- Changed Go files remain within the repository's 500-line file limit; the largest changed test file is 495 lines.
- Formatting and whitespace checks are clean.

### Assessment
The implementation is focused, minimal, and correctly closes the missing-manifest failure described by #524. The change is suitable for merge.

### Observations
- Removing the old `(nil, nil)` branch is safe for the current production provider because both S3 and filesystem stores return wrapped `manifest.ErrObjectNotFound` for confirmed absence.
- The mock provider was updated to mirror that production contract, avoiding a test-only behavior that real providers cannot produce.
- No existing PR review comments or inline review threads were present before this review.

### Verification
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction ./internal/manifest`: passed.
- `make fmt-check`: passed.
- `git diff --check main...HEAD`: passed.
- PR CI checks: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke all passed.
- Local `make test` reached the test suite but failed only in `internal/e2e_harness` because this environment has no usable container runtime and `$XDG_RUNTIME_DIR` is unavailable; this is environmental and unrelated to the PR changes.

### Summary
**Approve.** No findings.

PR #535, opened 2026-09-06
fix(manifest): refuse zero-stamped manifests that list entries for every schema (#522)

### LINKED ISSUE #522: Federated read path and manifest-reconcile accept zero-stamped manifests for any schema
Follow-up from the fourth review round of #518 (#371), sibling of #520.

**The lenience.** `manifest.LoadOrCreateForSchema` (`internal/manifest/manifest.go`) rejects a loaded manifest whose `schema_id` names another schema, but lets `schema_id: 0` through for any schema. The federated read path (`QuerySource.load`, `internal/manifest/query_source.go`) inherits that, and `manifest-reconcile`'s `ResolverManifestStore.Load` has the same carve-out from #481 (`docs/manifest-reconcile.md`, "`schema_id` 为 0（旧版 manifest 未落此字段）保持兼容、照常加载"). Both treat a zero stamp as "written before the field existed".

**Why it protects nothing.** `SchemaID` has been on `manifest.Manifest` since `internal/manifest/manifest.go` was first committed (968afbc), and `LoadOrCreate` has always stamped a created manifest. Every writer (cdc-flush, cdc-init, compactor, manifest-reconcile) loads then saves, preserving the stamp. So no Forma writer has ever produced a `schema_id: 0` object; the only source is a hand-edited or hand-made manifest.

**What it costs.** The two-probe check in `forma.ValidateManifestPathTemplate` / `ValidateManifestRead` (#300) renders schema IDs 1 and 2 only. A template such as `{{if lt .SchemaID 3}}manifest/{{.SchemaID}}.json{{else}}manifest/shared.json{{end}}` passes it and collides every schema from 3 upward onto one object. A zero-stamped manifest at that shared path is then served under whichever schema resolves it: the parquet scan does not filter by schema and the projection stamps every row as the requested schema, so the read serves another schema's rows under this identity. A stamped manifest at the same path is caught by the mismatch check; the zero stamp is the blind spot. `manifest-reconcile` would likewise classify and (with `--repair`) splice under the wrong schema.

#518 closed this for the writer with a destructive step: `cdc-init` refuses a non-empty zero-stamped manifest with `cdc.ErrManifestUnstamped` and restamps an empty one (`internal/cdc/init_publish.go`, `loadInitManifest`). The read path and reconcile were left as they were because that behaviour predates the PR.

**Proposed fix.**

1. Drop the zero-stamp carve-out from `LoadOrCreateForSchema`: a loaded manifest with `schema_id: 0` that lists any entry is rejected with `forma.ManifestSchemaMismatchError` (or a dedicated `ErrManifestUnstamped` sentinel promoted to the root package, so cdc-init's message and the read path's classification agree). An empty zero-stamped manifest can keep loading, stamped in memory, as cdc-init does.
2. `ResolverManifestStore.Load` in `internal/reconcile` follows the shared helper (this is the same consolidation #520 proposes), and the "`schema_id` 为 0 保持兼容" paragraph in `docs/manifest-reconcile.md` is replaced with the new rule.
3. `docs/federated-query/design.md` gets a line in the manifest-read section naming the rule.
4. Tests: the `TestLoadOrCreateForSchema` "pre-stamp manifest is accepted for any schema" case flips; a `QuerySource` test with a zero-stamped manifest at a shared path under a template that collides only outside the probe IDs fails the read instead of serving rows; a reconcile test does the same for `--repair`.

Because the carve-out has no writer behind it, removing it is not a migration for any deployment that only ever wrote manifests with Forma tools. If a hand-made zero-stamped manifest exists in some environment, the remedy is a one-field edit after confirming ownership, which is what cdc-init's error already instructs.

iceboundrock:
Closes #522.

## What

The zero-stamp carve-out in `manifest.LoadOrCreateForSchema` is gone. The schema-identity rule now lives in one exported helper, `manifest.VerifySchemaStamp`, and every consumer goes through it: the federated read path, `cdc-flush`, `cdc-init`, `manifest-reconcile` (via `LoadOrCreateForSchema`) and the compactor's manifest provider (directly, since it loads without create semantics).

| stamp | entries | outcome |
| --- | --- | --- |
| requested schema | any | load |
| another schema | any | `forma.ManifestSchemaMismatchError` (unchanged) |
| `0` | ≥ 1 | **new:** `forma.ManifestUnstampedError` |
| `0` | 0 | load, stamped for the requested schema in memory (what cdc-init already did) |

## Error surface

`forma.ManifestUnstampedError{RequestedSchemaID, Path, Entries}` unwraps to **both** `forma.ErrManifestUnstamped` and `forma.ErrManifestSchemaMismatch`. An unstamped manifest is a manifest whose identity does not match the requested schema, so every existing classification applies unchanged: `httpapi` redacts it under the `manifest_schema_mismatch` class as a 500, the federated engine keeps it non-degradable, and path resolution does not relabel it `ErrFederatedReadFailed`. `errors.Is(err, forma.ErrManifestUnstamped)` still tells the remedy apart (set `schema_id` on the object rather than fix the template). `httpapi.errorSchemaID` gains a branch so the redacted body still carries the requested schema id.

`cdc.ErrManifestUnstamped` and cdc-init's private copy of the rule are deleted; its tests now match the root sentinel and the message keeps the `N entries listed under schema_id 0` phrasing they assert on.

## Scope note

The compaction provider is included although the issue names only the read path and reconcile: it carried the same carve-out and swaps a schema's tiers on the strength of the stamp, and folding it in means there is exactly one enforcement site.

## Tests

- `TestLoadOrCreateForSchema`: the "pre-stamp manifest is accepted for any schema" case flips; an empty zero-stamped manifest loads stamped without a save.
- `TestQuerySource_UnstampedManifestAtSharedPathFailsRead`: the issue's scenario — a template that collides only outside the two probe IDs (asserted to pass `ValidateManifestPathTemplate`), a zero-stamped manifest at the shared path — fails the read with both sentinels and returns no paths. `TestQuerySource_EmptyUnstampedManifestFallsBack` covers the empty case.
- `TestReconcile_RepairRefusesUnstampedSharedManifest`: same shape through a real `ResolverManifestStore` with `--repair`; the schema reports the error, nothing is spliced or saved, the shared object is byte-identical.
- `TestManifestProvider_LoadManifest_RejectsUnstampedWithEntries` / `_StampsEmptyUnstamped` for the compactor.
- `httpapi`: class, schema-id and 5xx tests cover the new carrier. `errors_test.go` pins the dual unwrap.

## Docs

`docs/manifest-reconcile.md` (the "`schema_id` 为 0 保持兼容" paragraph and the cdc-init bullet), `docs/federated-query/design.md` §4.3.1 level 2, and `docs/error-handling.md` under `ErrManifestSchemaMismatch`.

## Migration

None for deployments that only ever wrote manifests with Forma tools. A hand-made zero-stamped manifest with entries now fails reads, compaction, flush appends and reconcile the same way cdc-init already did; the remedy is the one-field edit the error names.

## Verification

`make fmt-check`, `make lint` and `./scripts/test_with_container_runtime.sh` (full `make test`, 34 packages) all pass locally.


iceboundrock:
## Review of #535 — refuse zero-stamped manifests with entries (closes #522)

### Summary
The PR removes the `schema_id: 0` carve-out from `manifest.LoadOrCreateForSchema`, centralises the rule in a new exported `manifest.VerifySchemaStamp`, and routes every consumer through it (federated read path, `cdc-flush` via `AppendFile(s)`/`ReplaceTierFiles`, `cdc-init` via `loadInitManifest`, `manifest-reconcile` via `LoadOrCreateForSchema`, compactor via a direct call). A zero-stamped manifest with entries is now refused for every schema with the new `forma.ManifestUnstampedError` (dual-unwrapping to both `ErrManifestUnstamped` and `ErrManifestSchemaMismatch`); an empty zero-stamped manifest loads stamped-in-memory. `cdc.ErrManifestUnstamped` and the private cdc-init rule are deleted. This matches the issue's proposed fix, including the scope extension to the compactor, which is justified in the PR body. No blocking issues found; three non-blocking observations below.

### Spec compliance (against #522 items 1–4)
1. **Drop the carve-out, reject non-empty zero stamps** — done. `internal/manifest/manifest.go:293` (`VerifySchemaStamp`): foreign stamp → `ManifestSchemaMismatchError`; zero + entries → `ManifestUnstampedError`; zero + empty → stamp in memory. `LoadOrCreateForSchema` (`manifest.go:263`) delegates, so the read path (`query_source.go:57`), cdc-flush writers, cdc-init and reconcile all inherit it. Compactor (`internal/compaction/provider.go:72`) calls `VerifySchemaStamp` directly, which is the correct adaptation since it loads without create semantics (missing must stay `ErrObjectNotFound` → `noop` per #524).
2. **`ResolverManifestStore.Load` follows the shared helper** — done (`internal/reconcile/manifests.go:40`). Unstamped now fails at load before any classification/diff, and `TestReconcile_RepairRefusesUnstampedSharedManifest` proves `--repair` splices/saves nothing and leaves the shared object byte-identical.
3. **Design doc line** — done (`docs/federated-query/design.md` level 2, plus `docs/error-handling.md` zero-stamp section and `docs/manifest-reconcile.md` rewrite). The docs correctly retire the old "schema_id 0 keeps compatibility / legacy format" claim and state the archaeology (stamped since first commit, zero = hand-made).
4. **Tests** — done and well-shaped: `TestLoadOrCreateForSchema` flips the pre-stamp case and pins dual-`Is`, no-save-on-load (`manifest_test.go:359`), `TestQuerySource_UnstampedManifestAtSharedPathFailsRead` uses a template that provably passes `ValidateManifestPathTemplate` yet collides at IDs 3/4 with `require.Nil(paths)`, `TestQuerySource_EmptyUnstampedManifestFallsBack` covers the empty case, plus compactor, reconcile, cdc-init (`forma.ErrManifestUnstamped`), and httpapi class/schema-id/5xx coverage. `errors_test.go:41` pins the dual unwrap.

### Findings (blocking: none)
No correctness, classification, or migration blockers:
- **Dual-unwrap is sound and load-bearing classifications hold.** `errors.go:123` `Unwrap() []error` (Go ≥1.20; go.mod is 1.26) makes `errors.Is(err, ErrManifestSchemaMismatch)` true for the new carrier, so: `internal/federated/parquet_source.go:84` keeps unstamped on the non-`ErrFederatedReadFailed` branch (no relabel, no degraded fallback via `degradableFederatedError` in `engine_degrade.go:82`, which keys on the mismatch sentinel); `internal/httpapi/error_response.go:49` maps it to `manifest_schema_mismatch` (redacted 500). The `errorSchemaID` branch (`error_response.go:99`) correctly returns only the requested ID — there is no foreign ID to leak. Verified by reading all three call sites, not just the PR description.
- **Pass-through in `query_source.go:59` is correct.** Switching from `errors.As(mismatch)` to `errors.Is(ErrManifestSchemaMismatch)` is required: the old `As` would not have matched the new carrier type and would have demoted it under a generic `load manifest for schema …` wrap. The new code returns both carriers unwrapped so the engine wrap is the only classification-relevant layer (plus the `Paths`-level `resolve manifest paths: %w`, which preserves `Is`).
- **Reconcile `As`-vs-`Is` split is intentional and safe.** `manifests.go:41` keeps `errors.As` for the foreign-stamp custom wording; unstamped (a different carrier type) falls to the generic `load manifest %s: %w`, which preserves its own remedy message ("confirm … set schema_id"). Routing unstamped into the mismatch wording would have prescribed the wrong remedy (fix the template), so the fall-through is the right behaviour. Both still satisfy `ErrorIs(ErrManifestSchemaMismatch)` for callers.
- **Empty-stamp in-memory mutation is safe.** `VerifySchemaStamp` mutates only a freshly loaded (or newly created) `*Manifest`, keeps the loaded etag, and every test asserts `saves == 0` on the refusing path and etag preservation on the stamping path. No aliasing or persistence-on-load. Nil-receiver risk is nil: all callers pass a non-nil manifest on success, same as the pre-existing dereference.
- **cdc-init consolidation is a pure deletion of duplication.** `init_publish.go:31` now delegates; the removed private sentinel/message is replaced by the root carrier with the same `N entries listed under schema_id 0` phrasing the guard tests assert on. `errors`/`zap` imports remain used (`errors.Join`, logger calls) — no unused imports. Compactor's `forma` import removal is correct (no direct reference left).
- **Migration claim holds.** `cdc-flush` (`AppendFile(s)`/`ReplaceTierFiles`) inherits the refusal with no code change, as claimed. "No migration for Forma-written deployments" follows from the no-writer-ever-stamped-zero premise the docs now record.

### Standards
- **File/function sizes:** all touched sources stay well under limits (largest: `error_response.go` 359 lines, `manifest.go` 313; `VerifySchemaStamp` ~20 lines, `errorSchemaID` ~31). No refactor trigger.
- **Error semantics (`AGENTS.md`, `docs/error-handling.md`):** correct class. New sentinels are plain read-path consistency errors (no `InvalidInputf` carrier), default to 500 with redaction, and are returned as typed carriers rather than bare `%w` sentinel wraps — consistent with `ManifestSchemaMismatchError` and outside `TestNoBareSentinelWraps`'s `ErrInvalidInput/NotFound/Conflict` scope. Message names path, requested schema, entry count, and remedy; carries no S3 keys or credentials, and the httpapi body never emits `Error()` text (fixed prose + structured `schema_id`).
- **Nesting/early-returns:** `VerifySchemaStamp`, `LoadManifest`, `loadInitManifest` are flat with early returns. Exported items have godoc. `gofmt` clean, `go vet` clean on touched packages.

### Observations (non-blocking nits, safe to defer)
1. **Lost stamp-synthesis log in cdc-init.** The old `loadInitManifest` logged `empty manifest carries no schema stamp; stamping it for this schema`; the shared helper stamps silently and no caller logs the transition. Consider a debug/info line at call sites when `before.SchemaID == 0 && len(files) == 0` (or documenting the silence). Observability only.
2. **Doubled prose in wraps.** Generic wraps now read e.g. `load manifest manifest/7.json: manifest manifest/7.json was loaded for schema 7 …` (reconcile `manifests.go:47`, cdc-init `init_publish.go:34`). Pre-existing pattern for the mismatch case, so consistent — just noting the duplication comes along for the new carrier too.
3. **Compactor comment precision.** `provider.go:49` says "nothing is synthesised here" (contrasting with reconcile's LoadOrCreate). Strictly, the empty-zero case now synthesises a stamp in memory. Suggest "no manifest is synthesised for a missing object" to avoid over-reading. One-line comment fix if touched again.

### Verification
- `go test . ./internal/manifest ./internal/compaction ./internal/cdc ./internal/reconcile ./internal/httpapi ./internal/federated -count=1` — all 7 packages **ok**.
- Targeted runs — all **PASS**: `TestManifestUnstampedErrorIsASchemaMismatch`, `TestLoadOrCreateForSchema`, `TestQuerySource_UnstampedManifestAtSharedPathFailsRead` / `EmptyUnstampedManifestFallsBack`, `TestManifestProvider_LoadManifest_RejectsUnstampedWithEntries` / `StampsEmptyUnstamped`, `TestPreflightDeltaTier_RefusesUnstampedManifestWithEntries`, `TestUpdateSchemaManifest_RefusesUnstampedManifestWithEntries`, `TestResolverManifestStore_UnstampedManifestWithEntriesFails` / `EmptyUnstampedManifestLoadsStamped`, `TestReconcile_RepairRefusesUnstampedSharedManifest`, httpapi `TestReadPathErrorsClassifyAs5xx/manifest_unstamped`, `TestErrorClass…`, `TestErrorSchemaID…`.
- `gofmt -l` on all touched Go files: clean. `go vet` on touched packages + root: clean. (Full `make test` / `make lint` / container E2E not re-run here; PR body reports them green and the affected-package subset above passes locally.)

### Summary / verdict
Approve-worthy. The single-enforcement-site design (with the compactor included) is the right scope, the dual-sentinel carrier preserves every mismatch-keyed classification while keeping the remedy distinguishable, and the tests pin exactly the issue's scenario (probe-blind collision + shared zero-stamped object) on both the read and `--repair` paths plus the empty-manifest convergence. No changes requested; the three observations are optional follow-ups.

iceboundrock:
## Response to review

Pushed e7908a9 (`fix(cdc-init): log the in-memory stamp of an empty unstamped manifest again`). Two of the three observations are addressed in it; the third is answered below without a change. No follow-up issue: none of the three reaches outside this PR.

### 1. Lost stamp-synthesis log in cdc-init — fixed

Confirmed as a regression against `main`: the old `loadInitManifest` logged `empty manifest carries no schema stamp; stamping it for this schema` before setting `m.SchemaID`, and routing it through `LoadOrCreateForSchema` dropped the line. `loadInitManifest` now takes the two halves of `LoadOrCreateForSchema` separately (`manifest.LoadOrCreate`, then `manifest.VerifySchemaStamp`) so it can see the pre-stamp state, and emits the same info line with the same `schema_id` / `manifest_path` fields. `VerifySchemaStamp` stays the one enforcement site; the split is only to observe it, the same shape the compactor already uses. Godoc on `loadInitManifest` explains why the two calls are not the wrapper.

Tests: `TestLoadInitManifestLogsEmptyUnstampedStamp` pins the line (exactly once, both fields) through a zap observer; `TestLoadInitManifestStaysSilentForStampedManifest` pins that an already-stamped manifest does not log.

The other stamping consumers (`AppendFile(s)` / `ReplaceTierFiles` for cdc-flush, reconcile `--repair`, the compactor) stay silent, and I am leaving that as is: the `manifest` package has no logger, and for those paths the stamp is a one-shot repair of a hand-made object that the next save persists, after which every load is checked. Before this PR those writers preserved the zero stamp on save (the carve-out returned the manifest untouched), so the silent path is a strict improvement, not a lost signal. cdc-init is the only site that ever logged it, and it does again.

### 2. Doubled prose in wraps — no change

Acknowledged, and it is the pre-existing shape: `load manifest <path>: manifest <path> was loaded for schema N …`. The outer `load manifest %s` / `load manifest %s for schema %d` wrap in reconcile and cdc-init is the operation context for every error that can come out of the load (store failure, JSON decode, missing object), and those do not carry the path themselves, so dropping it from the generic wrap to de-duplicate the two manifest carriers would cost the other errors their location. Special-casing the unstamped carrier the way reconcile special-cases the mismatch carrier would add a second bespoke wording for no new information. Leaving the duplication.

### 3. Compactor comment precision — fixed

`provider.go` now reads "no manifest is synthesised for a missing object" instead of "nothing is synthesised here", so it no longer contradicts the in-memory stamp `VerifySchemaStamp` applies to an empty zero-stamped manifest.

### Verification

`gofmt`, `go vet`, `go test -count=1` on `internal/cdc`, `internal/compaction`, `internal/manifest`, `internal/reconcile`, plus `make fmt-check` and `make lint` all pass locally on e7908a9.

iceboundrock:
## Review

### Findings

No blocking or non-blocking findings identified.

### Specification

- Conforms to #522 required decision table: matching stamps load; foreign non-zero stamps fail with `ManifestSchemaMismatchError`; zero-stamped manifests with entries fail with the new `ManifestUnstampedError`; empty zero-stamped manifests are stamped in memory for the requested schema.
- `ManifestUnstampedError` unwraps to both `ErrManifestUnstamped` and `ErrManifestSchemaMismatch`, preserving the existing HTTP redaction class and non-degradable federated-read behavior while retaining a distinct remediation signal.
- Enforcement is centralized in `manifest.VerifySchemaStamp` and applied to manifest reads, CDC flush/init, reconcile, and compaction. The cdc-init split of load and verification is intentional so the in-memory stamp remains observable in the log.
- Documentation updates correctly describe the behavior change and the absence of a migration for Forma-written manifests.

### Standards

- Error handling preserves typed/sentinel evidence through contextual wrapping; no message-based classification was introduced.
- Exported sentinel and carrier have GoDoc comments. Changed Go files and functions remain within repository size limits, with flat control flow and focused responsibilities.
- `git diff --check` and `make fmt-check` pass.

### Assessment

Approved from code-review perspective. The change closes the zero-stamp collision path without weakening existing mismatch classification or allowing a partial federated answer.

### Observations

- Coverage exercises the critical collision shape outside the template validator two probe IDs, including no-path escape on reads and no mutation during reconcile repair.
- Unit coverage also covers cdc-init, compaction, HTTP classification/schema-id redaction, and dual-sentinel matching.
- Residual coverage note: there is no separate cdc-flush scenario test for a populated zero-stamped manifest, though its `AppendFile` path uses the shared helper directly and the shared behavior is covered.

### Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test . ./internal/manifest`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction ./internal/reconcile`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/httpapi ./internal/federated`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc -run "Test(UpdateManifest|PreflightDeltaTier|UpdateSchemaManifest|LoadInitManifest)" -count=1 -v`
- `make fmt-check`
- GitHub CI: Lint, Test, Benchmark Smoke, and k6 Smoke pass. E2E Smoke Tests are still pending at review time.

### Summary

No changes requested. Merge once the remaining E2E Smoke check succeeds.

iceboundrock:
## Review of #535

### Findings

1. **[P1] Rejecting `schema_id == 0` also rejects manifests with the field omitted, with no migration path** (`internal/manifest/manifest.go:304-309`, `internal/manifest/query_source.go:57-62`)

   `encoding/json` unmarshals both an omitted `schema_id` and an explicit `"schema_id": 0` into the same `Manifest.SchemaID == 0` value. `VerifySchemaStamp` therefore rejects every non-empty manifest that omits the field. This is a compatibility regression against the base behavior: the base `docs/error-handling.md` explicitly documented zero-stamped/pre-stamp manifests as readable, and the base query-source test covered that behavior. The PR changes the documentation but does not provide a migration or a way to distinguish a legacy omitted field from the unsafe hand-edited case.

   The claim that current Forma writers have always stamped the field does not make already-persisted or manually produced manifests safe to reject without a migration decision. Existing deployments can now receive a non-degradable 500 from federated reads, and reconcile/CDC consumers can fail on the same object. Either preserve field-presence information and apply the new rule only to an explicit zero, or provide an explicit, documented migration/validation path and call out the breaking compatibility change.

2. **[P2] A requested schema ID of zero bypasses the new rule** (`internal/manifest/manifest.go:293-295`)

   `VerifySchemaStamp` checks `m.SchemaID == schemaID` before checking whether the manifest is unstamped. When `schemaID == 0`, a non-empty zero-stamped manifest is accepted, contradicting the stated “refused for every schema” rule. This is not only theoretical: `internal/reconcile/db.go:25-50` enumerates registry values without rejecting zero, and the CDC/init enumeration paths likewise convert database IDs without a positive-ID guard; the schema registry DDL has no positive check. Reconcile or CDC can therefore process schema 0 and read/write an unowned manifest.

   Check the zero-stamp branch before the equality branch and reject non-positive requested schema IDs at this boundary (or enforce the invariant at every entry point), with a regression test for `schemaID == 0`.

### Specification

- The positive-schema portion of #522 is implemented: matching stamps load, foreign non-zero stamps return `ManifestSchemaMismatchError`, non-empty zero stamps return `ManifestUnstampedError`, and empty zero-stamped manifests are stamped in memory.
- The enforcement is centralized in `manifest.VerifySchemaStamp` and is applied to the federated read path, CDC flush/init, reconcile, and compaction.
- The dual unwrap to `ErrManifestUnstamped` and `ErrManifestSchemaMismatch` correctly preserves mismatch classification, non-degradable federated behavior, HTTP redaction, and the requested schema ID.
- The specification is not fully satisfied for requested schema ID zero, and the migration claim is incomplete because omitted `schema_id` cannot be distinguished from explicit zero after decoding.

### Standards

- No code-size or function-size violations were found in the touched Go files.
- Exported sentinels, carriers, and `VerifySchemaStamp` have documentation comments.
- Error wrapping preserves typed/sentinel evidence and uses contextual `%w` wrapping.
- Control flow is flat and the changed areas are cohesive.
- No formatting or `go vet` issues were found in the reviewed packages.

### Assessment

**Changes requested as a post-merge follow-up.** The core collision scenario is closed for positive schema IDs, but the persisted-format compatibility regression and the schema-zero bypass leave correctness gaps in the rule being introduced.

### Observations

- The tests cover the important probe-blind template collision, read failure, reconcile repair refusal, compaction, CDC init, HTTP classification, empty manifests, and dual-sentinel matching.
- There is no regression test for a manifest with `files` but omitted `schema_id`, which is the compatibility boundary introduced by this PR.
- There is no test for `VerifySchemaStamp` with `schemaID == 0`.
- There is no direct populated-zero-stamp CDC flush test; the shared helper coverage makes the intended positive-ID behavior clear, but not the boundary behavior above.

### Verification

- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test . ./internal/manifest ./internal/cdc ./internal/compaction ./internal/reconcile ./internal/httpapi ./internal/federated -count=1`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet . ./internal/manifest ./internal/cdc ./internal/compaction ./internal/reconcile ./internal/httpapi ./internal/federated`
- Passed: `make fmt-check`
- Passed: `git diff --check 4779bb8d2481908c77d09818c44ba8efedd057cc e1fe516c887f79ac4e5f50de7526bacccf59fc73`
- Partially passed: `make test` completed the unit packages but failed at `internal/e2e_harness/TestE2EHarnessMinimal` because this environment has no usable container runtime / `$XDG_RUNTIME_DIR`; the failure is environmental.
- GitHub PR checks were green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Summary

The PR correctly consolidates schema-stamp enforcement and closes the intended positive-ID cross-schema contamination path. Follow-up is still required for the omitted-field compatibility/migration decision and for handling `schemaID == 0` before treating this rule as complete.

iceboundrock:
## Response to the post-merge review

The PR was already merged when this review landed, so the two findings are handled as follow-ups. Verification below is against `main` at cf4953b plus the repository history.

### 1. [P1] Omitted `schema_id` is rejected like an explicit zero — no change, with reasoning

The factual premise is right: `encoding/json` decodes an omitted `schema_id` and an explicit `"schema_id": 0` to the same `Manifest.SchemaID == 0`, and the base docs did describe zero-stamped manifests as readable. The conclusion does not follow, for three reasons.

**No Forma writer has ever omitted the field.** `internal/manifest/manifest.go` was added in 968afbc (#7, "Add CDC and Federated Query feature"); `LoadOrCreate` set `SchemaID: schemaID` on the created manifest in that first version, and the `schema_id` tag has never carried `omitempty`, so every manifest a Forma writer saved has always serialised a non-zero field. `git log -p -S'Manifest{'` over non-test files shows only that one construction site across the whole history (re-touched by 8b4dfb1, still stamped). Every other writer (cdc-flush, compactor, reconcile) loads before it saves and preserves the loaded stamp. There is no persisted "legacy omitted-field" population for a migration to cover. The base doc line that said otherwise was the error #522 corrected; it described a compatibility that never had a producer.

**Preserving field presence would reopen the hole.** #522 is about a hand-made or hand-edited manifest at a path a colliding template renders; the report names the zero stamp, not the field's presence, and a hand-made object is at least as likely to omit the field as to write `0` into it. Treating "omitted" as safe and "explicit zero" as unsafe would let the reported scenario through unchanged while refusing only the variant nobody produces. Neither shape proves which schema owns the entries, which is the property the rule checks.

**The migration path exists and is the one the error names.** The refusal message and `docs/manifest-reconcile.md` state the remedy: confirm ownership and set `schema_id` on the object. It is a one-field edit per affected object, it is only ever needed for hand-made objects, and the PR body's Migration section and `docs/error-handling.md` call the behaviour change out. There is no CHANGELOG in this repository to add an entry to.

So: no code change, and the "breaking compatibility" framing does not hold because there was no compatible producer. If a deployment does turn out to hold Forma-written manifests without the field, that would be a new fact worth an issue, but the history says there cannot be one.

### 2. [P2] Requested schema ID `0` bypasses the rule — confirmed, follow-up filed and fixed

Confirmed. `VerifySchemaStamp` compares `m.SchemaID == schemaID` before it checks for a zero stamp, so a request for schema `0` against a populated zero-stamped manifest takes the equality branch. Reachability is as described, with one correction and two narrowings:

- The compactor already rejects `SchemaID == 0` at the flag and in `RunOnce`, and `manifest-reconcile --schema-id` rejects `<= 0`, so those two CLI entry points were never open.
- The read path cannot request `0` from an HTTP caller (schemas are named, resolved through the registry) and the OLTP repository rejects `schemaID <= 0`.
- What is open: `reconcile.RegistrySchemaEnumerator`, `cdc.getSchemaIDsToInit` and `cdc.getUnflushedSchemaIDs` enumerate every row without a positivity guard, and the registry DDL has no `CHECK (schema_id > 0)`. An out-of-band registry or change_log row with `schema_id = 0`, plus a hand-made manifest at the path the template renders for `0`, reaches the gap.

Filed as #536 and fixed in #537: `VerifySchemaStamp` refuses a non-positive requested ID before comparing or writing any stamp, as a plain error (an invariant violation by the caller, not a property of the manifest, so it does not classify as a schema mismatch). `TestVerifySchemaStamp_RefusesNonPositiveRequestedSchemaID` covers requested `0` and `-1` against populated-unstamped, empty-unstamped and stamped manifests, asserts no stamp is written and no save happens through `LoadOrCreateForSchema`, and is red without the guard. Hardening the three enumerators and the registry DDL is recorded in #536 as a separate decision rather than folded into the boundary fix.

### Observations

- No cdc-flush scenario test for a populated zero-stamped manifest: `AppendFile`/`AppendFiles` reach `VerifySchemaStamp` only through `LoadOrCreateForSchema`, and `TestAppendFile_RejectsForeignManifestWithoutSave` already pins the refuse-without-save shape for the mismatch carrier on that path. Not adding a parallel test; the helper coverage is the coverage.

Issue #511, opened 2026-09-01
refactor(federated): duckdb_query.go sits at exactly the 500-line cap and engine.Query is 101 lines — split before the next edit


iceboundrock:
Surfaced by the #381 work and confirmed in the PR #509 review (§3.1, §5.1.3). Neither item is a correctness problem; both are size debt that has reached the point of blocking ordinary edits.

Related: #451 tracks four *other* files already past the cap; #452 is the size-debt epic. This issue covers the two `internal/federated` cases #381 ran into, which are different in kind — one is exactly *at* the cap, the other is a function cap.

## 1. `internal/federated/duckdb_query.go` is at exactly 500 lines, with zero headroom

`coding-standard.md` caps source files at 500 lines. This file is at **500** — compliant, but with nothing left.

That is not theoretical. During #381 the file had to gain a `validateKeysetCursor` call in `StreamDuckDBFederatedQuery` (the fourth seam onto the keyset renderer), and paying for those lines meant compressing the explanatory comment above the sibling call in `ExecuteDuckDBFederatedQuery`. The rationale that was cut is load-bearing: it explained that the safe-identifier regex is the **only** injection guard on a name interpolated as an identifier rather than bound as a parameter. That reasoning still exists at `safeSQLIdentifier`'s definition in `internal/federated/keyset.go`, so nothing is unexplained today — but a cap that forces comment deletion to make room for code is a cap that will start costing correctness commentary.

The next change to this file must split it first. Natural seams to consider: the scan-source resolution pre-flight, the execution-plan bookkeeping helpers, and the streaming path.

## 2. `DBFederatedQueryEngine.Query` is 101 lines, one over the function cap

`coding-standard.md` caps functions at 100 lines. `Query` (`internal/federated/engine.go`) is **101**. It is pre-existing, and #381 *reduced* it from 105 while adding a validation gate — so the branch moved it in the right direction without getting under the line.

`engine.go` itself is at 491 lines, so an extraction has room to land in the same file. The obvious candidate is the execution-plan boilerplate — the `IncludeExecutionPlan` / `ExecutionPlan` initialisation and note-appending block — which is bookkeeping rather than routing logic and reads as a unit.

## Why this is worth its own issue rather than a drive-by

[...]

PR #537, opened 2026-09-06
fix(manifest): refuse a non-positive requested schema ID in VerifySchemaStamp (#536)

### LINKED ISSUE #536: VerifySchemaStamp accepts a populated zero-stamped manifest when the requested schema ID is 0
Follow-up to #522 / #535, raised in the post-merge review on #535 (finding 2).

## Problem

`manifest.VerifySchemaStamp` (`internal/manifest/manifest.go`) is the single schema-identity rule introduced by #535. It checks `m.SchemaID == schemaID` before it checks for a zero stamp:

```go
func VerifySchemaStamp(m *Manifest, path string, schemaID int16) error {
	if m.SchemaID == schemaID {
		return nil
	}
	if m.SchemaID != 0 { /* mismatch */ }
	if len(m.Files) > 0 { /* unstamped */ }
	m.SchemaID = schemaID
	return nil
}
```

When the *requested* schema ID is `0`, a zero-stamped manifest that lists entries satisfies the equality branch and is accepted. That contradicts the rule #535 states ("refused for every schema") and the invariant `docs/error-handling.md` leans on in two places (zero-stamp section, and the `schema_id` `omitempty` reasoning): schema IDs are always positive, so `0` is never a schema and never a legitimate request.

## Reachability

A requested ID of `0` cannot come from an HTTP caller directly (schemas are named, and resolved through the registry), and the OLTP repository already rejects `schemaID <= 0`. It can still reach the manifest boundary through enumeration, because nothing enforces positivity on the way in:

- `schema_registry.schema_id` is `SMALLINT UNIQUE NOT NULL` with no `CHECK (schema_id > 0)` (`cmd/tools/init_db.go`).
- `reconcile.RegistrySchemaEnumerator` and `cdc.getSchemaIDsToInit` select every registry row when no `--schema-id` filter is given; only the filter path is guarded (`> 0`).
- `cdc.getUnflushedSchemaIDs` selects `DISTINCT schema_id` from `change_log`; a row with `schema_id = 0` written out of band would be flushed.
- The compactor rejects `SchemaID == 0` at both the flag and `RunOnce`, and `manifest-reconcile --schema-id` rejects `<= 0`, so those two CLI entry points are already closed.

So the gap needs an out-of-band write (a hand-inserted registry or change_log row) *and* a hand-made zero-stamped manifest at the path the template renders for `0`. Narrow, but it is exactly the class of hand-made state #522 exists to refuse, and the fix is a boundary guard, not a new rule.

## Proposed fix

1. In `VerifySchemaStamp`, refuse a non-positive requested schema ID before any stamp comparison, with a plain (operator-visible, non-4xx) error naming the path and the requested ID. This keeps one enforcement site; every consumer (`LoadOrCreateForSchema`, cdc-init, compactor) inherits it.
2. Regression tests: `VerifySchemaStamp` / `LoadOrCreateForSchema` with `schemaID == 0` (and a negative ID) against a populated zero-stamped manifest, an empty zero-stamped manifest, and a stamped manifest — all refused, nothing stamped, nothing saved.
3. Update the godoc and the zero-stamp bullet in `docs/error-handling.md`.

Not in scope here, worth a separate decision: a `CHECK (schema_id > 0)` on the registry DDL and positivity guards in the three enumerators listed above. The boundary guard makes those defence in depth rather than the only line.

iceboundrock:
Closes #536. Follow-up to #535 (post-merge review, finding 2).

## What

`manifest.VerifySchemaStamp` checked `m.SchemaID == schemaID` before it checked for a zero stamp, so a request for schema `0` against a zero-stamped manifest with entries took the equality branch and was accepted. Schema IDs are always positive, so `0` (or a negative ID) is never a schema. The helper now refuses a non-positive requested ID before comparing or writing any stamp:

```
manifest <path>: requested schema id 0 is not a schema (schema IDs are positive); refusing to verify the stamp
```

It is a plain error, not a `ManifestSchemaMismatchError` / `ManifestUnstampedError` carrier: an invalid request is an invariant violation by the caller (an unguarded enumerator or a hand-inserted registry row), not a property of the manifest. Every consumer inherits it through the single enforcement site (`LoadOrCreateForSchema`, cdc-init, compactor).

## Federated entry guard (review follow-up)

The plain error is unclassified, and the federated read path relabels every unclassified parquet-source failure as `ErrFederatedReadFailed`, which `AllowPartialDegradedMode` absorbs into a Postgres-only answer. So `Query` and `ExecuteFederatedPaginatedQuery` now share one entry guard, `validateFederatedQueryTarget`, that refuses a nil query or `SchemaID <= 0` before routing, with the same `schema id must be positive` invariant the OLTP repository enforces. No parquet source is consulted, no DuckDB pass runs, and no degraded fallback answers.

## Reachability

Not from an HTTP caller (schemas are named and resolved through the registry) and not from the OLTP path (the repository rejects `schemaID <= 0`). It is reachable through `reconcile.RegistrySchemaEnumerator`, `cdc.getSchemaIDsToInit` and `cdc.getUnflushedSchemaIDs`, which enumerate every row without a positivity guard, given an out-of-band registry or change_log row with `schema_id = 0`. Hardening those enumerators and adding `CHECK (schema_id > 0)` to the registry DDL is #538; this PR closes the consumer-side boundaries.

## Tests

`TestQuery_NonPositiveSchemaIDIsNotDegradable` and `TestExecuteFederatedPaginatedQuery_NonPositiveSchemaIDRefused` (`internal/federated/engine_schema_id_test.go`): schema `0` and `-1` through both engine entry points with `AllowPartialDegradedMode: true` are refused before routing: no Postgres-only fallback, no DuckDB pass, no parquet-source call, no `ErrFederatedReadFailed` classification. Red without the entry guard (the degraded fallback answered), green with it.

`TestVerifySchemaStamp_RefusesNonPositiveRequestedSchemaID`: requested `0` and `-1` against a populated zero-stamped, an empty zero-stamped, and a stamped-for-7 manifest are all refused, neither sentinel matches, the stamp is never written; through `LoadOrCreateForSchema` the store is not saved. Red without the guard, green with it.

## Docs

`docs/error-handling.md`, zero-stamp bullet under `ErrManifestSchemaMismatch`, including why the engine stops the plain error at entry instead of classifying it downstream.

## Verification

`make fmt-check`, `make lint`, and `go test -count=1` on `.`, `internal`, `internal/manifest`, `internal/cdc`, `internal/compaction`, `internal/reconcile`, `internal/httpapi`, `internal/federated`, `cmd/...`, `factory` all pass locally.

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -757,6 +757,13 @@ being read. It is raised before any path reaches a scan.
   `VerifySchemaStamp` refuses a non-positive request before comparing any
   stamp, as a plain error (an invariant violation by the caller, not a
   property of the manifest), so it does not classify as a schema mismatch.
+  Because it is unclassified, the federated engine must not let it reach the
+  parquet source at all: `resolveParquetPaths` relabels every unclassified
+  source failure as `ErrFederatedReadFailed`, which `AllowPartialDegradedMode`
+  absorbs into a Postgres-only answer. So `Query` and
+  `ExecuteFederatedPaginatedQuery` refuse a non-positive `SchemaID` at entry,
+  before routing (PR #537 review) — the same `schema id must be positive`
+  invariant the OLTP repository enforces — and no degraded fallback runs.
 - **Where the zero-stamp rule is enforced.** The same single site,
   `manifest.VerifySchemaStamp`, reached through `LoadOrCreateForSchema` by the
   federated read path, `cdc-flush`, `cdc-init` and `manifest-reconcile`, and
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -751,6 +751,12 @@ being read. It is raised before any path reaches a scan.
   `schema_id` on the object, rather than fix the template. An empty
   zero-stamped manifest has nothing another schema could own and loads stamped
   for the requested schema in memory; the consumer's next save persists it.
+  The *requested* schema ID must be positive too (#536): a request for schema
+  0 is never a schema, and without the guard it would satisfy the stamp
+  equality against a zero-stamped manifest and admit its entries.
+  `VerifySchemaStamp` refuses a non-positive request before comparing any
+  stamp, as a plain error (an invariant violation by the caller, not a
+  property of the manifest), so it does not classify as a schema mismatch.
 - **Where the zero-stamp rule is enforced.** The same single site,
   `manifest.VerifySchemaStamp`, reached through `LoadOrCreateForSchema` by the
   federated read path, `cdc-flush`, `cdc-init` and `manifest-reconcile`, and
--- a/internal/federated/engine.go
+++ b/internal/federated/engine.go
@@ -130,13 +130,34 @@ func NewDBFederatedQueryEngine(pgSource PostgresFederatedSource, dirtyIDFetcher 
 	return e
 }
 
+// validateFederatedQueryTarget is the entry guard shared by Query and
+// ExecuteFederatedPaginatedQuery. A non-positive schema ID can never name a
+// schema (schema IDs are always positive), so a request carrying one is a
+// caller invariant violation (an unguarded enumerator or a hand-inserted
+// registry row), not a state of the read surface. It is refused here, before
+// routing, because nothing below classifies it: the manifest source reports
+// it as a plain error (#536), resolveParquetPaths relabels unclassified source
+// failures as ErrFederatedReadFailed, and AllowPartialDegradedMode would then
+// absorb it into a Postgres-only answer — the caller's error served as a
+// hot-only page (PR #537 review). The Postgres leg already rejects the same
+// ID, so the guard makes the two routes agree rather than adding a rule.
+func validateFederatedQueryTarget(fq *model.FederatedAttributeQuery) error {
+	if fq == nil {
+		return fmt.Errorf("federated query cannot be nil")
+	}
+	if fq.SchemaID <= 0 {
+		return fmt.Errorf("federated query for schema id %d cannot be served: schema id must be positive", fq.SchemaID)
+	}
+	return nil
+}
+
 // Query implements FederatedQueryEngine. Hot-only requests delegate directly
 // to Postgres; otherwise the routing policy decides between Postgres and the
 // DuckDB federated path, falling back to Postgres on DuckDB failure when
 // opts.AllowPartialDegradedMode is set.
 func (e *DBFederatedQueryEngine) Query(ctx context.Context, tables model.StorageTables, fq *model.FederatedAttributeQuery, opts *model.FederatedQueryOptions) (*model.PersistentRecordPage, error) {
-	if fq == nil {
-		return nil, fmt.Errorf("federated query cannot be nil")
+	if err := validateFederatedQueryTarget(fq); err != nil {
+		return nil, fmt.Errorf("validate federated query: %w", err)
 	}
 	if e == nil || e.pgSource == nil {
 		return nil, fmt.Errorf("postgres federated source is not available")
--- a/internal/federated/engine_schema_id_test.go
+++ b/internal/federated/engine_schema_id_test.go
@@ -1,0 +1,94 @@
+package federated
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/lychee-technology/forma/internal/model"
+)
+
+// nonPositiveSchemaQuery is coldTierQuery() re-targeted at an ID that can
+// never name a schema.
+func nonPositiveSchemaQuery(schemaID int16) *model.FederatedAttributeQuery {
+	fq := coldTierQuery()
+	fq.SchemaID = schemaID
+	return fq
+}
+
+// TestQuery_NonPositiveSchemaIDIsNotDegradable pins the PR #537 review
+// finding: a request for schema 0 or a negative ID is a caller invariant
+// violation, and VerifySchemaStamp reports it as a plain error (#536) that
+// resolveParquetPaths would relabel ErrFederatedReadFailed — degradable. With
+// AllowPartialDegradedMode the invalid request would then be served as a
+// Postgres-only page. The engine refuses it at entry instead: no source is
+// consulted, no DuckDB pass runs, and no Postgres-only fallback answers.
+func TestQuery_NonPositiveSchemaIDIsNotDegradable(t *testing.T) {
+	restore := initTestDescriptors()
+	defer restore()
+
+	for _, schemaID := range []int16{0, -1} {
+		t.Run(fmt.Sprintf("schema_%d", schemaID), func(t *testing.T) {
+			// The source answers the way the manifest package does for this
+			// ID: a plain, unclassified error. If the guard regressed, this is
+			// what the degraded fallback would absorb.
+			src := &fakeParquetSource{pathsErr: fmt.Errorf(
+				"manifest manifest/%d.json: requested schema id %d is not a schema (schema IDs are positive); refusing to verify the stamp",
+				schemaID, schemaID)}
+			duck := &fakeDuckDBExecutor{rows: &singleDuckDBRow{rowID: uuid.New()}}
+			pg := &fakePostgresFederatedSource{page: &model.PersistentRecordPage{TotalRecords: 3}}
+			engine := NewDBFederatedQueryEngine(pg, &fakeDirtyIDFetcher{}, duck, nil,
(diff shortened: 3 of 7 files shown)


iceboundrock:
## Review

**Outcome: changes requested**

### Findings

- **P3: A non-positive schema ID can still be treated as degradable in the federated read path.** `VerifySchemaStamp` now returns a plain invariant error for `schemaID <= 0` at `internal/manifest/manifest.go:301-303`, but `QuerySource.load` only preserves the non-degradable classification for `ErrManifestSchemaMismatch` (`internal/manifest/query_source.go:57-62`). `resolveParquetPaths` therefore wraps this new error with `ErrFederatedReadFailed` (`internal/federated/parquet_source.go:76-87`), and `degradableFederatedError` does not exclude it (`internal/federated/engine_degrade.go:69-87`). With `AllowPartialDegradedMode` enabled, `mayDegradeToPostgres` can then route the invalid request to `degradeToPostgresOnly` (`internal/federated/keyset.go:246-255`, `internal/federated/engine.go:197-202`).

  This is reachable by an internal/direct federated caller even if normal HTTP schema resolution prevents it. Depending on the injected Postgres source, it can either return a misleading hot-only answer or replace the intended invariant failure with an unrelated fallback error. The new error is explicitly documented as a caller invariant violation, so it should not be absorbed by partial-degraded mode. Either reject non-positive IDs before federated routing, or preserve a dedicated non-degradable classification through the parquet-source layer and exclude it from `degradableFederatedError`. Add an engine-level regression test with `schemaID` 0 or -1 and `AllowPartialDegradedMode: true` asserting that no Postgres-only fallback is returned.

### Specification Check

- The PR correctly closes the stated #536 case: requested IDs 0 and negative values are rejected before stamp comparison or mutation.
- The tests cover zero-stamped populated manifests, zero-stamped empty manifests, stamped manifests, and the `LoadOrCreateForSchema` no-save path.
- The remaining issue is propagation of the new invariant across the federated error-classification contract, not the manifest guard itself.

### Standards Check

- `VerifySchemaStamp` remains small and flat; no file or function size limit is exceeded.
- The new error includes path, requested ID, expected positivity, and operator-useful context.
- Existing error wrapping and typed sentinel conventions are preserved in the manifest package.
- No duplication, naming, formatting, or documentation-standard issues found.

### Assessment

The manifest-level fix is minimal and correctly prevents the original schema-0 admission bug. Because the new error can be downgraded into a partial result in one internal read path, I recommend addressing the finding before approval.

### Observations

- No other behavioral, concurrency, mutation, or test-coverage findings were identified.
- The documentation accurately describes the intended plain-error semantics, which makes the federated degradation gap worth fixing explicitly.

### Verification

- `make fmt-check`: passed.
- `make lint`: passed, including the infra module lint.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/manifest ./internal/cdc ./internal/compaction ./internal/reconcile ./internal/federated -count=1`: passed.
- `./scripts/test_with_container_runtime.sh`: passed with rootless Podman; complete unit test suite passed.
- GitHub PR checks: Benchmark Smoke, E2E Smoke Tests, Lint, Test, and k6 Smoke all passed.
- The initial direct `make test` attempt lacked a configured container runtime and failed during `TestE2EHarnessMinimal`; rerunning through the repository-provided runtime wrapper passed.

### Summary

One P3 finding remains around preserving non-degradable semantics for invalid schema IDs under `AllowPartialDegradedMode`. Otherwise, the implementation, focused tests, documentation, formatting, lint, local tests, and CI checks are in good shape.

iceboundrock:
## Review response

**P3 (non-positive schema ID degradable under `AllowPartialDegradedMode`): confirmed and fixed in d128f20.**

Verified the path as described: `QuerySource.load` only passes `ErrManifestSchemaMismatch` through unwrapped, `resolveParquetPaths` relabels every other source failure as `ErrFederatedReadFailed`, and `degradableFederatedError` has no exemption for a plain error, so with degraded mode on the invalid request was served as a Postgres-only page.

Fix takes the first of the two options you offered: reject before federated routing.

- `Query` and `ExecuteFederatedPaginatedQuery` now share one entry guard, `validateFederatedQueryTarget` (`internal/federated/engine.go`), which refuses a nil query or `SchemaID <= 0` before the hot-only gate, routing, or any parquet-source call. It is a plain error carrying the same `schema id must be positive` invariant the OLTP repository already enforces, so the two routes agree instead of adding a rule.
- Chose this over a new non-degradable classification through the parquet-source layer because the manifest error is deliberately unclassified (caller invariant, not manifest state), and a dedicated sentinel for "the caller asked for a non-schema" would give an error class to a request that should never reach the read surface at all.
- The nil check was folded into the same helper so `Query` does not grow: it was already at 110 lines before this PR (tracked in #511 via #452), and it stays at 110. The nil-query test now matches the wrapped message by substring.

**Regression tests** (`internal/federated/engine_schema_id_test.go`): schema 0 and -1 through both entry points with `AllowPartialDegradedMode: true`, asserting no Postgres-only fallback (`pg.queryCalls == 0`), no DuckDB pass, no parquet-source call, no execution plan recorded, and no `ErrFederatedReadFailed` classification. Red without the guard (the failure is the `pg.queryCalls` assertion, i.e. the fallback did answer), green with it.

**Docs**: the #536 paragraph in `docs/error-handling.md` now states why the plain error must be stopped at the engine entry rather than classified downstream.

**Follow-up**: the source-side hardening #536 deferred (registry `CHECK (schema_id > 0)` and positivity guards in the three enumerators) is now #538. Your finding is the argument for it: consumer-side guards catch the ID only where someone remembered to check.

**Verification**: `make fmt-check`, `make lint` (root and infra), `go test -count=1` on `.`, `./internal`, `./internal/federated`, `./internal/manifest`, `./internal/cdc`, `./internal/compaction`, `./internal/reconcile`, `./internal/httpapi`, `./cmd/...`, `./factory` all pass locally.

iceboundrock:
## Review

**Outcome: approve**

### Findings

No findings identified in the final head `d128f20`.

The original schema-0 admission issue is fixed at `manifest.VerifySchemaStamp` (`internal/manifest/manifest.go:302`), before stamp comparison or mutation. The previously identified degradation gap is also closed: `Query` and `ExecuteFederatedPaginatedQuery` share `validateFederatedQueryTarget` and reject `SchemaID <= 0` before routing, source resolution, DuckDB execution, or Postgres-only fallback (`internal/federated/engine.go:144-151`, `internal/federated/pagination.go:26-28`).

### Specification Check

- Non-positive requested schema IDs are rejected as caller invariant violations, not classified as manifest mismatches.
- Zero-stamped manifests with entries, zero-stamped empty manifests, and manifests stamped for another schema remain covered by the existing identity rules.
- Both federated entry points refuse schema IDs `0` and `-1` even when `AllowPartialDegradedMode` is enabled.
- The regression tests assert no Postgres fallback, DuckDB pass, parquet-source call, execution plan, or `ErrFederatedReadFailed` classification for invalid IDs.
- The documentation explains why the plain invariant error is stopped at the engine boundary rather than classified downstream.

### Standards Check

- Error messages include the manifest path or federated schema ID and the expected positive-ID invariant.
- New errors are wrapped with operation context at both federated entry points.
- The implementation remains flat and minimal; changed source files remain within the repository size limits.
- No naming, formatting, duplication, or documentation issues were found.

### Assessment

This is a focused and correct fix. The manifest guard prevents invalid schema ownership checks and the federated entry guard preserves the intended non-degradable invariant semantics across both public engine seams.

### Observations

- No behavioral, concurrency, mutation, or test-coverage findings remain for this PR.
- The deferred source-side hardening described in #538 remains a sensible follow-up, but it is not required for this consumer-side fix.

### Verification

- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/federated ./internal/manifest -count=1`: passed.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test . ./internal/... ./cmd/... ./factory -count=1`: all non-container packages passed; `TestE2EHarnessMinimal` could not start because this local environment has no configured Docker/Podman endpoint.
- `make fmt-check`: passed.
- `make lint`: could not complete locally because golangci-lint installation exhausted the local disk quota; this was an environment limitation, not a reported lint finding.
- GitHub checks passed: Benchmark Smoke, E2E Smoke Tests, Lint, Test, and k6 Smoke.

### Summary

Approved. The implementation satisfies the stated specification and resolves the prior P3 degradation finding with appropriate regression coverage and documentation.

PR #539, opened 2026-09-06
fix(compaction): resolve leading-slash manifest paths verbatim, matching reads and reconcile (#516)

### LINKED ISSUE #516: manifest paths with a leading slash resolve to different keys in compaction than in federated reads and reconcile
Split out of the PR #515 review (fifth-round observation), where it was correctly judged pre-existing and out of scope for #417.

## Problem

A manifest path with a leading slash resolves to **different S3 keys** depending on which component reads it.

| Consumer | Leading-slash relative path `/1/x.parquet` resolves to | Source |
|---|---|---|
| Federated read, `manifest.QuerySource.Paths` | `s3://<bucket>//1/x.parquet` → key `/1/x.parquet` (verbatim) | `internal/manifest/query_source.go` |
| Federated missing-object probe, `QuerySource.MissingIn` | key `/1/x.parquet` (only the `s3://<bucket>/` prefix is cut) | `internal/manifest/query_source.go` |
| `manifest-reconcile`, `normalizeKey` | key `/1/x.parquet` (verbatim, deliberately: its godoc says stripping the slash "would break the match against the S3 listing") | `internal/reconcile/classify.go` |
| Compaction rewrite, `objectURI` and `bucketRelativeKey` | `s3://<bucket>/1/x.parquet` → key `1/x.parquet` (**slash trimmed**) | `internal/compaction/rewrite.go`, `internal/compaction/verify_inputs.go` |

So the read path, the probe and reconcile agree that the manifest path *is* the key, while compaction silently rewrites it to a sibling key.

## Reachability

This is not only a hand-built-manifest shape. Every `cdc.Build*Path` helper trims a trailing slash off the prefix and then formats `%s/%d/...`, so an **empty data prefix** yields `/1/<uuid>.parquet`, and the CDC flush stores that literal key both in S3 and in the manifest (reconcile's `normalizeKey` godoc records exactly this case). The compactor's own `canRewrite` requires a non-empty `DataPrefix`, but that guards the keys it *mints*, not the keys it *consumes*: the flusher's prefix is separate configuration, so a manifest produced by an empty-prefix flush is a valid compaction input.

## Consequence today

Not silent, but wrong and misleading:

- With the #347 checksum gate on (the default), `verifySourceChecksums` GETs key `1/<uuid>.parquet`, which does not exist, so the pass fails with the probe-failure error. Terminal (not in `isRetryable`), but the error and log name a key that was never in the manifest, so the operator is pointed at the wrong object.
- With `--skip-input-checksum-verify`, `MergeToTmp` hands DuckDB `s3://<bucket>/1/<uuid>.parquet` and the merge fails on a missing object.
- Either way a schema whose manifest carries leading-slash keys can never be rewritten, and `deleteObjects` (via the same helper) would target the sibling key too.

PR #515's `bucketRelativeKey` deliberately mirrors `objectURI` so the gate hashes exactly the object the merge reads; that closes the gate-vs-merge gap inside compaction but preserves the compaction-vs-reader gap described here.

## Proposal

1. **Pick one contract and record it** in `docs/federated-query/design.md` (manifest section): the manifest `path` is the bucket-relative key **verbatim**, matching what the writers store, the read path scans, and reconcile lists. That is the only choice that does not invalidate existing manifests and objects.
2. **Make compaction follow it**: drop the leading-slash trim from `objectURI` and `bucketRelativeKey` (the empty-key refusal for `""`, `"/"`, `s3://<bucket>/` from #417 stays as is).
3. **Pin it with a normalization test** that feeds one leading-slash entry through `QuerySource.Paths`, `QuerySource.MissingIn`, `reconcile.normalizeKey` and `Compactor.objectURI` / `bucketRelativeKey` and asserts they all name the same key. An e2e in `internal/e2e_harness/production` with an empty CDC prefix (flush → federated read → rewrite → reconcile) would make it a contract across the real components; the existing `compaction_rewrite_e2e_test.go` is the closest template.

If instead the decision is that leading-slash keys are unsupported, the fix moves to the writer (`cdc.Build*Path` refusing or normalizing an empty prefix) and to a load-time rejection, with a migration note for existing deployments. Either way the four consumers must end up agreeing.

## Out of scope

Foreign-bucket (`s3://other/...`) paths: already refused by #417 / PR #515.

iceboundrock:
Fixes #516.

## Problem

A relative manifest path is the bucket-relative object key, verbatim, for three of the four consumers: `manifest.QuerySource.Paths` renders `s3://<bucket>/<path>`, `QuerySource.MissingIn` cuts exactly that prefix back off, and `manifest-reconcile` compares the path against the listing unchanged via `cdc.NormalizeObjectKey`. Compaction alone trimmed a leading slash in `objectURI` and `bucketRelativeKey`.

The writers' `cdc.Build*Path` helpers strip only a trailing `/` from the data prefix, so an empty prefix yields keys like `/1/<uuid>.parquet`. The flush stores that literal key and lists it as-is. Compaction then hashed, merged and deleted the sibling `1/<uuid>.parquet`, which was never written, and every rewrite of such a schema failed on a probe error naming the wrong object.

Review found the same trim on the producer side: `cdc.CopyTmpToFinal` built its S3 `CopySource` from a slash-trimmed tmp key, so under an empty prefix the copy named `bucket/1/_tmp/<uuid>.parquet` while the exporter had written `/1/_tmp/<uuid>.parquet`. Every empty-prefix promotion through `cdc-flush` and `cdc-init` failed on that copy.

A second review round found that the same header was never URL-encoded: S3 URL-decodes `x-amz-copy-source` and the SDK v2 serializer sets it verbatim, so a data prefix carrying a reserved byte (`?`, `#`, `%`, `+`, whitespace, non-ASCII) produced a copy source S3 would misread (a `?` starts a `versionId` query) or reject. `--s3-prefix` has no character validation, so every promotion under such a prefix failed on `CopyObject`.

## Change

- `Compactor.objectURI` renders a relative path as-is: `s3://<bucket>//1/x.parquet`, the object the read path scans.
- `Compactor.bucketRelativeKey` keeps the key verbatim. The empty-key refusal stays: `""`, a bare relative `/`, and `s3://<bucket>/` still resolve to no key; foreign buckets are still refused (#417). The bare-`/` refusal is relative-only: the own-bucket URI `s3://<bucket>//` resolves to the key `/`, as it does in `MissingIn` and reconcile (third review round).
- `cdc.CopyTmpToFinal` names the tmp key the exporter wrote in `CopySource` (`<bucket>//1/_tmp/<uuid>.parquet` under an empty prefix), so the object the exporter wrote is the one that gets promoted.
- `cdc.CopyTmpToFinal` percent-encodes that key for the header through `encodeCopySourceKey`: `/` and RFC 3986 unreserved characters pass through, so every key `Build*Path` mints under an ordinary prefix is byte-identical on the wire and a leading slash survives; everything else is escaped as `%XX`. The destination `Key` and the tmp cleanup keep the raw key.
- `docs/federated-query/design.md` §4.3.1 (manifest source level) records the contract: relative `path` is the bucket-relative key verbatim across all four consumers and the writers' promotion step, absolute `s3://` entries pass through, foreign buckets are refused, and `""` / `/` are refused as empty keys. It separates the logical key, which no consumer or producer changes, from the header's required transport encoding.
- The e2e helper `assertRewrittenBase` rendered the base URI with the same trim; it now follows the contract. `TestCompactionRewriteEquivalence` still passes on the harness. The remaining production-harness renderers that trim are swept in #540.

## Tests

- `TestManifestPathContract_ConsumersAgreeOnKey` feeds `/1/lead.parquet`, `p/1/plain.parquet`, `s3://bkt//1/abs.parquet`, and `s3://bkt//` through all four consumers and asserts one key and one URI per input.
- `TestVerifySourceChecksums_LeadingSlashRelativeKeyVerbatim` asserts the #347 checksum gate GETs the verbatim key.
- `TestCompactor_Rewrite_LeadingSlashSourcesResolveVerbatim` drives a full rewrite and asserts the gate GETs, DuckDB source URIs, and deletes all name the leading-slash objects.
- `TestDeleteObjects_UsesBucketRelativeKey` updated: a leading-slash relative path deletes the verbatim key, `s3://bkt//` deletes the key `/`, and a relative `/` is skipped with a WARN. `TestRejectForeignSources_EmptyKeyRefused` asserts `s3://bkt//` passes the scope gate.
- `TestCopyTmpToFinal_LeadingSlashTmpKeyCopiesVerbatim` asserts a `/1/_tmp/...` tmp key produces `CopySource` `test-bucket//1/_tmp/...`, copies to the final key, and deletes the verbatim tmp key. `TestCopyTmpToFinal_PrefixedTmpKeyCopySource` pins the ordinary `<bucket>/<key>` shape.
- `TestCopyTmpToFinal_ReservedCharsTmpKeyEncodedCopySource` asserts a leading-slash tmp key holding ` ?#%+é` produces `CopySource` `test-bucket//a%20b%3Fc%23d%25e%2Bf%C3%A9/1/_tmp/file.parquet` while the destination `Key` and the tmp delete carry the raw key. `TestEncodeCopySourceKey` pins the encoder: identity for `/` and unreserved characters, `%XX` for `?`, `#`, `%`, `+`, quotes, sub-delims, control and non-ASCII bytes.
- The compaction test double `fakeObjectS3.CopyObject` now decodes `CopySource` the way S3 does.

The issue's optional empty-prefix e2e was tried and dropped: MinIO rejects leading-slash object keys outright (`PutObject` 400 InvalidArgument; DuckDB COPY 400), so the harness cannot write the object such a test needs. AWS S3 accepts them, which is where the bug is reachable. Ruling recorded on the issue: https://github.com/Lychee-Technology/forma/issues/516#issuecomment-5562932033. The same limitation applies to the promotion path: a real `CopyObject` against a leading-slash key is only exercisable on AWS, so its coverage is the unit assertion on the request shape.

## Verification

- `make fmt-check`, `make lint`: clean
- `make test` (with the Podman socket for the harness smoke): all packages ok
- `go test ./internal/cdc ./internal/compaction` after the `CopyTmpToFinal` fixes: ok
- `go vet -tags=e2e ./internal/e2e_harness/production/`: clean
- `go test -tags=e2e ./internal/e2e_harness/production -run 'TestCompactionRewriteEquivalence|TestCDCFlush'` against MinIO after the encoding change: pass (the harness prefix encodes to itself, so the wire bytes are unchanged)


iceboundrock:
## Review

### Findings

1. **[P1] Preserve leading slashes in the `CopySource` key** (`internal/cdc/helpers.go:152`)

   The new contract makes `/1/...` a valid bucket-relative object key, but `CopyTmpToFinal` still constructs the S3 copy source with `strings.TrimPrefix(tmpKey, "/")`. For a writer using an empty prefix, the exported object is `/1/_tmp/<uuid>.parquet`; the copy request instead names `bucket/1/_tmp/<uuid>.parquet`, which is a different key. This helper is used by both CDC flush and `cdc-init`, so the empty-prefix scenario documented by this PR cannot publish its final object even though the subsequent compaction read/checksum paths now preserve the slash. Please keep the key verbatim when constructing `CopySource` (for example, `bucket + "/" + tmpKey`) and add a test asserting that `/1/_tmp/...` produces `bucket//1/_tmp/...` and is copied successfully.

### Spec Check

- The compaction changes correctly implement the #516 decision that relative manifest paths are bucket-relative keys verbatim.
- `QuerySource.Paths`, `QuerySource.MissingIn`, `cdc.NormalizeObjectKey`, `objectURI`, and `bucketRelativeKey` are covered consistently for leading-slash, ordinary relative, and own-bucket absolute paths.
- Foreign-bucket and empty-key refusal behavior remains intact.
- The producer-side `CopyTmpToFinal` path is not covered by the stated contract and is the gap identified above.

### Standards Check

- No changed Go file exceeds the 500-line limit; the new test file is 124 lines.
- The changed functions remain below the 100-line limit and use contextual error handling where applicable.
- `gofmt` and lint checks pass.

### Assessment

**Request changes.** The compaction normalization itself is correct, but the P1 finding prevents the documented empty-prefix producer path from working and should be fixed or explicitly removed from the supported contract before merge.

### Observations

- The dropped MinIO E2E is a reasonable limitation because the harness rejects leading-slash object keys.
- The new unit tests provide good coverage for the compaction-side regression, but they stage the leading-slash source directly and therefore do not exercise writer export, tmp-to-final promotion, or a real S3 `CopySource` request.
- No other correctness or coding-standard findings were identified in the touched compaction code.

### Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/compaction ./internal/reconcile ./internal/manifest ./internal/cdc`: pass.
- The three new leading-slash compaction tests: pass.
- `make fmt-check`: pass.
- PR CI run `34066695147`: Test, Lint, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all passed.

### Summary

One P1 finding requires preserving a leading slash in `CopyTmpToFinal`'s S3 `CopySource` and adding promotion coverage. Aside from that producer-side gap, the PR's compaction fix, tests, documentation, and standards compliance look correct.

iceboundrock:
## Response to review

**P1, `CopyTmpToFinal` CopySource trim.** Confirmed and fixed in 28f4b48.

Verified against the writers before changing it: `flush_batch.go` and `init.go` both hand DuckDB `s3://<bucket>/<tmpKey>` with the key verbatim, so under an empty prefix the tmp object really is `/1/_tmp/<uuid>.parquet`, and the trimmed `CopySource` named a sibling that was never written. The trim dates from the initial import with no recorded reason, and nothing else relied on it: the compaction rewrite never reaches it with a leading-slash tmp key (`canRewrite` refuses an empty `DataPrefix`), and the e2e harness's own `init_keep_delta.go` already builds `CopySource` as `bucket + "/" + key`.

Change:

- `internal/cdc/helpers.go`: `CopySource` is now `bucket + "/" + tmpKey`. Comment on the function records why the slash must survive.
- `internal/cdc/helpers_test.go`: `TestCopyTmpToFinal_LeadingSlashTmpKeyCopiesVerbatim` asserts a `/1/_tmp/file.parquet` tmp key yields `CopySource` `test-bucket//1/_tmp/file.parquet`, the destination `Key` is the final key, and the cleanup deletes the verbatim tmp key. `TestCopyTmpToFinal_PrefixedTmpKeyCopySource` pins the ordinary `<bucket>/<key>` shape so the change cannot regress prefixed keys.
- `internal/cdc/mocks_test.go`: `recordingS3Client` now records `CopySource` and `Key` per copy.
- `docs/federated-query/design.md` §4.3.1: the manifest path contract now states that the producers' tmp→final promotion names the tmp key verbatim too, and that no consumer *or producer* trims it.

On the "copied successfully" part of the ask: the unit test asserts the request shape only. A real `CopyObject` against a leading-slash source key is not exercisable on the harness, for the same reason the issue's empty-prefix e2e was dropped (MinIO rejects leading-slash keys with 400). AWS S3 accepts them, which is where both the original bug and this one are reachable. The PR body now says so.

Verification after the fix:

- `go test ./internal/cdc ./internal/compaction`: ok
- `make fmt-check`, `make lint`: clean
- `go vet -tags=e2e ./internal/e2e_harness/production/`: clean

No other findings in the review needed a change. No follow-up issue was opened: the fix is inside this PR's contract and there is no remaining out-of-scope item.

iceboundrock:
## Review (current head: 28f4b48)

### Findings

1. **[P2] URL-encode `CopySource` while preserving the logical key** (`internal/cdc/helpers.go:159`)

`CopySource` is an S3 request-header value that the API requires to be URL-encoded. The AWS SDK v2 serializer writes this field verbatim, and `--s3-prefix` has no character validation (`cmd/tools/tool_flags.go:126-130`). A valid prefix containing a reserved character, such as `?`, `#`, `%`, whitespace, or non-ASCII text, therefore creates a literal tmp object but sends an ambiguous or invalid copy source. For example, a `?` in the prefix can be interpreted as a `versionId` query delimiter instead of part of the source key. The new tests at `internal/cdc/helpers_test.go:73` and `internal/cdc/helpers_test.go:84` also pin the unencoded wire value.

Encode the source key for the `CopySource` header while retaining the exact logical key, including its leading slash; leave the destination `Key` unencoded. Add a request-shape test for a leading-slash tmp key containing reserved characters. Update the documentation wording from byte-for-byte `CopySource` construction to distinguish the logical S3 key from its required HTTP encoding.

### Spec Check

The compaction changes implement the #516 decision for ordinary and leading-slash manifest keys: `QuerySource.Paths`, `MissingIn`, reconcile normalization, checksum verification, merge URIs, and deletion now resolve the same object. Own-bucket absolute paths, foreign-bucket rejection, and the retained empty-key policy are covered consistently.

The producer side now preserves the leading slash logically, but the documented claim that `CopySource` carries the key verbatim is not valid for keys requiring URL escaping. The transport representation must be encoded without changing the decoded source key.

### Standards Check

Changed Go code remains within the repository file/function-size limits, error handling remains contextual, and the new tests are focused. No other coding-standard violations were found.

### Assessment

**Request changes.** The leading-slash regression is otherwise fixed, but the P2 source-header encoding defect leaves CDC flush, cdc-init, and compactor promotion unreliable for supported prefixes containing reserved characters.

### Observations

The MinIO limitation makes the omitted empty-prefix end-to-end test reasonable; the new unit coverage exercises the relevant logical paths well. There are additional production E2E helper renderers that still trim leading slashes (`internal/e2e_harness/production/compaction_inflight_reader_e2e_test.go:44` and `compaction_evolution_e2e_test.go:139`), so a future S3-backed empty-prefix test could query a sibling key unless those helpers are aligned too.

### Verification

- `git diff --check main...origin/pr-539`: pass.
- `make fmt-check`: pass.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc ./internal/compaction`: pass.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/manifest ./internal/reconcile`: pass.
- All PR CI checks passed: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Summary

One P2 finding remains: URL-encode the `CopySource` request-header value while preserving the leading-slash source key. Apart from that transport-level defect and the noted E2E-helper coverage gap, the implementation, specification, tests, and standards compliance are sound.

iceboundrock:
## Response to review (head 28f4b48)

**P2, `CopySource` URL encoding.** Confirmed and fixed in dcf5e2c.

Verified before changing it: `service/s3@v1.96.0`'s `awsRestxml_serializeOpHttpBindingsCopyObjectInput` does `encoder.SetHeader("X-Amz-Copy-Source").String(*v.CopySource)` with no escaping, and the SDK's own field doc says "The value must be URL-encoded". So the header carried whatever bytes the prefix held. The bug predates this PR (the trimmed form had the same gap), but it sits in the function this PR rewrote and the §4.3.1 wording I added described the header as byte-for-byte, which was wrong for keys that need escaping.

Change:

- `internal/cdc/helpers.go`: `CopyTmpToFinal` builds `CopySource` as `bucket + "/" + encodeCopySourceKey(tmpKey)`. The encoder passes `/` and RFC 3986 unreserved characters (`ALPHA DIGIT - . _ ~`) through and escapes every other byte as `%XX`. That keeps the logical key intact, leading slash included, makes every key `Build*Path` mints under an ordinary prefix byte-identical on the wire, and escapes `?` (`versionId` delimiter), `+` (S3 decodes it as a space), `%`, whitespace and non-ASCII. `url.PathEscape` was not used because it escapes `/` and leaves `+` and other sub-delims alone. The destination `Key` and the tmp cleanup keep the raw key.
- `internal/cdc/helpers_test.go`: `TestCopyTmpToFinal_ReservedCharsTmpKeyEncodedCopySource` asserts a leading-slash tmp key holding ` ?#%+é` yields `CopySource` `test-bucket//a%20b%3Fc%23d%25e%2Bf%C3%A9/1/_tmp/file.parquet` while `Key` and the delete carry the raw key. `TestEncodeCopySourceKey` pins the encoder table: identity for `/` and unreserved, `%XX` for `?`, `#`, `%`, `+`, quotes, sub-delims, control and non-ASCII bytes. The two existing shape tests are unchanged and still pass, since their keys encode to themselves.
- `internal/compaction/rewrite_test.go`: `fakeObjectS3.CopyObject` now `url.PathUnescape`s the source the way S3 does before looking the bytes up, so the double cannot silently pass an unencoded header.
- `docs/federated-query/design.md` §4.3.1: the contract now says no consumer or producer changes the key, and names transport encoding as the one transformation that is not a change of key: the decoded source S3 acts on is always the logical key, and `Key` is never encoded.

**Observation, e2e helpers that still trim.** Not changed here; opened #540. A sweep finds 15 sites in 10 production-harness files, not just the two named, and they cannot change any outcome today (MinIO refuses leading-slash keys and every Env uses a non-empty prefix). Aligning two of fifteen inside this PR would leave the harness inconsistent with itself, so #540 proposes one shared `ParquetURI` helper plus byte-for-byte path comparison for the two manifest-stamp tests.

**Not in scope, stated for the record.** The reserved-character prefix also reaches DuckDB as an `s3://` URI inside the export `COPY` and the manifest reads. The SQL literal is escaped (`sqlutil.EscapeLiteral`), so quoting is safe, but how DuckDB's httpfs treats `?` or `#` inside an `s3://` key was not verified here. If an operator hits that, it is a separate issue on prefix validation at startup, not on the copy header.

Verification after the fix:

- `go test ./internal/cdc ./internal/compaction`: ok (all eight `CopyTmpToFinal` tests and `TestEncodeCopySourceKey` run and pass under `-v`)
- `make fmt-check`, `make lint`: clean
- `go vet -tags=e2e ./internal/e2e_harness/production/`: clean
- `go test -tags=e2e ./internal/e2e_harness/production -run 'TestCompactionRewriteEquivalence|TestCDCFlush'` against MinIO via the Podman socket: pass, so the real `CopyObject` path still promotes with the encoded header on an ordinary prefix

PR body updated to cover the encoding change and #540.

iceboundrock:
## Review

### Findings
1. **[P2] Preserve the key of an own-bucket absolute URI whose key is `/`.** `bucketRelativeKey` now rejects `key == "/"` indiscriminately at `internal/compaction/verify_inputs.go:137`. That makes `s3://bkt//` fail compaction input validation even though it is an own-bucket absolute URI for the valid S3 key `/`. The federated read path passes that URI through and probes `/` (`internal/manifest/query_source.go:97-100,124-131`), manifest-reconcile normalizes it to `/` (`internal/cdc/init_delta.go:75-82`), and `objectURI` likewise passes it through. The result is a new cross-consumer inconsistency and a rejected rewrite for a manifest entry the other consumers accept. Reject the bare relative `/` only; retain `s3://<bucket>/` as the empty-key rejection. Add `s3://bkt//` to `TestManifestPathContract_ConsumersAgreeOnKey` and a rewrite/delete assertion.

### Specification
The finding contradicts the PR's documented manifest-path contract: absolute `s3://` entries pass through unchanged, while only the relative `""` and bare `/` shapes are refused. The leading-slash relative-key change and the `CopySource` transport-encoding rule otherwise match the stated contract.

### Coding Standards
No other standards issues found. The new encoding helper is focused, named clearly, below project size limits, and the changed error paths continue to add context.

### Assessment
**Request changes** for the P2 path-contract regression above. After that distinction is restored, the implementation is sound.

### Observations
- `CopyTmpToFinal` correctly preserves the raw tmp key for destination and cleanup while encoding only `x-amz-copy-source`. AWS SDK v2 serializes `CopySource` verbatim, so the new percent-encoding is necessary and correctly handles leading slashes, reserved bytes, control bytes, and UTF-8.
- The new tests cover ordinary, leading-slash, and reserved-character keys. They omit the own-bucket absolute single-slash key case that exposes the finding.
- PR CI is green and the PR is mergeable at review time.

### Verification
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc ./internal/compaction ./internal/manifest ./internal/reconcile`
- `make fmt-check`
- `git diff --check cdd67949643201870a75a6d76086b1d5b56a5b97 dcf5e2c13c8ca237f6014a5622984cd1ad8642b9`

All commands above passed.

### Summary
One P2 finding: compaction must distinguish the rejected relative bare `/` from the valid absolute URI `s3://<bucket>//` whose object key is `/`. No other blocking issues found.

iceboundrock:
## Response to review (head dcf5e2c)

**P2, `s3://<bucket>//` refused as an empty key.** Confirmed and fixed in 7c8db7c.

Verified before changing it: on `main`, `bucketRelativeKey("s3://bkt//")` cut the bucket prefix, was left with `/`, and passed it on as the key. This PR added the `key == "/"` refusal after the prefix cut, so it caught the absolute form as well as the relative one. `QuerySource.MissingIn` probes `/` for that URI, `cdc.NormalizeObjectKey` returns `/`, and `objectURI` passes the URI through, so compaction was the one consumer refusing an entry the others resolve. That contradicts the §4.3.1 rule that absolute entries pass through unchanged, and it is a regression against `main`, not a pre-existing gap.

Change:

- `internal/compaction/verify_inputs.go`: `bucketRelativeKey` now branches on shape. An absolute URI is refused only for a foreign bucket or an empty key after the cut (`s3://<bucket>/`); `s3://<bucket>//` yields `/`. The `""` and bare-`/` refusal applies to relative paths only. The doc comment says the refusal is relative-only and why.
- `internal/compaction/path_contract_test.go`: `TestManifestPathContract_ConsumersAgreeOnKey` gains `s3://bkt//` and asserts all four consumers agree on key `/` and URI `s3://bkt//`.
- `internal/compaction/verify_foreign_test.go`: `TestRejectForeignSources_EmptyKeyRefused` asserts `s3://bkt//` passes the scope gate while `s3://bkt/`, `/` and `""` are still refused. `TestDeleteObjects_UsesBucketRelativeKey` adds `s3://bkt//` (deleted as key `/`) and a relative `/` (skipped with a WARN), so the delete side of the rule is pinned in both directions.
- `docs/federated-query/design.md` §4.3.1: states the refusal is relative-only and that `s3://<bucket>//` resolves to `/` in compaction as in `MissingIn` and reconcile.

PR body updated to match.

Verification after the fix:

- `go test ./internal/compaction ./internal/manifest ./internal/reconcile ./internal/cdc`: ok
- `make fmt-check`, `make lint`: clean

No other findings in the review needed a change. No follow-up issue was opened: the fix is inside this PR's contract, and #540 already tracks the harness renderers noted in the previous round.

iceboundrock:
## Review

### Findings
- **P3, non-blocking:** `internal/e2e_harness/production/compaction_inflight_reader_e2e_test.go:44` still renders a manifest path with `strings.TrimPrefix(f.Path, "/")`. For the AWS-valid empty-prefix case, a listed `/1/file.parquet` becomes `s3://<bucket>/1/file.parquet`, which addresses the sibling object rather than `s3://<bucket>//1/file.parquet`. Therefore this in-flight-reader test cannot exercise the verbatim leading-slash contract in that configuration. The PR description explicitly assigns the remaining production-harness renderers to #540; that follow-up should include this line.

No blocking findings.

### Spec
- The implementation matches the documented manifest-path contract in `docs/federated-query/design.md` §4.3.1: relative entries remain verbatim; own-bucket absolute entries preserve their key; foreign buckets and zero-length keys are refused.
- `s3://bkt//` consistently resolves to key `/`, while relative `/` remains rejected.
- `CopyTmpToFinal` now preserves the logical source key and percent-encodes only the `x-amz-copy-source` transport value. Destination and cleanup keys remain raw, as required.

### Standards
- Changed production functions are below the project size limits, use contextual error wrapping, and retain flat control flow.
- The focused contract tests cover leading-slash relative paths, own-bucket absolute paths, scope rejection, checksum reads, merge URIs, deletion keys, and reserved-byte CopySource encoding.

### Assessment
The fix is correct and narrowly scoped. It removes the compactor/promotion disagreement with the established read and reconcile behavior, including reserved-character source keys. The remaining risk is limited to the production harness coverage gap above; MinIO cannot create the AWS-valid leading-slash object needed for a full end-to-end reproduction.

### Observations
- The changed compaction fake now URL-decodes `CopySource`, so the rewrite-level tests model S3 source resolution rather than merely checking a request string.
- Other existing production-harness slash-trimming renderers remain outside this PR and are appropriately tracked by #540.

### Verification
- Reviewed head commit `7c8db7c2010aad5e798d9f5af8da2a4cd2dfc8bb`.
- `TMPDIR=$PWD/.tmp GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc ./internal/compaction` passed locally.
- All PR CI checks pass: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

### Summary
PR #539 correctly preserves manifest object keys verbatim across compaction and tmp-to-final promotion, and safely encodes CopySource header transport bytes. No blocking issues found; carry the non-blocking production-harness cleanup into #540.

iceboundrock:
## Response to review (head 7c8db7c)

**P3, non-blocking, `compaction_inflight_reader_e2e_test.go:44` still trims.** No code change in this PR; already tracked.

Verified against #540 before answering: that issue's table lists this exact site (`compaction_inflight_reader_e2e_test.go`, line 44, "inline URI render over manifest `Files`") among the 15 trimming sites in 10 production-harness files, and the proposed fix routes it through the shared verbatim-key `ParquetURI` helper along with the rest. The reason it stays out of this PR is unchanged from the previous round: the line cannot alter any outcome today (MinIO refuses leading-slash keys and every Env uses a non-empty prefix), and fixing one of fifteen sites here would leave the harness inconsistent with itself.

No new follow-up issue was opened, since #540 already covers the finding in full. No other findings in this round needed a change.

iceboundrock:
## Review (current head: `7c8db7c`)

### Findings

No blocking findings.

- **P3, non-blocking:** production E2E helpers outside this PR still render or compare manifest keys after trimming a leading slash. An AWS-backed empty-prefix scenario could therefore address the sibling key rather than the verbatim key. This is already comprehensively tracked by [#540](https://github.com/Lychee-Technology/forma/issues/540), which covers all 15 sites across 10 files; it does not affect the production implementation in this PR.

### Specification

- The implementation now satisfies the #516 manifest-path contract: relative paths remain verbatim bucket-relative keys through compaction URI rendering, checksum verification, and deletion.
- Own-bucket absolute `s3://<bucket>//` correctly resolves to key `/`, while the separate relative bare `/` and empty-key cases remain refused.
- `CopyTmpToFinal` preserves the logical source key, including a leading slash, and percent-encodes only the `x-amz-copy-source` transport value. Destination and cleanup keys remain raw.

### Coding Standards

- Changed Go files and functions remain within the repository size limits.
- The encoder is focused, tests cover ordinary, leading-slash, own-bucket absolute, and reserved-byte cases, and changed error paths retain context.
- No additional standards violations found.

### Assessment

The fix is correct and suitably scoped. It removes the compactor and tmp-to-final promotion disagreement with the read and reconcile paths, without weakening cross-bucket or empty-key safeguards. No request changes.

### Observations

- The compaction S3 test double URL-decodes `CopySource`, so rewrite tests model S3 source resolution rather than only asserting a request string.
- MinIO cannot create AWS-valid leading-slash object keys, so full empty-prefix E2E coverage is not currently possible. The focused unit coverage is appropriate for that harness limitation.

### Verification

Reviewed `7c8db7c2010aad5e798d9f5af8da2a4cd2dfc8bb`.

- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc ./internal/compaction ./internal/manifest ./internal/reconcile`: pass.
- `make fmt-check`: pass.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet -tags=e2e ./internal/e2e_harness/production/`: pass.
- `git diff --check main...HEAD`: pass.
- PR CI is green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Summary

PR #539 correctly preserves manifest object keys verbatim across compaction and tmp-to-final promotion, while safely encoding `CopySource` header transport bytes. No blocking issues found; retain the non-blocking production-harness cleanup in #540.

PR #541, opened 2026-09-07
fix(reconcile): fail closed on an unfollowable ListObjectsV2 page (#521)

### LINKED ISSUE #521: manifest-reconcile: truncated ListObjectsV2 page without a continuation token is treated as a complete listing
## Context

The third review round on #518 found that `cdc.ListObjectKeys` returned the keys seen so far when a `ListObjectsV2` page was `IsTruncated` without a `NextContinuationToken`, and forwarded an empty token unchecked. #518 fixes that for cdc-init (`cdc.ErrIncompleteObjectListing`, repeated-token guard). The same fail-open branch exists in `manifest-reconcile` and is pre-existing, so it was left out of that PR.

## Where

- `internal/reconcile/deps.go` `S3ObjectStore.ListObjects`: `if !aws.ToBool(out.IsTruncated) || out.NextContinuationToken == nil { return objs, nil }`, and an empty token is forwarded as a real one.
- The e2e harness paginators (`internal/e2e_harness/production/env.go`, `artifacts.go`, `state_snapshot_helpers_e2e_test.go`) stop on `NextContinuationToken == nil` without looking at `IsTruncated`, so a malformed page is silently treated as the last one there too. Test-only code; lower stakes, but the same shape.

## Why it matters

`reconcile.Reconciler` consumes the listing in two places (`internal/reconcile/reconcile.go`):

- The per-schema diff (`diffSchema`): a short listing under-reports orphans (fewer `--gc` candidates, fewer `--repair` adoptions — a silent no-op rather than data loss) and over-reports **dangling** manifest entries (listed in the manifest, not seen in the listing). The dangling path is guarded by a single-key re-probe before anything is spliced out, and a single-key listing cannot realistically be truncated, so the manifest-splice risk is mitigated but the report is wrong and the run says "ok".
- The candidate re-probe itself (`ListObjects(ctx, key)`) shares the loop.

The exposure is lower than cdc-init's purge (which deletes everything it did not see listed), but a tool whose job is to report drift should not report a clean state from a page it could not follow.

## Proposed change

- Fail closed in `S3ObjectStore.ListObjects` the same way #518 does for cdc-init: error on a truncated page without a non-empty token, and on a repeated token. Consider lifting the loop into one shared paginator (`cdc.ListObjectKeys` already carries the guard; reconcile needs `Size`/`LastModified`, so a `ForEachObject(ctx, client, bucket, prefix, fn)` helper would serve both) rather than keeping two copies of the rule.
- Regression tests mirroring `TestListObjectKeys_UnfollowableTruncationFailsClosed` in `internal/cdc/init_wiring_test.go`.
- Tighten the e2e harness paginators to check `IsTruncated` as well, or route them through the shared helper.

Origin: review of #518, third round.

iceboundrock:
Fixes #521.

## Problem

`S3ObjectStore.ListObjects` in `internal/reconcile/deps.go` returned the objects seen so far when a `ListObjectsV2` page was `IsTruncated` without a `NextContinuationToken`, and forwarded an empty token unchecked. The reconciler then diffed a schema against a short listing: fewer orphans reported (silent `--gc` / `--repair` no-op), spurious dangling entries, and a run that still said "ok". #518 closed the same branch for cdc-init only. The three production e2e harness paginators stopped on a nil token without looking at `IsTruncated`, so a malformed page passed as the last one there too. The two federated e2e harness listings (`ListParquetFiles`, `deleteS3Prefix`) issued a single request and never paginated at all.

## Change

- **`internal/cdc/list_objects.go` (new)** — `cdc.ForEachObject(ctx, client, bucket, prefix, fn)` is now the one ListObjectsV2 paginator. It carries the #518 fail-closed rule: a truncated page without a non-empty token, or a repeated token, fails with `cdc.ErrIncompleteObjectListing` and is not delivered to the callback. A callback error stops the listing and is returned unwrapped. `cdc.S3ListClient` is the one-method surface it takes; `S3InitClient` embeds it. `ErrIncompleteObjectListing` and `ListObjectKeys` move here from `init_wiring.go`, the latter as a thin wrapper.
- **`internal/reconcile/deps.go`** — `ListObjects` collects `ObjectInfo{Key, Size, LastModified}` through the shared paginator; the fail-open branch is gone. `s3ObjectAPI` embeds `cdc.S3ListClient`.
- **production e2e harness** — `listS3Keys`, `dumpS3Listing`, and `snapshotS3Inventory` route through `cdc.ListObjectKeys` / `cdc.ForEachObject`.
- **federated e2e harness** (`internal/e2e_harness/federated/s3_operations.go`) — `ListParquetFiles` and `deleteS3Prefix` delegate to package-level `listParquetKeys` / `deleteAllUnderPrefix`, which take the narrow listing surface and go through the shared paginator. File counts now cover every page and cleanup deletes every page; delete failures stay best-effort as before.

`reconcile.go` needs no change: both consumers of the listing already propagate a lister error out of the run.

## Tests

- `TestS3ObjectStore_ListObjectsUnfollowableTruncationFailsClosed` (reconcile) mirrors `TestListObjectKeys_UnfollowableTruncationFailsClosed`: truncated without token, truncated with empty token, repeated token → `errors.Is(err, cdc.ErrIncompleteObjectListing)`, nil result, exact call count. Before the fix the first case returned `nil` error and the other two forwarded the bad token.
- `TestForEachObject_*` (cdc): metadata pass-through across pages with token forwarding, callback error stops the listing, fail-closed rule without delivering the malformed page.
- `TestListParquetKeys_*` and `TestDeleteAllUnderPrefix_*` (federated, untagged unit tests with a scripted two-page client): every page visited with the `.parquet` filter applied, every listed key deleted, token forwarding, and fail-closed on an unfollowable page with nothing counted or deleted.
- Existing `TestListObjectKeys_UnfollowableTruncationFailsClosed`, `TestPreflightDeltaTier_IncompleteListingStopsBeforeAnySwapOrPurge`, and `TestS3ObjectStore_ListObjectsPaginates` pass unchanged.

## Verification

- `make fmt-check`, `make lint` (root and infra) — clean
- `go vet -tags=e2e ./internal/e2e_harness/production/` and `./internal/e2e_harness/federated/` — clean
- `go test ./internal/cdc ./internal/reconcile ./internal/e2e_harness/federated` — ok
- `scripts/test_with_container_runtime.sh` (full `make test`, rootless Podman) — all packages ok (first two commits); CI's Test job covers the third

Implementation plan: https://github.com/Lychee-Technology/forma/issues/521#issuecomment-5563574747

### DIFF
--- a/internal/cdc/init_wiring.go
+++ b/internal/cdc/init_wiring.go
@@ -2,10 +2,7 @@ package cdc
 
 import (
 	"context"
-	"errors"
-	"fmt"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/lychee-technology/forma/internal/manifest"
 )
@@ -15,7 +12,7 @@ import (
 // inventory (#371). *s3.Client satisfies it.
 type S3InitClient interface {
 	S3FullClient
-	ListObjectsV2(ctx context.Context, params *s3.ListObjectsV2Input, optFns ...func(*s3.Options)) (*s3.ListObjectsV2Output, error)
+	S3ListClient
 }
 
 // applyInitS3Wiring sets every S3-derived field of an init run context: the
@@ -70,48 +67,4 @@ func newDeltaTierSeams(client S3InitClient, bucket string) (
 		return DeleteObjectKey(ctx, client, bucket, key)
 	}
 	return list, del
-}
-
-// ErrIncompleteObjectListing marks a listing whose pages cannot be followed
-// to the end: a truncated page without a continuation token, or a token the
-// listing has already used. The inventory built on such a listing would be
-// short, and a short inventory under --replace-delta would purge less than
-// the delta tier holds, so the listing fails closed instead (#371 review).
-var ErrIncompleteObjectListing = errors.New("object listing incomplete")
-
-// ListObjectKeys returns every object key under prefix, following
-// continuation tokens. A page that claims to be truncated but carries no
-// usable token, or hands back a token already used, fails with
-// ErrIncompleteObjectListing rather than returning the keys seen so far.
-func ListObjectKeys(ctx context.Context, client S3InitClient, bucket, prefix string) ([]string, error) {
-	var keys []string
-	var token *string
-	seen := map[string]struct{}{}
-	for {
-		out, err := client.ListObjectsV2(ctx, &s3.ListObjectsV2Input{
-			Bucket:            aws.String(bucket),
-			Prefix:            aws.String(prefix),
-			ContinuationToken: token,
-		})
-		if err != nil {
-			return nil, fmt.Errorf("list objects under %s: %w", prefix, err)
-		}
-		for _, obj := range out.Contents {
-			keys = append(keys, aws.ToString(obj.Key))
-		}
-		if !aws.ToBool(out.IsTruncated) {
-			return keys, nil
-		}
-		next := aws.ToString(out.NextContinuationToken)
-		if next == "" {
-			return nil, fmt.Errorf("list objects under %s: page %d is truncated but carries no continuation token: %w",
-				prefix, len(seen)+1, ErrIncompleteObjectListing)
-		}
-		if _, repeated := seen[next]; repeated {
-			return nil, fmt.Errorf("list objects under %s: page %d repeats continuation token %q: %w",
-				prefix, len(seen)+1, next, ErrIncompleteObjectListing)
-		}
-		seen[next] = struct{}{}
-		token = aws.String(next)
-	}
 }
--- a/internal/cdc/init_wiring_test.go
+++ b/internal/cdc/init_wiring_test.go
@@ -248,14 +248,6 @@ func (c *malformedListClient) DeleteObject(_ context.Context, in *s3.DeleteObjec
 	return &s3.DeleteObjectOutput{}, nil
 }
 
-func listPage(truncated bool, token *string, keys ...string) *s3.ListObjectsV2Output {
-	out := &s3.ListObjectsV2Output{IsTruncated: aws.Bool(truncated), NextContinuationToken: token}
-	for _, k := range keys {
-		out.Contents = append(out.Contents, types.Object{Key: aws.String(k)})
-	}
-	return out
-}
-
 // A page that claims to be truncated but carries no usable continuation
 // token, or hands back a token the listing already used, cannot be followed
 // to the end. The listing fails closed with the named sentinel instead of
--- a/internal/cdc/list_objects.go
+++ b/internal/cdc/list_objects.go
@@ -1,0 +1,103 @@
+package cdc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+)
+
+// S3ListClient is the one-method surface a prefix listing needs. *s3.Client
+// satisfies it, as does every wider client interface that carries
+// ListObjectsV2 (S3InitClient, reconcile's object API, the e2e harness's
+// cluster client).
+type S3ListClient interface {
+	ListObjectsV2(ctx context.Context, params *s3.ListObjectsV2Input, optFns ...func(*s3.Options)) (*s3.ListObjectsV2Output, error)
+}
+
+// ErrIncompleteObjectListing marks a listing whose pages cannot be followed
+// to the end: a truncated page without a continuation token, or a token the
+// listing has already used. The inventory built on such a listing would be
+// short — cdc-init's --replace-delta purge would delete less than the delta
+// tier holds (#371 review), and manifest-reconcile would under-report
+// orphans and over-report dangling entries while still saying "ok" (#521) —
+// so the listing fails closed instead.
+var ErrIncompleteObjectListing = errors.New("object listing incomplete")
+
+// ForEachObject is the shared ListObjectsV2 paginator: it calls fn for every
+// object under prefix, following continuation tokens until the listing is
+// exhausted. A page that claims to be truncated but carries no usable
+// token, or hands back a token already used, is not delivered to fn; the
+// listing fails with ErrIncompleteObjectListing rather than presenting the
+// objects seen so far as the whole prefix. An error from fn stops the
+// listing at that page and is returned unwrapped, so callers keep the
+// context they attached themselves.
+//
+// Every production and harness paginator routes through here so the
+// fail-closed rule has exactly one copy (#521).
+func ForEachObject(ctx context.Context, client S3ListClient, bucket, prefix string, fn func(types.Object) error) error {
+	var token *string
+	seen := map[string]struct{}{}
+	for {
+		out, err := client.ListObjectsV2(ctx, &s3.ListObjectsV2Input{
+			Bucket:            aws.String(bucket),
+			Prefix:            aws.String(prefix),
+			ContinuationToken: token,
+		})
+		if err != nil {
+			return fmt.Errorf("list objects under %s: %w", prefix, err)
+		}
+		next, err := nextListingToken(out, prefix, len(seen)+1, seen)
+		if err != nil {
+			return err
+		}
+		for _, obj := range out.Contents {
+			if err := fn(obj); err != nil {
+				return err
+			}
+		}
+		if next == nil {
+			return nil
+		}
+		token = next
+	}
+}
+
+// nextListingToken validates one page's pagination state and returns the
+// token to forward, or nil when the page is the last one. It records the
+// token in seen so a later page that hands it back again is refused.
+func nextListingToken(out *s3.ListObjectsV2Output, prefix string, page int, seen map[string]struct{}) (*string, error) {
+	if !aws.ToBool(out.IsTruncated) {
+		return nil, nil
+	}
+	next := aws.ToString(out.NextContinuationToken)
+	if next == "" {
+		return nil, fmt.Errorf("list objects under %s: page %d is truncated but carries no continuation token: %w",
+			prefix, page, ErrIncompleteObjectListing)
+	}
+	if _, repeated := seen[next]; repeated {
+		return nil, fmt.Errorf("list objects under %s: page %d repeats continuation token %q: %w",
+			prefix, page, next, ErrIncompleteObjectListing)
+	}
+	seen[next] = struct{}{}
+	return aws.String(next), nil
+}
+
+// ListObjectKeys returns every object key under prefix, following
+// continuation tokens through ForEachObject and inheriting its fail-closed
+// rule: an unfollowable page fails with ErrIncompleteObjectListing rather
+// than returning the keys seen so far.
+func ListObjectKeys(ctx context.Context, client S3ListClient, bucket, prefix string) ([]string, error) {
+	var keys []string
+	err := ForEachObject(ctx, client, bucket, prefix, func(obj types.Object) error {
+		keys = append(keys, aws.ToString(obj.Key))
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return keys, nil
+}
--- a/internal/cdc/list_objects.go
+++ b/internal/cdc/list_objects.go
@@ -36,8 +36,8 @@ var ErrIncompleteObjectListing = errors.New("object listing incomplete")
 // listing at that page and is returned unwrapped, so callers keep the
 // context they attached themselves.
 //
-// Every production and harness paginator routes through here so the
-// fail-closed rule has exactly one copy (#521).
+// Every production and harness ListObjectsV2 caller routes through here so
+// the fail-closed rule has exactly one copy (#521).
 func ForEachObject(ctx context.Context, client S3ListClient, bucket, prefix string, fn func(types.Object) error) error {
 	var token *string
 	seen := map[string]struct{}{}
--- a/internal/cdc/list_objects_test.go
+++ b/internal/cdc/list_objects_test.go
@@ -1,0 +1,99 @@
+package cdc
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/stretchr/testify/require"
+)
+
+// scriptedListClient answers ListObjectsV2 with a fixed page sequence and
+// records the token each call forwarded.
+type scriptedListClient struct {
+	pages  []*s3.ListObjectsV2Output
+	tokens []*string
+}
+
+func (c *scriptedListClient) ListObjectsV2(_ context.Context, in *s3.ListObjectsV2Input, _ ...func(*s3.Options)) (*s3.ListObjectsV2Output, error) {
+	c.tokens = append(c.tokens, in.ContinuationToken)
+	if len(c.tokens) > len(c.pages) {
+		return nil, errors.New("unexpected ListObjectsV2 call")
+	}
+	return c.pages[len(c.tokens)-1], nil
+}
+
+// ForEachObject hands the callback every listed object with its metadata
+// intact, across pages, forwarding each page's continuation token (#521:
+// reconcile needs Size and LastModified, so the shared paginator must not
+// flatten objects to keys).
+func TestForEachObject_VisitsEveryObjectAcrossPages(t *testing.T) {
+	mod := time.Date(2026, 9, 6, 12, 0, 0, 0, time.UTC)
+	client := &scriptedListClient{pages: []*s3.ListObjectsV2Output{
+		{
+			Contents: []types.Object{
+				{Key: aws.String("data/7/a.parquet"), Size: aws.Int64(10), LastModified: aws.Time(mod)},
+			},
+			IsTruncated:           aws.Bool(true),
+			NextContinuationToken: aws.String("page-2"),
+		},
+		{
+			Contents: []types.Object{
+				{Key: aws.String("data/7/b.parquet"), Size: aws.Int64(20), LastModified: aws.Time(mod)},
+			},
+		},
+	}}
+
+	var seen []types.Object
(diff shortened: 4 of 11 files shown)


iceboundrock:
## Review: fix(reconcile) fail closed on an unfollowable ListObjectsV2 page (#521)

**Verdict: approve — no blocking findings.** The change does exactly what #521 specifies, the shared-paginator design matches the posted implementation plan, and all checks I ran pass.

### Spec check (vs #521 and the plan at #521#issuecomment-5563574747)

- [x] `S3ObjectStore.ListObjects` fails closed on a truncated page without a non-empty token and on a repeated token, via the #518 guard. The fail-open branch (`!IsTruncated || NextContinuationToken == nil`) and the unchecked empty-token forward are gone (`internal/reconcile/deps.go:66-80`).
- [x] The loop is lifted into one shared paginator (`cdc.ForEachObject`, `internal/cdc/list_objects.go:41`) rather than a second copy of the rule, as the issue proposed. `ListObjectKeys` is now a thin wrapper over it (`list_objects.go:93-103`) with unchanged signature and error text, so the existing `TestListObjectKeys_UnfollowableTruncationFailsClosed` keeps passing unmodified.
- [x] All three e2e harness paginators route through the shared helper (`env.go:listS3Keys` → `ListObjectKeys`; `artifacts.go:dumpS3Listing` and `state_snapshot_helpers_e2e_test.go:snapshotS3Inventory` → `ForEachObject`), so they inherit the `IsTruncated` check instead of stopping on a nil token alone.
- [x] Regression tests mirror the cdc-init table exactly (truncated-without-token, truncated-with-empty-token, repeated token → `errors.Is(err, cdc.ErrIncompleteObjectListing)`, nil result, exact call count) at the reconcile layer, plus direct `ForEachObject` coverage (metadata pass-through across pages with token forwarding, callback-error stop, malformed page never delivered).
- [x] `reconcile.go` needs no change, as claimed: both `ListObjects` consumers propagate the error (`reconcile.go:156-160` per-schema diff → `s.Err`; `reconcile.go:324-327` re-probe → caller-mapped `s.Err`/exit 1, never an "ok" report). Verified by reading both call sites.

### Findings

None blocking. No correctness, safety, or scope issues found.

### Standards check (`coding-standard.md`, `docs/error-handling.md`)

- File/function sizes: `list_objects.go` 103 lines, `ForEachObject` ~26 lines, `nextListingToken` ~15, `ListObjectKeys` ~10; `deps.go` 137, both test files <200. All within the 500/100 limits.
- Error handling: every S3 error and every guard failure is wrapped with context (`list objects under %s`, page number, bucket); `errors.Is` targets the named sentinel, never string comparison. The one bare `return err` (`list_objects.go:58-60`, callback error passthrough) is the documented exception — it preserves the caller's own context and `errors.Is` identity, and every call site re-wraps (`list s3 objects in bucket %s`, `list env s3 keys`, `list s3 objects`). Not a bare-return violation.
- Error semantics: `ErrIncompleteObjectListing` is a plain `errors.New` sentinel for a read-path consistency failure (operator-visible), not a `forma.InvalidInputf/NotFoundf/Conflictf` carrier — correct per `docs/error-handling.md`. The `TestNoBareSentinelWraps` guard (root package, forma sentinels) does not apply here.
- Nesting/structure: flat loop with early returns, max two levels; `nextListingToken` extraction keeps the paginator readable. No duplication introduced — net effect is deduplication (three harness loops + one reconcile loop collapse onto one guard).
- Godoc: all four exported items (`S3ListClient`, `ErrIncompleteObjectListing`, `ForEachObject`, `ListObjectKeys`) carry doc comments explaining the fail-closed contract.
- Import direction: `reconcile` → `cdc` and `e2e_harness/production` → `cdc` already existed; no cycle. `S3InitClient = S3FullClient + S3ListClient` is conflict-free (`S3FullClient` has no `ListObjectsV2`).

### Observations (non-blocking, no action required)

1. **Error-text change on the reconcile path is safe.** Outer prefix changed from `list s3 objects under <bucket>/<prefix>` to `list s3 objects in bucket <bucket>: list objects under <prefix>: ...`. Both bucket and prefix are still present in the chain, and no caller matches on the string (all use `errors.Is`/wrapping). Flagging only so the author confirms the churn is intentional.
2. **Double prefix in the final chain is verbose but harmless** (e.g. `list schema 7 objects under data/7/: list s3 objects in bucket bkt: list objects under data/7/: page 1 ...`). Each layer adds its own context per the wrapping standard; leaving as-is.
3. **Test-helper coupling:** `list_objects_test.go` reuses `listPage` defined in `init_wiring_test.go` (same package). Works, but a future move/rename of one file breaks the other implicitly. Optional: relocate `listPage` to a shared `_test.go` helper file.
4. **`ForEachObject` callback-error coverage at cdc level tests only the stop-on-first-object case** (one call, no second page fetch). Empty-string and repeated-token cases at the cdc level are covered transitively through the unchanged `TestListObjectKeys_*` wrapper test and the reconcile table — sufficient; no additional test requested.
5. **No nil-`fn` guard** — a nil callback would panic. All four call sites pass non-nil closures; a guard would be defensive bloat. Noted and dismissed.
6. **Termination edge:** the repeated-token guard catches cyclic pagination but an adversarial S3 minting infinite fresh tokens would loop until ctx cancel — same trust model as #518, and S3 is not adversarial here. No change.

### Verification

On the PR worktree (`2e095bc`), with the repo-pinned toolchain (`GOTOOLCHAIN=go1.26.0+auto`):

- `go test -count=1 ./internal/cdc/ ./internal/reconcile/` — **ok** (cdc 3.4s, reconcile 0.9s).
- Targeted: `TestForEachObject_*` (3), `TestListObjectKeys_UnfollowableTruncationFailsClosed` (3 subcases), `TestPreflightDeltaTier_IncompleteListingStopsBeforeAnySwapOrPurge`, `TestApplyInitS3WiringWiresDeltaTierSeams` — **all pass**.
- Targeted: `TestS3ObjectStore_ListObjectsPaginates`, `TestS3ObjectStore_ListObjectsUnfollowableTruncationFailsClosed` (3 subcases) — **all pass**.
- `go vet -tags=e2e ./internal/e2e_harness/production/` — clean; `gofmt -l` over all 8 touched files — clean.
- CI on the PR: Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke — **all SUCCESS**. Full `make test` via container runtime was not re-run locally (CI's Test job already covers it on this SHA).

### Summary

A faithful, minimal fix for #521: one shared fail-closed paginator replaces five ad-hoc loops, the reconcile store and harness inherit the #518 guard, regression tests mirror the cdc-init table, and the "no `reconcile.go` change needed" claim checks out against both `ListObjects` consumers. Standards and error-semantics compliant. **Approve.**

iceboundrock:
## Response to the review observations

One code change (observation 3, commit `ee80deb`); the rest are confirmations with no change. No follow-up issue opened: none of the six points describes work that is both out of scope here and necessary.

1. **Error-text change on the reconcile path — intentional.** Before: `list s3 objects under <bucket>/<prefix>` built in one place. Now the bucket is added by `S3ObjectStore.ListObjects` and the prefix by the shared paginator, because only the paginator knows which page failed and why. The only string match on the old text was the reconcile test, which now asserts the new chain (`internal/reconcile/deps_test.go:156`). No production caller matches on the message; all use `errors.Is`.

2. **Double prefix in the chain — left as is.** The three layers each add what they alone know: `reconcile.go` names the schema, `deps.go` names the bucket, the paginator names the prefix and page. Collapsing the prefix would mean either dropping the bucket wrap in `deps.go` or teaching the paginator to print the bucket, and both trade a redundant token for a layer that no longer wraps with its own context. Kept per the wrapping standard.

3. **`listPage` coupling — fixed in `ee80deb`.** The helper now lives in `list_objects_test.go` next to the `ForEachObject` tests it builds pages for, with a comment noting that `init_wiring_test.go` reuses it for the `ListObjectKeys` and pre-flight cases. Same package, so it stays one definition; the dependency now points from the older wiring tests toward the paginator's test file rather than the other way round. `go test ./internal/cdc/ ./internal/reconcile/`, `gofmt`, and `make lint` are clean after the move.

4. **Callback-error coverage at the cdc level — no change.** The direct `ForEachObject` test covers the stop-on-first-object path; the empty-token and repeated-token cases run through `nextListingToken` regardless of which wrapper is on top, and both `TestListObjectKeys_UnfollowableTruncationFailsClosed` and the reconcile table pin them. Adding a third copy at the `ForEachObject` level would not exercise any new branch.

5. **No nil-`fn` guard — no change.** All four call sites pass literal closures. A guard would turn a programmer error into a runtime error with no caller able to act on it.

6. **Termination edge (endless fresh tokens) — no change.** The repeated-token guard is there to catch a cyclic listing, which is the failure mode a broken or misbehaving endpoint can actually produce. An endpoint minting unbounded distinct tokens is the same trust boundary #518 accepted, and the loop stays bounded by the caller's context either way. Not worth a page cap that would also fail legitimate very large prefixes.

Working copy of this response: none beyond this comment.

iceboundrock:
## Review: Changes Requested

**Artifact: PR review.** This comment is the complete review record; review work was performed in the detached PR-head worktree at `/tmp/opencode/forma-pr541`.

### Findings

1. **P2 - federated harness listings still bypass the shared fail-closed paginator.** `internal/e2e_harness/federated/s3_operations.go:191-208` (`ListParquetFiles`) and `:247-262` (`deleteS3Prefix`) each issue exactly one `ListObjectsV2` request and ignore both `IsTruncated` and the continuation token. This conflicts with the implementation plan requirement that every production and harness paginator route through `cdc.ForEachObject` / `cdc.ListObjectKeys`. Once a prefix exceeds the S3 page size, file-count assertions omit later objects and cleanup leaves later-page objects behind, allowing stale test data to affect later cases. Route both functions through the shared paginator and add a normal multi-page regression test for each behavior.

### Specification

- The #521 production requirement is met for `reconcile.S3ObjectStore.ListObjects`: a truncated page with a missing, empty, or repeated continuation token returns an error matching `cdc.ErrIncompleteObjectListing`; it does not return a partial inventory.
- The reconciler already propagates lister errors before diff, repair, or GC, so no `reconcile.go` change is required.
- The broader implementation-plan requirement to centralize every production and harness paginator is not met because of finding 1.

### Standards

- Exported additions have GoDoc. `internal/cdc/list_objects.go` is 103 lines; changed functions remain below the 100-line limit. The narrow `S3ListClient` interface appropriately avoids duplicating the pagination loop.
- Non-blocking standards observation: `ForEachObject` returns callback errors unchanged at `internal/cdc/list_objects.go:59`, which conflicts with the literal mandatory error-context rule in `coding-standard.md:22-33`. This is explicitly documented API behavior in both the implementation plan and code, and `errors.Is` would remain usable if it were wrapped. Either document an exception to the standard or wrap it with object/listing context; this is not the reason for the requested change.

### Assessment

Request changes. The core reconciliation fix is correct and well-tested, but the stated single fail-closed paginator guarantee remains incomplete until the federated harness paths are migrated.

### Observations

- The new shared-paginator tests cover multi-page metadata forwarding, callback termination, and refusing a malformed page before delivering its contents.
- Existing `ListObjectKeys` tests cover empty and repeated-token malformed pages; the new reconcile regression table mirrors those cases.
- The second commit moves the shared `listPage` test helper next to the paginator tests; no concern found there.

### Verification

- Inspected PR metadata, both commits, full diff, issue #521 implementation plan, affected callers, and existing tests.
- `git diff --check 687acc9..ee80deb31fe232a7a7e61cafd025805cf87e761b` passed.
- `make fmt-check` passed.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/cdc ./internal/reconcile` passed.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet -tags=e2e ./internal/e2e_harness/production/` passed.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/e2e_harness/federated -run "^$"` passed (compile-only; no tests run).
- At review time, GitHub `Lint`, `Test`, `Benchmark Smoke`, and `k6 Smoke` were passing; `E2E Smoke Tests` was pending.

### Summary

One P2 finding: route the two remaining federated S3 listing paths through the shared paginator and cover normal multi-page behavior. No other blocking correctness issue found.

iceboundrock:
## Response to the changes-requested review

**Artifact: review response.** One code change (finding 1, commit `b14ac85`); the standards observation is answered below with no change. No follow-up issue: neither point is both out of scope and unaddressed.

### Finding 1 (P2) — federated harness listings — fixed in `b14ac85`

Confirmed: `ListParquetFiles` and `deleteS3Prefix` in `internal/e2e_harness/federated/s3_operations.go` each made one `ListObjectsV2` call and read neither `IsTruncated` nor the token. They were the last two `ListObjectsV2` callers outside `cdc.ForEachObject` (grep over non-test Go files now shows only the paginator itself).

Change:

- `ListParquetFiles` delegates to a package-level `listParquetKeys(ctx, client cdc.S3ListClient, bucket, prefix)` that runs `cdc.ForEachObject` and applies the `.parquet` filter per object.
- `deleteS3Prefix` delegates to `deleteAllUnderPrefix(ctx, client s3PrefixClient, bucket, prefix)`, where `s3PrefixClient` is `cdc.S3ListClient` plus `DeleteObject`. It lists the whole prefix through `cdc.ListObjectKeys` first, then deletes each key, so a delete during pagination can never disturb the listing. Individual delete failures stay best-effort, as the only caller (`ClearAllData`, which already discards the error) expects; the listing error is now returned wrapped rather than swallowed by a short first page.
- The harness field stays `*s3.Client`; the narrow interfaces exist so a unit test can drive the helpers without MinIO.

Tests (`s3_operations_test.go`, untagged so they run under `make test`, not only under `-tags=e2e`): `TestListParquetKeys_FollowsEveryPage` (two pages, filter applied on both, tokens `[nil, "page-2"]`), `TestDeleteAllUnderPrefix_DeletesEveryPage` (all three keys across two pages deleted), and one fail-closed case for each helper (`errors.Is(err, cdc.ErrIncompleteObjectListing)`, nothing counted, nothing deleted).

The `ForEachObject` godoc now reads "every production and harness ListObjectsV2 caller routes through here", since these two were single-shot listings rather than paginators and the earlier wording let them slip past the plan's enumeration.

Verification on `b14ac85`: `gofmt -l` clean on touched files; `go test -count=1 ./internal/cdc ./internal/reconcile ./internal/e2e_harness/federated` ok; `go vet -tags=e2e ./internal/e2e_harness/federated/` clean; `make lint` (root and infra) clean. `s3_operations.go` is 302 lines, the test file 96.

### Standards observation — callback error returned unwrapped — no change

The bare `return err` at `list_objects.go:59` returns the *caller's own* closure error, not an error from a lower layer. The rule in `coding-standard.md` exists so context is not lost on the way up; here the callback already attached whatever context it wanted, and every caller wraps the `ForEachObject` result again (`list s3 objects in bucket %s`, `list parquet files in bucket %s`, `list objects to delete in bucket %s`, and the production harness wraps). Wrapping inside the paginator would prepend `list objects under <prefix>` a second time to an error that is about the object, not the listing. The exception is documented in the `ForEachObject` godoc ("An error from fn stops the listing at that page and is returned unwrapped, so callers keep the context they attached themselves") and pinned by `TestForEachObject_CallbackErrorStopsListing` via `require.ErrorIs`. Taking the "document an exception" branch of the observation, which the godoc already does.

Working copy of this response: none beyond this comment.

iceboundrock:
## Review: final follow-up for #521

**Artifact: PR review.** This is the consolidated review record for the current PR head `b14ac85659b7bb0f085af23c510140057b0f93b2`. Review work was performed in the detached worktree `/tmp/opencode/forma-pr541-final-review`.

### Findings

No blocking, non-blocking, or standards findings. The earlier P2 finding about the federated harness single-page listings is resolved by `b14ac85`; this review is against that commit, not the earlier two-commit state.

### Specification

- `cdc.ForEachObject` is now the sole `ListObjectsV2` caller in non-generated Go code. The full-source call-site check found no production or harness bypasses.
- The paginator validates a truncated page before delivering any of its objects. Missing/empty continuation tokens and repeated tokens return an error wrapping `cdc.ErrIncompleteObjectListing`, so consumers receive no partial listing.
- `reconcile.S3ObjectStore.ListObjects` preserves `Key`, `Size`, and `LastModified` while inheriting the guard. Its callers already propagate lister errors before reporting reconciliation results, repair, or GC.
- Both federated helpers now cover all pages: `listParquetKeys` filters `.parquet` keys across every page; `deleteAllUnderPrefix` completes the listing before attempting best-effort deletions, so deletes cannot interfere with pagination.
- Unit coverage exercises multi-page token forwarding, metadata pass-through, callback termination, malformed-page fail-closed behavior, parquet filtering, and all-page cleanup.

### Standards

- Changed source files and functions are within the repository 500-line/100-line limits.
- Exported additions have GoDoc; interfaces remain narrow and avoid a second pagination implementation.
- S3 and listing failures are contextualized and preserve sentinel matching through `%w`. The callback-error passthrough is explicit API behavior, documented and tested; it preserves callback-owned context.
- Error classification is appropriate: incomplete listing is an operator-visible plain sentinel, not a public write-validation carrier.

### Assessment

No code changes requested. The implementation is a minimal, complete resolution of #521 and its related harness coverage gap.

### Observations

- The reconciliation error chain now separates bucket context (`S3ObjectStore`) from prefix/page context (shared paginator). This is a harmless, useful wording change; callers use wrapped-error identity rather than error-text matching.
- The only remaining merge gate at review time is the CI `E2E Smoke Tests` job, which is still running.

### Verification

- Reviewed PR metadata, all three commits, full diff, existing review discussion, repository guidance, and all current `ListObjectsV2` call sites.
- `make fmt-check` passed.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/cdc ./internal/reconcile ./internal/e2e_harness/federated` passed.
- `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet -tags=e2e ./internal/e2e_harness/production ./internal/e2e_harness/federated` passed.
- GitHub CI: Lint, Test, k6 Smoke, and Benchmark Smoke are successful; E2E Smoke Tests is in progress.

### Summary

No findings. #541 centralizes every current production and harness S3 listing behind the fail-closed paginator, retains the reconciler metadata contract, and adds targeted multi-page regression coverage. Working copy of this review: `/tmp/opencode/forma-pr541-final-review`.,workdir:/home/ruoshi/code/lychee/forma,timeout:120000}]}},{

PR #542, opened 2026-09-07
fix(sqlgen): reject bigint predicate operands outside int64 on both routes (#502)

### LINKED ISSUE #502: bigint predicate operands beyond int64 still fail only on the DuckDB route (CAST('1e+30' AS BIGINT) conversion error)
Follow-up from PR #500 review finding P2 (refs #384). Pre-existing on `main`; #384/#500 resolved the same failure shape for the integer/smallint and numeric classes (their operands now cast to DOUBLE, the storage width), but the **bigint** class keeps `CAST(? AS BIGINT)` — correctly, since bigint EAV/bound columns are physically BIGINT and must compare exactly above 2^53 (#281/#357).

## Symptom

A bigint-typed predicate whose operand does not fit int64 — e.g. `total gt:1e30` — renders `CAST('1e+30' AS BIGINT)` on the DuckDB route and raises a Conversion Error, while the Postgres route compares the operand against `NUMERIC` and answers normally. One route fails, the other answers: the failure mode #384 eliminated for the narrow classes.

## Constraint

The operand cannot simply widen to DOUBLE: BIGINT columns must keep exact comparison above 2^53 (the #357 boundary suite pins this). Options, per the PR #500 review:

1. **Consistent rejection at shared normalization** (recommended): an operand outside int64 range for a bigint-typed attribute is `forma.InvalidInputf` on BOTH routes. Symmetric with the #384 write-side contract (bigint values beyond int64 can no longer be written), and `gt:1e30` has an exact equivalent within range (`gt:9223372036854775807` / `lte` inversions). Note this changes today's PG-route behavior (it currently answers, and can even match pre-#384 historical over-range NUMERIC values — which are #205-ceiling territory and NULL on every DuckDB tier anyway).
2. A width-preserving wider cast (HUGEINT covers ±1.7e38) — keeps both routes answering, but introduces a third comparison domain and still errors past 1e38; more machinery for a corner with no legal data on either side of the comparison.

Acceptance: a regression test pinning that an out-of-int64 bigint operand gets the same answer (or the same 400) on the hot Postgres route and the DuckDB tiers.

iceboundrock:
Closes #502

## Problem

A bigint predicate operand that does not fit int64 (e.g. `total gt:1e30`) diverged by route. The Postgres binders compared it against `NUMERIC` and answered; the DuckDB route rendered `CAST('1e+30' AS BIGINT)` and raised a Conversion Error. `BIGINT` cannot widen to `DOUBLE` the way #384 widened integer/smallint/numeric, because bigint columns must compare exactly above 2^53 (#281/#357).

## Fix (option 1 from the issue: consistent rejection at shared normalization)

- New `sqlgen.checkBigIntOperandRange`, called from all three predicate binders (`ConvertPgMainValue`, `parsePgEavValue`, `parseDuckDBRawParam`). A float64 parse result for a bigint attribute outside `[-2^63, 2^63)`, including `Inf`/`NaN` spellings, is refused with `forma.InvalidInputf`, so every route returns the same 400.
- The bound mirrors the write funnel's `checkDeclaredIntegerFit` (#384), so read and write agree on what "fits bigint". The bound is on magnitude, not integrality: a fractional literal past int64 (`1.5e30`) is rejected too, as the write side rejects a non-integral out-of-range value.
- Unaffected: every integral literal within int64 (already bound as exact int64 by `TryParseNumber`, #357) and fractional in-range operands such as `lt:1.5`.
- Every rejected literal has an exact in-range equivalent (`gt:1e30` ≡ `gt:9223372036854775807`), so no query becomes inexpressible.

## Out of scope (unchanged)

- Batch attr-value anchor (`buildAttrValuesAnchor`): a single-route, Postgres-only anchor with no DuckDB twin. It binds caller-supplied values through `TryParseNumber` (relation-enrichment lookup), so an out-of-range float64 simply compares against `value_numeric`; there is no cross-route divergence to fix.
- `BuildListPredicate`: no production caller; it still strict-casts and bypasses the shared binders. Tracked as a follow-up (see the review thread below).
- Historical over-range `NUMERIC` rows (#205).

## Tests

- `internal/sqlgen/bigint_operand_range_test.go`: all three binders reject the same literal set (`1e30`, `9223372036854775808`, `-1e19`, `0x1p70`, `1.5e30`, `-9.5e18`, `Inf`, `-Inf`, `NaN`) with `ErrInvalidInput`, accept the boundary literals (`9223372036854775807`, `-9223372036854775808`, `9.223372036854775807e18`, fractional in-range), and the check is bigint-only. Entrypoint coverage through `ToDualClauses`, `ToSQLClauses` and `BuildDuckClause`.
- `internal/e2e_harness/production/bigint_operand_range_e2e_test.go` (`-tags=e2e`): seeds a bound bigint at `MaxInt64` plus an EAV-only bigint, then drives the same rejection probes and boundary controls through hot Postgres, warm delta (real flush) and cold base (real compaction).

## Docs

`docs/federated-query/design.md` §6.4 records the read-side contract next to the existing bigint `BIGINT`-cast note.

Verification notes follow as a PR comment.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -848,7 +848,17 @@ applied on both sides:
   shortest round-trip form (`strconv.FormatFloat(v, 'g', -1, 64)`), so the
   DuckDB side recovers the identical float64 the PG side binds — `%.15g`
   dropped the 16th-17th significant digits. `bigint` stays `BIGINT` with
-  exact int64/decimal-string binds (#281/#357).
+  exact int64/decimal-string binds (#281/#357), so it cannot take the DOUBLE
+  widening: an integral `bigint` operand outside int64 range (`gt:1e30`,
+  `equals:9223372036854775808`, and the `Inf`/`NaN` spellings) is instead
+  rejected as user-facing invalid input by every predicate binder
+  (`sqlgen.checkBigIntOperandRange`, #502) — the same bound the write funnel
+  enforces, so no legal data sits on either side of such a comparison and
+  every rejected literal has an exact in-range equivalent
+  (`gt:9223372036854775807`). Before #502 the Postgres route answered against
+  `NUMERIC` while the DuckDB route raised a Conversion Error on
+  `CAST('1e+30' AS BIGINT)`. Fractional `bigint` operands in range are
+  unaffected.
 * **Bool**: both engines compare the `value_numeric <> 0` truthiness — the PG
   EAV EXISTS predicate renders `(x.value_numeric <> 0) =/!= <bool>` — and
   parse operands under one shared rule (`ParseBool` spellings, else any
--- a/internal/e2e_harness/production/bigint_operand_range_e2e_test.go
+++ b/internal/e2e_harness/production/bigint_operand_range_e2e_test.go
@@ -1,0 +1,104 @@
+//go:build e2e
+
+package production
+
+import (
+	"context"
+	"errors"
+	"math"
+	"testing"
+
+	"github.com/lychee-technology/forma"
+)
+
+// This file is the #502 acceptance suite: a bigint predicate operand that
+// does not fit int64 must get the same answer on the hot Postgres route and on
+// every DuckDB tier. The answer is a rejection — shared normalization refuses
+// the operand as invalid input on both routes — because BIGINT columns must
+// keep exact comparison above 2^53 (#281/#357) and so cannot widen to DOUBLE
+// the way #384 widened the narrow classes. Pre-#502 the Postgres route
+// compared the operand against NUMERIC and answered, while the DuckDB route
+// rendered CAST('1e+30' AS BIGINT) and raised a Conversion Error.
+
+// rejectionProbe is one filter that must fail with forma.ErrInvalidInput on
+// every route.
+type rejectionProbe struct {
+	name   string
+	filter Filter
+}
+
+// runRejectionProbes asserts every probe fails with a user-facing invalid
+// input error on the given route. A nil error is the pre-#502 Postgres
+// behaviour (answering); a non-sentinel error is the pre-#502 DuckDB
+// behaviour (Conversion Error surfacing as an opaque failure).
+func runRejectionProbes(ctx context.Context, t *testing.T, env *Env, label string,
+	base Query, probes []rejectionProbe) {
+	t.Helper()
+	for _, p := range probes {
+		q := base
+		q.Filters = []Filter{p.filter}
+		_, err := env.Query(ctx, q)
+		if err == nil {
+			t.Errorf("%s/%s: out-of-int64 bigint operand must be rejected, got an answer", label, p.name)
+			continue
+		}
+		if !errors.Is(err, forma.ErrInvalidInput) {
+			t.Errorf("%s/%s: rejection must be user-facing invalid input, got %v", label, p.name, err)
+		}
+	}
+}
+
+// TestBigIntOperandRangeParityBothDialects seeds a bound bigint (`amount`,
+// bigint_01) at MaxInt64 and an EAV-only bigint (`total`, attr 15), then
+// drives the same probes through hot Postgres, warm delta (real flush) and
+// cold base (real compaction): the out-of-int64 operands are rejected
+// identically everywhere, and the boundary literals keep answering the same
+// row set — the rejection must not clip MaxInt64 itself.
+func TestBigIntOperandRangeParityBothDialects(t *testing.T) {
+	cluster := SharedCluster(t)
+	env := NewEnv(t, cluster)
+	ctx := context.Background()
+	wide := DefaultSchemaFixtures()[1]
+
+	rowMax := CreateEvent(wide, map[string]any{"title": "r-max", "amount": int64(math.MaxInt64), "total": int64(7)})
+	rowSmall := CreateEvent(wide, map[string]any{"title": "r-small", "amount": int64(5), "total": int64(9)})
+	mustApplyEvents(ctx, t, env, "bigint operand range creates", rowMax, rowSmall)
+
+	rejected := []rejectionProbe{
+		// The issue's headline: an EAV-only bigint compared against 1e30.
+		{"total_gt_1e30", Filter{Attr: "total", Op: "gt", Value: "1e30"}},
+		// The bound bigint takes the pg-main binder on Postgres and the same
+		// DuckDB cast on the other route.
+		{"amount_gt_1e30", Filter{Attr: "amount", Op: "gt", Value: "1e30"}},
+		{"amount_lt_negative_1e19", Filter{Attr: "amount", Op: "lt", Value: "-1e19"}},
+		// The first integer past MaxInt64, in bare digits.
+		{"total_equals_2p63", Filter{Attr: "total", Op: "equals", Value: "9223372036854775808"}},
+	}
+	controls := []widthProbe{
+		// MaxInt64 stays addressable in bare and exponent spelling (#357).
+		{"amount_gte_maxint64", Filter{Attr: "amount", Op: "gte", Value: "9223372036854775807"}, []*Event{rowMax}},
+		{"amount_gte_maxint64_exponent", Filter{Attr: "amount", Op: "gte", Value: "9.223372036854775807e18"}, []*Event{rowMax}},
+		// MinInt64 as a bound answers (everything is above it).
+		{"amount_gt_minint64", Filter{Attr: "amount", Op: "gt", Value: "-9223372036854775808"}, []*Event{rowMax, rowSmall}},
+		// The exact in-range equivalent of the rejected gt:1e30 answers empty.
+		{"total_gt_maxint64_empty", Filter{Attr: "total", Op: "gt", Value: "9223372036854775807"}, nil},
+		{"total_lte_maxint64", Filter{Attr: "total", Op: "lte", Value: "9223372036854775807"}, []*Event{rowMax, rowSmall}},
+	}
+
+	hotBase := Query{Schema: wide, PreferHot: true, Limit: 100}
+	runRejectionProbes(ctx, t, env, "hot-pg", hotBase, rejected)
+	runWidthProbes(ctx, t, env, "hot-pg", hotBase, false, controls)
+
+	if _, err := env.RunFlush(ctx); err != nil {
+		t.Fatalf("flush: %v", err)
+	}
+	duckBase := Query{Schema: wide, Limit: 100}
+	runRejectionProbes(ctx, t, env, "warm-duck", duckBase, rejected)
+	runWidthProbes(ctx, t, env, "warm-duck", duckBase, true, controls)
+
+	if _, err := env.RunCompaction(ctx, wide); err != nil {
+		t.Fatalf("compaction: %v", err)
+	}
+	runRejectionProbes(ctx, t, env, "cold-duck", duckBase, rejected)
+	runWidthProbes(ctx, t, env, "cold-duck", duckBase, true, controls)
+}
--- a/internal/sqlgen/bigint_operand_range_test.go
+++ b/internal/sqlgen/bigint_operand_range_test.go
@@ -35,6 +35,11 @@ func TestBigIntOperandRangeAllBinders(t *testing.T) {
 		{"2p63_bare", "9223372036854775808", nil},
 		{"negative_1e19", "-1e19", nil},
 		{"hex_float_2p70", "0x1p70", nil},
+		// The bound is on magnitude, not integrality: a fractional literal
+		// past int64 is rejected too, exactly as the write funnel rejects a
+		// non-integral out-of-range value. Contrast fractional_in_range.
+		{"fractional_out_of_range", "1.5e30", nil},
+		{"negative_fractional_out_of_range", "-9.5e18", nil},
 		{"inf", "Inf", nil},
 		{"negative_inf", "-Inf", nil},
 		{"nan", "NaN", nil},
--- a/internal/sqlgen/bigint_operand_range_test.go
+++ b/internal/sqlgen/bigint_operand_range_test.go
@@ -1,0 +1,135 @@
+package sqlgen
+
+import (
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/stretchr/testify/require"
+)
+
+// TestBigIntOperandRangeAllBinders pins the #502 contract on the three
+// predicate binders — ConvertPgMainValue (PG main + hybrid), parsePgEavValue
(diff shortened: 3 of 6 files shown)


iceboundrock:
## Verification notes

All runs on the branch head (`9b6912f`), Go toolchain `go1.26.0+auto`, rootless Podman via `DOCKER_HOST=unix:///run/user/1000/podman/podman.sock` with `TESTCONTAINERS_RYUK_DISABLED=true`.

| Gate | Result |
|---|---|
| `make fmt-check` | pass |
| `make lint` (golangci-lint, root + infra module) | pass, exit 0 |
| `make test` (full unit suite, container runtime for `internal/testdb` etc.) | pass, every package `ok`, exit 0 |
| Production E2E, `-tags=e2e`, `go test ./internal/e2e_harness/production -run 'TestBigIntOperandRangeParityBothDialects|TestBigintFilterBoundaryBothDialects|TestEAVBigintFilterBoundaryBothDialects|TestEAVIntegerWidthParityBothDialects'` | pass, exit 0 (63s) |

E2E per-test:

```
--- PASS: TestBigintFilterBoundaryBothDialects (15.73s)
--- PASS: TestEAVBigintFilterBoundaryBothDialects (6.92s)
--- PASS: TestBigIntOperandRangeParityBothDialects (11.67s)
--- PASS: TestEAVIntegerWidthParityBothDialects (26.04s)
ok   github.com/lychee-technology/forma/internal/e2e_harness/production   63.359s
```

Red-then-green: the new `internal/sqlgen/bigint_operand_range_test.go` was run before the binder change and failed with nil errors on every rejected literal for all three binders; it passes after `checkBigIntOperandRange` was wired in.

What the new E2E pins (the issue's acceptance criterion): with a bound bigint seeded at `MaxInt64` and an EAV-only bigint, the probes `total gt:1e30`, `amount gt:1e30`, `amount lt:-1e19` and `total equals:9223372036854775808` fail with `forma.ErrInvalidInput` on hot Postgres, warm delta (after a real flush) and cold base (after a real compaction). Boundary controls (`gte:9223372036854775807`, `gte:9.223372036854775807e18`, `gt:-9223372036854775808`, `gt/lte:9223372036854775807` on the EAV bigint) keep returning the same row set on every tier, so the rejection does not clip `MaxInt64` itself.

Pre-existing bigint suites (`TestBigintFilterBoundaryBothDialects`, `TestEAVBigintFilterBoundaryBothDialects`) and the #384 width-parity suite were re-run alongside to confirm no regression in in-range behaviour.

iceboundrock:
## Review of #542 — reject bigint predicate operands outside int64 on both routes (#502)

### Summary

This PR implements option 1 from #502 (consistent rejection at shared normalization), matching the implementation plan posted on the issue. A new `checkBigIntOperandRange` in `internal/sqlgen/dualpath_sql_helpers.go` is called from the `float64` arm of all three predicate binders — `ConvertPgMainValue` (PG main + hybrid), `parsePgEavValue` (PG EAV EXISTS), `parseDuckDBRawParam` (DuckDB) — so an integral bigint operand outside `[-2^63, 2^63)` (including `Inf`/`NaN` spellings) is `forma.InvalidInputf` on every route instead of Postgres answering against `NUMERIC` while DuckDB raises `CAST('1e+30' AS BIGINT)` Conversion Error. 6 files, +292/−1. No blocking findings; verdict is **approve**.

### Spec check (against #502 + the plan comment)

- [x] **Chosen option**: option 1 (consistent rejection), as ruled in the plan. The HUGEINT alternative is correctly not taken; the code comment and `design.md` §6.4 both record why (no legal data beyond int64 since #384, exact in-range equivalent always exists).
- [x] **All three binders wired**: `ConvertPgMainValue`, `parsePgEavValue`, `parseDuckDBRawParam`. The hybrid walker is covered transitively via `ConvertPgMainValue` (`predicate_normalizer.go:144`), and the entrypoint test pins `ToDualClauses`, the PG-only generator, and `BuildDuckClause` independently — including the per-attribute ordering (bound `amount` fails in `pg main generation` first, EAV-only `total` in `pg sql generation`).
- [x] **Bound mirrors the write funnel**: `v >= math.MinInt64 && v < math.MaxInt64` is exactly the `#384 checkDeclaredIntegerFit` bigint bound (`internal/transform/typed_value.go:119`), including the documented `-2^63`-image edge (`-9223372036854775809` rounds to `-2^63` and is accepted, same as writes). `NaN` fails both halves, so it is rejected too.
- [x] **What is unchanged is correct**: integral in-range literals in any spelling still bind exact `int64` via `TryParseNumber` (#357, never reach the check); fractional in-range stays `float64` (pre-existing contract, pinned); integer/smallint/numeric still accept `1e30` as `float64` (bigint-only fence, pinned by `TestBigIntOperandRangeIsBigIntOnly`).
- [x] **Acceptance shape**: the new production E2E (`bigint_operand_range_e2e_test.go`) drives the issue's probes (`total gt:1e30`, `amount gt:1e30`, `amount lt:-1e19`, `total equals:9223372036854775808`) through hot Postgres → warm delta (real flush) → cold base (real compaction) asserting `ErrInvalidInput` everywhere, with boundary controls proving `MaxInt64` itself is not clipped. Out-of-scope statements (`buildAttrValuesAnchor`, `BuildListPredicate`, historical over-range `NUMERIC` rows) match the plan; see F1/F2 below for wording/residual notes.
- [x] **Behavior change disclosed**: the PG route previously *answered* these predicates (and could match pre-#384 over-range `NUMERIC` rows); it now 400s. The issue explicitly blesses this (option 1 note, #205-ceiling territory), and the PR body + docs state it.

### Standards check

- [x] **File/function sizes** (`coding-standard.md`): `dualpath_sql_helpers.go` 241 lines, `predicate_normalizer.go` 434, `duckdb_type_mapper.go` 277, new test files 135/104 — all under 500; `checkBigIntOperandRange` is ~10 lines, well under 100. No refactor trigger.
- [x] **Error semantics** (`docs/error-handling.md`): rejection is `forma.InvalidInputf("value %s out of range for bigint attribute '%s' (operand must fit [%d, %d])")` — names the literal as given, the attribute, and the expected range on a bigint-typed attribute. Upstream `%w` wraps (`dualpath_sql_generator.go:45,55`) preserve the sentinel + published message (entrypoint test asserts both `ErrorIs(ErrInvalidInput)` and the full tail). `TestNoBareSentinelWraps` passes (verified locally).
- [x] **Structure**: early-return guard, flat, max 1 level of nesting; called from three `float64` arms with no duplication beyond the one-line call each (no DRY violation).
- [x] **Docs**: `docs/federated-query/design.md` §6.4 records the read-side contract next to the existing `bigint stays BIGINT` sentence; accurate and minimal.

### Findings

No blocking findings. Three non-blocking notes:

- **F1 (info, docs wording) — `buildAttrValuesAnchor` rationale**: the PR body says it "binds stored FK values". Strictly, `internal/postgres_persistent_repository_by_attr_values.go:100-113` binds caller-supplied `values []string` through `TryParseNumber` (used by relation-enrichment lookup). The *outcome* (leave as-is) is still correct: that path is Postgres-only with no DuckDB twin, so an out-of-range float64 just compares against `value_numeric` with no cross-route divergence to fix. Suggest tweaking the wording to "single-route PG-only anchor, no DuckDB twin" if touched again.
- **F2 (info, residual) — `BuildListPredicate` still strict-casts**: `duckdb_sql_generator.go:39-82` binds the raw string with `CAST(? AS BIGINT)` for bigint element types, so an out-of-range element would still Conversion-Error there. Confirmed no production callers (only `duckdb_sql_generator_test.go`), so out-of-scope is correct today; if it is ever wired to caller literals it needs the same `checkBigIntOperandRange` call.
- **F3 (suggestion, test gap, non-blocking) — pin a fractional out-of-range literal**: the bound check is magnitude-based, so `1.5e30` is also rejected (consistent with option 1 and with writes, which reject non-integral out-of-range). The unit rejected set pins integral spellings + `Inf`/`NaN` but no fractional-out-of-range case. Consider adding e.g. `{"fractional_out_of_range", "1.5e30", nil}` to `TestBigIntOperandRangeAllBinders` to lock the magnitude (not integrality) semantics.

### Observations

- The `-9223372036854775809` → `-2^63`-image acceptance is a faithful mirror of the write side, not a hole: both routes accept the same image, so parity holds. The code comment states this; no action.
- Fractional in-range bigint operands (`lt:1.5`, `9007199254740993.5`) keep the pre-existing cross-type comparison behavior on both engines — unchanged by this PR, correctly out of scope.
- `duckdb_type_mapper.go` comment-only change is the right call: the `BIGINT` cast stays (exact comparison above 2^53 per #281/#357), and the binder now guarantees nothing uncastable reaches it.

### Verification

Ran on the PR head (`9b6912f`) in the review worktree, `GOTOOLCHAIN=go1.26.0+auto`:

| Gate | Result |
|---|---|
| `internal/sqlgen` targeted: `-run 'TestBigIntOperandRange'` (`AllBinders` 12 subtests, `IsBigIntOnly` 3, `Entrypoints` 2) | pass |
| `internal/sqlgen` full package `-count=1` | `ok` |
| Root `TestNoBareSentinelWraps` | pass |
| `make fmt-check` | pass |
| `make lint` (root + infra module) | exit 0 |

I did not re-run the container-backed production E2E locally; I rely on the author's posted verification notes on this PR (full unit suite + `TestBigIntOperandRangeParityBothDialects` with the #384/`boundary` companions passing in 63s on rootless Podman, plus the red-then-green unit run before the binder change). The red half is also evident by inspection: without the check, the `float64` arms return `(v, nil)` unconditionally. Unit + entrypoint coverage above is sufficient to pin the contract without re-running the 63s E2E.

### Assessment

Meets the #502 acceptance criterion with the agreed option-1 semantics, symmetric read/write bounds, and parity pinned at binder, entrypoint, and three-tier E2E levels. Error carrier, messaging, docs, and file/function standards all check out. **Approve — no changes required** (F3 is an optional hardening of the test matrix).

iceboundrock:

Thanks for the review. All three findings handled; branch head is now `13122c6`.

**F1 (wording, `buildAttrValuesAnchor`)** — accepted. The PR body now describes it as a single-route, Postgres-only anchor that binds caller-supplied values through `TryParseNumber` (relation-enrichment lookup), so an out-of-range float64 compares against `value_numeric` with no DuckDB twin and no cross-route divergence to fix. The plan comment on #502 carries the same inaccuracy ("binds stored FK values, never caller literals"); I have posted a correction there rather than edit the decision record in place. Outcome unchanged: left as is.

**F2 (`BuildListPredicate` residual)** — no change in this PR; tracked as follow-up #543. Beyond the strict `CAST(? AS BIGINT)` you noted, that helper also bypasses `TryParseNumber` and the shared binders entirely, so it predates #355/#357/#384 P2b as well as #502. It has no production caller (only `duckdb_sql_generator_test.go`; production list attributes are pivoted per element by `eavElementCastExpr` and filtered by the normal normalizer). #543 proposes deleting it, or routing its operand through `parseDuckDBRawParam` + `ToDuckDBParam` if a caller is ever planned, with acceptance criteria for either path.

**F3 (fractional out-of-range literal)** — accepted, in `13122c6`. `TestBigIntOperandRangeAllBinders` now pins `1.5e30` and `-9.5e18` as rejected on all three binders, with a comment stating the bound is on magnitude, not integrality, matching the write funnel. Both spellings must contain the literal as given in the published message.

Verification for `13122c6`:

| Gate | Result |
|---|---|
| `go test ./internal/sqlgen -run TestBigIntOperandRange -count=1` | pass (14 + 3 + 2 subtests) |
| `gofmt -l` on the edited test | clean |

The change is test-only in a file that already passed `make lint` and the full `make test` at `9b6912f`; no production code moved.

iceboundrock:
## Review: #542 bigint predicate range parity

### Summary

Reviewed the full PR range at `13122c6` (two commits, six files, `+297/-1`). The change implements #502 option 1: reject out-of-int64 `bigint` predicate operands during shared normalization, rather than allowing Postgres to compare them as `NUMERIC` while DuckDB fails its `BIGINT` cast. No blocking findings.

### Findings

None.

### Spec Check

- [x] The three relevant binders all enforce the rule: `ConvertPgMainValue` for Postgres main/hybrid, `parsePgEavValue` for Postgres EAV, and `parseDuckDBRawParam` for DuckDB.
- [x] The guard is bigint-only and uses the same float64 bound as `transform.checkDeclaredIntegerFit`: `[-2^63, 2^63)`. Exact in-range integral spellings remain `int64`; in-range fractional operands remain unchanged.
- [x] Rejections use `forma.InvalidInputf`, preserving a published `ErrInvalidInput` error for a consistent HTTP 400 on every route. The message identifies the supplied literal, bigint attribute, and accepted range.
- [x] Unit coverage spans all binders, generator entry points, EAV-only and column-bound bigint attributes, limits, exponent and hexadecimal forms, `Inf`/`NaN`, bigint-only controls, and both fractional out-of-range signs.
- [x] The production E2E exercises the required rejection controls through hot Postgres, warm delta after flush, and cold base after compaction, with MaxInt64 boundary controls.
- [x] `docs/federated-query/design.md` records the read-side contract beside the numeric-width rules.

### Standards Check

- [x] Modified production files and new tests remain below the 500-line limit; the new helper is small, single-purpose, and flat.
- [x] No duplicated range logic: the shared helper is called at each necessary float64 binder boundary.
- [x] Error construction follows `docs/error-handling.md`; no bare sentinel wrapping is introduced.
- [x] `git diff --check` and `make fmt-check` pass.

### Observations

- The deliberate `-9223372036854775809` to `-2^63` float64-image acceptance remains aligned with the existing write funnel and is documented in the helper. It preserves route and read/write parity.
- `BuildListPredicate` remains a non-production strict-cast residual, correctly outside this PR scope and already tracked in #543.

### Verification

Run locally on detached PR head `13122c6`, with `GOTOOLCHAIN=go1.26.0+auto`:

| Check | Result |
| --- | --- |
| `go test ./internal/sqlgen -count=1` | pass |
| `go test ./internal/sqlgen -run TestBigIntOperandRange -count=1 -v` | pass: 14 binder cases, 3 type-fence cases, 2 entry-point cases |
| `go test . -run TestNoBareSentinelWraps -count=1` | pass |
| `make fmt-check` | pass |
| `go test ./internal/e2e_harness/production -tags=e2e -run ^TestBigIntOperandRangeParityBothDialects$ -count=1 -timeout=5m` | pass in 9.752s, rootless Podman |

GitHub checks at review time: Lint, Test, Benchmark Smoke, and k6 Smoke pass; E2E Smoke Tests is still pending.

### Assessment

The implementation meets #502s selected consistent-rejection contract and preserves exact bigint behavior at the valid boundaries across Postgres and DuckDB tiers. Error semantics, documentation, tests, and repository standards are satisfied. **Approve: no changes required.**

PR #544, opened 2026-09-07
fix(cdc): --replace-delta retires the tiers of a schema with zero live rows (#519)

### LINKED ISSUE #519: cdc-init --replace-delta cannot retire the delta tier of a schema with zero live rows
Follow-up from #518 review finding F1 (#371).

## Problem

`cdc-init --replace-delta` purges the delta tier only after a base swap, and a schema with **zero live rows** takes no swap: `prepareSchemaInit` returns early, the manifest is left untouched, and the delta tier stays in place with only a warning (`internal/cdc/init.go`, "schema has no live rows; delta tier left in place because there is no base swap to publish"). This was a deliberate scoping decision in the #371 plan ("pre-existing gap, out of scope").

Consequence for an operator whose schema was emptied (every row dead) and then had an attribute type changed:

- Re-running `cdc-init --replace-delta` never retires the old-typed base and delta; they stay listed forever.
- The next init once rows exist still refuses (`cdc.ErrDeltaTierPresent`) until `--replace-delta` is passed again, so the flag is not idempotent for empty schemas. That part is safe (loud, never silent).
- There is no tool path to retire the stale tiers in the meantime: `manifest-reconcile --repair` classifies the uncovered tombstones in such a delta as lost deletes and appends them back (see the "why inline delete" paragraph in `docs/manifest-reconcile.md`), so hand-editing the manifest and deleting S3 objects is the only remedy.
- Federated reads over the emptied schema pin the stale columns; a value the pin cannot convert fails the read with a Conversion Error even though the result would be empty.

## Proposed fix

When `--replace-delta` is set and the schema has zero live rows, publish an empty swap under the same CAS path (`init_publish.go`): base tier replaced by the empty set, delta tier emptied, 412 re-splice as today, then purge the delta objects after the save commits. Dry-run reports it like any other purge. Without the flag the behavior stays unchanged (refuse when the delta inventory is non-empty).

Open question to settle in the plan: whether the old base entries of an emptied schema should be purged too, or only unlisted (compaction GC currently owns base retirement).

## Scope

Small: `runSchemaInit`'s `state == nil` branch plus a unit test in `internal/cdc/init_delta_test.go` and a production e2e case. Low priority; the residual is loud and bounded.

iceboundrock:
Closes #519 (follow-up of the #518 review finding F1, #371).

## What changed

`initSchema` returned early when a schema had zero live rows, so `cdc-init --replace-delta` over a schema whose every row was deleted and flushed left its stale base and delta tiers listed and the delta objects in place. The next plain init kept refusing with `ErrDeltaTierPresent`.

With `--replace-delta` that branch now goes through `finishEmptySchema` (`internal/cdc/init_delta.go`):

1. Export **one zero-row base object** through the shared `exportBatch` path. The base export SQL accepts an empty row-id list (`FALSE` row filters, full projected column set); snapshot exports still refuse it.
2. Publish it as the **sole base entry** (`RowCount 0`, no row-id range) with the delta tier emptied, under the same CAS publish as a populated schema (412 re-splice, ambiguous-save confirmation).
3. Purge the inventoried delta objects only after the save commits.

Dry-run reports 1 file / 0 rows and touches nothing. A schema that lists nothing and has no delta objects gets nothing exported and no manifest minted. Without the flag nothing changes. Old base objects are unlisted, never deleted (#416 / #203), and are ordinary orphans for `manifest-reconcile --gc`.

## Why a zero-row base instead of an empty swap

The plan posted on #519 proposed publishing an empty entry set. The production e2e caught that this **resurrects deleted rows**: `manifest.QuerySource.Paths` treats a manifest with zero entries like an absent one and falls back to the per-schema glob, which scans the unlisted old base without the purged tombstones (all 6 deleted rows came back). Compaction already keeps the invariant for an all-tombstone merge (a `RowCount 0` base entry so the manifest never empties), so init now follows the same rule. `updateSchemaManifest` never publishes an empty entry set, flag or no flag.

The object takes compaction's `base-{uuid}.parquet` key shape: an init-shaped `{min}_{max}_{uuid}` name needs a row-id range, and a range-less stem is unclassifiable for `--gc`. Because the manifest keeps one in-prefix base entry, the #463 empty-manifest guard does not trip and `--gc` needs no `--allow-empty-manifest-schema`.

Rulings 2 and 3 of the plan comment are superseded accordingly; a comment on #519 records the deviation.

## Verification

```
go test ./internal/cdc/ ./internal/compaction/ ./internal/manifest/ -count=1   # ok
go test . ./cdc ./cmd/... ./factory ./internal/... -count=1                     # ok (e2e_harness with DOCKER_HOST set)
go test -v ./internal/e2e_harness/production/ -tags=e2e \
  -run 'TestInitReplaceDeltaRetiresEmptiedSchema|TestCDCFlushAndInit' -count=1  # PASS under rootless Podman
make fmt-check && make lint                                                     # clean
```

New unit tests: `internal/cdc/init_empty_schema_test.go` (export → save → delete ordering, dry-run, failed save, failed export, nothing-to-retire, unlisted-delta-only), zero-row base SQL plan in `init_exporter_sql_test.go`.

### DIFF
--- a/docs/manifest-reconcile.md
+++ b/docs/manifest-reconcile.md
@@ -521,8 +521,20 @@ delta 真正退役的是 cdc-init 本身：
   对象。删除失败会记录 key 并让该 schema 返回带 key 列表的错误，但已发布的 manifest
   不回滚：读者此时只会看到新 base，残留对象成为未列孤儿，可由下文规则处理。
   `--dry-run` 只打印 `would delete N delta objects`。
-- **无 live 行的 schema 不换 base，也不动 delta。** 这类 schema 没有可发布的 swap，
-  带 `--replace-delta` 时只记一条 warning。
+- **无 live 行的 schema 用一个 0 行 base 完成 swap（#519）。** 所有行都已删除的 schema 没有
+  行可导出，但 manifest 可能仍列着旧类型的 base 与 delta。带 `--replace-delta` 时，这类
+  schema 先导出**一个 0 行的 base 对象**（`{prefix}/{schemaID}/base-<uuid>.parquet`，
+  与 compaction 全墓碑合并写出的形态相同，manifest 条目 `row_count: 0`、无 row-id 范围），
+  再走同一条 CAS 路径发布 swap：base 层替换为仅此一条、delta 层清空，保存成功后再删除
+  清理集中的 delta 对象；`--dry-run` 同样只报告（计 1 个文件、0 行）。manifest **绝不会
+  变成空集**：零条目的 manifest 会让读路径退回按 schema 前缀的 glob 扫描，把未列出的旧
+  base 连同其已被墓碑删除的行一起“复活”。旧 base 对象**只下架、不删除**（与任何重跑一致，
+  #416 起 init 从不删 base），由 `manifest-reconcile --gc` 作为普通未列孤儿回收——manifest
+  仍有一条本前缀内的 base 条目，#463 的空 manifest 防护不会触发，无需 `--allow-empty-manifest-schema`；
+  `--repair` 的晋升守卫（#292）对没有存活行的 schema 一律拒绝晋升，所以这些对象不会被
+  重新列回 manifest。manifest 本身不存在（或没有任何条目）、也没有任何 delta 对象的 schema
+  没有东西可下架，不会为此导出 0 行 base、也不会生成 manifest。不带 `--replace-delta`
+  时行为不变：delta 盘点非空即拒绝，否则不做任何事。
 - **清理集只认命名空间，不认 `tier` 标签。** 可删除的 key 必须是本 schema delta 命名
   空间里的 delta 形态对象：`{delta-prefix}/{schemaID}/<uuid>.parquet`（`--delta-prefix ""`
   时前缀部分不限，但 `/{schemaID}/<uuid>.parquet` 尾部仍然必须匹配）。manifest 里标为
--- a/internal/cdc/duckdb_exporter.go
+++ b/internal/cdc/duckdb_exporter.go
@@ -3,6 +3,7 @@ package cdc
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -123,20 +124,25 @@ func (e *DuckExporter) executeExportSQL(ctx context.Context, cfg CDCConfig, logM
 }
 
 func buildExportSQLPlan(spec exportModeSpec, pgConnStr string, s3TmpPath string, cfg CDCConfig, schemaID int16, snapshotTS int64, rowIDs []uuid.UUID, attrCache forma.SchemaAttributeCache) (exportSQLPlan, error) {
-	if len(rowIDs) == 0 {
-		modeName := "base"
-		if spec.useChangeLog {
-			modeName = "snapshot"
-		}
-		return exportSQLPlan{}, fmt.Errorf("export %s: no row ids provided", modeName)
+	if len(rowIDs) == 0 && spec.useChangeLog {
+		return exportSQLPlan{}, errors.New("export snapshot: no row ids provided")
 	}
 
 	pgEsc := sqlutil.EscapeLiteral(pgConnStr)
 	s3Esc := sqlutil.EscapeLiteral(s3TmpPath)
 	entityMain, eavData := resolveMainAndEAVTableNames(cfg)
-	rowList := quoteUUIDList(rowIDs)
-	mFilter := fmt.Sprintf("ltbase_row_id IN (%s)", rowList)
-	eFilter := fmt.Sprintf("row_id IN (%s)", rowList)
+	// A base export with no row ids is a legitimate request (#519): a schema
+	// with zero live rows still needs one zero-row base object carrying the
+	// full projected column set, so its manifest never lists nothing (an
+	// empty manifest sends readers to the legacy glob fallback, see
+	// manifest.QuerySource.Paths). The FALSE filters keep the projection
+	// intact while selecting no rows.
+	mFilter, eFilter := "FALSE", "FALSE"
+	if len(rowIDs) > 0 {
+		rowList := quoteUUIDList(rowIDs)
+		mFilter = fmt.Sprintf("ltbase_row_id IN (%s)", rowList)
+		eFilter = fmt.Sprintf("row_id IN (%s)", rowList)
+	}
 
 	plan := exportSQLPlan{}
 	if spec.useChangeLog {
@@ -144,7 +150,7 @@ func buildExportSQLPlan(spec exportModeSpec, pgConnStr string, s3TmpPath string,
 		if changeLog == "" {
 			changeLog = "change_log"
 		}
-		clFilter := fmt.Sprintf("row_id IN (%s)", rowList)
+		clFilter := fmt.Sprintf("row_id IN (%s)", quoteUUIDList(rowIDs))
 		plan.changeLogQuery = fmt.Sprintf(
 			"SELECT schema_id, row_id, changed_at, deleted_at FROM %s WHERE schema_id = %d AND flushed_at = 0 AND changed_at <= %d AND %s",
 			changeLog, schemaID, snapshotTS, clFilter,
--- a/internal/cdc/duckdb_exporter_sql_test.go
+++ b/internal/cdc/duckdb_exporter_sql_test.go
@@ -214,8 +214,10 @@ func TestBuildExportSQL_UsesCustomTableNames(t *testing.T) {
 // TestExportToTmpWrapsPlanErrors pins the caller-side context on both export
 // entry points: a plan failure must name the operation the caller was
 // performing, not arrive as a bare pass-through (#276, coding-standard.md).
-// Empty rowIDs is the cheapest plan failure — it returns before any DB or S3
-// use, so a zero-value DuckExporter is enough.
+// The cheapest plan failures return before any DB or S3 use, so a zero-value
+// DuckExporter is enough: empty rowIDs for the snapshot (a base export with
+// no row ids is a legitimate zero-row export since #519) and a missing
+// attribute cache for the base.
 func TestExportToTmpWrapsPlanErrors(t *testing.T) {
 	exporter := &DuckExporter{}
 	cfg := CDCConfig{}
@@ -230,9 +232,9 @@ func TestExportToTmpWrapsPlanErrors(t *testing.T) {
 
 	t.Run("base export", func(t *testing.T) {
 		err := exporter.ExportBaseFileToTmp(context.Background(), cfg,
-			"postgres://user@host/db", "s3://bucket/base/7/_tmp/f.parquet", 7, nil, testAttrCache())
+			"postgres://user@host/db", "s3://bucket/base/7/_tmp/f.parquet", 7, nil, nil)
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "export base file to tmp for schema 7")
-		require.Contains(t, err.Error(), "no row ids provided")
+		require.ErrorIs(t, err, ErrSchemaAttrCacheUnavailable)
 	})
 }
--- a/internal/cdc/init.go
+++ b/internal/cdc/init.go
@@ -275,7 +275,8 @@ func getSchemaIDsToInit(ctx context.Context, db *sql.DB, schemaRegistryTable str
 // initSchema exports all existing data for a single schema to S3 base files.
 // The delta tier is inventoried first (#371): a non-empty delta tier refuses
 // the schema unless the run may replace it, and with that permission the
-// inventory is purged only after the manifest swap has committed.
+// inventory is purged only after the manifest swap has committed — for a
+// schema with zero live rows that swap is an empty one (#519).
 func initSchema(ctx context.Context, runCtx *initRunContext, schemaID int16) (int64, int, error) {
 	inventory, err := preflightDeltaTier(ctx, runCtx, schemaID)
 	if err != nil {
@@ -287,14 +288,11 @@ func initSchema(ctx context.Context, runCtx *initRunContext, schemaID int16) (in
 		return 0, 0, err
 	}
 	if state == nil {
-		if !inventory.empty() {
-			// No live rows means no base swap, and the purge is ordered
-			// strictly after the swap, so the delta tier stays in place.
-			runCtx.logger.Warn("schema has no live rows; delta tier left in place because there is no base swap to publish",
-				zap.Int16("schema_id", schemaID),
-				zap.Int("delta_objects", len(inventory.purgeKeys())))
+		state, err = finishEmptySchema(ctx, runCtx, schemaID, inventory)
+		if err != nil || state == nil {
+			return 0, 0, err
 		}
-		return 0, 0, nil
+		return state.rowsExported, state.filesCreated, nil
 	}
 	state.deltaPurge = inventory.purgeKeys()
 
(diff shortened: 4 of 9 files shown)


iceboundrock:
## Review of #544 — `--replace-delta` retires emptied schema via zero-row base

### Summary
This PR correctly fixes #519 (follow-up of #518 finding F1 / #371): `initSchema` no longer early-returns on zero live rows under `--replace-delta`. Instead `finishEmptySchema` (`internal/cdc/init_delta.go:309`) exports one zero-row base through the shared `exportBatch` path and publishes it as the sole base entry with delta emptied under the existing CAS `publishAndPurge`, then purges delta objects. The deviation from the #519 plan (empty entry set → zero-row base) is load-bearing and well-evidenced: an empty manifest falls back to the per-schema glob (`internal/manifest/query_source.go:86-92`, pinned by `TestQuerySource_EmptyManifestFallsBackToGlob`) and resurrects deleted rows, while a `RowCount 0` base keeps the compaction invariant (`internal/compaction/merge_sql.go:104-106`). E2E caught the resurrection (all 6 rows), fix verified end-to-end per PR body. Overall design is sound; one crash-level finding (F1) needs a guard before merge, rest are nits/observations.

### Spec compliance (vs #519 + plan deviation comment)
- **Deviation justified and documented**: plan Rulings 2/3 superseded explicitly on #519 and in PR body. Zero-row base (`BuildMergedBasePath`, `internal/cdc/manifest.go:89-91`) avoids `ClassUnknown` range-less init stem (`internal/reconcile/classify.go:68-78`) and keeps one in-prefix base entry so #463 guard (`internal/reconcile/gc.go:33-36`, `manifestEntriesInPrefix > 0 → nil`) does not trip — docs bullet (`docs/manifest-reconcile.md:524`) correctly says no `--allow-empty-manifest-schema` needed. Verified against code.
- **#371 ordering preserved**: `export → save → delete` asserted in `TestFinishEmptySchema_ReplaceDeltaExportsZeroRowBaseThenSwapsThenPurges` (`internal/cdc/init_empty_schema_test.go:92`: `export, save, delete unlisted, delete listed`), failed-save deletes nothing, failed-export publishes nothing. Same `publishAndPurge` as populated path (`internal/cdc/init.go:320-325`), including 412 re-splice + `confirmAmbiguousSwap` — correct reuse.
- **#416/#203 respected**: old base unlisted, never deleted; e2e asserts `oldBaseKeys` present, `deltaKeys` absent, `assertBaseRows(...,0)` reads physical parquet (not just manifest), federated result empty, plain init succeeds again with 3 rows. Idempotency claim holds.
- **`updateSchemaManifest` never publishes empty set** (`internal/cdc/init_publish.go:86-88` + doc `78-85`), flag or no flag, pinned by `TestUpdateSchemaManifest_EmptyEntriesNeverPublished`. Correct — empty would re-trigger glob fallback.
- **Base export `FALSE` filter** (`internal/cdc/duckdb_exporter.go:140-145`, `internal/cdc/export_sql_builder.go:66-88`): `ltbase_schema_id = 1 AND ltbase_deleted_at IS NULL AND FALSE` + `schema_id = 1 AND FALSE`, full projection unchanged (`selectClause` equality test). Snapshot path still refuses empty (`errors.New("export snapshot: no row ids provided")`), base path now legitimate. `TestExportToTmpWrapsPlanErrors` update (base empty + nil cache → `ErrSchemaAttrCacheUnavailable`) is correct.
- **Nothing-to-retire + unlisted-delta-only cases**: `listed==0 && purge==0 → nil,nil` (no manifest minted, `IsNotFound` asserted); unlisted-only still swaps so `--repair` cannot re-adopt. Matches revised Ruling 3. `--repair` promotion guard (`internal/reconcile/promote.go:56-57`, `liveCount==0` refuses init-shaped) cited correctly — old init-shaped bases cannot return.
- **Without flag unchanged**: `finishEmptySchema` early `nil,nil` when `!replaceDelta`, plain re-init still `ErrDeltaTierPresent` in e2e. Good.

### Findings
**F1 — `finishEmptySchema` nil-derefs when `manifestStore == nil` (should-fix, regression vs main).**
`internal/cdc/init_delta.go:313` calls `countListedManifestEntries → loadInitManifest → manifest.LoadOrCreate(ctx, nil, ...)` which panics on `st.Load` (`internal/manifest/manifest.go:53`). Reproduced locally on this branch with `manifestStore:nil, replaceDelta:true, empty inventory`:
```
panic: runtime error: invalid memory address ... manifest.Load (.../manifest.go:53)
  at loadInitManifest (init_publish.go:38) → countListedManifestEntries (init_delta.go:346) → finishEmptySchema (init_delta.go:313)
```
On `main` this path was a no-op (`return 0,0,nil`) and never touched the store. Reachable when `ManifestTemplate == ""` (so `applyInitS3Wiring` leaves `manifestStore nil`, `internal/cdc/init_wiring.go:37-46`) + zero live rows + `--replace-delta`: `preflightDeltaTier` returns empty inventory without the `manifestStore==nil` gate (`init_delta.go:133-142` only gates non-empty), then `finishEmptySchema` panics instead of no-op/error. `PathResolver.Resolve` with empty template still succeeds (defaults to `manifest/{{.SchemaID}}.json`, `resolver.go:19-20`), so it proceeds to the nil load. Suggested fix: early `if runCtx.manifestStore == nil { return nil, nil }` (nothing manifest-backed to retire; consistent with `updateSchemaManifest` nil-guard) or return a wrapped error naming the missing template like the preflight `manifestStore==nil` branch. Add a unit test for nil-store + flag.

**F2 — Stale doc comment in `init.go` (nit).**
`internal/cdc/init.go:275-279`: “for a schema with zero live rows that swap is an empty one (#519)” — now it is a one-file zero-row base swap, never empty. Update to match `init_publish.go:78-85` wording to avoid future confusion with the superseded plan.

**F3 — Bare `return err` in helper (nit, standards).**
`internal/cdc/init_delta.go:347-348`: `return 0, err` from `loadInitManifest` is a bare pass-through. Outer caller wraps (`retire tiers ...: %w`), and `loadInitManifest` already carries `load manifest %s for schema %d`, so no info lost — but `coding-standard.md` asks for wrap at every return. Either `fmt.Errorf("load manifest entries ...: %w", err)` or a comment noting intentional pass-through.

No other blockers. No bare-sentinel wraps (write path uses `fmt.Errorf(...: %w)` with schema context; read-path drift errors stay plain — consistent with `docs/error-handling.md`).

### Standards checks (`coding-standard.md` as authority)
- **File ≤500, func ≤100**: `init_delta.go` 365, `init.go` 486, `duckdb_exporter.go` 372, `init_publish.go` 198, `init_empty_schema_test.go` 226 — all pass. Longest touched funcs: `exportBatch` ~47 lines (`init.go:444`), `buildExportSQLPlan` ~59 (`duckdb_exporter.go:126`), `finishEmptySchema` ~28, `purgeDeltaTier` ~33 (nesting ≤3, early returns, flat). New logic correctly placed in `init_delta.go` keeping `init.go` net-neutral.
- **Error wrapping**: `finishEmptySchema` wraps both export and publish failures with `schema %d`; `preflight`/`refuseUnpurgeable` name counts/keys; `F3` only exception. `TestNoBareSentinelWraps` passes (verified).
- **No AI attribution** in commits/PR text — conforms to `AGENTS.md`.
- **Test placement**: unit tests next to code, e2e carries `//go:build e2e` — conforms.

### Assessment
- **Request one change (F1)** before merge: nil-store guard + regression test. F2/F3 can ride along in same push.
- Otherwise **LGTM**: spec deviation is correct and proven by e2e, shared-path reuse minimizes drift, coverage of ordering/failure/dry-run/nothing-to-retire/unlisted-only is thorough, docs accurately describe GC/repair interaction.
- Do not merge until F1 fixed; no other merge blockers. No scope creep requested.

### Observations (non-blocking, positive + edge notes)
- Key-shape choice is the subtle win: `base-{uuid}` → `ClassBaseMerged` (GC-eligible) vs range-less init stem → `ClassUnknown` (never reclaimed). Comment in `finishEmptySchema` states this explicitly — good operator-facing reasoning.
- Dry-run semantics are coherent: `exportBatch` dry-run counts 1 file/0 rows without touching S3, `updateSchemaManifest` early-returns on `dryRun`, `purgeDeltaTier` only logs. Unit + e2e both assert `RowsExported==0 && FilesCreated==1` and state unchanged.
- `attrCache` taken from `runCtx.attrCaches[schemaID]` (pre-flight #193) without existence check — nil cache fails closed via `ErrSchemaAttrCacheUnavailable` in plan builder. Acceptable; a missing-cache error already names schema.
- 412/ambiguous-save for zero-row path has no dedicated unit test, but shares `publishInitTiers`/`swapCommitted` (path-set equality + delta-empty) with populated path — coverage by existing #371 tests is sufficient; consider noting in PR body if reviewers ask.
- E2E does not exercise an attribute-type-change pin scenario explicitly (issue consequence 4), only tombstone retirement + empty federated read. Retirement is the prerequisite for unpinning; pin behavior itself is covered by #518 tests. Sufficient for #519 scope.

### Verification (reviewer-run on `pr-544-review` @ `72da447`)
- `go test ./internal/cdc -run 'TestFinishEmptySchema|TestUpdateSchemaManifest_EmptyEntriesNeverPublished|TestBuildBaseExportSQL_EmptyRowIDs|TestExportToTmpWrapsPlanErrors' -v -count=1` — **PASS** (8/8).
- `go test ./internal/cdc ./internal/compaction ./internal/manifest -count=1` — **PASS**.
- `go test . ./cdc ./cmd/... ./factory ./internal/cdc ./internal/compaction ./internal/manifest ./internal/reconcile ./internal/federated ./internal/sqlgen -count=1` — **PASS**; full `./internal/...` only fails at `internal/e2e_harness` `TestE2EHarnessMinimal` due to no container runtime in this env (testcontainers `NewDockerClient` panic), same as upstream without `DOCKER_HOST` — unrelated to this change.
- `make fmt-check` — clean. `make lint` (v1.64.8, root + `infra/`) — clean.
- `go test . -run TestNoBareSentinelWraps` — **PASS**.
- Production e2e `TestInitReplaceDeltaRetiresEmptiedSchema` **not re-run here** (needs Docker/rootless Podman + `-tags=e2e`); author reports PASS under rootless Podman with `TestCDCFlushAndInit`, plus `assertBaseRows(0)` physical-row check. Trust but verify in CI (`ci.yml` production harness + nightly full suite).
- F1 panic reproduced via temporary `zz_tmp_nilstore_test.go` (removed after) — see stack above.

### Summary
Approve direction, block on F1 nil-store guard. Once `finishEmptySchema` handles `manifestStore==nil` (no-op or explicit error) with a test, plus F2 wording fix, this is ready to squash-merge with the existing clean message (`fix(cdc): --replace-delta retires the tiers of a schema with zero live rows (#519)`). No AI attribution present. Thanks for catching the glob-resurrection in e2e and documenting the superseded rulings on #519 — that note is exactly what made this reviewable.

iceboundrock:
## Response to review — fixes in `b48534f`

### F1 — nil `manifestStore` panic in `finishEmptySchema` (fixed)

Confirmed by reading and reproduced: with `ManifestTemplate == ""` the store is nil, `PathResolver{}` still resolves the default path, and `manifest.LoadOrCreate` dereferences the nil store at `manifest.go:53`. The pre-flight's storeless refusal only covers a non-empty inventory, so an empty one fell through to the load.

Fix: `finishEmptySchema` returns `nil, nil` (with an Info line) when `runCtx.manifestStore == nil`, before any manifest access. No-op rather than error, for two reasons:

- The only way to reach the branch storeless is with an empty inventory (a non-empty one is already refused by the pre-flight with the "needs a configured manifest template" error). Empty inventory plus no manifest means there is nothing to retire.
- It matches the populated path: `updateSchemaManifest` no-ops on a nil store, so a storeless `--replace-delta` run over a populated schema exports and publishes nothing manifest-wise. The emptied schema now behaves the same instead of panicking.

Regression test `TestFinishEmptySchema_NoManifestStoreIsNoOp` (`internal/cdc/init_empty_schema_test.go`) mirrors production wiring (nil store, zero-value resolver), runs the pre-flight first to prove it lets the empty inventory through, and asserts nil state, no error, no export/save/delete events. It panicked with the reviewer's stack before the guard and passes after.

### F2 — stale `initSchema` doc comment (fixed)

`internal/cdc/init.go` now says the zero-row swap publishes one zero-row base object as the whole base tier, never an empty entry set, and points at `finishEmptySchema`.

### F3 — bare pass-through in `countListedManifestEntries` (fixed)

Wrapped as `count listed manifest entries: %w`. `loadManifestDeltaEntries` in the same file had the identical pass-through from #371, so it is wrapped too (`load manifest delta entries: %w`). No test asserts on the exact prefix; the `resolve manifest path` assertions are on a different return.

### Observations

- 412 / ambiguous-save on the zero-row path: agreed that `publishInitTiers` and `confirmAmbiguousSwap` are shared with the populated path and covered by the #371 tests; no dedicated test added.
- Attribute-type-change pin scenario in e2e: agreed it is #518's coverage; retirement is the prerequisite and is what this PR asserts.
- No follow-up issue is needed from this review; nothing raised is out of scope or left unaddressed.

### Verification

```
go test ./internal/cdc ./internal/compaction ./internal/manifest ./internal/reconcile -count=1   # ok
go test . -run TestNoBareSentinelWraps -count=1                                                 # ok
make fmt-check && make lint                                                                     # clean
```

`init_delta.go` is at 374 lines, `init.go` at 488; the production e2e was not re-run for this push since the change is a guard before any S3 access.

iceboundrock:
## Review of #544

### Summary
Reviewed the latest head, `b48534f08bfbd7b80e1987af4dffe69d2fc52175`. This correctly fixes #519: an emptied schema under `--replace-delta` exports and lists one zero-row base before delta objects are purged, rather than publishing an empty manifest that would trigger the legacy glob fallback and resurrect unlisted rows.

### Findings
No blocking or non-blocking correctness findings in the current head. The prior nil-`manifestStore` panic is fixed by the early no-op in `finishEmptySchema` (`internal/cdc/init_delta.go:317-320`) and is covered by `TestFinishEmptySchema_NoManifestStoreIsNoOp`.

### Spec Compliance
- The revised #519 design is implemented consistently: zero-row base export uses `FALSE` source filters with the normal schema projection (`internal/cdc/duckdb_exporter.go:126-183`), then follows the shared export -> CAS manifest swap -> delta purge ordering (`internal/cdc/init_delta.go:339-345`).
- The zero-row entry uses `BuildMergedBasePath` (`internal/cdc/init_delta.go:366-374`), so it remains manifest-listed, avoids glob fallback, and leaves superseded bases as GC-eligible merged-base orphans.
- No manifest is minted when there are neither existing manifest entries nor delta objects (`internal/cdc/init_delta.go:326-330`); an unlisted delta object still causes a zero-row swap and post-commit purge.
- `--replace-delta` with no manifest template is safely a no-op only for an empty inventory. The existing preflight still refuses a non-empty inventory without a manifest-backed swap.
- The production E2E covers initial export, all-row deletion plus flush, plain-init refusal, dry-run immutability, zero-row swap, object-retention/deletion expectations, federated empty result, and a subsequent successful plain init.

### Standards
- Touched production files remain below the 500-line limit; the added helpers are below the 100-line function limit.
- Error paths in the changed helpers carry context; the previous bare pass-throughs in the manifest-count/load helpers are now wrapped.
- The change reuses the existing export and publish paths instead of duplicating CAS or purge behavior.
- Tests are colocated with the CDC code, and the production test has the required `e2e` build tag.

### Assessment
LGTM. The implementation is narrowly scoped, preserves the write-once and save-before-delete invariants, and handles the subtle empty-manifest fallback correctly. It is ready to merge once required CI checks complete successfully.

### Observations
- Keeping a physical zero-row parquet file is load-bearing, not cosmetic: it preserves the manifest-driven read set while retaining a full footer schema for validation.
- The merged-base key shape is the correct choice for a range-less object and aligns init with the compactor's all-tombstone invariant.
- No additional follow-up issue is warranted from this review.

### Verification
Reviewer-run at `b48534f`:

```text
GOCACHE=/home/ruoshi/code/lychee/forma/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto GOTMPDIR=/home/ruoshi/code/lychee/forma/.gocache/tmp go test ./internal/cdc ./internal/compaction ./internal/manifest -count=1  # PASS
GOCACHE=/home/ruoshi/code/lychee/forma/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto GOTMPDIR=/home/ruoshi/code/lychee/forma/.gocache/tmp go test . -run TestNoBareSentinelWraps -count=1  # PASS
make fmt-check  # PASS
git diff --check origin/main...HEAD  # PASS
```

GitHub Actions currently reports `Lint` and `Test` passing. `E2E Smoke Tests` was still pending when reviewed; the production E2E was not rerun locally because it requires the container runtime.

Issue #531, opened 2026-09-06
sqlgen/federated: dedup-column set and attr placeholder are duplicated across packages with no sync guard


iceboundrock:
Follow-up from the review of #530 (finding F4), which fixed #512.

## Problem

Two definitions exist twice, once per package, with no mechanism keeping them equal:

| `internal/sqlgen/parquet_attr_column.go` | `internal/federated/keyset.go` |
| --- | --- |
| `federatedDedupColumns` (`rn`, `source_tier_priority`) | `keysetRejectedColumns` (`rn`, `source_tier_priority`) |
| `parquetAttrPlaceholder = "attr"` | `parquetAttrFallbackColumn = "attr"` |

`sqlgen.ValidateUnregisteredParquetAttrColumn` (filters, #512) and `federated.validateKeysetCursor` (cursors, #509) implement the same fold rule against these sets. The split is forced by layering: `federated` imports `sqlgen`, not the other way round. Both sites comment their counterpart, but if the dedup machinery ever grows a column (or the placeholder ever changes in `ParquetAttrColumn`), one guard can be updated and the other silently left behind, reopening the #354 silent-wrong-answer class on whichever path was missed.

## Proposed fix

Cheapest: export the `sqlgen` definitions (`FederatedDedupColumns`, `ParquetAttrPlaceholder`) and have `federated` alias them instead of redeclaring. Alternative if exporting is undesirable: a small test in `internal/federated` asserting set equality with the `sqlgen` set, so drift fails CI.

Also worth considering while there: `parquetAttrPlaceholder` should be the single source for the `"attr"` literal inside `ParquetAttrColumn` itself, which currently returns the bare string.

## Scope

Maintainability only. No behaviour change today, both sets are identical.

iceboundrock:
## Implementation plan (#531)

Artifact: implementation plan, produced before touching code. Working copy lives in the session scratchpad; this comment is the record.

### What is duplicated today

| `internal/sqlgen/parquet_attr_column.go` | `internal/federated/keyset.go` |
| --- | --- |
| `federatedDedupColumns` (var, unexported) | `keysetRejectedColumns` (var, unexported) |
| `parquetAttrPlaceholder = "attr"` (const) | `parquetAttrFallbackColumn = "attr"` (const) |
| `ParquetAttrColumn` returns the bare literal `"attr"` | — |

Both packages judge the same folded name against the same two facts; `federated`
already imports `sqlgen`, so the dependency direction allows a single source.

[...]

PR #548, opened 2026-09-08
fix(sqlgen): reject case-variant parquet system columns (#532)

### LINKED ISSUE #532: sqlgen: ValidateParquetAttrColumns reserved-column check is case-sensitive, so a registered Created_At binds to created_at in DuckDB
Follow-up from the review of #530 (finding F5), pre-existing, not introduced by that PR.

## Problem

`ValidateParquetAttrColumns` (`internal/sqlgen/parquet_attr_column.go`) is the registration-time guard that rejects an attribute whose folded parquet column lands on a reserved system column. It looks the folded name up verbatim:

```go
col := ParquetAttrColumn(name)
if _, ok := reservedParquetColumns[col]; ok { ... }
```

`ParquetAttrColumn` preserves case, and `reservedParquetColumns` keys are lower-case, so `Created_At`, `Row_ID`, `RN`, `Source_Tier_Priority` and any other case variant pass registration. DuckDB resolves unquoted identifiers case-insensitively, so at read time:

- a registered `Created_At` attribute filter or projection binds to the unified-CTE system column `created_at`, not to the attribute's parquet column: the attribute silently reads the creation timestamp (the #354 silent-wrong-answer family);
- a registered `Row_ID` collides with the export column `row_id` in the parquet SELECT list, or shadows it, depending on the binder; either way the attribute is unusable and the failure mode is not a registration error;
- a registered `RN` binds to the dedup rank.

The two newer guards already lower-case before the lookup: `federated.validateKeysetCursor` (#509) and `sqlgen.ValidateUnregisteredParquetAttrColumn` (#512). Registration is the one site left with the verbatim lookup, and it is the site both newer guards rely on ("registered names never reach the check: `ValidateParquetAttrColumns` already rejects them at registration"). That reliance is only sound once this check is case-insensitive too.

## Proposed fix

Lower-case the folded name before the `reservedParquetColumns` lookup, and do the same for the intra-schema collision map (`colToAttr`), since `contact_name` and `Contact_Name` also resolve to the same DuckDB column. Add registration tests for the case variants. The error message should keep the caller's spelling.

## Compatibility

Any schema already registered with a case-variant reserved name is broken today (it reads the wrong column), so refusing it at re-registration is not a regression. Worth a note in the release notes. The CDC writer and federated reader call `ValidateParquetAttrColumns` as defence in depth, so an existing schema with such an attribute would start failing there as well; that is the correct fail-closed outcome, but the operator-facing message should say what to rename.

iceboundrock:
Closes #532.

## Summary

- Normalize folded parquet column names before reserved-system-column and intra-schema collision checks, using `duckdbFoldIdentifier` — an ASCII-only fold matching what DuckDB actually does, pinned against the engine by `TestDuckDBIdentifierFoldIsASCIIOnly`. Go's `strings.ToLower` is the wrong primitive: it merges `Á`/`á`, `Ж`/`ж`, `cafÉ`/`café` and U+212A/`k`, which DuckDB keeps distinct, and maps U+0130 onto `i`, reading the legitimate `row_İd` as the reserved `row_id`.
- Preserve caller spelling in every diagnostic and retain the byte-stable `ParquetAttrColumn` naming contract.
- Cover case-variant system names and aliases at both the validator and file-schema registration seams.
- Name the resolved lower-case column in both diagnostics, so a case-variant rejection is greppable against `reservedParquetColumns` and the collision message never claims a shared parquet column that the byte-stable fold does not produce.
- Record the registration guard's case-insensitivity, and its ASCII boundary, in `docs/federated-query/design.md` §4.4, which already states a premise ("registered names never reach it") that only this change makes true for case variants.

## Compatibility / release note

A schema using a case-variant reserved name, such as `Created_At`, is already broken: DuckDB resolves its unquoted identifier as the `created_at` system column. This change fails closed during registration, CDC export, and federated projection rather than allowing a silent wrong answer. No parquet column naming or persisted data changes.

Two properties of that fail-closed behavior are worth stating for operators, both pre-existing for exact-case names and only widened here:

- **A single offending schema fails the whole registry.** `validateSchemaAttributeCache` runs inside the per-schema load loop and every caller returns on first error, so one legacy schema carrying `Created_At` fails registry construction for the entire directory or registry table at startup, not just its own entity.
- **Retired ledger entries are in scope.** The validator deliberately runs over the full cache before retired entries are stripped (#342). A *retired* case-variant name produces no wrong answers today — no consumer ever sees it — yet it will now block boot, and "rename the attribute" is not an actionable remedy for a ledger record that exists to preserve a historical name. Tracked in #549.

## Scope

The same case-insensitivity premise backs two guards this PR does not change, both still folding with `strings.ToLower`: `ValidateUnregisteredParquetAttrColumn` (#512/#530) and `federated.validateKeysetCursor` (#381/#509). Both are runtime guards over caller-supplied strings, where being narrower than DuckDB refuses a query rather than failing registration, and both predate this PR. Tracked in #550.

## Verification

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/schemameta`
- `make fmt-check`
- `git diff --check`
- `make lint`
- `./scripts/test_with_container_runtime.sh` (complete `make test` suite using rootless Podman) — re-run green after the ASCII-fold commit

The red-to-green regression evidence and full verification record are posted on #532.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -185,6 +185,20 @@ resolved lower-case column beside it. This is what makes the premise stated
 under the filter rule below — "registered names never reach it" — hold for
 case variants and not merely for exact-case ones.
 
+That case-insensitivity is **ASCII-only**, because DuckDB's is: the engine
+folds `A`–`Z` and nothing else, so `Á` and `á`, `Ж` and `ж`, and `cafÉ` and
+`café` are distinct identifiers to it, and U+212A KELVIN SIGN does not
+resolve onto `k`. The guard therefore keys on `duckdbFoldIdentifier`, not on
+`strings.ToLower`: Go's Unicode case mappings would merge every one of those
+pairs and reject a schema DuckDB serves correctly, and because they also map
+U+0130 (`İ`) onto `i`, they would read the legitimate attribute `row_İd` as
+the reserved `row_id`. Since a registration failure fails the whole registry,
+a false rejection here is a boot failure, so the boundary is pinned against
+the engine itself rather than its documentation
+(`TestDuckDBIdentifierFoldIsASCIIOnly`). The two runtime guards below still
+fold with `strings.ToLower` and are narrower than DuckDB in the same way;
+that is tracked in #550.
+
 Keyset cursor columns obey the same contract as every other column reference,
 and it is one contract, not a per-seam one (#381). A single validator,
 `federated.validateKeysetCursor`, binds every entry point onto the keyset
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -174,7 +174,16 @@ generator (read side) and cannot diverge: the logical WHERE clause is applied
 both against raw `read_parquet` output (physical columns) and against the
 `visible` CTE (unified columns), so both must expose the same names (#260).
 The fold is lossy, so both sides fail fast if two attributes of one schema
-collide on the same folded column.
+collide on the same folded column. `sqlgen.ValidateParquetAttrColumns`, the
+registration guard behind that, compares **case-insensitively** for the same
+reason the keyset and filter guards below do: DuckDB resolves an unquoted
+identifier without regard to case, so `Created_At` reaches the `created_at`
+system column and `Contact_Name` reaches the same column as `contact_name`
+(#532). The emitted names stay byte-stable either way — `ParquetAttrColumn`
+preserves the caller's spelling, and so do the diagnostics, which name the
+resolved lower-case column beside it. This is what makes the premise stated
+under the filter rule below — "registered names never reach it" — hold for
+case variants and not merely for exact-case ones.
 
 Keyset cursor columns obey the same contract as every other column reference,
 and it is one contract, not a per-seam one (#381). A single validator,
--- a/internal/schemameta/lookup_test.go
+++ b/internal/schemameta/lookup_test.go
@@ -161,6 +161,24 @@ func TestFileSchemaRegistryRejectsReservedFoldedColumnAtLoad(t *testing.T) {
 	assert.Contains(t, err.Error(), "reserved")
 }
 
+// TestFileSchemaRegistryRejectsCaseVariantReservedColumnAtLoad pins #532 at
+// the registration seam: Created_At resolves to the created_at system column
+// in DuckDB and must not be accepted into a schema cache.
+func TestFileSchemaRegistryRejectsCaseVariantReservedColumnAtLoad(t *testing.T) {
+	dir := t.TempDir()
+	writeJSONFile(t, fmt.Sprintf("%s/created.json", dir), map[string]any{
+		"type": "object",
+	})
+	writeJSONFile(t, fmt.Sprintf("%s/created_attributes.json", dir), map[string]any{
+		"Created_At": map[string]any{"attributeID": float64(1), "valueType": "text"},
+	})
+
+	_, err := NewFileSchemaRegistryFromDirectory(dir)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "Created_At")
+	assert.Contains(t, err.Error(), "reserved")
+}
+
 // TestFileSchemaRegistryRejectsFoldedAttrCollisionAtLoad: two attributes
 // whose folded parquet columns collide are likewise rejected at load.
 func TestFileSchemaRegistryRejectsFoldedAttrCollisionAtLoad(t *testing.T) {
@@ -177,4 +195,22 @@ func TestFileSchemaRegistryRejectsFoldedAttrCollisionAtLoad(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "contact.name")
 	assert.Contains(t, err.Error(), "contact_name")
+}
+
+// TestFileSchemaRegistryRejectsCaseVariantFoldedAttrCollisionAtLoad pins
+// #532: DuckDB treats these two physical parquet aliases as one column.
+func TestFileSchemaRegistryRejectsCaseVariantFoldedAttrCollisionAtLoad(t *testing.T) {
+	dir := t.TempDir()
+	writeJSONFile(t, fmt.Sprintf("%s/fold.json", dir), map[string]any{
+		"type": "object",
+	})
+	writeJSONFile(t, fmt.Sprintf("%s/fold_attributes.json", dir), map[string]any{
+		"contact_name": map[string]any{"attributeID": float64(1), "valueType": "text"},
+		"Contact_Name": map[string]any{"attributeID": float64(2), "valueType": "text"},
+	})
+
+	_, err := NewFileSchemaRegistryFromDirectory(dir)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "contact_name")
+	assert.Contains(t, err.Error(), "Contact_Name")
 }
--- a/internal/sqlgen/parquet_attr_column.go
+++ b/internal/sqlgen/parquet_attr_column.go
@@ -93,6 +93,13 @@ var federatedDedupColumns = map[string]struct{}{
 // "Created_At") is admitted and left to fail at the binder or, on the
 // federated route, at the PG EAV payload. Caller fault, so the error is a
 // forma.InvalidInputf carrier that keeps the caller's spelling.
+//
+// The case folding below is still strings.ToLower, so this guard is narrower
+// than DuckDB for non-ASCII names — it can refuse a filter the engine would
+// have bound. That over-rejects a query rather than answering one wrongly,
+// and unlike ValidateParquetAttrColumns it cannot fail registration, so it
+// is tracked separately in #550 rather than changed here; see
+// duckdbFoldIdentifier.
 func ValidateUnregisteredParquetAttrColumn(attr, folded string) error {
 	if strings.EqualFold(folded, parquetAttrPlaceholder) && !strings.EqualFold(attr, parquetAttrPlaceholder) {
 		return forma.InvalidInputf(
@@ -113,13 +120,45 @@ func ValidateUnregisteredParquetAttrColumn(attr, folded string) error {
 	return nil
 }
 
+// duckdbFoldIdentifier lower-cases a folded column name the way DuckDB
+// resolves an unquoted identifier: ASCII only. strings.ToLower is the wrong
+// primitive here — it applies Unicode case mappings DuckDB does not, so it
+// merges identifiers DuckDB keeps distinct and would reject valid schemas
+// (#532). Verified against the pinned DuckDB (see
+// TestDuckDBIdentifierFoldIsASCIIOnly, which asserts this function against
+// the engine rather than against the documentation): "Á"/"á", "Ж"/"ж",
+// "ẞ"/"ß" and "cafÉ"/"café" are distinct columns, and U+212A KELVIN SIGN
+// does not resolve onto "k". The reserved half matters too: Go maps U+0130
+// LATIN CAPITAL LETTER I WITH DOT ABOVE onto "i", so a Unicode fold would
+// collapse the legitimate attribute "row_İd" onto the reserved "row_id" and
+// fail registry construction for every schema in the directory.
+func duckdbFoldIdentifier(col string) string {
+	var folded []byte
+	for i := 0; i < len(col); i++ {
+		c := col[i]
+		if c < 'A' || c > 'Z' {
+			continue
+		}
+		if folded == nil {
+			folded = []byte(col)
+		}
+		folded[i] = c + ('a' - 'A')
+	}
+	if folded == nil {
+		return col
+	}
+	return string(folded)
+}
+
 // ValidateParquetAttrColumns rejects attribute sets whose folded parquet
 // column names land on a reserved system column or collide with each other
 // (the fold is lossy: "contact.name" and "contact_name" both become
 // contact_name). DuckDB resolves unquoted identifiers case-insensitively, so
-// comparisons use lower-cased folded names while errors preserve the caller's
-// spelling and name the resolved column beside it, so a case-variant
-// rejection stays legible against the reserved set an operator can grep.
+// comparisons run on duckdbFoldIdentifier keys — ASCII-only, matching the
+// engine, so non-ASCII names DuckDB keeps distinct are not merged here —
+// while errors preserve the caller's spelling and name the resolved column
+// beside it, so a case-variant rejection stays legible against the reserved
+// set an operator can grep.
 // Schema registration calls it so an unusable schema is
 // rejected before it accepts hot-tier writes; the CDC writer and the
 // federated reader call it again as defense in depth. Plain operator
@@ -139,7 +178,7 @@ func ValidateParquetAttrColumns(cache forma.SchemaAttributeCache) error {
 	colToAttr := make(map[string]foldedAttr, len(names))
 	for _, name := range names {
 		col := ParquetAttrColumn(name)
-		key := strings.ToLower(col)
+		key := duckdbFoldIdentifier(col)
 		if _, ok := reservedParquetColumns[key]; ok {
 			return reservedParquetColumnError(name, col, key)
 		}
--- a/internal/sqlgen/parquet_attr_column.go
+++ b/internal/sqlgen/parquet_attr_column.go
@@ -118,7 +118,9 @@ func ValidateUnregisteredParquetAttrColumn(attr, folded string) error {
 // (the fold is lossy: "contact.name" and "contact_name" both become
 // contact_name). DuckDB resolves unquoted identifiers case-insensitively, so
 // comparisons use lower-cased folded names while errors preserve the caller's
-// spelling. Schema registration calls it so an unusable schema is
+// spelling and name the resolved column beside it, so a case-variant
+// rejection stays legible against the reserved set an operator can grep.
+// Schema registration calls it so an unusable schema is
 // rejected before it accepts hot-tier writes; the CDC writer and the
 // federated reader call it again as defense in depth. Plain operator
 // error, never forma.ErrInvalidInput. Attributes are checked in sorted
@@ -130,21 +132,55 @@ func ValidateParquetAttrColumns(cache forma.SchemaAttributeCache) error {
 	}
 	sort.Strings(names)
 
-	colToAttr := make(map[string]string, len(names))
+	type foldedAttr struct {
+		name string
+		col  string
+	}
+	colToAttr := make(map[string]foldedAttr, len(names))
 	for _, name := range names {
 		col := ParquetAttrColumn(name)
 		key := strings.ToLower(col)
 		if _, ok := reservedParquetColumns[key]; ok {
-			return fmt.Errorf(
-				"attribute %q folds to parquet column %q, which is reserved for system columns; rename the attribute",
-				name, col)
+			return reservedParquetColumnError(name, col, key)
 		}
 		if prev, ok := colToAttr[key]; ok {
-			return fmt.Errorf(
-				"attributes %q and %q both map to parquet column %q; attribute names must remain distinct after identifier folding",
-				prev, name, col)
+			return foldedColumnCollisionError(prev.name, prev.col, name, col, key)
 		}
-		colToAttr[key] = name
+		colToAttr[key] = foldedAttr{name: name, col: col}
 	}
 	return nil
+}
+
+// reservedParquetColumnError explains a reserved-column rejection. When the
+// folded name is already lower case it is itself the reserved entry and the
+// message says so plainly; when the caller's case differs, the message must
+// also name the reserved column the identifier resolves onto, because that
+// lower-cased name — not the caller's spelling — is what an operator finds in
+// reservedParquetColumns (#532).
+func reservedParquetColumnError(name, col, key string) error {
+	if col == key {
+		return fmt.Errorf(
+			"attribute %q folds to parquet column %q, which is reserved for system columns; rename the attribute",
+			name, col)
+	}
+	return fmt.Errorf(
+		"attribute %q folds to parquet column %q, which DuckDB resolves case-insensitively onto the reserved system column %q; rename the attribute",
+		name, col, key)
+}
+
+// foldedColumnCollisionError explains an intra-schema collision. Two
+// attributes that fold to the same string share one parquet column and the
+// message says exactly that; two that differ only in case keep distinct
+// parquet columns — ParquetAttrColumn is byte-stable — and are merged only by
+// DuckDB's case-insensitive identifier matching, so the message names both
+// columns rather than claiming a single one that no parquet footer holds.
+func foldedColumnCollisionError(prevName, prevCol, name, col, key string) error {
+	if prevCol == col {
+		return fmt.Errorf(
+			"attributes %q and %q both map to parquet column %q; attribute names must remain distinct after identifier folding",
+			prevName, name, col)
+	}
+	return fmt.Errorf(
+		"attributes %q and %q map to parquet columns %q and %q, which DuckDB resolves case-insensitively onto the same column %q; attribute names must remain distinct after identifier folding",
+		prevName, name, prevCol, col, key)
 }
--- a/internal/sqlgen/parquet_attr_column.go
+++ b/internal/sqlgen/parquet_attr_column.go
@@ -116,7 +116,9 @@ func ValidateUnregisteredParquetAttrColumn(attr, folded string) error {
 // ValidateParquetAttrColumns rejects attribute sets whose folded parquet
 // column names land on a reserved system column or collide with each other
 // (the fold is lossy: "contact.name" and "contact_name" both become
-// contact_name). Schema registration calls it so an unusable schema is
+// contact_name). DuckDB resolves unquoted identifiers case-insensitively, so
+// comparisons use lower-cased folded names while errors preserve the caller's
+// spelling. Schema registration calls it so an unusable schema is
 // rejected before it accepts hot-tier writes; the CDC writer and the
 // federated reader call it again as defense in depth. Plain operator
 // error, never forma.ErrInvalidInput. Attributes are checked in sorted
@@ -131,17 +133,18 @@ func ValidateParquetAttrColumns(cache forma.SchemaAttributeCache) error {
 	colToAttr := make(map[string]string, len(names))
 	for _, name := range names {
 		col := ParquetAttrColumn(name)
-		if _, ok := reservedParquetColumns[col]; ok {
+		key := strings.ToLower(col)
+		if _, ok := reservedParquetColumns[key]; ok {
 			return fmt.Errorf(
 				"attribute %q folds to parquet column %q, which is reserved for system columns; rename the attribute",
 				name, col)
 		}
-		if prev, ok := colToAttr[col]; ok {
+		if prev, ok := colToAttr[key]; ok {
 			return fmt.Errorf(
 				"attributes %q and %q both map to parquet column %q; attribute names must remain distinct after identifier folding",
 				prev, name, col)
 		}
-		colToAttr[col] = name
+		colToAttr[key] = name
 	}
 	return nil
 }
--- a/internal/sqlgen/parquet_attr_column_test.go
+++ b/internal/sqlgen/parquet_attr_column_test.go
@@ -1,6 +1,7 @@
 package sqlgen
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/lychee-technology/forma"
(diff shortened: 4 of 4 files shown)


iceboundrock:
# Review — PR #548 `fix(sqlgen): reject case-variant parquet system columns (#532)`

Artifact: code review. Working copy: local checkout of `fix/532-case-insensitive-parquet-columns` at `350b5bf`; this comment is the durable record.

**Verdict: approve.** The change is correct, minimal, and is exactly the fix #532 specified. Three non-blocking items below, none of which need to hold the merge.

---

## Findings

### F1 — Both error messages now fire on a case-insensitive match but describe a case-sensitive world (medium, diagnostics)

`internal/sqlgen/parquet_attr_column.go:137-146`. The comparison key is folded, the message is not, and the gap is visible to the operator:

- Reserved: `attribute "Created_At" folds to parquet column "Created_At", which is reserved for system columns; rename the attribute`. `Created_At` is **not** in `reservedParquetColumns` — an operator who greps the reserved set for the string the message just handed them finds nothing and cannot tell why the schema was refused. AGENTS.md requires the message to name "the expected state"; the expected state here is DuckDB's case-insensitive identifier equivalence, and the message never mentions it. #532 asked for the same thing in its Compatibility note ("the operator-facing message should say what to rename") — the *what* is there, the *why* is not.
- Collision: `attributes "Contact_Name" and "contact_name" both map to parquet column "contact_name"`. This is literally false for `Contact_Name`, whose parquet column is `Contact_Name` — the byte-stable naming contract documented in this very file (`ParquetAttrColumn`, lines 17-32). An operator who opens the parquet footer finds two distinct columns and concludes the message is wrong, when the real claim is that DuckDB resolves both onto one.

There is in-repo precedent for surfacing the normalized name: `federated.validateKeysetCursor` puts the lower-cased `key` into its remedy (`internal/federated/keyset.go:186`, "cursor on %s under its own name"). Suggested shape:

```go
"attribute %q folds to parquet column %q, which DuckDB resolves case-insensitively onto the reserved system column %q; rename the attribute"   // name, col, key
"attributes %q and %q both resolve to parquet column %q — DuckDB matches unquoted identifiers case-insensitively, so the folded names %q and %q are one column; attribute names must remain distinct after identifier folding"
```

Behavior is right; this is diagnostic quality only.

### F2 — `docs/federated-query/design.md` §4.4 is not updated (low, durable record)

§4.4 documents case-insensitivity explicitly for the two *newer* guards — the keyset set ("matched on the folded name **case-insensitively**", ~L196-201) and the filter guard ("admits an identity-up-to-case fold", ~L304-310) — but the paragraph covering the registration guard still says only "The fold is lossy, so both sides fail fast if two attributes of one schema collide on the same folded column" (~L176-177). Worse, L309-310 states the load-bearing premise "Registered names never reach it: `ValidateParquetAttrColumns` already rejected any whose fold collides with a reserved column" — **this PR is what makes that sentence true for case variants**, and nothing in `docs/` records it. AGENTS.md: anything that becomes a durable repository convention belongs in `docs/`, the issue comment records the thinking. One sentence in §4.4 closes it.

### F3 — The fail-closed blast radius is wider than the release note states (low, operability)

The PR body says the change "fails closed during registration, CDC export, and federated projection". Two things it does not say:

1. **One offending schema blocks every schema.** `validateSchemaAttributeCache` is called inside the per-schema load loop and every caller returns on first error — `internal/schemameta/loader.go:81`, `loader.go:360`, `internal/schemameta/file_registry.go:192`, `file_registry.go:389`. A single legacy schema carrying `Created_At` therefore fails registry construction for the whole directory / registry table at startup, not just its own entity.
2. **Retired ledger entries are in scope, and for them the remedy is unactionable.** The validator deliberately runs over the **full** cache before `activeAttributeCache` strips retired entries (#342). A *retired* `Created_At` is the one sub-case that is not "already broken today": `activeAttributeCache` removes it before any consumer sees it, so it produces no wrong answers — yet after this PR it hard-blocks boot, and "rename the attribute" is not a real option, since the entry exists precisely to preserve the historical name for #341 orphan detection and matches a column already sitting in flushed parquet.

This is pre-existing design for exact-case names and the PR only widens the trigger set, so it is not a defect in the change. It is worth a sentence in the release note so an operator hitting it at startup knows what they are looking at.

---

## Spec check (against #532)

| #532 asks for | Status |
| --- | --- |
| Lower-case the folded name before the `reservedParquetColumns` lookup | Done — `parquet_attr_column.go:136-137` |
| Do the same for the intra-schema `colToAttr` collision map | Done — `:142`, `:147` |
| Add registration tests for the case variants | Done — 2 in `internal/sqlgen`, 2 in `internal/schemameta` at the file-registry seam |
| Error message keeps the caller's spelling | Done — `name`/`col` are unfolded; asserted explicitly by the new sqlgen test |
| Do not change `ParquetAttrColumn` / persisted names | Done — untouched; `TestParquetAttrColumn` byte-stability pin unchanged |
| Operator-facing message should say what to rename | Partial — names the attribute, does not explain the case-insensitive equivalence (F1) |

Scope discipline is correct: the shared placeholder / dedup-column duplication is left to #531, as the plan comment on #532 said it would be.

Coverage of the seams the issue named is complete. I grepped every other consumer of the fold — `ParquetAttrColumn` callers, `reservedParquetColumns`, `federatedDedupColumns`, `keysetRejectedColumns` — and there is no remaining site that looks a folded name up verbatim against a lower-case set. `ValidateUnregisteredParquetAttrColumn` (#512) and `validateKeysetCursor` (#509) were already normalized; registration was the last one, exactly as #532 claimed.

## Standards check

- **Size**: `parquet_attr_column.go` is 150 lines, longest function 24 — well inside the 500/100 limits. Net production delta is +7/-4.
- **Error class**: plain `fmt.Errorf`, no sentinel wrap. Correct per `docs/error-handling.md` — a registration/read-path consistency failure is operator-visible, not a 4xx carrier. The file's own docstring already pins this ("Plain operator error, never `forma.ErrInvalidInput`") and it stays true. `TestNoBareSentinelWraps` unaffected.
- **Structure**: early returns preserved, no added nesting.
- **Comments**: the `ValidateParquetAttrColumns` docstring is updated in place to state the new invariant, at the density this file uses. Good — the reason survives the diff.
- **Git conventions**: single commit, conventional-commit subject with the issue ref, empty body, no AI attribution anywhere in the commit, title or description. Clean.
- **Non-code artifacts**: implementation plan and verification record are both posted on #532 in full, per AGENTS.md. Nothing load-bearing lives only in a local file.

## Observations

- **O1 — the invariant three other comments already asserted is now actually true.** `federated/keyset.go:75-77` ("rn and source_tier_priority are both in that reserved set"), `federated/cold_columns.go:47-48` ("the #260 reserved-column guard rejects such schemas at registration"), and `ValidateUnregisteredParquetAttrColumn`'s "registered names never went through it" premise were all case-blind before this commit. The change retroactively earns three existing comments.
- **O2 — a free side effect worth knowing about.** Lower-casing `colToAttr` also catches a case-variant collision against the `attr` placeholder: `Attr` and `[]` (which folds to `attr`) now collide. Correct, untested, no action needed.
- **O3 — `strings.ToLower` is Unicode-aware; DuckDB's identifier folding is ASCII.** A pair such as `K` (U+212A KELVIN SIGN) and `k` would be rejected here while DuckDB treats them as distinct. Over-rejection, i.e. fail-closed, and vanishingly unlikely. Noting it only so it is not rediscovered as a bug.
- **O4 — one assertion in the new sqlgen test is currently a duplicate.** `TestValidateParquetAttrColumns_CaseVariantReservedSystemColumn` asserts both `Contains(err, attr)` and `Contains(err, ParquetAttrColumn(attr))`, but all four inputs (`Created_At`, `Row_ID`, `RN`, `Source_Tier_Priority`) fold to themselves, so the second assertion can never fail independently. A case where the fold **and** the case both act — `Row.ID` → `Row_ID` → `row_id`, or `Created At` → `Created_At` — would exercise the two axes together and give that assertion weight. The existing `TestValidateParquetAttrColumns_ReservedSystemColumn` covers the fold axis alone, so this is the one combination still unpinned.
- **O5 — no over-rejection guard for a case variant.** `TestValidateParquetAttrColumns_ValidCachePasses` pins that camelCase does not fold onto snake_case system columns (`createdAt`), which is the right instinct, but no case exercises a mixed-case name that *should* pass, e.g. a lone `Contact_Name` with no sibling. One line pins that the normalization narrows nothing.
- **O6 — no repository fixture is affected.** I checked every `*_attributes.json` under `cmd/`. The mixed-case names in them (`createdAt`, `leadId`, `userId`, `nextFollowUpAt`, `durationMinutes`, `updatedAt`, `yearOfProduction.year`) fold to themselves and lower-case to `createdat`, `leadid`, … — none lands in `reservedParquetColumns`. Nothing in-tree starts failing.

## Verification (independent, re-run locally)

- **Red-to-green reproduced from scratch.** Reverted *only* `internal/sqlgen/parquet_attr_column.go` to `origin/main`, leaving the new tests in place, and ran `-run CaseVariant` in both packages. All four fail with `An error is expected but got nil`:
  - `TestValidateParquetAttrColumns_CaseVariantReservedSystemColumn` (first failure on `Created_At`)
  - `TestValidateParquetAttrColumns_CaseVariantAttrCollision`
  - `TestFileSchemaRegistryRejectsCaseVariantReservedColumnAtLoad`
  - `TestFileSchemaRegistryRejectsCaseVariantFoldedAttrCollisionAtLoad`

  Restored the file, re-ran: all green. Each test genuinely constrains the production change, and both the validator seam and the file-registry seam are covered independently.
- `go test ./internal/sqlgen ./internal/schemameta` — **ok**.
- `make fmt-check` — clean. `make lint` — clean, root and `infra/` modules.
- `git diff --check origin/main...HEAD` — clean.
- `go test . ./cdc ./cmd/... ./factory ./internal/...` — everything green except `internal/e2e_harness`, which panicked on `checked path: $XDG_RUNTIME_DIR` inside testcontainers. That is the known container-runtime gap, not this change.
- `./scripts/test_with_container_runtime.sh` — full suite green under rootless Podman, including `internal/e2e_harness`, `internal/e2e_harness/federated`, `internal/e2e_harness/production`, `internal/federated`, `internal/compaction`, `internal/reconcile`. This independently confirms the author's verification record on #532.
- Branch is 1 commit ahead of `origin/main` and 0 behind; `MERGEABLE` / `CLEAN`; all five checks green (Lint, Test, E2E Smoke Tests, k6 Smoke, Benchmark Smoke).

## Summary

Eleven lines of production change that close the last verbatim reserved-column lookup in the codebase. The fix is the one #532 asked for, applied at the one place left holding a case-sensitive comparison, with the comparison key normalized and the emitted column name and every diagnostic left byte-identical — so the `ParquetAttrColumn` contract and every parquet file already on S3 are untouched. Tests pin both halves of the guard (reserved and intra-schema collision) at both seams (validator and registry load), and I confirmed all four are red without the production hunk.

The three items above are all follow-up material, not merge blockers: F1 is a message-clarity improvement that could ride along in this PR or a small follow-up; F2 is one sentence in `docs/federated-query/design.md` §4.4 and I would slightly prefer it lands with this change, since the doc currently asserts a premise only this commit satisfies; F3 is a release-note addition. None of them changes the behavior being merged, which is correct.

Ready to merge on your word — I have not merged it.

iceboundrock:
# Review response — PR #548

Artifact: review response. Working copy: `fix/532-case-insensitive-parquet-columns` at `a061beb`; this comment is the durable record.

Disposition: F1, F2, O4, O5 addressed in `a061beb`. F3 split — release note in the PR body, the substantive half filed as #549. O1, O2, O3, O6 need no action; reasoning below.

---

## Addressed

### F1 — diagnostics now name the resolved column (fixed)

Both messages were wrong in the way described: the reserved message handed the operator a string absent from `reservedParquetColumns`, and the collision message asserted a shared parquet column that `ParquetAttrColumn` never produces.

Fixed, but branched rather than reworded unconditionally, because the suggested single shape reads redundantly on the pure-fold path (`attributes "contact.name" and "contact_name" map to parquet columns "contact_name" and "contact_name" ... onto the same column "contact_name"`). Where the folded name is already lower case, or where the two folds are byte-identical, the original wording is exact and stays; the case-variant path gets the explanation. Error construction moved into `reservedParquetColumnError` and `foldedColumnCollisionError` to keep `ValidateParquetAttrColumns` flat.

What the four paths now emit:

```
attribute "created.at" folds to parquet column "created_at", which is reserved for system columns; rename the attribute
attribute "Created_At" folds to parquet column "Created_At", which DuckDB resolves case-insensitively onto the reserved system column "created_at"; rename the attribute
attributes "contact.name" and "contact_name" both map to parquet column "contact_name"; attribute names must remain distinct after identifier folding
attributes "Contact_Name" and "contact_name" map to parquet columns "Contact_Name" and "contact_name", which DuckDB resolves case-insensitively onto the same column "contact_name"; attribute names must remain distinct after identifier folding
```

The `ValidateParquetAttrColumns` docstring now states that the diagnostics carry the resolved name beside the caller's spelling, so the reason survives the diff.

### F2 — `docs/federated-query/design.md` §4.4 updated (fixed)

The registration-guard paragraph now states the case-insensitive comparison explicitly, gives both worked examples, pins that the emitted names and diagnostics stay byte-stable, and says in as many words that this is what makes the "registered names never reach it" premise at L309-310 hold for case variants rather than only exact-case ones. The finding was right that the doc asserted a premise only this commit satisfies.

### O4 — the duplicate assertion now has weight (fixed)

`Row.ID` (→ `Row_ID` → `row_id`) and `Created At` (→ `Created_At` → `created_at`) added to `TestValidateParquetAttrColumns_CaseVariantReservedSystemColumn`, so the fold and the case act together and `Contains(ParquetAttrColumn(attr))` can fail independently of the caller-spelling assertion. A third assertion pins the resolved lower-case name, which is what F1 is actually about. The collision half gets the same treatment in a new `TestValidateParquetAttrColumns_CaseVariantFoldedAttrCollision` (`Contact.Name` vs `contact_name`), where the two attributes hold genuinely distinct parquet columns.

### O5 — over-rejection guard added (fixed)

`Contact_Phone` and `Attr` added to `TestValidateParquetAttrColumns_ValidCachePasses`: a mixed-case name that neither hits the reserved set nor has a case-variant sibling still passes, and — per O2 — a lone `Attr` is not refused just because it lower-cases onto the placeholder string.

### F3 — split (release note here, substance in #549)

Both sub-items verified against the code: `validateSchemaAttributeCache` is called at `loader.go:81`, `loader.go:360`, `file_registry.go:192`, `file_registry.go:389`, all inside per-schema loops that return on first error; and it runs over the full cache before `activeAttributeCache` strips retired entries (`validation.go:47-57`).

The release note is in the PR body now, both bullets, marked as pre-existing and only widened here.

The retired-entry half is more than a release note: after this PR a retired `Created_At` hard-blocks boot with "rename the attribute", which is not a remedy for a ledger record whose whole purpose is preserving the historical name for #341 orphan detection against a column already in flushed parquet. That is out of scope for #532 and filed as **#549**, with three options costed (retired-aware diagnostics; aggregate registry errors; narrowing the guard) and a preference for the first.

## Not actioned

- **O1** — agreed as an observation. The three comments (`keyset.go:75-77`, `cold_columns.go:47-48`, `ValidateUnregisteredParquetAttrColumn`'s premise) are now true as written; none needs an edit.
- **O2** — the `Attr` / `[]` side effect is correct behavior. It is now covered from the passing direction (O5 above); the rejecting direction is `[Attr]`, already pinned by the `ValidateUnregisteredParquetAttrColumn` tests at the filter seam.
- **O3** — `strings.ToLower` being Unicode-aware where DuckDB folds ASCII is real, but it over-rejects (`K` U+212A vs `k`), the guard is fail-closed, and an ASCII-only fold helper would have to land at three sites in two packages to be worth having (`ValidateParquetAttrColumns`, `ValidateUnregisteredParquetAttrColumn`, `federated.validateKeysetCursor`). Not filed: an issue for a rejection that cannot produce a wrong answer and has never been observed is backlog noise. If the shared-helper work in #531 goes ahead, that is the place to fix it in one move — noting it here so the reasoning is on record.
- **O6** — confirmed independently; no fixture changes needed.

## Verification of the response commit

- `go test ./internal/sqlgen ./internal/schemameta` — ok.
- `./scripts/test_with_container_runtime.sh` — full suite green under rootless Podman, exit 0, including `internal/e2e_harness`, `internal/e2e_harness/federated`, `internal/e2e_harness/production`, `internal/federated`, `internal/compaction`, `internal/reconcile`.
- `make fmt-check` — clean. `make lint` — clean, root and `infra/` modules.
- `parquet_attr_column.go` is 186 lines, longest function 25 — inside the 500/100 limits.
- Grepped for other assertions on the two changed message strings: none outside the file itself.

Still not merged.

iceboundrock:
# Review - PR #548 `fix(sqlgen): reject case-variant parquet system columns (#532)`

Artifact: code review. Working copy: local checkout of `fix/532-case-insensitive-parquet-columns` at `a061beb`; this comment is the durable review record.

## Findings

### F1 - [P1] Match DuckDB identifier folding, not Go Unicode folding

`internal/sqlgen/parquet_attr_column.go:142` derives the collision/reserved lookup key with `strings.ToLower`. Go applies Unicode case mappings; DuckDB documents that identifier comparison is ASCII-based: `col_A` equals `col_a`, while non-ASCII letters are not folded ([DuckDB identifiers](https://duckdb.org/docs/current/sql/dialect/keywords_and_identifiers.html#case-sensitivity-of-identifiers)).

Consequently, a valid cache containing distinct attributes `Á` and `á` is rejected: `ParquetAttrColumn` preserves both spellings, DuckDB treats the two identifiers as distinct, but `strings.ToLower` collapses them to one map key. The same applies to other Unicode mappings such as `K` -> `k`. This is an externally visible registration regression, not merely a conservative runtime guard: one such schema aborts registry construction and can prevent startup for every schema. No attribute-name rule I found restricts names to ASCII.

Use an ASCII-only identifier-folding helper for this comparison, and qualify `docs/federated-query/design.md:177-186` accordingly. Add a counterexample proving `Contact_Name` / `contact_name` is refused while `Á` / `á` remains accepted. A DuckDB-backed assertion would pin the dependency behavior directly.

## Spec Check

- #532 requested normalization before the reserved-column lookup and the intra-schema collision map. Both paths now normalize and correctly reject ASCII case variants.
- The intended contract is DuckDB-equivalent comparison; F1 means the implementation is not equivalent for permitted non-ASCII identifiers.
- Registration coverage exists at the shared validator and directory-registry seams. The CDC and federated projection callers re-use that validator; their existing defense-in-depth behavior remains intact.
- The follow-up diagnostics, documentation, regression combinations, and passing-case coverage requested by the earlier review are present in `a061beb`.

## Standards Check

- Production file is 186 lines; functions remain far below the 100-line limit.
- Error classification remains correct: registration failures are contextual plain errors, not public 4xx sentinel carriers.
- Formatting and diff whitespace are clean.
- F1 is a correctness issue in the chosen normalization primitive; no other coding-standard violations found in the changed production code.

## Assessment

**Request changes.** The ASCII case-variant fix is otherwise focused and well-tested, but F1 can reject valid schemas and block startup. It needs correction before merge.

## Observations

- The response commit correctly fixed the prior diagnostic issue: case-variant errors now retain physical spelling and name the resolved lower-case system column.
- The design record now documents the registration guard and its relation to downstream guards. It must state the ASCII boundary once F1 is fixed.
- The retired-ledger operability concern is appropriately split to #549 and is not a merge blocker for this PR.

## Verification

Local, independently run at `a061beb`:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/schemameta -count=1` - passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc -count=1` - passed.
- `make fmt-check` - passed.
- `git diff --check 350b5bf^..a061beb` and `git diff --check origin/main...HEAD` - clean.

GitHub CI is green for Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke. I did not re-run the container suite locally because the existing green CI and PR verification cover it.

## Summary

The PR closes the intended ASCII case-insensitive reserved-name and collision gap, keeps the byte-stable parquet naming contract, and incorporates the previous review feedback. Replace Unicode `strings.ToLower` with DuckDBs ASCII-only identifier fold, add the non-ASCII passing regression test, and update the wording in the design document; then the change is ready for approval.

iceboundrock:
# Review response - F1 fixed in `a1db047`

Artifact: review response. Working copy: local checkout of `fix/532-case-insensitive-parquet-columns` at `a1db047`; this comment is the durable record.

## F1 - [P1] Match DuckDB identifier folding, not Go Unicode folding: **confirmed and fixed**

Verified against the pinned DuckDB (`duckdb-go/v2 v2.5.6`, engine `v1.4.5`) rather than the documentation, since the whole guard rests on the dependency's behavior. Two attribute names share a parquet column exactly when DuckDB refuses to create a table holding both as quoted columns, so that refusal is the oracle:

| pair | DuckDB merges | `strings.ToLower` merges |
|---|---|---|
| `col_A` / `col_a` | yes | yes |
| `Contact_Name` / `contact_name` | yes | yes |
| `Á` / `á` | no | **yes** |
| `Ж` / `ж` | no | **yes** |
| `ẞ` / `ß` | no | **yes** |
| `cafÉ` / `café` | no | **yes** |
| `K` (U+212A KELVIN SIGN) / `k` | no | **yes** |
| `İ` (U+0130) / `i` | no | **yes** |

DuckDB folds `A`–`Z` and nothing else. Every non-ASCII mapping Go performs is one DuckDB does not, so the finding holds exactly as written.

One correction to the finding's framing, in the direction of making it worse rather than milder. The finding presents this as a collision-map defect (`Á` and `á` reported as sharing a column). The reserved half is reachable too, which the finding did not claim: Go's simple case mapping sends U+0130 `İ` to `i`, so the legitimate attribute `row_İd` folded onto the reserved `row_id` and was rejected as a system column. Both halves of the guard were narrower than the engine, not just the collision half.

Fixed by keying both lookups on a new `duckdbFoldIdentifier` — an ASCII-only byte fold — instead of `strings.ToLower`.

### Coverage

- `TestValidateParquetAttrColumns_NonASCIICaseVariantsPass`: the counterexample requested. `Á`/`á`, `Ж`/`ж`, `cafÉ`/`café` and `row_İd` all pass in one cache, and the same test then adds `Contact_Name`/`contact_name` to that cache and requires the error, so the pass is the boundary being respected rather than the guard being switched off.
- `TestDuckDBIdentifierFoldIsASCIIOnly`: the DuckDB-backed assertion suggested. It runs the table above against the engine and requires `duckdbFoldIdentifier` to agree with it pair by pair, so a DuckDB upgrade that changes identifier folding fails the build instead of silently invalidating the guard's premise. `internal/sqlgen` tests already link the driver, so this sits beside the code it pins.

Red-to-green: with `key := strings.ToLower(col)` restored, `TestValidateParquetAttrColumns_NonASCIICaseVariantsPass` fails on `attributes "cafÉ" and "café" map to parquet columns ... onto the same column "café"`. All previously added ASCII rejection tests stay green either way.

### Documentation

`docs/federated-query/design.md` §4.4 now states the ASCII boundary explicitly, names the reason it is pinned against the engine (a registration failure fails the whole registry, so a false rejection here is a boot failure), and points at the two seams that still differ.

## Out of scope, tracked in #550

The same premise backs two guards this PR does not touch, both still folding with `strings.ToLower`:

- `internal/sqlgen/parquet_attr_column.go:102` — `ValidateUnregisteredParquetAttrColumn` (#512/#530), plus the three `strings.EqualFold` calls in the same function.
- `internal/federated/keyset.go:175` — `validateKeysetCursor` (#381/#509).

Kept out of this PR deliberately, on a severity split rather than a file boundary. Those two are runtime guards over caller-supplied strings: being narrower than DuckDB means refusing a query the engine would have bound — a 4xx `forma.InvalidInputf` carrier — and nothing there can fail registration or block startup. The registration guard is the one where narrowness is a boot failure, and it is the one this PR introduced (`350b5bf`); the other two predate it. Neither can produce a silent wrong answer: an over-strict fold only over-rejects.

Fixing them also means exporting the helper across the `sqlgen`/`federated` package boundary and changing the behavior of two guards with their own test suites and their own error-carrier semantics, which is a change worth reviewing on its own. #550 carries the measurement table, the `İ`-reachability note, and the proposed fix. `ValidateUnregisteredParquetAttrColumn` carries an in-code note pointing at it, so the file does not read as self-contradictory next to the new helper.

## Other review points

- **Spec check, "the intended contract is DuckDB-equivalent comparison"** — agreed, and that is now literally what the test asserts, against the engine rather than a restatement of it.
- **Observations, "it must state the ASCII boundary once F1 is fixed"** — done, §4.4.
- **Observations, #549 retired-ledger split** — agreed, not a merge blocker; unchanged here.

## Verification at `a1db047`

- `go test ./internal/sqlgen ./internal/schemameta ./internal/cdc -count=1` — passed.
- `./scripts/test_with_container_runtime.sh` (full `make test`, rootless Podman) — passed, all packages.
- `make lint` (root + infra modules) — clean.
- `make fmt-check` — clean.
- `git diff --check` — clean.

`internal/sqlgen/parquet_attr_column.go` is 225 lines; the new helper is 15 lines. Error classification is unchanged: the registration guard still returns plain contextual errors.

iceboundrock:
# Review - PR #548 `fix(sqlgen): reject case-variant parquet system columns (#532)`

Artifact: code review. Working copy: local checkout of `fix/532-case-insensitive-parquet-columns` at `a1db047`; this comment is the durable review record.

## Findings

No blocking or non-blocking findings. I independently reviewed the latest head, including the three commits in this PR and every production caller of `ValidateParquetAttrColumns`.

## Spec Check

- The registration guard normalizes the `ParquetAttrColumn` result before both the reserved-system-column lookup and the intra-schema collision map (`internal/sqlgen/parquet_attr_column.go:167-190`).
- It uses an ASCII-only fold, which matches the pinned DuckDB engine behavior, rather than Go Unicode folding (`:123-151`). This correctly rejects `Created_At` and `Contact_Name` / `contact_name` while admitting distinct non-ASCII DuckDB identifiers such as `Á` / `á` and `row_İd`.
- Physical parquet names remain byte-stable: `ParquetAttrColumn` is unchanged, while errors preserve caller spelling and identify the lower-case resolved column (`:199-225`). No persisted data migration is needed.
- The guard is reached at registration (`internal/schemameta/validation.go:53`) and is retained as defense in depth for CDC export (`internal/cdc/export_sql_builder.go:124`) and DuckDB projection (`internal/sqlgen/duckdb_schema_projection.go:119`).
- Validator and file-registry seams are covered for reserved-name and collision cases; the engine-backed test also detects a future DuckDB identifier-folding contract change (`internal/sqlgen/parquet_attr_column_test.go:157-218`).

## Standards Check

- `parquet_attr_column.go` is 225 lines; its changed functions are well below the 100-line limit.
- Errors remain contextual plain errors, correctly classified as operator-visible registration/read-path consistency failures rather than public 4xx carriers.
- The implementation is flat, deterministic (sorted attributes), and adds no unnecessary exported API or compatibility path.
- `docs/federated-query/design.md:165-200` records the case-insensitive registration invariant, its ASCII boundary, persisted-name compatibility, and the remaining runtime-guard scope.
- The PR follows repository artifact requirements: plan, prior findings, response, and verification records are present on the linked issue/PR.

## Assessment

**Approve.** This is a focused, correct fail-closed fix for the case-variant system-column and collision gap. The latest change addresses the prior Unicode-folding regression by matching the actual DuckDB behavior rather than relying on a broader language-library transformation.

## Observations

- The retained `strings.ToLower` / `EqualFold` behavior in unregistered-filter and keyset runtime guards is explicitly documented and tracked in #550. It can over-reject a caller query, but cannot admit an ambiguous registered schema or block registry construction; it is not a blocker for this PR.
- The wider retired-ledger startup-operability consequence is documented in the PR release note and tracked in #549. It predates this change for exact-case names; this PR only widens the rejected invalid-name set.
- All five current GitHub CI checks are green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke. The PR is `MERGEABLE` with `CLEAN` merge state.

## Verification

Independently run at `a1db047`:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/schemameta -count=1` - passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/sqlgen ./internal/schemameta -count=1` - passed.
- `make fmt-check` - passed.
- `git diff --check origin/main...HEAD` - clean.
- Worktree was clean before and after verification.

## Summary

The PR fixes the intended registration-time gap without changing parquet naming or stored data. It rejects ASCII case variants exactly where DuckDB merges them, preserves useful diagnostics, and prevents Unicode false rejections that could otherwise block startup. Ready to merge; no merge action was taken.

Issue #550, opened 2026-09-08
sqlgen/federated: runtime filter and keyset guards fold identifiers with Unicode ToLower, not DuckDB's ASCII fold


iceboundrock:
## Summary

`sqlgen.ValidateParquetAttrColumns` now folds identifiers the way DuckDB does — ASCII only, via `sqlgen.duckdbFoldIdentifier` (#532 / #548). The two runtime guards that share the same "DuckDB resolves unquoted identifiers case-insensitively" premise still fold with Go's `strings.ToLower`, which applies Unicode case mappings the engine does not:

- `internal/sqlgen/parquet_attr_column.go:102` — `ValidateUnregisteredParquetAttrColumn`, the unregistered-filter-attribute guard (#512 / #530). Also the three `strings.EqualFold` calls in the same function, which fold the same way.
- `internal/federated/keyset.go:175` — `validateKeysetCursor`, the keyset cursor column guard (#381 / #509).

## Evidence

Measured against the pinned DuckDB (v1.4.5) by creating a table with both names as quoted columns; "merged" means DuckDB refused with `Column with name ... already exists`:

| pair | DuckDB merges | `strings.ToLower` merges |
|---|---|---|
| `col_A` / `col_a` | yes | yes |
| `Á` / `á` | no | **yes** |
| `Ж` / `ж` | no | **yes** |
| `ẞ` / `ß` | no | **yes** |
| `cafÉ` / `café` | no | **yes** |
| `K` (U+212A) / `k` | no | **yes** |
| `İ` (U+0130) / `i` | no | **yes** |

DuckDB's fold is strictly `A`–`Z`. Every non-ASCII mapping Go performs is one DuckDB does not.

## Impact

Both guards are narrower than the engine, so the failure mode is a false rejection of a valid query, not a silent wrong answer:

- A filter or cursor on a non-ASCII attribute whose Unicode fold collides with a reserved or dedup column is refused. The reserved set is all-ASCII, so the reachable path is `U+0130` → `i`: a filter on `row_İd`, `ver_İd`-style names, or any attribute whose Unicode fold lands on a reserved name is rejected as a system-column collision.
- These are `forma.InvalidInputf` carriers on the read path, so the user sees a 4xx on a query that DuckDB would have bound correctly. Unlike the registration guard, nothing here can block startup — which is why it was split out of #548 rather than fixed there.

[...]

PR #551, opened 2026-09-08
refactor(sqlgen): single-source the dedup columns and attr placeholder (#531)

### LINKED ISSUE #531: sqlgen/federated: dedup-column set and attr placeholder are duplicated across packages with no sync guard

iceboundrock:
Closes #531.

## Problem

Two definitions existed twice, once per package, with nothing keeping them equal:

| `internal/sqlgen/parquet_attr_column.go` | `internal/federated/keyset.go` |
| --- | --- |
| `federatedDedupColumns` (`rn`, `source_tier_priority`) | `keysetRejectedColumns` (same two) |
| `parquetAttrPlaceholder = "attr"` | `parquetAttrFallbackColumn = "attr"` |

`sqlgen.ValidateUnregisteredParquetAttrColumn` (filters, #512) and
`federated.validateKeysetCursor` (cursors, #509) apply the same fold rule to
them. Both sites commented their counterpart, but a column added to the dedup
machinery could be taught to one guard and silently left out of the other,
reopening the #354 silent-wrong-answer class on whichever path was missed.

## Change

`sqlgen` owns both definitions — the only direction layering allows, since
`federated` imports `sqlgen` and not the reverse.

- `parquetAttrPlaceholder` becomes the exported const `ParquetAttrPlaceholder`,
  and `ParquetAttrColumn` returns it instead of the bare `"attr"` literal (the
  issue's "also worth considering" item).
- `federatedDedupColumns` stays unexported behind
  `FederatedDedupColumns() map[string]struct{}`, which returns a fresh copy per
  call. This is the one deviation from the issue's sketch: an exported map
  *variable* would trade the drift hazard for an aliasing one, since any
  importer could edit a set that decides whether a query is refused.
- `federated` derives `keysetRejectedColumns` from that accessor and compares
  against `sqlgen.ParquetAttrPlaceholder`, deleting both local declarations.

No behaviour change. The two sets are identical today, and the lookup rules are
preserved exactly on both paths: fold, then `strings.ToLower` for the map
lookups; case-insensitive comparison for the placeholder. The
`strings.ToLower`-vs-DuckDB narrowness in both guards is untouched and stays
tracked in #550.

## Tests

- `internal/sqlgen`: `FederatedDedupColumns()` returns a copy — mutating what it
  hands back changes neither a later call nor what the filter guard refuses; and
  `ParquetAttrColumn`'s empty-name fallback is the placeholder const. The literal
  `"attr"` is still asserted separately, because that value is a byte-stable
  contract with parquet already flushed to S3.
- `internal/federated`: `TestKeysetCursorRejectsEverySqlgenDedupColumn` drives
  the cursor guard from `sqlgen.FederatedDedupColumns()` rather than from a
  literal list, under every spelling the fold and DuckDB's case-insensitive
  resolution reach a column by. `TestKeysetCursorPlaceholderTracksSqlgen` does
  the same for the placeholder rule.

The drift guard was verified to bite rather than merely pass: with a third
column added to the sqlgen set **and** `federated` regressed to a hardcoded
local set (the pre-PR shape), the new test fails with

```
cursor column "tier_rank" folds onto dedup column "tier_rank" and must be refused
```

Both changes were then reverted; the committed tree is the wired version.

All pre-existing cursor and filter tests pass unchanged, which is the
no-behaviour-change evidence.

## Gates

`make fmt-check`, `make lint`, and the full unit suite via
`./scripts/test_with_container_runtime.sh` — all green. Both touched source
files stay well under the 500-line limit (244 and 274).

### DIFF
--- a/internal/federated/keyset.go
+++ b/internal/federated/keyset.go
@@ -73,10 +73,15 @@ var keysetSystemColumns = map[string]struct{}{
 // caught by mistake: schema registration rejects an attribute whose fold
 // collides with a reserved parquet column, and rn and source_tier_priority are
 // both in that reserved set (sqlgen.ValidateParquetAttrColumns).
-var keysetRejectedColumns = map[string]struct{}{
-	"rn":                   {},
-	"source_tier_priority": {},
-}
+//
+// The set is DERIVED from sqlgen, not redeclared here (#531). The same columns
+// are refused on the filter path by
+// sqlgen.ValidateUnregisteredParquetAttrColumn, and layering forbids the
+// reverse import, so sqlgen owns the definition and a column added there
+// reaches this guard too. The map below is this package's own copy:
+// FederatedDedupColumns hands out a fresh one per call, so neither package can
+// edit the other's guard.
+var keysetRejectedColumns = sqlgen.FederatedDedupColumns()
 
 // safeSQLIdentifier matches a bare, unquoted DuckDB identifier. Cursor
 // attributes are interpolated into SQL as identifiers, not bound as
@@ -86,15 +91,6 @@ var keysetRejectedColumns = map[string]struct{}{
 // attribute like "contact.annualIncome" passes, while a quote, semicolon,
 // parenthesis or leading digit does not.
 var safeSQLIdentifier = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
-
-// parquetAttrFallbackColumn is the placeholder column name ParquetAttrColumn
-// substitutes when the fold empties an attribute name. It is a legitimate
-// attribute name in its own right, so the cursor rule can only reject a name
-// that FOLDS ONTO it — a folded "attr" whose raw attribute was something else,
-// whether the fold emptied the name (the empty string, "[]", a bare pair of
-// backticks) or merely stripped it down to the literal placeholder, as
-// "[attr]" and a backtick-wrapped attr do.
-const parquetAttrFallbackColumn = "attr"
 
 // validateKeysetCursor is THE keyset cursor contract, and the only validation
 // entry point. Both seams call it — the engine gate (engine.go) and the
@@ -151,14 +147,18 @@ func validateKeysetCursor(cursor *model.KeysetCursor, orders []model.AttributeOr
 		// name the code generator emits, so a rule applied to the raw name
 		// guards a string that never reaches SQL.
 		folded := sqlgen.ParquetAttrColumn(col.Attribute)
-		if strings.EqualFold(folded, parquetAttrFallbackColumn) && !strings.EqualFold(col.Attribute, parquetAttrFallbackColumn) {
-			// The literal "attr" is ParquetAttrColumn's placeholder: it
-			// substitutes it whenever the fold empties the name — "", "[]",
+		if strings.EqualFold(folded, sqlgen.ParquetAttrPlaceholder) &&
+			!strings.EqualFold(col.Attribute, sqlgen.ParquetAttrPlaceholder) {
+			// sqlgen.ParquetAttrPlaceholder is the column ParquetAttrColumn
+			// substitutes whenever the fold empties the name — "", "[]",
 			// "``" — and a name like "[attr]" or "`attr`" strips down onto it
-			// directly. That placeholder exists for the export writer, which
-			// needs some column name for every attribute; here either route
-			// would silently retarget the cursor at a real attribute called
-			// "attr".
+			// directly. It is read from sqlgen rather than restated here so
+			// the guard cannot outlive the literal it guards (#531); it is
+			// also a legitimate attribute name in its own right, which is why
+			// the rule can only reject a name that FOLDS ONTO it. That
+			// placeholder exists for the export writer, which needs some
+			// column name for every attribute; here either route would
+			// silently retarget the cursor at a real attribute called "attr".
 			//
 			// Both comparisons are case-insensitive, for the same reason the
 			// two map lookups below are: DuckDB resolves unquoted identifiers
--- a/internal/federated/keyset_test.go
+++ b/internal/federated/keyset_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/lychee-technology/forma"
 	"github.com/lychee-technology/forma/internal/model"
+	"github.com/lychee-technology/forma/internal/sqlgen"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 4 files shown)


iceboundrock:
# Review: refactor(sqlgen): single-source the dedup columns and attr placeholder (#531)

## Summary

Maintainability-only refactor, no behaviour change. `sqlgen` now owns both shared definitions (`ParquetAttrPlaceholder`, `federatedDedupColumns` behind `FederatedDedupColumns()`), `federated` derives its cursor guard from them, and both local redeclarations are deleted. The drift hazard from #531 is closed structurally for the dedup set and the placeholder, with tests that drive the cursor guard from the `sqlgen` source rather than a literal list. **No blocking findings — looks merge-ready.**

## Spec check (against #531)

- [x] Dedup set single-sourced: `federatedDedupColumns` stays unexported in `sqlgen`, exposed via `FederatedDedupColumns()`; `federated.keysetRejectedColumns` is derived from that accessor. No second declaration remains (verified by grep — zero hits for `parquetAttrFallbackColumn` / duplicate set).
- [x] Placeholder single-sourced: `parquetAttrPlaceholder` → exported `ParquetAttrPlaceholder`, and `ParquetAttrColumn` returns the const instead of the bare `"attr"` literal — this covers the issue's "also worth considering" item.
- [x] Layering direction correct: `federated` imports `sqlgen`, never the reverse (verified — no `internal/federated` reference under `internal/sqlgen/`).
- [x] Deliberate deviation from the issue sketch is justified: the issue suggested exporting the definitions, possibly as variables; the PR exports a const (placeholder) plus a copy-returning accessor (dedup set) instead of an exported map variable, with the aliasing rationale documented on the accessor and in the commit message. I agree with the deviation — an exported mutable map guarding query refusal would be worse than the drift it replaces — and the copy semantics are pinned by `TestFederatedDedupColumnsIsACopy`.
- [x] Scope respected: the `strings.ToLower`-vs-DuckDB narrowness on both guards is untouched and still tracked in #550. Diff confirms no fold-logic change.

## Standards check (`coding-standard.md`, error semantics, layering)

- [x] File sizes: `parquet_attr_column.go` 244 lines, `keyset.go` 274 lines — well under the 500-line limit. No function approaches 100 lines (`FederatedDedupColumns` ~7, test additions each <40, `validateKeysetCursor` body unchanged in shape).
- [x] Error semantics (`docs/error-handling.md`): preserved exactly. Filter guard still returns `forma.InvalidInputf` carriers; cursor guard still returns plain read-path `fmt.Errorf`. No bare sentinel wraps introduced, no status-class change.
- [x] Godoc on both new exported items (`ParquetAttrPlaceholder`, `FederatedDedupColumns`) present and explains the cross-package contract plus the copy rationale.
- [x] No duplication added (>15 similar lines): the change strictly removes duplication (net −20 lines in `keyset.go` declarations).
- [x] Early-return / flat structure untouched; comments reference the counterpart guard and the tracking issues (#531, #354, #550).

## Findings

No blocking findings.

- **F1 (checked, not an issue) — `keysetRejectedColumns` is a snapshot copy at package init, not a live view.** `var keysetRejectedColumns = sqlgen.FederatedDedupColumns()` copies once. This is sound here because the source is immutable at runtime (unexported `var`, no mutator API, only read via the accessor), so init-copy ≡ live for all practical purposes, and it avoids per-request allocation in the cursor hot path. The comment on the var already states it is "this package's own copy". No action.
- **F2 (checked, not an issue) — mutation surface within `federated` still exists in theory.** The var could still be reassigned from inside `federated`, but that was true before and the scope is now one package with a single assignment (verified — exactly one `keysetRejectedColumns =` site). No action.

## Observations (nits, non-blocking, safe to merge as-is)

- **O1:** `FederatedDedupColumns` doc says "returns the dedup-machinery column set, lower-cased the way both guards look it up". Strictly, the stored literals are already lowercase; it is the guards that lowercase the lookup key. Optional reword to "stored lower-cased, matching the guards' lookup keys" if touched again.
- **O2:** `TestFederatedDedupColumnsIsACopy` pins a literal `want` set (`rn`, `source_tier_priority`), so adding a dedup column requires updating that test alongside the source. That friction is intended (it forces attention on the byte-level contract), and it pairs correctly with the auto-adapting federated test that needs no update. Noting only so a future adder is not surprised.
- **O3:** Remaining `"attr"` string literals are all legitimate: the const definition itself, the byte-stability pin (`require.Equal(t, "attr", ParquetAttrPlaceholder)` — correctly kept as a value assertion against already-flushed parquet), and pre-existing table-driven error-string expectations. No stray literal feeds logic.

## Verification

- **CI (all green at review time):** Lint, Test, E2E Smoke Tests, Benchmark Smoke, k6 Smoke — 5/5 pass.
- **Local, targeted:** `go test ./internal/sqlgen -run 'TestParquetAttrColumnUsesPlaceholderConst|TestFederatedDedupColumnsIsACopy|TestParquetAttrColumn' -v` — pass; `go test ./internal/federated -run 'TestKeysetCursorRejectsEverySqlgenDedupColumn|TestKeysetCursorPlaceholderTracksSqlgen|TestPaginatedQueryTakesKeyset' -v` — pass.
- **Local, full packages:** `go test ./internal/sqlgen ./internal/federated -count=1` — both `ok` (no regressions in pre-existing cursor/filter tests, supporting the no-behaviour-change claim).
- **Hygiene:** `gofmt -l` on all four touched files — clean; `go vet ./internal/sqlgen/ ./internal/federated/` — clean.
- **Structural grep:** no residual `parquetAttrFallbackColumn`; single `keysetRejectedColumns` assignment; no `sqlgen → federated` import.
- **Mutation bite:** the PR body reports the new federated test fails with the pre-PR hardcoded local set plus an injected third column (`cursor column "tier_rank" folds onto dedup column "tier_rank" and must be refused`), then reverted. Author-verified; not re-run locally.

## Assessment

Closes #531 as scoped, via the stronger of the issue's two proposed mechanisms (structural single-sourcing, not just an equality test — plus the driven-from-source tests as a second net). Correct layering, correct error classes, tested copy semantics, byte-stability of the `"attr"` contract preserved, #550 cleanly excluded. Nothing load-bearing left in chat or local files beyond this comment. Approve / merge at the author's convenience (not merging per repo PR rules — awaiting explicit instruction).

iceboundrock:
# Review response

One item actioned (O1), four confirmed no-action, one follow-up issue opened for
a gap the review did not reach. Pushed as `c379935`.

## O1 — actioned: `FederatedDedupColumns` doc misattributed the lower-casing

Verified before changing: the stored literals in `federatedDedupColumns` are
already lower case, and it is each guard that lower-cases the *lookup key* —
`internal/sqlgen/parquet_attr_column.go:128` and
`internal/federated/keyset.go:175`, both `key := strings.ToLower(folded)`. The
old wording ("returns the dedup-machinery column set, lower-cased the way both
guards look it up") reads as if the accessor normalises what it hands back,
which would mislead the next person adding a column into thinking a
mixed-case entry is fine.

Reworded, and the rewrite says *why* the guards fold rather than only that they
do, so the next reader does not have to derive the DuckDB reason a second time:

```go
// FederatedDedupColumns returns the dedup-machinery column set. Its entries
// are stored lower-cased, matching the keys both guards look up: each folds
// the caller's name and lower-cases the result before the map lookup, because
// DuckDB resolves unquoted identifiers case-insensitively.
//
// The map is a fresh copy on every call: the set is package state that decides
// whether a query is refused, so handing out the live map would let any
// importer edit the guard rather than read it.
```

Comment only — no behaviour, API, or test change.

## F1, F2, O2, O3 — no action, and I re-verified each rather than taking them as read

- **F1 (init-time snapshot, not a live view).** Confirmed. `grep` finds exactly
  one assignment to `keysetRejectedColumns` (`keyset.go:84`) and no mutator on
  `federatedDedupColumns`; the source is immutable at runtime, so init-copy is
  observationally identical to a live read and saves an allocation per cursor
  validation. Making it a per-call read would cost the hot path to buy nothing.
- **F2 (in-package reassignment still possible in theory).** Confirmed and
  agreed it is not worth closing. Go has no in-package immutability for maps
  short of hiding the var behind an accessor *within* `federated` too, which
  would add indirection to guard against this package sabotaging itself. The
  pre-PR shape had the identical exposure over a wider surface.
- **O2 (`TestFederatedDedupColumnsIsACopy` pins a literal `want`).** Agreed the
  friction is the point, and I am deliberately leaving it. That test is the
  byte-level contract; the auto-adapting federated test is the drift net. Both
  is correct — driving both from the source would leave nothing asserting *what*
  the set contains.
- **O3 (remaining `"attr"` literals).** Re-grepped: the const definition, the
  byte-stability pin `require.Equal(t, "attr", ParquetAttrPlaceholder)`, and
  pre-existing error-string expectations in table tests. The pin must stay a
  literal — asserting the const against itself would assert nothing, and the
  value is a contract with parquet already flushed to S3.

## Out of scope, now tracked: #552

While verifying F1/F2 I found a **third** copy of the dedup names that the
review did not reach, in the same file: `reservedParquetColumns`
(`parquet_attr_column.go:43`) restates `"rn"` and `"source_tier_priority"` as
literals. It is not redundant — it is the *only* guard covering the
**registered**-attribute path:

- `ValidateParquetAttrColumns` (registry construction, CDC export, schema
  projection) checks against `reservedParquetColumns` alone and never consults
  `federatedDedupColumns`.
- `ValidateUnregisteredParquetAttrColumn` consults both, but `normalizeDuckPayload`
  gates it behind `if !hasMeta`, and its own comment says registered names were
  "already checked at registration by `ValidateParquetAttrColumns`".

So a future third dedup column added to `federatedDedupColumns` would reach both
guards this PR wires — and still let an attribute folding onto it register
successfully, after which the `hasMeta` branch never rechecks it and the clause
silently filters on the dedup rank. That is the #354 family on the very path
#531 set out to close.

No live defect: `rn` and `source_tier_priority` are in both sets today, so the
hazard is drift, not a current wrong answer — the same standing as #531 itself.

Deliberately **not** folded into this PR. It changes a set that gates schema
registration, where a composition slip rejects schemas that register today, and
this PR is reviewed and approved as a no-behaviour-change refactor. Filed as
**#552** with the composition sketch, the identity constraint on the current
set, and the drive-from-source test shape.

## Gates after the change

`make fmt-check`, `make lint` (root + infra) and
`go test ./internal/sqlgen ./internal/federated -count=1` — all green; both
packages `ok`, no test touched.

Not merging — awaiting explicit instruction per the repo PR rules.

iceboundrock:
## Review

### Findings

No blocking, non-blocking, or test-coverage findings.

### Spec

Meets #531:

- `sqlgen` is the single owner of the dedup-column definition and the Parquet attribute placeholder, preserving the permitted dependency direction (`federated` -> `sqlgen`).
- `ParquetAttrColumn` now returns `ParquetAttrPlaceholder`, eliminating its former independent `"attr"` literal.
- `federated` derives its reject set from `sqlgen`; the accessor returns a fresh map, so importers cannot mutate `sqlgen` package state.
- The cross-package tests pin both drift-sensitive contracts without changing the existing filter/cursor behavior.

### Standards

- Exported API has Go doc comments.
- Touched production files remain below the 500-line limit (`parquet_attr_column.go`: 248; `keyset.go`: 274); the added accessor is 7 lines.
- No layer inversion, duplicated source of truth, bare sentinel wrap, or error-classification regression was introduced.
- The change is appropriately scoped to the maintenance issue.

### Assessment

The fresh-copy accessor is preferable to an exported mutable map: it removes the drift risk without exposing a mutable guard. Initializing `keysetRejectedColumns` from it is correct because both packages are compiled and initialized together; a later source change to the owned set rebuilds the derived copy.

### Observations

- The existing `strings.ToLower` versus DuckDB non-ASCII identifier-folding limitation remains unchanged, as intentionally scoped to #550.
- Existing cursor and filter contracts remain covered, while the new tests ensure future additions to the `sqlgen` dedup set cannot silently omit the cursor guard.

### Verification

- Reviewed all 2 commits and the complete `main...HEAD` diff (4 files, +136/-27).
- `make fmt-check` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/federated` passed.
- CI is fully green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

### Summary

No findings. The PR satisfies #531 and is ready to merge.

PR #553, opened 2026-09-09
fix(entity): make Update a guarded read-modify-write (#457)

### LINKED ISSUE #457: bug: Update is an unguarded read-modify-write — concurrent updates to one row silently lose the whole document
## Problem

`Update` is an unguarded read-modify-write: the merge base is read outside the write transaction, there is no optimistic-concurrency check on write-back, and the EAV write replaces the whole attribute set. Two concurrent updates to one row silently lose the entire document of the first committer.

## Current behavior

`entityCRUDService.Update` (`internal/entity_crud_service.go`):
1. reads the current record via `repository.GetPersistentRecord` — which runs on `r.pool` **outside any transaction** (`loadMainRecord`/`fetchAttributes` use `r.pool.QueryRow`/`r.pool.Query`, and read `entity_main` and `eav_data` as two separate statements — a second correctness gap, below);
2. builds `mergedData = StripComputedFields(mergeMaps(existingData, req.Updates))`;
3. calls `repository.UpdatePersistentRecord`, which only now opens its transaction and `replaceEAVAttributes` **deletes every current-schema EAV row and reinserts the whole document**.

`UpdatePersistentRecord` (`internal/postgres_persistent_repository.go`) takes **no** `pg_advisory_xact_lock` (the create/delete paths deliberately take `lockRowVersion`), and `buildUpdateMainStatement` (`internal/postgres_persistent_repository_main_table.go`) writes `WHERE ltbase_schema_id = $ AND ltbase_row_id = $` with **no** `ltbase_updated_at = <expected>` predicate. The version is advanced (`GREATEST($1, ltbase_updated_at + 1)`) but never *asserted*.

Two related asymmetries in the same path:
- **Torn read → torn persist.** `GetPersistentRecord` reads `entity_main` then `eav_data` in two untransacted statements (possibly on two pooled connections). A concurrent commit between them yields a record mixing version N main columns with N+1 EAV rows. Because `Update` merges onto exactly this record, the torn state is then *persisted*. The list path does not have this problem (`runOptimizedQuery` fetches main + `JSON_AGG`'d EAV in one statement) — only the single-row path splits them.
- **Response echoes unstored input.** `Update` returns `Attributes: mergedData` (the pre-write merge), whereas `Create` returns `transformer.FromPersistentRecord(record)` (the storage round-trip). So the API confirms values it may not have stored; the divergence is invisible until the next `GET`.

## Reproduction

Two `PUT /api/v1/lead/{row}` on the same row, `{"pipeline":"buy"}` and `{"status":"open"}`. Both read the same pre-write snapshot; both `replaceEAVAttributes` delete-all-then-reinsert their own full document. The second committer wins entirely; the first update's field is gone. No error, both return 200. The stale document is also what CDC flushes, propagating the loss to warm/cold tiers.

## Root cause

The update's merge base is read outside the update's transaction, and the write-back neither locks the row nor asserts the version it read. Create/delete take `pg_advisory_xact_lock`; update takes nothing equivalent, and the `lockRowVersion` comment justifies its absence only against *deletes*, not against other updates.

## Impact

Silent lost update on the primary store under any concurrent write to one row — and because EAV update is delete-all-then-reinsert, the loss is the whole document, not one column. Torn single-row reads. API success reports that do not match stored state.

- #274 (closed) is about version *ordering* across tiers, not mutual exclusion between two updates — INDEPENDENT.
- Not #294/#312/#355 (searched `lost update`, `concurrent update`, `read-modify-write`, `advisory lock`, `optimistic`). Epic #452's Phase 3 write-path list does not contain it.

## Acceptance criteria

**Maintainer ruling (2026-08-24):** take the repo's existing mutual-exclusion precedent — the same `(schemaID, rowID)` advisory lock create/delete use (`lockRowVersion`) — **and read the merge base inside that locked write transaction.** This fixes lost updates and the torn read together (the read shares one snapshot with the write; each waiter re-reads post-lock, so last-committer-wins becomes per-merged-document) and makes the storage-round-trip response nearly free. The optimistic version predicate is deferred: a failed assert needs a caller-facing conflict protocol (409 mapping / retry) that does not exist yet.

- [ ] `Update` serializes on the `lockRowVersion` `(schemaID, rowID)` advisory lock, matching create/delete.
- [ ] The merge base is read inside the write transaction (one snapshot), so a concurrent update to a disjoint field cannot silently drop a committed field and single-row reads are not torn.
- [ ] `Update` response reflects stored state (storage round-trip), matching `Create`.

- [ ] Concurrency test: two overlapping updates to disjoint fields of one row both survive.
- [ ] Torn-read test: main + EAV of a returned single-row record belong to one version.

iceboundrock:
Closes #457.

`Update` read its merge base outside the write transaction and wrote back
without a lock or a version assert, so two concurrent updates to one row both
merged onto the same snapshot — and because the EAV write is
delete-all-then-reinsert, the second committer dropped the first committer's
entire document, silently, with both requests returning 200.

Per the maintainer ruling on the issue, this takes the repo's existing
mutual-exclusion precedent rather than an optimistic version predicate:

- `MergePersistentRecord` opens the write transaction, takes the same
  `(schema_id, row_id)` advisory lock create and delete take, reads the row
  **inside** that transaction, hands it to the caller's merge function, and
  writes the result before commit. A waiter is released only after the
  previous holder commits and re-reads that commit, so last-committer-wins
  operates per merged document.
- The transaction stays at READ COMMITTED deliberately: a snapshot isolation
  level would freeze the waiter's view at `BEGIN`, before the lock was
  granted, and hand the merge the very stale base this exists to prevent.
- The single-row read is now one statement (main projection + `JSON_AGG`'d
  EAV, the shape the list path already uses), so main columns and EAV rows
  can never come from two snapshots — on the update path or on a plain `Get`.
- `Update` answers `transformer.FromPersistentRecord(stored)`, matching
  `Create`, instead of echoing the pre-write merge.
- The unguarded `UpdatePersistentRecord` is deleted so no second update path
  survives.

The optimistic `ltbase_updated_at` predicate stays deferred: a failed assert
needs a caller-facing conflict protocol (409 mapping / retry) that does not
exist yet.

## Regression coverage

- `TestMergePersistentRecordConcurrentDisjointUpdatesAllSurvive` — eight
  concurrent writers, disjoint fields, all survive. Verified to fail with
  "attribute 1 was lost" on 3/3 runs when the advisory lock is removed.
- `TestGetPersistentRecordIsNeverTorn` — a reader loops while a writer flips a
  main column and an EAV attribute together; the pair must always agree.
  Verified to fail on 3/3 runs when the two-statement read is restored.
- `TestMergePersistentRecordLocksThenReadsInsideTransaction` — pins lock
  before read, read inside the transaction, and the effective version (#274)
  reaching both `entity_main` and `change_log`.
- `TestEntityManager_Update_ResponseIsStorageRoundTrip` and
  `..._MergeBaseComesFromTheWriteTransaction` — the service seam.

Gates run locally: `make fmt-check`, `make lint`, `make test` (against a real
PostgreSQL, so the integration tests executed rather than skipped),
`make check-infra`, plus `go test -race ./internal/...`. All pass.

## Out of scope

`entityBatchService.batchUpdateAtomic` has the same shape — merge base read
outside the batch transaction, no per-row lock, and a response that echoes
the pre-write merge. Filed as #554; #457's acceptance criteria name `Update`
only, and a multi-row merge callback needs its own lock-ordering design.
Until #554 lands, this PR guards the single-row `Update` entry point (and, by
delegation, best-effort `BatchUpdate`) — not every update API: an atomic
batch racing a guarded `Update` on one row can still clobber it, and the
atomic-batch response still echoes the pre-write merge. `lockRowVersion`'s
godoc says so explicitly.


iceboundrock:
# Review of #553 — guarded Update (#457)

## Spec (against #457 + maintainer ruling 2026-08-24)

#457 requires, via mutual exclusion (not an optimistic predicate): (1) `Update` serializes on the same `(schema_id, row_id)` advisory lock create/delete use; (2) the merge base is read inside the write transaction so disjoint concurrent updates both survive and single-row reads are not torn; (3) the `Update` response is the storage round-trip, matching `Create`.

This PR meets all three:

- `internal/postgres_persistent_repository_merge.go:34` — `MergePersistentRecord` does `BeginTx` → `lockRowVersion` → `loadRecordWithAttributes` on the tx → merge → write → commit. Lock-before-read is the required order.
- `internal/entity_crud_service.go:219` — `Update` delegates the entire read-modify-write to `MergePersistentRecord`, with the 404 authored in `mergeUpdateRecord` (`internal/entity_crud_service.go:245`) where the schema/row names live. Correct layering.
- `internal/entity_crud_service.go:229` — response is `transformer.FromPersistentRecord(ctx, stored)`, matching `Create` (`internal/entity_crud_service.go:130`). The pre-write-merge echo is gone.
- `internal/postgres_persistent_repository.go` — the unguarded `UpdatePersistentRecord` is deleted; `rg` confirms no code references remain (only a historical comment in a test). No second update path survives.
- `internal/postgres_persistent_repository_single_row.go:33` — the single-row read is one statement (main projection + `JSON_AGG`'d EAV, `COALESCE` to `'[]'`), byte-for-byte the same JSON shape the list path emits (`internal/advanced_query_template.go:115`) and decoded through the same `model.ParseAttributesJSON`. `Get` (`internal/postgres_persistent_repository.go:287`) and the merge base share the loader via `singleRowQuerier`, so the torn-read fix covers both paths.
- READ COMMITTED rationale in the `MergePersistentRecord` godoc is correct: a waiter must take a fresh per-statement snapshot after the lock is granted; a snapshot isolation level would freeze the view at `BEGIN`.
- Deferring the optimistic `ltbase_updated_at` predicate is per the ruling (no 409/retry protocol exists yet), and the `#274` `GREATEST(...)+RETURNING` version plumbing is preserved — the mock pin asserts the effective version reaches both `entity_main` and `change_log`.

Required regression coverage from #457 is present: disjoint-concurrent-updates and torn-read pins (see Verification).

## Standards

- File/function sizes: all touched files ≤ 367 lines; `MergePersistentRecord` is 67 lines, `loadRecordWithAttributes` 29, `mergeUpdateRecord` 45, `Update` 53. Within the 500/100 limits.
- Error handling uses wrapped context everywhere on the new path (`lock row version for update`, `load merge base for %s`, `update main row for %s`, `replace eav attributes for %s`, `upsert change log for %s`, `commit transaction`), and new messages name the row/expected state. The bare `return nil, err` for the merge callback is intentional and correct: the callback authors published carriers (`NotFoundf`/`InvalidInputf`), the service seam owns the 404 text, and the outer `Update` wrap (`failed to update persistent record: %w`) preserves `errors.Is` classification while `ResolvePublicMessage` still resolves the inner publication.
- Godoc exists on all new exported surface (`MergePersistentRecord`, `PersistentRecordMerge`, the interface method). `lockRowVersion`'s comment was updated so it no longer claims updates need no lock.
- `gofmt` clean on all touched files; `go vet ./internal/...` clean (run locally, see Verification).

## Findings

1. (Suggestion, non-blocking) Isolation level is load-bearing but implicit. `MergePersistentRecord` calls `BeginTx(ctx, pgx.TxOptions{})` and relies on the cluster default being READ COMMITTED — the godoc explains why that matters, but nothing enforces it. If a deployment ever defaulted to REPEATABLE READ/SERIALIZABLE, the waiter's `SELECT` would snapshot at `BEGIN` (pre-lock) and the stale-base bug would return silently, plus serialization failures. Consider `pgx.TxOptions{IsoLevel: pgx.ReadCommitted}` to make the deliberate choice code. (Insert/Delete share the `TxOptions{}` shape, but for them the level was never load-bearing; for Merge it is.)
2. (Nit, non-blocking) The `merge.go` comment "A prefix here would land in that body" is imprecise under the #313 semantics: a plain `fmt.Errorf("...: %w", mergeErr)` prefix stays in `Error()`/logs only — `ResolvePublicMessage` walks past the plain wrapper to the inner publication. Only `WrapPublicf` would move text into the body. The bare return itself is fine (avoids double context; the service wrap already adds operator context), but the comment teaches the wrong rule. Suggest rewording to "kept bare so the service seam owns the published message and the log line".
3. (Doc gap, non-blocking) The `PersistentRecordMerge` contract (`internal/model/interfaces.go:45`) does not state that merge must carry `CreatedAt`/`DeletedAt` (and typed-column maps) forward — only `UpdatedAt` is stamped by the repository. All current callers do (service copies both; tests copy `CreatedAt`), and the DB `UPDATE` does not rewrite `CreatedAt` anyway, but a future direct caller that returns a sparse record would get a corrupt response object. One sentence in the interface godoc would close it.

No blocking findings. I verified there is no silent upsert (a merge result for an absent row falls into `updateMainRow`'s `ErrNoRows` → `NotFound`), no changelog write on the not-found path (write happens only after a successful merge return), and the `#294` tolerate-and-preserve scope (`knownAttrIDs`) still gates the EAV replace inside the same transaction.

## Observations

- Correctness win beyond the acceptance text: `Update` vs `Delete` vs `Create` now all serialize on the same advisory key, so the update window where a concurrent delete could interleave between base-read and write is closed too.
- Best-effort `BatchUpdate` inherits the fix (it delegates per-op to `crud.Update`). The remaining unguarded shape is exactly what the PR body discloses: `batchUpdateAtomic` reads bases via `GetPersistentRecord` outside the batch tx (`internal/entity_batch_service.go:296`) and `BatchUpdatePersistentRecords` (`internal/postgres_persistent_repository_batch.go:66`) takes no `lockRowVersion`, so batch-vs-single and batch-vs-batch on one row still have the old whole-document race. Acceptable as out of scope for #457 (criteria name `Update` only), but the follow-up issue should cover both the service-layer base read and the repo-layer lock — and note the atomic-batch success records still echo pre-write `mergedData` (`internal/entity_batch_service.go:335`), the same echo bug this PR fixes for single `Update`.
- Perf: the single-row `LEFT JOIN LATERAL ... JSON_AGG` scans the same EAV rows the old second statement fetched; no new index requirement, and it matches the list path's per-row cost shape. Correctness justifies it.
- Deleted-code hygiene is good: `fetchAttributes`/`loadMainRecord` and their dedicated tests are removed, with their scan-coverage carried over into `TestGetPersistentRecordScansEveryColumnKind`.

## Verification (run locally on this branch)

- Targeted unit + mock suite — PASS:
  `TestMergePersistentRecord{LocksThenReadsInsideTransaction, WhenRowMissingHandsNilAndWritesNothing, ConcurrentDisjointUpdatesAllSurvive, NoSchemaCache_FailsBeforeDelete}`, `TestGetPersistentRecord{ReadsMainAndEAVInOneStatement, ScansEveryColumnKind, SingleStatementNotFound}`, `TestEntityManager_Update_{MergesAndPreserves, InvalidOptionalValue..., ResponseIsStorageRoundTrip, MergeBaseComesFromTheWriteTransaction}`, `TestInsertNilRecordAndMergeNilFunctionAreRejected`
- Real-Postgres integration (local PG available, so not skipped) — PASS:
  `TestMergePersistentRecordConcurrentDisjointUpdatesAllSurvive` (8 writers), `TestGetPersistentRecordIsNeverTorn` (300 paired reads), `TestSameMillisecondWritesStayStrictlyOrderedIntegration` (#274), `TestChangeLogWritesOnUpdateAndDeleteIntegration`
- `gofmt -l` on all touched files: clean. `go vet ./internal/ ./internal/model/`: clean.
- Did not re-run the claimed 3/3 fail-on-old-code mutation checks or the full `make test`/`make lint` gates; trusting the PR description on the former.

## Summary

Approve direction. The change implements exactly the maintainer-ruled mutual-exclusion fix, removes the unguarded path rather than leaving a second one, fixes the torn single-row read for both `Get` and `Update` with one shared loader, aligns the `Update` response with `Create`, and pins each behavior with the regression test #457 asked for. The three findings above are follow-ups, not merge blockers. Please file/link the `batchUpdateAtomic` follow-up (lock + in-tx base + response echo) so the disclosed residual race is tracked.

iceboundrock:
# Response to the review (f990e23)

All three findings taken. One commit: `fix(storage): pin READ COMMITTED on the merge transaction`.

## 1. Isolation level implicit → pinned, and asserted

`MergePersistentRecord` now opens with `pgx.TxOptions{IsoLevel: pgx.ReadCommitted}` instead of `pgx.TxOptions{}`. The reasoning in the finding is the reasoning: for Insert/Delete the level was never load-bearing, but here a cluster defaulting to REPEATABLE READ would snapshot the waiter's `SELECT` at `BEGIN` — before the lock was granted — and silently restore the exact stale base this PR exists to remove. A comment cannot stop that; a `TxOptions` field can.

The godoc changed with it, since "deliberately stays at the default" was no longer true:

> The transaction pins READ COMMITTED explicitly rather than inheriting the cluster default, because here the level is load-bearing: […] READ COMMITTED takes a fresh snapshot per statement, so the post-lock read sees the previous holder's commit.

It is also pinned in a test rather than only in code. `TestMergePersistentRecordLocksThenReadsInsideTransaction` now expects `mock.ExpectBeginTx(pgx.TxOptions{IsoLevel: pgx.ReadCommitted})`. This is a real assertion, not a formality: pgxmock's plain `ExpectBegin()` does **not** match a `BeginTx` carrying options, so reverting the repository to `TxOptions{}` — or changing the level to anything else — fails with `call with transaction options '{read committed ...}' was not expected`. Confirmed by running it against the un-pinned code before the fix landed. The two other merge-path mocks (`...WhenRowMissingHandsNilAndWritesNothing`, `TestMergePersistentRecord_NoSchemaCache_FailsBeforeDelete`) were updated to the same expectation; the delete/batch-delete mocks keep plain `ExpectBegin()`, since those paths still take `TxOptions{}` and the level genuinely is not load-bearing there.

## 2. `merge.go` comment taught the wrong #313 rule → reworded

Correct on the semantics: `ResolvePublicMessage` walks past a plain `fmt.Errorf("...: %w", …)` wrapper to the inner publication, so a prefix there would have stayed in `Error()` and the log, not the response body. Only `WrapPublicf` moves text into the body. The bare return is kept — the reason is just different from the one the comment gave:

```go
// Deliberately a bare return rather than the usual wrap: merge is the
// caller's own body, which has already attached its context, and its error
// is frequently a published forma carrier (NotFoundf/InvalidInputf). Kept
// bare so the service seam owns both the published message and the log
// line; a wrap here would only duplicate context the caller already set.
```

## 3. `PersistentRecordMerge` contract silent on completeness → documented

Extended the interface godoc (`internal/model/interfaces.go`). While writing it the asymmetry turned out to be sharper than "carry `CreatedAt`/`DeletedAt` forward", so the doc states each field's actual failure mode rather than a blanket rule:

- **EAV attributes** — `replaceEAVAttributesScoped` deletes every in-scope `attr_id` then reinserts what the result carries, so a dropped attribute is *deleted from storage*, not merely missing from the answer.
- **Typed columns** — `buildUpdateMainStatement` only emits assignments for the maps present, so a dropped column *keeps its stored value* but is missing from the answered record.
- **`DeletedAt`** — always assigned (`ltbase_deleted_at = $n`, nil → NULL), so dropping it clears a stored tombstone. That column is the soft-delete filter on the CDC export path (`ltbase_deleted_at IS NULL`, `internal/cdc/export_sql_builder.go:69`), so this is not cosmetic.
- **`CreatedAt`** — never written, only echoed, so an omitting result answers a zero timestamp.
- **`UpdatedAt`** — the one field the repository stamps itself.

Both current callers already satisfy this; the doc is there for the next direct caller.

## Follow-up filed

#554 (`bug: batchUpdateAtomic is an unguarded read-modify-write, same lost-update shape as #457`) was already open and covers both halves the review asked for: the service-layer base read outside the batch transaction (`internal/entity_batch_service.go:296`) and the missing repo-layer `lockRowVersion` in `BatchUpdatePersistentRecords` (`internal/postgres_persistent_repository_batch.go:66`), with a `BatchMergePersistentRecords` shape and deterministic `(schemaID, rowID)` lock ordering sketched.

The response-echo half was **not** in it, so it has been added as a scope addendum: https://github.com/Lychee-Technology/forma/issues/554#issuecomment-5628240838 — `successful[i].Attributes = mergedData` (`internal/entity_batch_service.go:335`) is the pre-write map, so the answered `ltbase_updated_at` is the service's client-side `nowMillis()` rather than the version `GREATEST($n, ltbase_updated_at + 1) RETURNING` actually stored (#274). Since #553 the batch path is the only writer answering with something it never read back. The PR body's "Out of scope" paragraph now names #554 and the echo explicitly.

## Gates re-run on f990e23

- `make fmt-check` — clean.
- `make lint` — clean (root + `infra/`).
- `make test` via `./scripts/test_with_container_runtime.sh` — all packages pass, `internal/e2e_harness` included. (A bare `make test` in this shell fails `TestE2EHarnessMinimal` with testcontainers' `checked path: $XDG_RUNTIME_DIR` panic, i.e. no `DOCKER_HOST` exported — environmental, which is what that script exists to set.)
- Targeted: `TestMergePersistentRecord*`, `TestGetPersistentRecord*`, `TestEntityManager_Update*`, `TestDeletePersistentRecord*`, `TestBatchDelete*`, `TestInsertNilRecordAndMergeNilFunctionAreRejected` — PASS, against a real local PostgreSQL (the integration tests executed rather than skipped).

No behavior change to the write path itself: the isolation level is what the code already assumed, now stated and enforced.

iceboundrock:
# Review of #553 (latest head `f990e23`)

## Findings

1. **P1 follow-up: atomic batch updates still have the lost-update race**

   `internal/entity_batch_service.go:296-342` reads each merge base through `GetPersistentRecord` before the write transaction, while `internal/postgres_persistent_repository_batch.go:66-105` opens a separate transaction without taking `lockRowVersion`. Two batch updates, or a batch update racing the now-guarded single-row `Update`, can therefore merge from the same stale full document; the later EAV delete-and-reinsert can silently erase the earlier writer's fields while both requests succeed. This is the same data-loss shape fixed here for single-row `Update`. The PR body discloses this and #554 tracks it, so I treat it as a scoped follow-up rather than a blocker for #457, but the repository should not be described as protecting all update entry points until #554 is fixed.

2. **P2, non-blocking: merge callback errors are returned without repository context**

   `internal/postgres_persistent_repository_merge.go:70-72` uses `return nil, err`. The repository standard requires errors to carry operation context. A plain `fmt.Errorf("merge record for %s: %w", rowID, err)` would preserve `errors.Is` and `ResolvePublicMessage` behavior for `forma.NotFoundf`/`InvalidInputf`, while giving direct repository callers an actionable row context.

3. **P2, non-blocking: commit message violates the repository attribution rule**

   Commit `f990e23` contains a `Claude-Session: https://claude.ai/...` footer. `AGENTS.md` explicitly prohibits AI attribution in commit messages; remove that footer before merge.

## Spec

The single-row acceptance criteria from #457 and the maintainer ruling are met:

- `MergePersistentRecord` takes the same `(schema_id, row_id)` advisory lock used by create/delete.
- The merge base is read after the lock, inside the write transaction.
- The transaction explicitly pins `READ COMMITTED`, which is required for a post-lock statement snapshot.
- `GetPersistentRecord` and the merge path share a one-statement main-table plus EAV loader, eliminating the torn two-statement read.
- `Update` returns the transformed storage round-trip, including PostgreSQL's effective `ltbase_updated_at`, rather than echoing the pre-write merge.
- The deferred optimistic predicate is consistent with the ruling because no 409/retry protocol exists yet.

## Standards

- Touched source files remain within the 500-line file and 100-line function limits.
- Exported merge API additions have documentation.
- `gofmt`, lint, vet, and diff whitespace checks are clean.
- The bare merge error and `Claude-Session` commit footer are the only standards issues found.

## Assessment

The single-row implementation is technically sound and addresses the stated #457 scope. Request the two non-functional cleanups above before merge, and keep #554 linked as the explicit follow-up for atomic batch updates. If #553 is intended to claim concurrency safety for every update API, then the batch race becomes a merge blocker; under the PR's explicitly limited scope, it is a tracked follow-up.

## Observations

- The new advisory lock also closes the update-versus-delete interleaving window, not just update-versus-update.
- The lateral `JSON_AGG` loader matches the existing list-path attribute shape and does not introduce a new index requirement.
- The new concurrency coverage primarily exercises the repository merge seam; there is no regression test for atomic batch versus single-row update, which is the scenario covered by #554.
- The batch path also still constructs successful response records from the pre-write merge, so its returned version can differ from the stored effective version; this is included in #554's scope addendum.

## Verification

- `make fmt-check`: PASS
- `make lint`: PASS
- `go vet ./internal/...`: PASS
- Targeted merge, single-row, update, batch, and version tests: PASS
- `./scripts/test_with_container_runtime.sh`: PASS with Podman, including the e2e harness
- `DOCKER_HOST=unix:///run/user/1000/podman/podman.sock TESTCONTAINERS_RYUK_DISABLED=true go test -race ./internal/...`: PASS
- `git diff --check`: PASS
- GitHub PR checks (Test, Lint, E2E Smoke, Benchmark Smoke, k6 Smoke): PASS
- A direct `make test` without `DOCKER_HOST` failed only because testcontainers could not discover a container runtime; rerunning through the repository's runtime wrapper passed.

## Summary

The guarded single-row read-modify-write fix is correct for #457, with no blocking functional finding in that stated scope. Please remove the attribution footer, add context to the merge callback error, and keep the batch lost-update/response-round-trip work tracked in #554.

iceboundrock:
# Response to the review of `f990e23`

Branch rewritten and force-pushed; head is now `d6065c6`. Two things changed: one new commit (`fix(storage): wrap the merge callback's error with the row it was merging`) and one message-only rewrite of the READ COMMITTED commit (`f990e23` → `630f9f2`). Tree at `630f9f2` is byte-identical to `f990e23`; the five earlier commits are untouched.

## 1. P1 follow-up — atomic batch race

Not fixed here, per the stated scope; #554 stays the tracked follow-up and already carries everything this finding and the observations ask for: the service-layer base read outside the batch transaction, the missing repo-layer `lockRowVersion` in `BatchUpdatePersistentRecords`, the pre-write response echo (scope addendum), and the batch-vs-single-`Update` regression test in its "Regression coverage to add" list. I've labeled it `bug`.

The "should not be described as protecting all update entry points" caveat did point at a real overclaim, though — in code, not only in prose. `lockRowVersion`'s godoc said "Updates take it too (#457)" and "Batch writers acquire these locks in input order", and neither is true of `BatchUpdatePersistentRecords`, which takes no lock at all. Reworded (`internal/postgres_persistent_repository.go:112`):

> Single-row updates take it too (#457) […] Batch insert and batch delete acquire these locks in input order […] Batch update does not take it yet and still has the #457 lost-update shape; that is #554.

The PR body's "Out of scope" paragraph now states the same limit explicitly: this guards the single-row `Update` entry point (and best-effort `BatchUpdate`, which delegates to it), not every update API.

## 2. P2 — merge callback error returned bare → wrapped

Taken. My previous comment's reasoning for the bare return ("a wrap here would only duplicate context the caller already set") does not hold up: the service wrap is `failed to update persistent record: %w` and names neither the row nor the step, and a validation or transform failure inside the callback carries no row id of its own. So the bare return was dropping context, not avoiding a duplicate.

`internal/postgres_persistent_repository_merge.go:70` is now

```go
record, err := merge(ctx, existing)
if err != nil {
    return nil, fmt.Errorf("merge record for %s: %w", rowID, err)
}
```

with the comment reduced to the one rule that matters there: plain wrap, never `WrapPublicf`, because `ResolvePublicMessage` (`client_error.go:47`) steps through a plain `Unwrap()` wrapper to the callback's published carrier, so the row context lands in `Error()` and the log while the 404/400 body stays the caller's message (#313).

Both halves are pinned rather than asserted in prose. `TestMergePersistentRecordWhenRowMissingHandsNilAndWritesNothing` now checks

```go
assert.ErrorContains(t, err, "merge record for "+rowID.String())
msg, ok := forma.ResolvePublicMessage(err)
require.True(t, ok)
assert.Equal(t, "entity not found: mock_schema/"+rowID.String(), msg)
```

so reverting to the bare return fails the first assertion, and switching to `WrapPublicf` (or otherwise leaking the prefix into the body) fails the third. `TestNoBareSentinelWraps` still passes — the wrap is over a variable, not a sentinel.

## 3. P2 — `Claude-Session` footer → removed

Stripped from `f990e23` with `git filter-branch --msg-filter` over `main..HEAD`, which is why the branch was force-pushed (`--force-with-lease` against `f990e23`). `git log main..HEAD --format=%B | grep -c Claude-Session` is now `0`; the new commit carries no footer either.

## Gates re-run on `d6065c6`

- `make fmt-check` — clean.
- `make lint` — clean (root + `infra/`).
- `git diff --check` — clean.
- `make test` via `./scripts/test_with_container_runtime.sh` — all packages pass, `internal/e2e_harness` included.
- With a local PostgreSQL 16 up, so executed rather than skipped: `TestMergePersistentRecordConcurrentDisjointUpdatesAllSurvive`, `TestGetPersistentRecordIsNeverTorn`, `TestSameMillisecondWritesStayStrictlyOrderedIntegration`, `TestChangeLogWritesOnUpdateAndDeleteIntegration` — PASS. Plus the merge/get/update mock suite and `TestNoBareSentinelWraps` — PASS.

iceboundrock:
# Review of #553 (head `d6065c6`)

## Findings

1. **P1 follow-up, scoped and non-blocking for #457: atomic batch updates still have the lost-update race.** `internal/entity_batch_service.go:296-342` reads each merge base through `GetPersistentRecord` before the write transaction, while `internal/postgres_persistent_repository_batch.go:66-105` writes without `lockRowVersion`. Two atomic batch updates, or an atomic batch update racing the guarded single-row `Update`, can therefore merge from the same stale document; the later EAV delete-and-reinsert can silently erase the earlier writer's fields while both requests succeed. The batch response also still echoes pre-write `mergedData` rather than the effective stored version. The PR body discloses this and #554 tracks the required in-transaction merge, deterministic locking, and response round-trip work. This is not a blocker under the stated single-row `Update` scope, but #553 should not be read as protecting every update entry point.

No additional merge-blocking or correctness findings were found in the stated scope.

## Spec

- `MergePersistentRecord` begins a `READ COMMITTED` transaction, takes the same `(schema_id, row_id)` advisory lock used by create/delete, reads the merge base after the lock through the transaction, runs the callback, replaces the in-scope EAV set, writes the main row and change log, and commits as one unit.
- `Update` delegates the complete read-modify-write to that repository operation and authors the missing-row 404 at the service layer.
- `GetPersistentRecord` and the merge path share a single main-table plus EAV statement, eliminating the previous torn two-statement read.
- `Update` transforms the committed/stored record for its response, including PostgreSQL's effective monotonic `ltbase_updated_at`.
- Deferring an optimistic version predicate is consistent with the maintainer ruling because the repository has no caller-facing 409/retry protocol.

## Standards

- Touched source files and functions remain within the repository's 500-line and 100-line limits.
- New exported interface/API documentation is present.
- New errors carry operation and row context, preserve `errors.Is`, and keep operator-only context out of published client messages.
- `gofmt`, lint, vet, and whitespace checks are clean.

## Assessment

The implementation is technically sound for the requested single-row `Update` behavior and addresses the #457 acceptance criteria. There are no merge blockers in that scope. Keep #554 linked as a required follow-up before claiming concurrency protection for all update APIs.

## Observations

- The shared advisory lock also closes the update-versus-delete interleaving window.
- The lateral `JSON_AGG` loader matches the existing list-path attribute shape and does not require a new index.
- The repository contract correctly documents that EAV attributes replace the current-schema set, while omitted typed columns retain their stored database value but are absent from the returned record; current service callers carry the complete merged document forward.
- The new concurrency tests cover repository-level disjoint updates and torn reads. A batch-versus-single update regression remains appropriately assigned to #554.

## Verification

- `make fmt-check`: PASS
- `make lint`: PASS
- `make check-infra`: PASS
- `go vet ./internal/...`: PASS
- Targeted merge, single-row, update, delete, batch-delete, and version tests: PASS
- `./scripts/test_with_container_runtime.sh`: PASS using the local Podman socket, including the E2E harness
- `go test -race ./internal/...`: PASS with the local Podman endpoint configured
- `git diff --check`: PASS
- GitHub PR checks: Test, Lint, Benchmark Smoke, and k6 Smoke PASS; E2E Smoke was still pending at the time of posting
- A direct `make test` without `DOCKER_HOST` hit the documented testcontainers runtime-discovery failure; rerunning through the repository wrapper passed

## Summary

Approve direction. No merge-blocking findings were found for #457's single-row scope. The only residual correctness risk is the explicitly disclosed atomic batch update path; #554 should remain the tracking issue for its lock, transaction-boundary, and response-round-trip fixes.

iceboundrock:
# Response to the review of `d6065c6`

One finding, no code change requested for this PR. Disposition below, with the alternatives I weighed before leaving it in #554.

## 1. P1 follow-up — atomic batch race stays in #554

Confirmed against `d6065c6`: `batchUpdateAtomic` reads each merge base on the pool at `internal/entity_batch_service.go:296`, before `BatchUpdatePersistentRecords` (`internal/postgres_persistent_repository_batch.go:66`) opens its transaction; that transaction takes no `lockRowVersion`; and `successful[i].Attributes = mergedData` at `:335` still answers the pre-write map. All three halves are in #554 (body + scope addendum), which is labeled `bug`.

I weighed four ways to answer this finding before settling on "leave it tracked":

| Approach | Outcome |
|---|---|
| **Defer to #554** (chosen) | Matches #457's acceptance criteria and the 2026-08-24 ruling, which name `Update` only. Post-merge state is strictly better than pre-merge: Update-vs-Update and the torn read are gone, and no new failure mode is introduced. |
| **Fix in this PR** (`BatchMergePersistentRecords`) | This is #554 in full, and it carries an unresolved design decision. Batch insert and batch delete lock in **input order** (`postgres_persistent_repository_batch.go:37`, `:149`); #554 sketches a **sorted** `(schemaID, rowID)` order for batch update so two overlapping updates cannot invert. A sorted batch update racing an input-order batch delete over the same rows can still invert and deadlock, so the lock discipline has to be chosen once for all three batch writers — and a deadlock abort needs a caller-facing answer (the 40P01 story #554 already flags). That is a ruling for the issue, not something to decide inside a PR scoped to `Update`. |
| **Add `lockRowVersion` at write time in `BatchUpdatePersistentRecords`** | False comfort. The merge base is still read before the transaction, so the stale merge is persisted exactly as today; the lock only adds wait time at the write. Rejected. |
| **Route atomic batch through per-row `MergePersistentRecord`** | Gives up the batch's all-or-nothing commit. Rejected. |

One fact that sharpens the residual risk, and that was not on #554 yet: **the HTTP API has no batch-update route.** `internal/httpapi/handlers.go` calls `manager.Update` (PUT), `manager.BatchCreate`, and `manager.BatchDelete` only; nothing in `internal/httpapi` or `cmd/` invokes `BatchUpdate`. So after this PR the whole HTTP write surface is guarded; the #554 race is reachable only through the Go library API (`forma.EntityManager.BatchUpdate` with `Atomic: true`, `storage.go:21`). Still a real bug — Forma is a library first — but a library-caller exposure, not an endpoint one. Recorded on #554 as an exposure note, and #457 now carries a comment recording the split so the issue is the complete decision record.

Wording stays as it is in the branch: `lockRowVersion`'s godoc (`internal/postgres_persistent_repository.go:122`) names batch update as unguarded and points at #554, and the PR body's "Out of scope" paragraph says the same. Nothing else in the tree or `docs/` claims all-updates protection (`git grep` over `#457`, `advisory`, `lost update`, `read-modify-write` in `docs/` and `README.md`).

## Verification since the review

The one check that was pending when the review was posted has since completed: GitHub PR checks on `d6065c6` are all green — Test, Lint, E2E Smoke Tests, Benchmark Smoke, k6 Smoke.

No new commits; head remains `d6065c6`.

Issue #555, opened 2026-09-12
bug: DuckDB hot-leg projection does not normalise iso8601-encoded date/datetime main columns


iceboundrock:
Found during the final review of #459 (write-path value/column fidelity); pre-existing, not introduced by that branch.

## Symptom

`internal/sqlgen/duckdb_schema_projection.go` (`buildPGSource`, the `a.isColumn` loop) special-cases only `bool` (`mainColBoolExpr`) and `uuid` (`CAST(m.<col> AS VARCHAR)`) bound columns. A `date`/`datetime` attribute bound to a text column with `encoding: iso8601` falls through to the generic branch:

```sql
COALESCE(ANY_VALUE(hot_vals.<attr>), m.text_NN) AS <attr>
```

`hot_vals.<attr>` is the EAV pivot's epoch-ms `BIGINT` (`TRY_CAST(value_numeric AS BIGINT)`, #173), while `m.text_NN` is the ISO-8601 `VARCHAR`. The two COALESCE arms disagree in type, so the hot leg of a federated read either fails to bind or implicitly casts the string, and either way the value does not match the epoch-ms convention every other tier projects (#194, #200).

## Why it matters now

#459 codifies `date`/`datetime` → text (`iso8601`) as an admitted pair in `schemameta.ValidateColumnBinding` because the write path, the Postgres read path and CDC export (`castDateMainValue`, #219) all round-trip it. The DuckDB hot-leg projection is the one surface that was not verified. No shipped fixture uses `iso8601` today, so nothing regresses, but the matrix should not admit a pair one tier cannot read.

## Fix sketch

Add an `iso8601` arm next to the `bool` one that normalises `m.text_NN` to epoch-ms `BIGINT` on the DuckDB side (e.g. `CAST(epoch_ms(TRY_CAST(m.text_NN AS TIMESTAMP)) AS BIGINT)`), mirror it in whatever the benchmark projection does for the same column, and add a federated e2e case that binds a datetime attribute with `iso8601`, writes a row, and reads it back through the hot leg.

Until then, `binding_compat.go` notes the pair as admitted with the DuckDB hot leg unverified (pointing here).

iceboundrock:
Status update after PR #556: the admitted date/datetime to text (iso8601) binding remains intentionally unresolved on the DuckDB hot-leg projection. PR #556 handles write-path fidelity and registration validation, but explicitly tracks this read-path normalization as a follow-up. Issue remains open.

[...]

PR #556, opened 2026-09-12
fix: enforce value/column fidelity on the write path and refuse non-round-trippable bindings (#459)

### LINKED ISSUE #459: bug: write path does not validate value/column type fidelity — narrow-numeric main columns silently wrap, valueType↔column mismatches 500 or drop values
## Problem

The write path does not validate that a value fits the physical column its attribute is bound to, nor that the attribute's `valueType` agrees with the column's storage encoding. The result is three silent-corruption / wrong-error-class shapes on the primary store, reachable on shipped schemas.

## Current behavior

**1. Narrow-numeric main columns silently wrap.** `storeInMainColumn` (`internal/transform/persistent_record.go`):

```go
case forma.MainColumnTypeSmallint:
    record.Int16Items[columnName] = int16(*attr.ValueNumeric)   // no range check
case forma.MainColumnTypeInteger:
    record.Int32Items[columnName] = int32(*attr.ValueNumeric)   // no range check
```

`populateTypedValue` funnels the whole numeric family through `numutil.Float64` with no bound check (`internal/transform/typed_value.go`), and narrowing an out-of-range `float64` in Go wraps. Verified on amd64: `int16(40000.0) = -25536`, `int32(3e9) = -2147483648`. Shipped fixture `internal/e2e_harness/production/schemas/e2e_wide_attributes.json` binds `count` (integer) → `integer_01` and `rank` (smallint) → `smallint_01`.
- Create: `e2e_wide.json` declares `count` as `type: integer` with **no maximum**, so JSON Schema passes and `3e9` wraps into `entity_main`.
- Update: `rank` *does* declare `min/max`, but updates are report-only by default (`Entity.ValidateUpdatesStrict` defaults false), so `40000` is logged and stored as `-25536`.

**2. `valueType`↔column-encoding mismatch on a shipped schema → redacted 500 (create path).** `cmd/server/schemas/log_attributes.json` binds `leadId` (`valueType: text`) to `uuid_02`. `log.json` constrains `leadId` only by `pattern: ^[a-zA-Z0-9_-]+$`. `POST /api/v1/log {…,"leadId":"abc"}`: JSON Schema passes; `populateTypedValue` stores `ValueText="abc"` (valueType is text, `toUUID` never runs); `storeInMainColumn`'s UUID branch (derived from the column *name prefix*) calls `uuid.Parse("abc")` → plain `fmt.Errorf` (no `forma.InvalidInputf` carrier) → `classifyManagerError` finds no sentinel → **HTTP 500, redacted body** for pure caller input.

**3. The mirror mismatch silently drops the value.** `text` bound to `smallint_01` reads `attr.ValueNumeric` (nil for a text value), the `if attr.ValueX != nil` guard has no `else`, so nothing is stored — 200 with the value echoed back (see shape 4).

**4. Update response echoes unstored input.** `Update` returns the caller's `mergedData`, not the storage round-trip (Create does the opposite), so shapes 1 and 3 are confirmed to the caller with the value they *sent*, not the value stored.

## Root cause

The type funnel is split three ways — `valueType` decides validation, the column *name prefix* decides storage encoding, and JSON Schema decides admissibility — and nothing checks the three agree or that a value fits the target column width. `internal/schemameta/validation.go` checks binding *collisions* only, not value/column compatibility.

## Impact

Outright corruption of `entity_main` (wrong stored integer), a reachable redacted 500 on caller input on a shipped schema (create path, not gated by report-only mode), and silent value drops — all confirmed to the caller as success where they do not error.

- #384 (open) shares the write-side non-validation root cause but is scoped to EAV-only integer/smallint (value preserved in `value_numeric`, lost only to DuckDB's `TRY_CAST`) and its read-side remedy cannot fix `entity_main` corruption. Cross-referenced there; this issue owns the write-side main-column + encoding-mismatch scope.
- #314 (closed) enforcement is real, but update-path validation is report-only by default, which is what makes shape 1 silent on updates.
- Not #402 (that is the `1e400` decode site in `schemavalidate`), #296, or #312. INDEPENDENT.

## Acceptance criteria

**Maintainer ruling (2026-08-24):** the write-side range/fidelity rule is **one shared contract with #384's write half** (both cases funnel through `populateTypedValue`) — *the value must fit its physical destination (bound column) or, EAV-only, its declared type, else `forma.InvalidInputf`.* Write the ruling once and reference it from #384; do not answer them separately (epic sequencing constraint).

- [ ] One `populateTypedValue` rule rejects a value that does not fit its physical destination (bound column) or EAV declared type, as a published `forma.InvalidInputf` — shared with #384's write half.
- [ ] Registration/validation rejects a `valueType`↔column-encoding binding that cannot round-trip (e.g. `text`→uuid/smallint column), rather than deferring to a runtime 500 / silent drop.
- [ ] A numeric value out of range for its bound narrow column is rejected as caller input, never wrapped into storage.
- [ ] The mismatch-driven error path uses a `forma.InvalidInputf` carrier, not a redacted 500.
- [ ] `Update` response reflects stored state.
- [ ] `validate-schema-consistency` is extended to detect `valueType`↔column-encoding mismatches across an already-deployed schema set (registration checks only guard *future* schemas; shipped schemas like `log_attributes.json`'s `leadId`→`uuid_02` carry the live shape today).

- [ ] Out-of-range value for `smallint`/`integer` main columns → 4xx, not stored-and-wrapped.
- [ ] `text`→uuid-column and `text`→numeric-column bindings → rejected at registration or a published 4xx, not 500 / silent drop.

iceboundrock:
Closes #459

## Problem

The write path checked a value against its *declared* `valueType` (#384) but never against where the value was physically going. Four shapes fell through:

1. A numeric value out of range for a **bound narrow column** silently wrapped: `integer`→`smallint_01` with `40000` stored `-25536`; `numeric`→`integer_01` with `3e9` wrapped in `int32()`.
2. A `text` attribute bound to a **`uuid_*` column** (the shipped `log` schema's `leadId`/`visitId`) answered a redacted **500** on `uuid.Parse` failure.
3. A `text` attribute bound to a **numeric column** left the typed slot empty and the value was **silently dropped** from the main table.
4. Nothing refused such bindings at registration or metadata load, and `validate-schema-consistency` did not report them across a deployed set.

## Contract

**Write-side fidelity rule (shared with #384; one funnel, `transform.populateTypedValue` → `checkStorageFit`).** A value must fit its physical destination, else `forma.InvalidInputf` (published 4xx):

1. EAV-only (no binding): fit the declared type — `smallint`/`integer`/`bigint` must be integral and inside `[-2^15,2^15)`, `[-2^31,2^31)`, `[-2^63,2^63)`; `numeric` unconstrained (#205). *(Already enforced by #384; unchanged.)*
2. Bound to a main column: rule 1 **and** the column's own width — any value serialized from the numeric slot into `smallint_*`/`integer_*`/`bigint_*` must be integral and inside that column's range regardless of declared type or encoding (so `numeric`→`integer_01` rejects `1.5` and `3e9`; `integer`→`smallint_01` rejects `40000`; a `date` bound to `integer_01` rejects its epoch millis). `double_*` is unconstrained (#205 owns the float64 ceiling).
3. Bound to `uuid_*`: the text value must parse as a UUID (a `uuid` valueType already guarantees this; a `text` valueType is checked here).
4. The typed slot must match the column family (text-slot value → text/uuid column; numeric-slot value → numeric/encoded column). A mismatch is refused, never dropped.

Messages name the value, the destination and the allowed range, e.g. `value 40000 out of range for bound column smallint_01 (smallint) (allowed [-32768, 32767])`, `text value "abc" is not a UUID: bound column uuid_02 requires one`.

**Registration matrix (`schemameta.ValidateColumnBinding`).** A binding must be one of:

| valueType | column type | encoding |
| --- | --- | --- |
| `text` | text | default |
| `uuid` | uuid | default |
| `smallint`/`integer`/`bigint`/`numeric` | smallint, integer, bigint, double | default (width enforced at write) |
| `date`/`datetime` | bigint | default or `unix_ms` |
| `date`/`datetime` | text | `iso8601` |
| `bool` | smallint | `bool_smallint` |
| `bool` | text | `bool_text` |
| `list` | — | never bindable |

Unknown encodings and unknown valueTypes are rejected. Retired entries are skipped (ledger only). System columns (`ltbase_*`) go through the same matrix via `ColumnType()`. The matrix is enforced at `RegisterSchema` and at every runtime metadata load; only the `validate-schema-consistency` tool loads leniently (`MetadataLoader.DeferColumnBindingCheck()`) so it can report every mismatch in one pass.

## Changes

- **`internal/transform`** — new `storage_fit.go`: `checkStorageFit` (declared-type fit from #384 + bound-column width + uuid parse + slot/family match) called once at the end of `populateTypedValue`; `checkDeclaredIntegerFit` folded into it. `storeInMainColumn` split into `storeWithEncoding`/`storeWithDefaultEncoding` returning `(bool, error)` so an empty slot is an error, never a silent drop. Tests: `storage_fit_test.go` (width, family mismatch, non-numeric-family values into narrow columns), `TestPersistentRecordTransformer_RejectsOutOfRangeBoundColumn`, `TestStoreInMainColumn_EmptySlotIsAnError`.
- **`internal/schemameta`** — new `binding_compat.go` (`ValidateColumnBinding`), hooked into `validateSchemaAttributeCacheOpts(name, cache, checkBindings)` for registration and load; `MetadataLoader.DeferColumnBindingCheck()` for the tool.
- **`internal`** — `entity_write_fidelity_test.go` pins the published 4xx (`errors.Is(err, forma.ErrInvalidInput)`, `forma.ResolvePublicMessage`) through the manager seam for create and update, and that no row is written.
- **`cmd/tools`** — `validate-schema-consistency` gains the `valueType/column-encoding binding mismatches` category (`schema=<name> <error>` per line); README entry added.
- **`cmd/server/schemas/log_attributes.json`** — `leadId` `uuid_02`→`text_02`, `visitId` `uuid_01`→`text_03`. Both are `text` attributes; the uuid binding was the shape that 500'd on write (and, incidentally, rendered a `COALESCE(VARCHAR, UUID)` the DuckDB hot leg refuses). Rebinding to text columns keeps the declared type, which is what callers send. Deployments carrying the old copy: run the tool, then the migration SQL in `docs/schema-consistency-migration.md` before upgrading — the server now refuses to load the old binding.
- **docs** — `docs/error-handling.md` gains "Value/column fidelity on the write path (#384, #459)"; `docs/schema-consistency-migration.md` describes the new check, the admitted pairs and the concrete migration SQL.

The AC "Update response reflects stored state" was already delivered by #457 (PR #553): `Update` returns `FromPersistentRecord(stored)` from `MergePersistentRecord`, pinned by `TestEntityManager_Update_ResponseIsStorageRoundTrip`. No change here.

## Deliberate exclusions / follow-ups

- `bigint`→`double_*` precision loss above 2^53 is not checked; #205 owns the float64 ceiling.
- `uuid`→text columns are refused at registration although the transform would round-trip them, because the DuckDB projection assumes a UUID-typed main column for a `uuid` attribute.
- `date`/`datetime`→text `iso8601` is admitted (write path, Postgres read and CDC export round-trip it) but the DuckDB hot-leg projection does not normalise it — pre-existing, tracked in #555; the matrix comment says so.
- `storeWithEncoding` still writes `unix_ms`/`bool_smallint` into `Int64Items`/`Int16Items` without consulting `ColumnType()`. The matrix refuses the mismatched pairs at registration and the main-table writer rejects an unknown key loudly (`unsupported column`), so it is a slot-placement inconsistency rather than a fidelity hole; worth aligning with `checkBoundColumnFit` in a follow-up.

## Verification

- `make fmt-check` clean; `make lint` (golangci-lint v1.64.8, root + infra) clean.
- `./scripts/test_with_container_runtime.sh` (rootless Podman): all 34 packages `ok`, exit 0, on the final HEAD.
- `go test ./internal/e2e_harness/... -timeout=5m -count=1`:
  ```
  ok  	github.com/lychee-technology/forma/internal/e2e_harness	7.660s
  ok  	github.com/lychee-technology/forma/internal/e2e_harness/federated	0.059s
  ok  	github.com/lychee-technology/forma/internal/e2e_harness/federated/benchmark	0.082s
  ok  	github.com/lychee-technology/forma/internal/e2e_harness/production	0.061s
  ```

### DIFF
--- a/cmd/server/schemas/log_attributes.json
+++ b/cmd/server/schemas/log_attributes.json
@@ -21,7 +21,7 @@
   "leadId": {
     "attributeID": 4,
     "valueType": "text",
-    "column_binding": {"col_name":"uuid_02"}
+    "column_binding": {"col_name":"text_02"}
   },
   "noteId": {
     "attributeID": 5,
@@ -56,6 +56,6 @@
   "visitId": {
     "attributeID": 10,
     "valueType": "text",
-    "column_binding": {"col_name":"uuid_01"}
+    "column_binding": {"col_name":"text_03"}
   }
 }
--- a/cmd/tools/README.md
+++ b/cmd/tools/README.md
@@ -10,6 +10,12 @@ Forma Tools CLI
 - `manifest-reconcile`：比对 S3 parquet 对象与 manifest 条目，报告孤儿/dangling，
   `--repair` 补录 #197 flush 孤儿，`--gc` 清理 #188 compaction 遗留。
   详见 `docs/manifest-reconcile.md`。退出码：0 一致 / 2 有差异 / 1 失败。
+- `validate-schema-consistency`：升级前校验 schema 元数据与 EAV 存储的一致性，
+  使用与服务端相同的元数据加载路径。检查项包括：
+  - 重复的 `attributeID` / `column_binding.col_name`；
+  - `valueType`↔column-encoding 绑定无法 round-trip 的情况（#459）；
+  - `eav_data` 中未知的 `attr_id`、存储列错位、list 属性下残留的标量行。
+  详见 `docs/schema-consistency-migration.md`。
 
 generate-attributes
 -------------------
--- a/cmd/tools/validate_schema_consistency.go
+++ b/cmd/tools/validate_schema_consistency.go
@@ -113,7 +113,9 @@ func runValidateSchemaConsistencyOut(ctx context.Context, args []string, out io.
 }
 
 func (v schemaConsistencyValidator) run(ctx context.Context) error {
-	loader := schemameta.NewMetadataLoader(v.pool, v.schemaTable, v.schemaDir)
+	// Deferred so checkColumnBindings can list every mismatch across the
+	// deployed set instead of the loader stopping at the first (#459).
+	loader := schemameta.NewMetadataLoader(v.pool, v.schemaTable, v.schemaDir).DeferColumnBindingCheck()
 	cache, err := loader.LoadMetadata(ctx)
 	if err != nil {
 		return fmt.Errorf("load schema metadata: %w", err)
@@ -171,6 +173,8 @@ func (v schemaConsistencyValidator) report(failures, notices []validationIssue, 
 
 func (v schemaConsistencyValidator) collectIssues(ctx context.Context, cache *schemameta.MetadataCache) ([]validationIssue, error) {
 	var issues []validationIssue
+
+	issues = append(issues, v.checkColumnBindings(cache)...)
 
 	unknownIssues, err := v.checkUnknownAttributeIDs(ctx, cache)
 	if err != nil {
--- a/cmd/tools/validate_schema_consistency_bindings.go
+++ b/cmd/tools/validate_schema_consistency_bindings.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/lychee-technology/forma/internal/schemameta"
+)
+
+// checkColumnBindings reports every active binding whose valueType cannot
+// round-trip through its main column (#459). Registration refuses such a
+// binding for new schemas, but a schema deployed before that guard still
+// carries it — and with it a write path that answers a redacted 500 (text→
+// uuid column) or silently drops the value (text→numeric column). The tool
+// loads through DeferColumnBindingCheck so the whole set is listed at once.
+func (v schemaConsistencyValidator) checkColumnBindings(cache *schemameta.MetadataCache) []validationIssue {
+	var issues []validationIssue
+	for _, schemaName := range cache.ListSchemas() {
+		schemaCache, ok := cache.GetSchemaCache(schemaName)
+		if !ok {
+			continue
+		}
+		for attrName, meta := range schemaCache {
+			if err := schemameta.ValidateColumnBinding(attrName, meta); err != nil {
+				issues = append(issues, validationIssue{
+					category: "valueType/column-encoding binding mismatches",
+					details:  fmt.Sprintf("schema=%s %v", schemaName, err),
+				})
+			}
+		}
+	}
+	return issues
+}
--- a/cmd/tools/validate_schema_consistency_bindings_test.go
+++ b/cmd/tools/validate_schema_consistency_bindings_test.go
@@ -38,8 +38,11 @@ func TestValidateSchemaConsistencyReportsBindingMismatches(t *testing.T) {
 			t.Fatalf("missing %q in output:\n%s", want, got)
 		}
 	}
-	if strings.Contains(got, "name") {
+	if strings.Contains(got, "attribute name (") {
 		t.Fatalf("compatible binding must not be reported:\n%s", got)
+	}
+	if n := strings.Count(got, "- valueType/column-encoding binding mismatches:"); n != 2 {
+		t.Fatalf("expected exactly two binding mismatch lines, got %d:\n%s", n, got)
 	}
 	if err := mock.ExpectationsWereMet(); err != nil {
 		t.Fatalf("EAV checks must still run: %v", err)
--- a/cmd/tools/validate_schema_consistency_bindings_test.go
+++ b/cmd/tools/validate_schema_consistency_bindings_test.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/pashagolub/pgxmock/v4"
+)
+
+// #459: registration guards only future schemas; a deployment already
+// carrying text→uuid_02 has the live 500/silent-drop shape today. The tool
+// lists every such binding across the deployed set as a failure, and still
+// runs the EAV checks (it loads through the deferred-binding loader).
+func TestValidateSchemaConsistencyReportsBindingMismatches(t *testing.T) {
+	schemaDir := t.TempDir()
+	writeSchemaConsistencyArtifacts(t, schemaDir, "contact",
+		`{"type":"object","properties":{"leadId":{"type":"string"},"rank":{"type":"integer"}}}`,
+		`{"leadId":{"attributeID":1,"valueType":"text","column_binding":{"col_name":"uuid_02"}},
+		  "rank":{"attributeID":2,"valueType":"bool","column_binding":{"col_name":"smallint_01"}},
+		  "name":{"attributeID":3,"valueType":"text","column_binding":{"col_name":"text_01"}}}`)
+	mock := newSchemaConsistencyMock(t, pgxmock.NewRows([]string{"schema_id", "attr_id", "record_count"}))
+
+	var out strings.Builder
+	validator := schemaConsistencyValidator{
+		pool: mock, schemaDir: schemaDir, schemaTable: "schema_registry_dev", eavTable: "eav_data_dev", out: &out,
+	}
+	err := validator.run(context.Background())
+	if err == nil || !strings.Contains(err.Error(), "failed with 2 issue(s)") {
+		t.Fatalf("expected two binding failures, got err=%v out=%q", err, out.String())
+	}
+	got := out.String()
+	for _, want := range []string{
+		"- valueType/column-encoding binding mismatches: schema=contact attribute leadId (valueType text) cannot round-trip through main column uuid_02",
+		"- valueType/column-encoding binding mismatches: schema=contact attribute rank (valueType bool) cannot round-trip through main column smallint_01",
+	} {
+		if !strings.Contains(got, want) {
+			t.Fatalf("missing %q in output:\n%s", want, got)
+		}
+	}
+	if strings.Contains(got, "name") {
+		t.Fatalf("compatible binding must not be reported:\n%s", got)
+	}
+	if err := mock.ExpectationsWereMet(); err != nil {
+		t.Fatalf("EAV checks must still run: %v", err)
+	}
+}
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -590,10 +590,16 @@ One rule, one funnel (`transform.populateTypedValue` → `checkStorageFit`):
 - EAV-only attribute: the destination is the declared `valueType`.
   `smallint`/`integer`/`bigint` must be integral and inside the type's
   range; `numeric` is unconstrained (#205 owns its float64 ceiling).
-- Column-bound attribute: the declared type **and** the column's own width.
+- Column-bound attribute: the declared type **and** the column's own width
+  (`double_*` is unconstrained; #205 owns the float64 ceiling, so
+  `bigint`→`double_01` rounds above 2^53 rather than refusing).
   `numeric`→`integer_01` refuses `1.5` and `3e9`; `integer`→`smallint_01`
   refuses `40000`. Before #459 these wrapped (`int16(40000) = -25536`) into
-  `entity_main`.
+  `entity_main`. The width check judges the slot the store actually writes:
+  a declared `bigint` carries an exact int64 sidecar and admits the full
+  int64 range, while a `numeric` value is stored from its float64 image, so
+  `numeric`→`bigint_01` refuses `9223372036854775807` (its image is 2^63,
+  which does not fit).
 - `text`→`uuid_*`: the value must parse as a UUID (published 4xx, not the
   redacted 500 `uuid.Parse` used to raise in `storeInMainColumn`).
 - A value whose typed slot does not match the column family is refused, never
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -582,6 +582,34 @@ already flushed does not re-enter the federated dirty set on a hand-delete, and
 DuckDB keeps serving the stale copy from `/delta/` or `/base/` while
 PostgreSQL-only reads see the value gone.
 
+## Value/column fidelity on the write path (#384, #459)
+
+One rule, one funnel (`transform.populateTypedValue` → `checkStorageFit`):
+**a value must fit its physical destination, else `forma.InvalidInputf`.**
+
+- EAV-only attribute: the destination is the declared `valueType`.
+  `smallint`/`integer`/`bigint` must be integral and inside the type's
+  range; `numeric` is unconstrained (#205 owns its float64 ceiling).
+- Column-bound attribute: the declared type **and** the column's own width.
+  `numeric`→`integer_01` refuses `1.5` and `3e9`; `integer`→`smallint_01`
+  refuses `40000`. Before #459 these wrapped (`int16(40000) = -25536`) into
+  `entity_main`.
+- `text`→`uuid_*`: the value must parse as a UUID (published 4xx, not the
+  redacted 500 `uuid.Parse` used to raise in `storeInMainColumn`).
+- A value whose typed slot does not match the column family is refused, never
+  dropped.
+
+The published message names the attribute, the value, the destination and
+the allowed range, e.g. `invalid value for attribute 'rank' (attrID=3): value
+40000 out of range for bound column smallint_01 (smallint) (allowed [-32768,
+32767])`.
+
+Registration (`schemameta.ValidateColumnBinding`) refuses a
+`valueType`↔column-encoding pair that cannot round-trip at all, so the runtime
+rule only ever sees width and UUID-shape questions on a well-formed schema;
+`validate-schema-consistency` lists such pairs across an already-deployed set
+(see `docs/schema-consistency-migration.md`).
+
 ## Read-path consistency errors
 
 Read operations return plain errors when persisted data and metadata disagree.
--- a/docs/schema-consistency-migration.md
+++ b/docs/schema-consistency-migration.md
@@ -485,6 +485,17 @@ width is refused at write time as invalid input); `date`/`datetime`→bigint
 (`unix_ms` or default) or text (`iso8601`); `bool`→smallint (`bool_smallint`)
 or text (`bool_text`); `list` never binds.
 
+Some refused pairs do store and read back losslessly on the Postgres path:
+`uuid`→text, `bool`→smallint/integer/bigint/double with the default encoding,
+and `date`/`datetime`→double with the default encoding. They are refused as a
+matter of policy, not because stored data is at risk: the filter rendering
+and the DuckDB projection key on the declared type and the encoding, so a
+`bool` with the default encoding is compared as text `'1'`/`'0'` and a `uuid`
+attribute is projected from a UUID-typed column. If a deployment carries one
+of these shapes, no row is corrupted; rebind with the explicit encoding
+(`bool_smallint`, `bool_text`, `unix_ms`) or to the column family the
+valueType names, and migrate the existing column values with SQL as above.
+
 ### Scalar rows under list attributes (`#372`)
 
 Example validator output:
--- a/docs/schema-consistency-migration.md
+++ b/docs/schema-consistency-migration.md
@@ -459,7 +459,25 @@ before deploying:
   and admissible value already has that shape.
 
 Existing rows in the old column are not moved by either change; migrate them
-with SQL first if they must survive.
+with SQL first if they must survive. For the shipped `log` schema, which #459
+rebound `leadId` from `uuid_02` to `text_02` and `visitId` from `uuid_01` to
+`text_03`:
+
+```sql
+-- shipped `log` schema: leadId uuid_02 -> text_02, visitId uuid_01 -> text_03
+UPDATE entity_main
+   SET text_02 = uuid_02::text
+ WHERE ltbase_schema_id = <log schema id> AND uuid_02 IS NOT NULL;
+UPDATE entity_main
+   SET text_03 = uuid_01::text
+ WHERE ltbase_schema_id = <log schema id> AND uuid_01 IS NOT NULL;
+-- NULL the old columns once the new binding is live
+```
+
+Parquet tiers (delta/base) are keyed by attribute name, not by main column, so
+they need no rewrite. The shipped `cmd/server/schemas/log_attributes.json`
+itself was rebound in #459; a deployment still carrying the old copy must apply
+this migration before upgrading, or the server refuses to load the schema.
 
 Admitted pairs: `text`→text; `uuid`→uuid; `smallint`/`integer`/`bigint`/`numeric`
 → smallint/integer/bigint/double (a value that does not fit the column's
--- a/docs/schema-consistency-migration.md
+++ b/docs/schema-consistency-migration.md
@@ -119,6 +119,10 @@ It validates:
 - every referenced `<schema>_attributes.json` parses successfully
 - each schema’s metadata has unique `attributeID` values
 - each schema’s metadata has unique `column_binding.col_name` values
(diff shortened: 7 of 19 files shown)


iceboundrock:
## Review: PR #556 — write-path value/column fidelity (#459)

Reviewed the 8 commits on `worktree-issue-459` against `main` (`f761a39` HEAD). Read every touched file, the #459 issue thread (plan, mid-execution ruling, closing comment) and the #384 contract it shares.

### Findings

**F1 — `internal/transform/storage_fit.go:147` — Important — `numeric`→`bigint_*` still wraps at the int64 boundary**

`checkIntegerFit`'s bigint arm returns nil as soon as `toInt64ExactForEAV(raw)` succeeds, on the assumption that the exact int64 is what will be stored. That holds only when the *declared* type is `bigint`, because `populateTypedValue` fills `attr.ValueInt64` for `ValueTypeBigInt` alone. For a `numeric` attribute bound to a `bigint_*` column, `ValueInt64` stays nil and `storeWithDefaultEncoding` writes `int64(*attr.ValueNumeric)`, where `ValueNumeric` is the float64 image `9.223372036854776e+18` (= 2^63). Go's float→int conversion of an out-of-range value is implementation-defined; on amd64 it is `MinInt64`.

Verified with a throwaway test in the package (removed afterwards):

```
raw=9223372036854775807 (int64)  set=true err=<nil>  ValueNumeric=9.223372036854776e+18 ValueInt64=<nil>
  stored bigint_01=-9223372036854775808
raw=9223372036854775807 (string) set=true err=<nil>  ValueNumeric=9.223372036854776e+18 ValueInt64=<nil>
  stored bigint_01=-9223372036854775808
raw=9.223372036854776e+18 (float64) → rejected: "value 9.223372036854776e+18 out of range for bound column bigint_01 (bigint)"
```

So the same value is refused when sent as a JSON float and wrapped when sent as a string or via the Go API as `int64` — exactly the "never wrapped into storage" shape #459 closes. Reachable for any `numeric`→`bigint_*` binding (admitted by the matrix); `smallint`/`integer`-declared values are capped by the declared-type check first, and `date`/`datetime` carry `ValueInt64`, so only `numeric` is exposed.

Suggested fix (either):
- In `checkBoundColumnFit`, pass `raw` to `checkIntegerFit` only when `attr.ValueInt64 != nil` (i.e. the exact sidecar is what `storeWithDefaultEncoding`/`unix_ms` will actually write); otherwise pass `nil` so the float-image bound check governs. One-line change, keeps #384's boundary-literal behaviour for declared `bigint`.
- Or, more faithfully, set `attr.ValueInt64` for the whole numeric family whenever `Int64Exact` succeeds, so a bound bigint column stores the exact value (this would also remove the float rounding for `numeric`→`bigint_*` between 2^53 and 2^63, e.g. `9223372036854775000` → `9223372036854774784` today). Needs a look at the EAV echo path (`ToPersistentRecord` already clears `ValueInt64` for EAV rows, so it may be safe).

Please add a case to `TestCheckStorageFit_FamilyMismatch` (or a new table) for `numeric`→`bigint_01` with `"9223372036854775807"` and `int64(math.MaxInt64)` asserting a published rejection (or exact storage, depending on which fix you take).

No other correctness findings. Things I specifically checked and found sound:

- Slot parity between `checkBoundColumnFit` and `storeWithEncoding`/`storeWithDefaultEncoding` is exact for every (encoding, column type) pair, so the new "no value to store" plain error in `storeInMainColumn` is unreachable from the public write path (`flattenToAttributes` refuses JSON `null` for known attributes; the empty-list marker row exists only for `list`, which the matrix never binds). Classifying it as an operator-visible invariant error is right.
- `*attr.ValueNumeric` is only dereferenced after `columnFitType` returned true, which implies the default-encoding slot check or the `unix_ms`/`bool_smallint` arm already proved it non-nil. No nil deref.
- Update path: `MergePersistentRecord` re-runs the merged record through the funnel, so report-only update validation cannot bypass it (pinned by `TestEntityManager_WriteFidelity`), and values read back from narrow columns are always in range, so no existing row becomes un-updatable.
- `DeferColumnBindingCheck` reaches only `loadAttributeMetadataFromFiles`; `RegisterSchema` and both `file_registry.go` sites stay strict. The tool's `checkColumnBindings` walks the active cache, so retired entries are skipped consistently with `validateSchemaAttributeCacheOpts`; issues are sorted before reporting, so map iteration order does not leak into output.
- Every `*_attributes.json` in the repo (20 files, 43 bindings: sample, server, e2e production, federated benchmark) passes the new matrix; `watch_attributes.json`'s `createdAt`→`ltbase_created_at` (`unix_ms`) and `yearOfProduction.year` `numeric`→`smallint_02` are admitted as intended.

### Spec check (#459 acceptance criteria)

| AC | Status |
| --- | --- |
| One `populateTypedValue` rule, shared with #384, rejects a value that does not fit its physical destination as `forma.InvalidInputf` | Met — `checkStorageFit` subsumes `checkDeclaredIntegerFit`; messages stay parallel (`declared type …` / `bound column … (…)`). F1 is a gap inside this rule. |
| Registration refuses non-round-trippable `valueType`↔column bindings | Met — `ValidateColumnBinding` at `RegisterSchema`, file registry, and runtime loader. |
| Out-of-range numeric for a bound narrow column → 4xx, never wrapped | Met for smallint/integer; **F1** leaves one bigint boundary case wrapping. |
| Mismatch path uses `InvalidInputf`, not redacted 500 | Met — `text`→uuid parse failure now published; `text`→numeric refused not dropped. |
| `Update` response reflects stored state | Already delivered by #457/PR #553; correctly recorded as no-op with the pinning test named. |
| `validate-schema-consistency` detects mismatches across a deployed set | Met — new category, lenient loader, exact-count test, README + migration doc. |
| Regression: out-of-range smallint/integer → 4xx | `TestCheckStorageFit_BoundColumnWidth`, `TestPersistentRecordTransformer_RejectsOutOfRangeBoundColumn`, `TestEntityManager_WriteFidelity`. |
| Regression: `text`→uuid / `text`→numeric bindings → registration reject or published 4xx | `TestValidateColumnBinding_Matrix`, `TestDirectoryRegistry_RejectsIncompatibleBinding`, `TestCheckStorageFit_FamilyMismatch`, manager-seam test. |

The 2026-08-24 maintainer ruling is written once (plan comment + `docs/error-handling.md`) and #384 points at it — matches the epic sequencing constraint. The mid-execution ruling (width cap applies to every numeric-slot write, not only the numeric family) is implemented in `46c2455` and covered by `TestCheckStorageFit_NonNumericFamilyColumnWidth`.

### Standards check

- File/function limits: all touched sources under 500 lines (`persistent_record.go` is at 488 — see O1); no function near 100 lines.
- Error wrapping: every new error path wraps with context; the published carriers are built via `forma.InvalidInputf` in `populateTypedValue`'s `handleConversionError`, so `checkStorageFit` returning plain `fmt.Errorf` is correct (the funnel wraps once). `TestNoBareSentinelWraps` passes.
- Error messages name value, destination and allowed range as `docs/error-handling.md` requires; the binding error names attribute, valueType, column, column type, encoding and the admitted set.
- Early returns / ≤3 nesting: `storeWithEncoding`/`storeWithDefaultEncoding` split is flatter than the original nested switch.
- Commit messages and PR body: no AI attribution; conventional prefixes; bodies explain the why.
- Non-code artifacts (plan, mid-execution ruling, PR link) are on #459 as required.

### Assessment

The design is right: one funnel, mirror of the store dispatch, registration matrix as a separate pure function, lenient loader only for the tool. The `46c2455` follow-up that made the width cap independent of valueType/encoding is the correct generalisation and closes the `date`→`integer_*` hole the first draft had. F1 is the same class of hole one level up (the exact-int64 shortcut assumes the store uses the sidecar) and is a small fix. Approve once F1 is addressed with a test.

### Observations (non-blocking)

- **O1** `internal/transform/persistent_record.go` is at 488/500 lines. The next touch will have to split it; `isSystemManagedColumn` + the read-side helpers (`readFromSystemColumn`, `readWithEncoding`, `readWithDefaultEncoding`) would move cleanly to a `persistent_record_read.go`.
- **O2** The matrix refuses some pairs that do round-trip on the Postgres path today: `uuid`→`text_*` (deliberately, documented — DuckDB projection), but also `bool`→`smallint_*`/`integer_*`/`bigint_*` with default encoding and `date`/`datetime`→`double_*` with default encoding. None ship in-repo, and a conservative matrix is defensible, but a deployment carrying one of these now fails to boot with no data at risk. Worth one sentence in `docs/schema-consistency-migration.md` listing them as "lossless on Postgres but refused, rebind with the explicit encoding", so an operator hitting it knows it is a policy refusal, not corruption.
- **O3** `bigint`/`integer` → `double_*` is admitted with no 2^53 check, so `bigint` `9007199254740993` stores as `9007199254740992`. The PR body already calls this out as #205's; fine to leave, but the `docs/error-handling.md` section says "the column's own width" without the double exception — a parenthetical "(`double_*` is unconstrained; #205)" there would match the PR text.
- **O4** `checkIntegerFit` messages use `%v` for the float, so epoch millis and `3e9` render as `1.7040672e+12` / `3e+09`. Pre-existing #384 format, but for the new date→narrow-column rejections the value is unrecognisable to the caller. `%.0f` for integral values would read better.
- **O5** `ColumnType()` defaults to `text` for an unknown column name, so `text`→`col_name: "foo"` passes the matrix and only fails at the main-table writer. Pre-existing; the matrix is not the place to fix it, just noting it does not validate column existence.
- **O6** `storeWithEncoding` still writes `unix_ms`/`bool_smallint` into `Int64Items`/`Int16Items` without consulting `ColumnType()` — acknowledged in the PR as a follow-up; agreed it is not a fidelity hole given the matrix + writer rejection.

### Verification

Run on `f761a39` in the `worktree-issue-459` worktree (`GOTOOLCHAIN=go1.26.0+auto`, `GOTMPDIR` in `.gocache/tmp`):

- `make fmt-check` — clean.
- `make lint` (golangci-lint v1.64.8, root + infra) — clean, exit 0.
- `go test ./internal/transform ./internal/schemameta ./cmd/tools -count=1` — all `ok`.
- `go test ./internal -run 'TestEntityManager_WriteFidelity|TestEntityManager_Update|TestBuildHybridConditionsBoundAuditColumn|FileSize'` — all PASS, including `TestEntityManager_Update_ResponseIsStorageRoundTrip`.
- `go test . -run TestNoBareSentinelWraps` — PASS.
- Binding audit over every `*_attributes.json` in the repo (jq) — 43 bindings, all admitted by the matrix.
- F1 reproduced with a throwaway `zz_review_probe_test.go` in `internal/transform` (output above); file deleted, tree clean.
- Did not re-run the full `./scripts/test_with_container_runtime.sh` or the e2e harness; relying on the PR's reported run for those.

### Summary

Solid implementation of the #459 contract with the ruling correctly written once and shared with #384. One important gap: **F1** — `numeric`→`bigint_*` accepts `9223372036854775807` (string or int64) via the exact-int64 shortcut and stores `MinInt64`, because the sidecar the shortcut assumes is only populated for declared `bigint`. Fix + regression test requested; the rest is non-blocking polish (O1–O6). Not merging.

iceboundrock:
## Response to the review (5649041553)

Addressed in `5ac29ac`. Item by item:

### F1 — `numeric`→`bigint_*` wraps at the int64 boundary — **fixed**

Reproduced first with the same probe shape (`int64(MaxInt64)` and `"9223372036854775807"` both stored `-9223372036854775808`; nil `ValueInt64` in both cases).

Of the two fixes offered I took the first, and slightly further than the one-liner: `checkIntegerFit` now takes the `EAVRecord` and no longer sees the raw literal at all. Its bigint arm passes on `attr.ValueInt64 != nil` (the sidecar is an int64 and is what the bigint/`unix_ms` store arms write when populated) and otherwise bound-checks the float64 image, which is what `storeWithDefaultEncoding` converts. The `raw` parameter is gone from `checkStorageFit` and `checkBoundColumnFit`, so the check cannot drift from the store again by looking at something the store never sees. Declared `bigint` keeps the #384 boundary-literal admission: `populateTypedValue` fills the sidecar from `numutil.Int64Exact(value)`, which is the same predicate `toInt64ExactForEAV` applied for the non-pointer shapes `numutil.Float64` admits, so nothing previously admitted for `bigint` is now refused (`TestPopulateTypedValue_DeclaredIntegerFit` and the existing `bigint declared, bigint column, max exact string ok` case still pass).

Why not the second option (populate `ValueInt64` for the whole numeric family): `numeric` is a float64-valued type by contract — `extractValueFromEAVRecord` returns `*ValueNumeric` for it and ignores the sidecar, and the DuckDB projection casts it as DOUBLE — so an exact int64 in the column would be invisible to every reader and the read-back would still round above 2^53. It would also make the main column disagree with what the API returns for the same row. The write rule is "fit what is actually written"; widening what is written for one valueType is a separate decision (it is #205's territory) and not what #459 asks for.

Tests added in `storage_fit_test.go`:
- `TestCheckStorageFit_FamilyMismatch`: `numeric`→`bigint_01` with `"9223372036854775807"` and `int64(math.MaxInt64)` rejected (message `value 9223372036854775808 out of range for bound column bigint_01 (bigint) …`), `MinInt64` and `2^62` accepted, `bigint`→`bigint_01` `MaxInt64` accepted via the sidecar.
- `TestCheckStorageFit_BigintColumnStoreParity`: for both `numeric` and `bigint` declared, over twelve boundary literals in string/int64/float64 shapes, every value the funnel admits is run through `storeInMainColumn` and the stored int64 must equal the sidecar (when present) or convert back to the float image exactly — i.e. no wrap is reachable for anything admitted.

### O1 — `persistent_record.go` at 488/500 — **follow-up #558**

Not touched by this fix. Filed #558 with the split you outlined (read-side helpers + `isSystemManagedColumn` → `persistent_record_read.go`); it should land before or with #559, which edits that file.

### O2 — refused-but-lossless pairs — **done**

Added a paragraph to `docs/schema-consistency-migration.md` under the admitted-pairs list naming `uuid`→text, `bool`→smallint/integer/bigint/double (default encoding) and `date`/`datetime`→double (default encoding) as policy refusals with no data at risk, and why: `convertPgBoolValue` renders a default-encoded bool operand as text `'1'`/`'0'` and the DuckDB projection assumes a UUID-typed column for a `uuid` attribute, so the query path — not storage — is what does not round-trip. Remedy stated: rebind with the explicit encoding or to the family the valueType names, then migrate with SQL as in the section above it.

### O3 — `double_*` exception in `docs/error-handling.md` — **done**

Added `(double_* is unconstrained; #205 owns the float64 ceiling, so bigint→double_01 rounds above 2^53 rather than refusing)` to the column-bound bullet, plus a sentence stating the sidecar rule from F1 so the doc explains why `numeric`→`bigint_01` refuses `9223372036854775807` while `bigint`→`bigint_01` admits it.

### O4 — `%v` exponent rendering — **done**

`formatFitValue`: integral values below 1e21 print with `'f', 0` (exact digits, so a caller who sent `9223372036854775807` sees the `9223372036854775808` its float image became — that is the more useful message for the F1 case), non-integral values below 1e21 print shortest-round-trip (`1.5`), and `|v| >= 1e21` keeps `'g'` so `1e300` does not become 301 digits. Applies to the declared-type messages too, since they share the helper. Pinned by `TestCheckStorageFit_MessageRendersPlainDigits` (`3000000000`, `1704067200000`, `1.5`, `1e+300`). No existing test asserted the old exponent form.

### O5 — `ColumnType()` defaults to text for an unknown name — **follow-up #557**

Agreed the matrix is not the place: `ColumnType()`'s fallback is relied on by read paths over already-loaded metadata. Filed #557 for a column-existence check next to `ValidateColumnBinding`, reported by `validate-schema-consistency` in the same pass; `text_99` is the sharper example since it classifies as text and is admitted although the table stops at `text_09`.

### O6 — `storeWithEncoding` ignores `ColumnType()` — **follow-up #559**

Already listed in the PR body as a follow-up; now has an issue (#559) with the parity-table test that would pin it. Not a fidelity hole for the reasons you and the PR body give.

### Verification on `5ac29ac`

- `make fmt-check` clean; `make lint` (root + infra) clean.
- `./scripts/test_with_container_runtime.sh` (rootless Podman): all 34 packages `ok`, exit 0.
- `go test ./internal -run 'TestEntityManager_WriteFidelity|TestEntityManager_Update|FileSize|Guard'` and `go test . -run TestNoBareSentinelWraps` pass.
- `storage_fit.go` is at 203 lines, `storage_fit_test.go` at 302; no function near 100.

iceboundrock:
## Review Summary

### Findings

No blocking or correctness findings identified in the current head `5ac29ac`.

The final follow-up correctly closes the earlier `numeric -> bigint_*` boundary case: `checkIntegerFit` now validates the EAV slot that the store actually consumes, while preserving exact-sidecar handling for declared `bigint` values. The added parity and boundary tests cover the previously wrapping `MaxInt64` shapes.

### Spec Check

- The write path has one `populateTypedValue -> checkStorageFit` fidelity funnel covering declared-type limits, bound-column width, UUID parsing, and slot/family mismatches.
- Narrow `smallint`/`integer`/`bigint` bindings reject out-of-range or fractional values as published `forma.InvalidInputf` errors rather than wrapping.
- Text-to-UUID and text-to-numeric mismatches are rejected as 4xx input errors and are not silently dropped.
- Registration and runtime metadata loading enforce the binding matrix; `validate-schema-consistency` deliberately loads leniently and reports all active mismatches.
- The update response requirement is correctly recorded as already delivered by PR #553, with no regression introduced here.
- Migration documentation, the shipped `log` metadata rebind, and validator output are consistent with the implementation.

### Standards Check

- `make fmt-check` passes.
- `make lint` passes for both the root and `infra/` modules.
- Touched source files remain within the repository's 500-line limit; `persistent_record.go` is 488 lines and should be split on the next related change.
- New errors are contextualized, and caller-facing write failures use the published invalid-input carrier.
- The implementation uses early returns and keeps the storage-fit logic separate from the binding policy matrix.
- Commit messages and the PR body contain no AI attribution.

### Assessment

The implementation is technically sound and the design is appropriately split between runtime value fidelity and metadata binding policy. No changes are required from this review before merge.

### Observations

- `persistent_record.go` is close to the 500-line limit; the planned read-side split in follow-up #558 remains advisable.
- `ColumnType()` falls back to `text`, so unknown or prefix-like column names can still pass the new matrix and fail later at the main-table writer; this is already tracked by follow-up #557.
- Special encodings still rely on the registration matrix to pair with the correct physical column type; aligning the writer dispatch with `ColumnType()` is tracked by follow-up #559.
- The DuckDB hot-leg projection for `date`/`datetime -> text (iso8601)` remains the documented residual risk in #555.

### Verification

- `make fmt-check` — passed.
- `make lint` — passed.
- `go test ./internal/transform ./internal/schemameta ./cmd/tools` — passed.
- `go test . ./cdc ./factory ./cmd/...` — passed.
- `git diff --check main...HEAD` — passed.
- GitHub PR checks — passed: Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.
- A local `go test ./internal/...` run reached all non-container packages successfully; `internal/e2e_harness` could not start because this environment lacks `$XDG_RUNTIME_DIR`/a configured container runtime. This is an environment limitation, not a code failure; the corresponding PR CI checks passed.

### Summary

Approve from a code-review perspective. The final head addresses the only correctness gap found in the earlier review, adds focused regression coverage, and passes the repository's formatting, lint, unit, and CI checks. No blocking findings remain.

PR #560, opened 2026-09-13
fix(batch): reject duplicate row ids in atomic BatchUpdate (#458)

### LINKED ISSUE #458: bug: atomic BatchUpdate with the same row_id twice silently discards the earlier op while reporting both successful
## Problem

An atomic `BatchUpdate` containing two operations on the same `row_id` deterministically discards the earlier operation, while reporting **both** as successful. The non-atomic (best-effort) path applies both correctly — so "atomic" here is weaker than best-effort, a contract inversion.

## Current behavior

`batchUpdateAtomic` (`internal/entity_batch_service.go`) reads `GetPersistentRecord(R)` for *every* operation up front, before any write, so two operations on the same row both merge onto the same base. `BatchUpdatePersistentRecords` (`internal/postgres_persistent_repository_batch.go`) then applies record[0] (delete known EAV ids, insert `base+a`) and record[1] (delete known EAV ids, insert `base+b`); the final row is `base+b`. `a` is lost. No deadlock or PK error surfaces because `replaceEAVAttributesScoped` deletes before inserting. `BatchResult.Successful` reports both operations with their own merged attributes.

The best-effort path (`executeBestEffortBatch` → `crud.Update` per op) re-reads before each op, so it applies both updates correctly.

## Reproduction

`BatchUpdate({Atomic:true, Operations:[{RowID:R, Updates:{"a":1}}, {RowID:R, Updates:{"b":2}}]})` → stored row has `b:2` only, `a` silently dropped, both operations reported successful.

## Root cause

Read-then-write over the whole batch with no intra-batch coalescing and no per-record re-read, combined with EAV replace-whole-attribute-set semantics.

## Impact

Silent, deterministic data loss (no concurrency needed) with a 200-shaped success report, on the path that advertises atomicity. `BatchUpdate` is library API (`forma.EntityManager`); it is not currently mounted by `internal/httpapi`, which bounds the blast radius to embedding callers.

INDEPENDENT. Not #401 (file-size only). Related in spirit to the update read-modify-write issue (same "merge base read before write") but a distinct trigger (same-batch duplicate keys) and distinct fix (intra-batch coalescing / last-op-wins semantics or an explicit duplicate-key rejection).

## Acceptance criteria

**Maintainer ruling (2026-08-24):** **reject** a duplicate `row_id` in an atomic batch at intake (a cheap map check → `forma.InvalidInputf`), rather than defining silent last-op-wins coalescing — duplicates in one batch are almost always a caller bug, and rejection is observable while coalescing launders the anomaly. Best-effort keeps its well-defined apply-in-order (natural last-wins) behavior; document it, no behavior change there. (Same posture as #459: reject ambiguous caller input rather than guess intent.)

- [ ] Atomic `BatchUpdate` rejects a duplicated `row_id` up front with a published `forma.InvalidInputf`.
- [ ] Best-effort's in-order apply (last-wins by execution order) is documented explicitly.
- [ ] Atomic batch is never weaker than best-effort for the same operations.

- [ ] `BatchUpdate` atomic with two ops on one row: assert stored state and reported result agree with the chosen contract.

iceboundrock:
Closes #458.

## Problem

An atomic `BatchUpdate` reads every base record before any write (`batchUpdateAtomic`), so two operations on the same `row_id` both merge onto the same base and the later write silently discards the earlier one — while `BatchResult.Successful` reports both. The best-effort path re-reads before each op and applies both, so "atomic" was weaker than best-effort for identical input.

## Change

Per the maintainer ruling on #458 (reject, don't coalesce):

- **`internal/entity_batch_update_validation.go`** (new): `validateAtomicUpdateOperations` does one list-order pre-pass over the operations. For each operation it checks the required fields (schema name, row id, updates) and then whether `(SchemaName, RowID)` was already targeted, returning a published `forma.InvalidInputf` for the first problem it meets. Because the pass is strictly in list order, a malformed operation always outranks a duplicate that comes later, and a duplicate never masks a missing field on the same operation. Keying on the pair keeps a cross-schema batch that happens to share a UUID valid.
- **`internal/entity_batch_service.go`**: `batchUpdateAtomic` calls the pre-pass right after resolving the atomic repository, before any registry lookup, read, or write. The per-operation required-field checks move out of the main loop since the pre-pass now owns them.
- **`types_batch.go`** (new), **`types.go`**, **`storage.go`**: `BatchOperation` and `EntityBatchOperator.BatchUpdate` doc comments state both contracts — best-effort applies in list order (last-wins by execution order, no behavior change); atomic refuses a repeated `(schema, row_id)`. The batch write types (`OperationType`, `EntityIdentifier`, `EntityOperation`, `BatchOperation`, `BatchResult`, `OperationError`) move from `types.go` into `types_batch.go`; `types.go` was already over the 500-line limit on `main` (501) and is now 454.

`BatchCreate`/`BatchDelete` atomic paths are unchanged: a duplicate create collides on the Postgres PK, a duplicate delete is idempotent; neither loses data silently.

## Tests

`internal/entity_batch_update_validation_test.go` (new; the #458 tests moved here so `entity_batch_service_test.go` stays under the file-size limit):

- `TestEntityManager_BatchUpdate_AtomicRejectsDuplicateRowID` — two ops on one row, atomic → `errors.Is(err, forma.ErrInvalidInput)`, published message mentions `operation[1]`, `operation[0]`, the row id, and `visit`; the mock repository saw zero `GetPersistentRecord` and zero `BatchUpdatePersistentRecords` calls; stored record untouched. Fails on `main` (both ops reported successful).
- `TestEntityManager_BatchUpdate_AtomicRequiredFieldsOutrankDuplicate` — a duplicate with `Updates == nil` reports `updates are required`; an earlier op with an empty schema or zero row id reports that instead of a later duplicate. Fails against the first revision of this PR.
- `TestEntityManager_BatchUpdate_BestEffortAppliesDuplicateRowIDInOrder` — same ops, best-effort → both succeed and the stored row carries both `status` and `feedback` (pins the documented contract).

`internal/entity_manager_mocks_test.go`: the mock repository now counts `getCalls` and `batchUpdateCalls`.

## Verification

- `make fmt-check` — clean
- `make lint` — clean (root + infra)
- whitespace check — clean
- `go test . ./internal ./internal/httpapi -count=1` — pass (includes `TestNoBareSentinelWraps`)


iceboundrock:
## Review

### Findings

1. **P2: The duplicate pre-pass can mask the existing per-operation validation error.** (`internal/entity_batch_duplicates.go:23-35`, called from `internal/entity_batch_service.go:273-275`)

   The helper runs before the ordered validation loop and only skips empty schema names and zero row IDs. For example, if operation 0 targets row `r`, operation 1 also targets `r` but has `Updates == nil`, the new code returns `duplicate row id` for operation 1. The previous validation order would return `operation[1]: updates are required for update operation`. Similarly, an operation with an empty schema can be masked when a later valid operation duplicates an earlier target, because the invalid operation is skipped by the pre-pass and the later duplicate is reported first.

   This contradicts the helper comment that the existing per-operation required-field message still wins for missing fields, and makes the error less actionable for malformed batches. Please preserve validation precedence by validating the per-operation required fields in list order before duplicate detection, or otherwise ensure the duplicate pass cannot outrank those validation errors while still doing all checks before registry lookup and data I/O.

### Spec

- The core #458 behavior is implemented: atomic `BatchUpdate` rejects repeated `(SchemaName, RowID)` targets before any record read or write, using a published `forma.InvalidInputf` carrier.
- Best-effort updates retain ordered execution and last-wins behavior, and the new test pins that contract.
- Keying by schema name keeps the requested cross-schema, same-UUID case valid. The built-in registries enforce the corresponding name/ID uniqueness invariant; custom `SchemaRegistry` implementations would need to preserve that invariant as well.
- Leaving atomic `BatchCreate` and `BatchDelete` unchanged matches the stated scope and ruling.

### Standards

- `make fmt-check` passes, and the changed helper/functions stay within the function-size limit.
- `types.go` is 515 lines on this branch, up from 501 on `main`, exceeding the repository's 500-line source-file limit. The new public contract documentation at `types.go:89-106` should be shortened or the file should be refactored so this change does not further exceed the limit.
- The new error correctly uses `forma.InvalidInputf`; `errors.Is(err, forma.ErrInvalidInput)` and the published message are covered by the regression test.

### Assessment

**Changes requested.** The intended lost-update case is fixed, but the validation-precedence issue should be corrected before approval. The file-size violation is also a standards failure introduced by the documentation change.

### Observations

- The atomic regression test verifies that the stored record remains unchanged, but it does not directly count repository reads/writes. A counting mock would make the documented "before any read or write" guarantee explicit.
- No additional supported-path correctness or error-semantic regressions were found.

### Verification

- PASS: `git diff --check main...origin/pr-560`
- PASS: `make fmt-check`
- PASS: `GOTOOLCHAIN=go1.26.0+auto go test ./internal -run 'TestEntityManager_Batch(Update|Create|Delete)' -count=1 -v`
- PASS: `GOTOOLCHAIN=go1.26.0+auto go test . ./internal ./internal/httpapi -count=1`
- PASS on GitHub: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke checks.
- Local `make lint` reached the infra lint stage but was blocked by the environment's `disk quota exceeded` error while compiling Pulumi dependencies; the corresponding GitHub Lint check passed.
- Local `go test ./... -count=1` passed for the non-container packages but could not complete `internal/e2e_harness` because `TestE2EHarnessMinimal` panicked while resolving the container host in this environment (`$XDG_RUNTIME_DIR` / container runtime setup).

### Summary

The PR fixes the reported atomic duplicate-update data-loss bug with a bounded implementation and appropriate public error classification. Please address the validation-precedence finding and the 500-line source-file limit before merge.

iceboundrock:
## Response to review (5649795894)

Addressed in 787d0d6. Item by item:

### Finding 1 — duplicate pre-pass masks per-operation validation (fixed)

Confirmed both scenarios against the first revision: `op[1]` duplicating `op[0]` with `Updates == nil` came back as `duplicate row id`, and an empty-schema `op[0]` was skipped so `op[2]`'s duplicate was reported ahead of it.

Two ways to restore precedence were on the table:

1. Keep the duplicate check as a separate pass and run the required-field loop ahead of it. This makes the *class* of error win (all required-field errors before any duplicate), which is not the same as list order: with `op[0]` a valid duplicate of a later op and `op[1]` malformed, it reports `op[1]` first even though `op[0]` is the earlier problem. It also duplicates the field checks in two places.
2. One list-order pass that does the field checks and then the duplicate check per operation. The lowest-index problem always wins, whichever kind it is, and the invariant is a single sentence.

Went with 2. `rejectDuplicateAtomicUpdateTargets` is now `validateAtomicUpdateOperations` (`internal/entity_batch_update_validation.go`), and `batchUpdateAtomic` no longer repeats the field checks inside its loop. Everything still runs after `atomicRepository()` and before any registry lookup or I/O, so the "before any read or write" guarantee is unchanged.

One deliberate consequence worth naming: on `main`, a malformed `op[1]` was only reported after `op[0]` had already been looked up and read, so a not-found on `op[0]` used to outrank a missing field on `op[1]`. Now input validation for the whole batch precedes any I/O, which is what "all checks before registry lookup and data I/O" asks for, and is the stronger contract.

Regression test: `TestEntityManager_BatchUpdate_AtomicRequiredFieldsOutrankDuplicate` covers the nil-`Updates` duplicate, the empty-schema earlier op, and the zero-row-id earlier op, asserting the exact published message and zero repository reads/writes. I ran it against the first-revision helper to confirm it fails on all three cases with the messages you described.

### Standards — `types.go` over 500 lines (fixed by refactor)

`types.go` was already 501 on `main`, so trimming the comment back to 500 would have hidden the violation rather than fixed it. The batch write contracts (`OperationType`, `EntityIdentifier`, `EntityOperation`, `BatchOperation`, `BatchResult`, `OperationError`) now live in `types_batch.go` (67 lines); `types.go` is 454. Pure move, no content changes, same package, so no call sites change.

The same limit would have caught `entity_batch_service_test.go` (545 after adding the precedence test), so the #458 tests moved to `entity_batch_update_validation_test.go` next to the code they cover; the service test file is back to 378.

### Observation — counting mock (done)

The mock repository already counted deletes and inserts, so adding `getCalls` and `batchUpdateCalls` was four lines. Both duplicate/precedence tests assert `reads == 0 && writes == 0`, making the "before any read or write" claim in the doc comment a tested property rather than a reading of the code.

### Spec note — custom `SchemaRegistry` implementations

Agreed that keying on `SchemaName` relies on the registry keeping name ↔ id one-to-one. Nothing to change here: `GetSchemaAttributeCacheByName` is the same lookup every write path already depends on, so a registry that broke that invariant would misroute plain `Update` calls long before this check mattered. No follow-up issue opened.

### Verification on this revision

- `make fmt-check` — clean
- `make lint` — clean, root and infra
- `go test . ./internal ./internal/httpapi -count=1` — pass
- Reverting the helper to the skip-based version fails the new precedence test on all three subcases; restoring it passes.

iceboundrock:
## Findings

No blocking findings. The implementation is consistent with the stated fix and I found no correctness, regression, or API-compatibility issue in the changed paths.

## Spec

- Atomic `BatchUpdate` rejects repeated `(schema, row_id)` targets with a published `forma.InvalidInputf` before registry lookup or repository reads/writes.
- Duplicate detection is keyed by schema name plus UUID, so the same UUID remains valid across different schemas as required.
- Required-field validation and duplicate detection run in one list-order pre-pass, so the first applicable malformed-input error is preserved and a duplicate does not mask missing fields on the same operation.
- Best-effort update behavior remains in-order and re-reads each row, preserving the documented merge behavior.
- The public `BatchOperation` and `EntityBatchOperator.BatchUpdate` contracts document both modes.
- `BatchCreate` and `BatchDelete` behavior is unchanged and appropriately out of scope.

## Standards

- Changed production files remain within the repository 500-line limit: `entity_batch_service.go` is 457 lines, `types.go` is 454 lines, and the new files are smaller.
- The helper is cohesive, uses a flat early-return structure, and returns the project error carrier expected by the write path.
- The type move is package-only organization; JSON tags and exported API identities remain unchanged.
- No whitespace or formatting issues found.

## Assessment

Approve. The fix is minimal, addresses the silent data-loss condition at intake, and keeps atomic execution stronger than best-effort for identical input.

## Observations

- The new regression tests cover atomic rejection, no repository read/write on rejection, list-order error precedence, and best-effort application of disjoint updates to the same row.
- The O(n) pre-pass uses the correct physical identity boundary represented by `(SchemaName, RowID)`.
- No additional follow-up is required for this PR.

## Verification

- PR CI for commit `6f3014d`: Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke all passed.
- Local targeted duplicate-update tests passed.
- Local `go test . ./internal ./internal/httpapi -count=1` passed.
- Local `go test ./internal/schemameta ./internal/transform -count=1` passed.
- Local `make fmt-check` passed.
- Local `git diff --check` passed.
- Full local `make test` and `make lint` were also attempted, but the environment hit `/tmp` disk quota while compiling/installing broader dependencies; this was an environment limitation, not a test assertion failure.

## Summary

The PR correctly rejects ambiguous duplicate targets in atomic batch updates, documents the intentional best-effort semantics, and adds focused regression coverage. No findings.

PR #561, opened 2026-09-13
refactor(transform): extract attribute_converter.go below the line cap (#375)

### LINKED ISSUE #375: internal/transform/attribute_converter.go at 491/500 lines — extract before next change
Standing extraction debt flagged in PR #373's post-merge review (standards section) and again in PR #374's review (observations): the file sits 9 lines under the 500-line cap, so the next change there must extract rather than grow.

Candidate seams, in rough order of fit:
- Move the required-policy block (`checkRequiredAttributes` + `relationRootsFor` + the missing-attribute selection) to a sibling file (e.g. `attribute_required_check.go`) or into `relation_roots.go`'s orbit — it is the newest and most self-contained region.
- The `ToEAVRecord`/`FromEAVRecord` scalar-conversion half and the `FromEAVRecords` assembly half are separable if a larger split is warranted.

While extracting, revisit the two judgement-call call sites deferred in the #373 review thread: the bare `return nil, err` at the `checkRequiredAttributes` call site (attribute_converter.go:211-213) and its sibling at :161 — the sweep is the agreed moment to reconsider both together.

Refs #315, PR #373, PR #374.

iceboundrock:
Closes #375.

`internal/transform/attribute_converter.go` sat at 490/500 lines. This splits it along two seams (plan and rationale on the issue: https://github.com/Lychee-Technology/forma/issues/375#issuecomment-5650075090):

- **`relationRootsFor` + `checkRequiredAttributes` → `attribute_required.go`.** That file already held the `isRequiredAttributeMissing` helpers these two are the only caller of, so the whole required policy now lives in one 178-line file. `SetRelationRoots` stays with the struct as the `RelationRootsAware` method.
- **Receiver-free scalar coercion helpers → new `scalar_conversion.go`** (`toTimeForEAV`, `toBoolForEAV`, `derefPointer`, `toTime`, `toBool`, `ToFloat64`, …), mirroring the existing `numeric_conversion.go` seam.

`attribute_converter.go` now holds only the converter type and its four conversion methods: **206 lines**.

### Deferred judgement calls from the #373 review

Both bare `return nil, err` sites in `FromEAVRecords` are now wrapped:

- `GetSchemaMetadata` → `load schema metadata for schema %d: %w`
- `checkRequiredAttributes` → `required-policy check for schema %d: %w`

AGENTS.md's rule is unconditional; the wrap adds the operation and schema ID rather than restating the callee's facts. Both remain plain errors (no sentinel, no HTTP status change). Every existing assertion on these messages is a `strings.Contains`, so no test changed.

### Commits

1. `3037a86` — the move plus the two wraps. Pure move: `git diff` on `attribute_converter.go` adds only the import line and the two wraps; the extracted tails are byte-identical.
2. `62d10fc` — comment-only fixes from the review (stale `ToFloat64Ok` name, incomplete caller map in the `scalar_conversion.go` header). Kept separate so commit 1 stays verifiable as an extraction.

### Follow-ups filed from the review

- #562 — `toBoolForEAV` integer→bool coercion differs by width and pointer-ness (pre-existing, preserved byte-for-byte here).
- #563 — `transform.ToFloat64` has no callers.

### Verification

- `make fmt-check`, `make lint` — green
- `make test` — all packages green except `internal/e2e_harness`, which needs a container runtime on this machine (pre-existing, unrelated)
- After commit 2: `go test ./internal/transform/ -count=1` ok, `go vet ./internal/transform/` clean, `gofmt -l` clean.

### DIFF
--- a/internal/transform/attribute_converter.go
+++ b/internal/transform/attribute_converter.go
@@ -1,20 +1,13 @@
 package transform
 
 import (
-	"encoding/json"
 	"fmt"
 	"slices"
-	"strconv"
-	"strings"
-	"time"
-
-	"github.com/lychee-technology/forma/internal/model"
 
 	"github.com/google/uuid"
 	"github.com/lychee-technology/forma"
-	"github.com/lychee-technology/forma/internal/numutil"
+	"github.com/lychee-technology/forma/internal/model"
 	"github.com/lychee-technology/forma/internal/schemameta"
-	"go.uber.org/zap"
 )
 
 // AttributeConverter provides conversion between model.EntityAttribute and model.EAVRecord
@@ -35,19 +28,6 @@ func NewAttributeConverter(registry forma.SchemaRegistry) *AttributeConverter {
 // empty set, leaves enforcement exactly as it was.
 func (c *AttributeConverter) SetRelationRoots(lookup RelationRootsLookup) {
 	c.relationRoots = lookup
-}
-
-// relationRootsFor resolves the relation roots of schemaID, translating the ID
-// to the schema name the lookup is keyed by.
-func (c *AttributeConverter) relationRootsFor(schemaID int16) (RelationRoots, error) {
-	if c.relationRoots == nil {
-		return nil, nil
-	}
-	schemaName, _, err := c.registry.GetSchemaByID(schemaID)
-	if err != nil {
-		return nil, fmt.Errorf("resolve schema name for id %d: %w", schemaID, err)
-	}
-	return c.relationRoots(schemaName), nil
 }
 
 // ToEAVRecord converts an model.EntityAttribute to an model.EAVRecord.
@@ -169,7 +149,7 @@ func (c *AttributeConverter) FromEAVRecords(records []model.EAVRecord) ([]model.
 	schemaID := records[0].SchemaID
 	cache, idToName, err := schemameta.GetSchemaMetadata(c.registry, schemaID)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("load schema metadata for schema %d: %w", schemaID, err)
 	}
 
 	presentAttrIndices := make(map[string]map[string]struct{}, len(records))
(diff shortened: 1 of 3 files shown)


iceboundrock:
## Review: refactor(transform) extract attribute_converter.go (#375)

### Spec
Closes #375 and follows the approved plan on the issue (5650075090) point-for-point:
1. `relationRootsFor` + `checkRequiredAttributes` -> `attribute_required.go`, next to the `isRequiredAttributeMissing` helpers they are the only caller of. Done, verbatim (verified byte-identical extraction).
2. Receiver-free scalar coercion helpers (`toTimeForEAV`, `toBoolForEAV`, `derefPointer`, `toTime`, `toBool`, `ToFloat64`, + `finiteFloat64ToBool`/`boolFrom*`/`isTrueStringForEAV`) -> new `scalar_conversion.go`, mirroring the `numeric_conversion.go` seam. Done, verbatim (tail-from-`toTimeForEAV` diffs IDENTICAL). The floated `ToEAVRecord`/`FromEAVRecords` split was correctly *not* taken.
3. Both deferred bare `return nil, err` sites from the #373 review wrapped with operation + schema ID (`load schema metadata for schema %d`, `required-policy check for schema %d`). Done; both stay plain errors, no sentinel, no status change. `grep 'return nil, err$'` on the new converter file returns nothing.

### Standards
- File cap (<=500): `attribute_converter.go` 206, `attribute_required.go` 178, `scalar_conversion.go` 203. All green. (Nit: PR body says 213 for the converter; `wc -l` on the head OID says 206. Either way well under the cap, but worth correcting the description.)
- Function cap (<=100): longest moved/c remaining function is `toBoolForEAV` at ~83 lines; `checkRequiredAttributes` ~74; all others <=32. Green.
- Error wrapping (AGENTS.md unconditional rule): the two deferred sites are now wrapped with the operation and schema ID rather than restating callee facts. No bare sentinel wraps introduced; `TestNoBareSentinelWraps` passes.
- Error semantics (`docs/error-handling.md`): correct that both wraps stay plain. `FromEAVRecords` serves the read path (`FromPersistentRecord`) as well as write, so a sentinel here would repeat the #301 inversion the moved comment block documents. The write path keeps its own sentinel-carrying validator (`validateRequiredAttributesFromInput`).
- Nesting/early-returns: move-only; no new nesting. No unrelated rewrite.

### Findings
No blockers. Nits / pre-existing items correctly carried over (do not fix in this PR):
1. (Nit, description) 213 vs 206 line count, see above.
2. (Nit, carried over) `scalar_conversion.go:199` comment still says `ToFloat64Ok` but the function is `ToFloat64`. Pre-existing stale name, moved verbatim; fine to fix opportunistically or leave.
3. (Nit, header) `scalar_conversion.go` header says helpers are "behind `ToEAVRecord` and `populateTypedValue`", but `toBoolForEAV` is also called from `attribute_value_extraction.go:98`. Header is accurate for the moved set's primary owners but slightly incomplete as a caller map.
4. (Pre-existing, preserved correctly) `toBoolForEAV` pointer/value asymmetry: value `int32`/`int64` use `boolFromNonZero` (`!= 0`) while `*int32`/`*int64` deref to `boolFromPositive` (`> 0`), so a negative value behind a pointer coerces differently than the same value inline. The move preserves this byte-for-byte (confirmed), so no behavior change is smuggled here — but it reads like a latent bug worth a follow-up issue with a test pinning intended semantics before any fix.

### Assessment
Approve-ready. Pure move plus the two agreed wraps; no logic change, no test change needed, scope matches the issue exactly. Do not merge per repo PR rules unless the owner asks.

### Observations
- Good seam choice: the required policy now lives in one file with its helpers, and the receiver-free coercion helpers leave the converter type file holding only the type + its four methods.
- Sentinel discipline is the load-bearing correctness point of the two wraps, and it is handled right — prefix-only, `strings.Contains` assertions (`missing required attribute ...`) unaffected, confirmed no test file references the two new prefixes.
- `gofmt -l` clean on the touched package; import pruning on the converter (`encoding/json`, `strconv`, `strings`, `time`, `numutil`, `zap` out) matches exactly what moved out.

### Verification
- CI on head `3037a86`: all 5 checks SUCCESS (Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke).
- Local (PR worktree): `go vet ./internal/transform/` clean; `gofmt -l internal/transform/` clean; `go test ./internal/transform/ -count=1` ok; `go test . ./internal/schemameta/ -count=1` ok; `go test . -run TestNoBareSentinelWraps` PASS.
- Move fidelity checked by extraction diff: scalar tail IDENTICAL, `relationRootsFor` IDENTICAL, `checkRequiredAttributes` IDENTICAL.

### Summary
Meets #375's spec and all coding-standard gates with zero behavior change. Four nits/observations above, none blocking. The only item I would act on beyond this PR is a follow-up issue for the `*int32`/`*int64` `boolFromPositive` vs `boolFromNonZero` asymmetry.

iceboundrock:
Response to the review (https://github.com/Lychee-Technology/forma/pull/561#issuecomment-5650569948). Each finding verified against the head before acting.

### 1. Description says 213, file is 206 — fixed

Confirmed: `wc -l` is 206. The 213 came from a pre-`goimports` count. PR body corrected.

### 2. `ToFloat64Ok` stale doc comment — fixed in `62d10fc`

Comment-only, pushed as a separate commit on top of the move so `3037a86` remains a byte-verifiable extraction.

While checking this I looked for the "legacy optimizer helper" the comment refers to: `transform.ToFloat64` has **no callers** anywhere in the module (only its definition, `numutil.ToFloat64` which it wraps, and an unrelated `cmd/sample` `FieldMapper` of the same name). It is under `internal/`, so no external caller is possible either. Deleting an exported symbol is out of scope for a move PR — filed as #563.

### 3. Header caller map incomplete — fixed in `62d10fc`

Rewritten to name the actual consumers rather than just the primary owners: the two write funnels (`ToEAVRecord`, `populateTypedValue`), the read path (`extractValueFromEAVRecord` → `toBoolForEAV`), and `numeric_conversion.go`'s reuse of `derefPointer`. Phrased as a role statement ("shared by the write funnels and the read path") so it does not go stale on the next caller.

### 4. `*int32`/`*int64` `boolFromPositive` vs `boolFromNonZero` — follow-up filed as #562

Confirmed preserved byte-for-byte. On closer reading the asymmetry is wider than pointer-vs-value:

| Shape | Rule | `-1` → |
|---|---|---|
| `int`, `int16` | `> 0` | false |
| `int32`, `int64` | `!= 0` | true |
| every `*intN` | `> 0` | false |
| `json.Number` | `!= 0` | true |
| floats | `> 0.5` | false |

So `int(-1)` and `int32(-1)` already disagree without any pointer involved, and `attribute_converter_test.go` pins both answers (`int(-1)→false`, `int32(-1)→true`). Reachability is limited to in-process Go callers — JSON decoding never produces Go ints, and the read path only sends `*float64` through the float branch — which is why it has never surfaced through the HTTP surface. Agreed it needs a ruling on intended semantics plus a test pinning value/pointer agreement before any fix; #562 lays out the table, reachability, and candidate rules (`!= 0` matching `toBool` and `json.Number`, or `== 1` matching what `boolToFloat64` actually writes).

### State

- Head is now `62d10fc` (two commits: the move + the comment fixes).
- Local after the second commit: `go test ./internal/transform/ -count=1` ok, `go vet` clean, `gofmt -l` clean. CI is re-running on the new head.
- Not merging per the PR rules.

iceboundrock:
## Review

### Findings
No actionable correctness, regression, security, performance, or maintainability findings in this PR.

### Spec
The implementation conforms to the approved #375 plan. Required-policy code is now colocated with its helper functions in `attribute_required.go`; receiver-free scalar coercion helpers are in `scalar_conversion.go`; `AttributeConverter` retains its type and four conversion methods. The two deferred `FromEAVRecords` errors add operation and schema context while retaining `%w`, so error classification and the read-path plain-error semantics are unchanged.

### Standards
The resulting source files are within the 500-line limit: `attribute_converter.go` 206 lines, `attribute_required.go` 178 lines, and `scalar_conversion.go` 205 lines. The touched methods/helpers remain below the 100-line limit. Naming, package placement, exported `ToFloat64` documentation, and contextual error handling meet `coding-standard.md` and `AGENTS.md`.

### Assessment
This is a well-scoped, behavior-preserving refactor. The extraction has no functional changes beyond the explicitly intended contextual error prefixes. No blocking changes requested.

### Observations
Existing transform tests cover scalar bool coercion, required-policy behavior (including relation roots and array contexts), and `FromEAVRecords` paths. There is no dedicated test for the two newly prefixed error paths; that is a low-risk coverage gap for this refactor, not a merge blocker, because the wrapping preserves the original cause and existing assertions use substring matching.

### Verification
- Reviewed both commits and the complete PR diff.
- `git diff --check` passed.
- `make fmt-check` passed; `gofmt -l` reported no changed files.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform -count=1` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go vet ./internal/transform` passed.
- At review time, CI `Lint`, `Test`, `Benchmark Smoke`, and `k6 Smoke` had passed; `E2E Smoke Tests` was still pending.

### Summary
No findings. The PR is ready from code review perspective once the pending E2E smoke check succeeds.

Issue #558, opened 2026-09-12
refactor(transform): persistent_record.go at 488/500 lines — move the read-side helpers out before the next edit


iceboundrock:
Surfaced by the review of PR #556 (#459, observation O1). Same shape as #375 (`attribute_converter.go` at 491) and #497 (`internal/cdc/init.go` at 500).

## Problem

`internal/transform/persistent_record.go` is at 488 of the 500-line cap after #459 split `storeInMainColumn` into `storeWithEncoding`/`storeWithDefaultEncoding`. The next change to the write or read dispatch will have to split the file first, and the follow-up that aligns `storeWithEncoding`'s slot selection with `ColumnType()` (see the companion O6 issue) touches exactly that code.

## Proposed split

Pure move, no behaviour change:

- `persistent_record.go` keeps the transformer type, `ToPersistentRecord`, `FromPersistentRecord`, `MergePersistentRecord` and the write-side `storeInMainColumn` / `storeWithEncoding` / `storeWithDefaultEncoding`.
- `persistent_record_read.go` takes `readFromMainColumn`, `readFromSystemColumn`, `readWithEncoding`, `readWithDefaultEncoding` and `isSystemManagedColumn` (shared with `storage_fit.go`; it is a read-side classification helper, so it can live with the readers).

Both files stay well under the cap; the file-size guard test in `internal/transform` (if any) needs no change, and the package's existing tests cover the moved functions.

iceboundrock:
Count update: #564 (#404) adds the `boolFromBoolText` read guard in `readWithEncoding`, taking `internal/transform/persistent_record.go` from 488 to 492 lines. The split proposed here (`persistent_record_read.go` taking the read helpers) is now the precondition for any further edit to the read dispatch; #565 (pg-main bool truthiness) is sqlgen-side and does not touch this file.

iceboundrock:
**Implementation plan** (working copy: none, this comment is the record)

Verified against `main` at 4b21950: `internal/transform/persistent_record.go` is 492 lines. Lines 1–288 are the type, `SetRelationRoots`, `newConverter`, `NewPersistentRecordTransformer`, `ToPersistentRecord`, `FromPersistentRecord` and the three `store*` writers; lines 289–492 are `readFromMainColumn`, `readFromSystemColumn`, `readWithEncoding`, `readWithDefaultEncoding` and `isSystemManagedColumn`. That is exactly the boundary the issue proposes, so the plan is the proposed split verbatim:

[...]

PR #564, opened 2026-09-13
fix: one bool truth table across write and read (#404)

### LINKED ISSUE #404: The two bool write funnels embody different truth tables (finite floats, negatives, strings)
Follow-up from PR #403's review (issue #322). Out of that PR's scope by ruling; both funnels' *non-finite* handling is fixed there — this issue is about what the two bool converters do with values they both still accept.

## The divergence

The two write-side bool converters disagree on finite floats and on strings:

| input | `toBool` (attribute_converter.go, used by `populateTypedValue`) | `toBoolForEAV` (used by `ToEAVRecord` and the EAV read path) |
|---|---|---|
| `float64(0.3)` | `v != 0` → **true** | `float64ToBool` = `> threshold` → **false** |
| `float64(-1)` | **true** | **false** |
| `"banana"` / `"NaN"` | `strconv.ParseBool` → **error** | `isTrueStringForEAV` → silently **false** |
| `int(-1)` | `v != 0` → **true** | `boolFromPositive` → **false** |

On the main write path `populateTypedValue` coerces first and its 0/1 lands in `ValueNumeric`, so its semantics win in storage today; `ToEAVRecord` re-converts an already-boolean `Value` and the divergence stays latent. But both converters are reachable independently (`ToEAVRecord` directly, `toBoolForEAV` on the read path via `extractValueFromEAVRecord`), and a same-package caller has no way to know the two spellings of "to bool" embody different truth tables.

## Why it matters

- A value accepted by one funnel and inverted by the other is silent data corruption waiting for a second entry point (exactly how #322's NaN divergence was found: `toBool` said true where `toBoolForEAV` said false).
- The silent `"anything" → false` string path in `toBoolForEAV` is the same "coerce instead of reject" shape PR #403's P1 closed for floats.

## Direction (needs a ruling)

Pick one truth table and make both funnels share it (a single `boolFromAny` helper), deciding explicitly:
- negative numbers: true (`!= 0`) or false (`> threshold`)?
- fractional (0.3): true or false?
- unparseable strings: reject (ParseBool, `toBool`'s behavior) or coerce-false (`toBoolForEAV`'s)? Rejecting matches the repo's write-validation posture; coercing-false on the *read* path may need to stay tolerant for stored legacy 0/1-adjacent values — the read tolerance can be scoped to `*float64` 0/1 rather than shared with the write path.

Found during PR #403's response review; details in that PR's review-response comments.

### LINKED ISSUE #562: transform: toBoolForEAV integer-to-bool coercion is inconsistent across widths and pointer-ness
Raised in the PR #561 review (https://github.com/Lychee-Technology/forma/pull/561#issuecomment-5650569948, finding 4). The move in #561 preserved this byte-for-byte; nothing here is a regression.

## What is inconsistent

`internal/transform/scalar_conversion.go` `toBoolForEAV` coerces integers to bool through two different helpers depending on the Go type, with no stated reason:

| Input shape | Helper | Rule | `-1` → |
|---|---|---|---|
| `int`, `int16` | `boolFromPositive` | `> 0` | `false` |
| `int32`, `int64` | `boolFromNonZero` | `!= 0` | `true` |
| `*int`, `*int16`, `*int32`, `*int64` | `boolFromPositive` | `> 0` | `false` |
| `json.Number` | inline | `!= 0` | `true` |
| `float32`/`float64` (and pointers) | `float64ToBool` | `> 0.5` | `false` |

So the same negative value coerces differently depending on width (`int(-1)` vs `int32(-1)`) and on pointer-ness (`int32(-1)` vs `&int32(-1)`). The sibling funnel `toBool` (populateTypedValue) uses `!= 0` for every integer and float it accepts, which is a third answer for floats (`0.3` → `true` there, `false` in `toBoolForEAV`).

The existing tests in `attribute_converter_test.go` pin the current behaviour (`int(-1)→false`, `int32(-1)→true`, `int64(-1)→true`, `json.Number("-1")→true`), so this is characterised, not accidental at the test level — but the code carries no rationale, and `boolFromNonZero`'s type set (`~int32 | ~int64`) being narrower than `boolFromPositive`'s looks like an artefact of how the cases were added rather than a design.

## Reachability

HTTP callers cannot reach the integer cases: JSON decoding yields `float64`/`json.Number`/`bool`/`string`. The read path (`extractValueFromEAVRecord`) sends a `*float64` through `finiteFloat64ToBool`, which is consistent for the `0.0`/`1.0` that `boolToFloat64` stores. The integer paths are reachable only by in-process Go callers passing typed values, so the blast radius is the embedded-library API, not the server.

## Proposed resolution

1. Decide the intended integer→bool rule and document it once. The natural candidates are `!= 0` (C-like, matches `toBool` and the `json.Number` case) or `== 1` (strict, matches what `boolToFloat64` actually writes). `> 0` is the odd one out: it is neither.
2. Collapse `boolFromPositive`/`boolFromNonZero` into one helper with one type set, so width and pointer-ness cannot diverge again.
3. Rewrite the pinned negative-int expectations in `attribute_converter_test.go` to the chosen rule, and add a table that asserts value and pointer shapes agree for every width.
4. Consider whether `toBool` should share the same helper for its integer cases; the float threshold disagreement between the two funnels is separately documented in `numeric_conversion.go` (`finiteBoolInput`) and `nonfinite_bool_test.go` and can stay out of scope here if it widens the change too much.

This is a behaviour change for Go callers, so it needs a deliberate ruling, not a drive-by fix.

iceboundrock:
Closes #404. Closes #562 (the integer-width/pointer-ness inconsistency it reports is the same divergence; `boolFromPositive`/`boolFromNonZero` are deleted here and the pinned negative-int expectations rewritten to the ruled rule).

## Ruling

Recorded on the issue: https://github.com/Lychee-Technology/forma/issues/404#issuecomment-5652081093 (numeric domain amended after review, see the follow-up comment there).

**Write side is strict.** One shared `transform.boolFromAny` replaces `toBool` and `toBoolForEAV`:

| input | verdict |
|---|---|
| Go `bool` / `*bool` | value |
| `string` / `*string` | `strconv.ParseBool`, else reject |
| the write path's numeric set (`numutil.Float64`: int, int16, int32, int64, float32, float64, pointers to each) | finite check (#322/#403), then exactly `0` → false, `1` → true, else reject ("not a boolean image; must be 0 or 1") |
| `json.Number` | decided on its decimal text: a literal denoting exactly 0 or 1 (`"1"`, `"1.0"`, `"1e0"`, `"-0"`) is accepted; one that merely rounds to an image in float64 (`"1.0000000000000001"`) is rejected |
| int8, uint* (rejected by the numeric funnel too) | reject; widening every funnel together is #566 |
| anything else | reject |

Both funnels (`populateTypedValue` and `ToEAVRecord`) now give the same verdict for every input; the rejections are `forma.InvalidInputf` (4xx) as before, and a rejected width carries the `numutil` cause and the value.

**Read side reads the persisted image as the nearest of 0/1.** The Go `float64ToBool` `> 0.5` threshold is the deliberate rule (absorbs float noise around either end), so every SQL reader now spells the same `> 0.5` through a new `sqlgen.BoolTruthiness(expr)`: PG EAV EXISTS LHS, DuckDB hot pivot, list element cast, hot `bool_smallint` column (EAV-joined and no-EAV projections alike), CDC export for main and EAV values. Previously the SQL legs compared `<> 0`, and `BuildPGSelectNoEAV` projected the raw column for the outer `CAST(attr AS BOOLEAN)` to read by its own table.

**Bound bools stay nullable end to end.** The CDC export emits the bare truthiness (no `CASE … ELSE FALSE`), so an unset optional bool is NULL in parquet as it is in `entity_main`; `equals:false` does not start matching it at the flush boundary.

**The DuckDB outer select casts a bound bool back to its column's shape.** It cast to BOOLEAN while aliasing to the `entity_main` column the reader scans by kind, so every DuckDB read of a schema with a bound `bool_smallint` failed the `sql.NullInt64` scan and a bound `bool_text` arrived as `"true"`/`"false"` in the `"1"/"0"` slot. Pre-existing (independent of #404); surfaced by the review-requested e2e and fixed here: `SMALLINT` 1/0, `VARCHAR` '1'/'0', NULL kept.

**`bool_text` is read by its `"1"`/`"0"` contract on every leg.** The Go main-column read now errors on off-contract text (plain consistency error naming the column) instead of silently reading `false`; the CDC exporter drops its private `LOWER(col) IN ('true','1')` set for `= '1'`, matching the DuckDB hot pivot.

**Adjacent:** `convertPgBoolValue` (pg-main pushdown) parses operands with the engine-shared `parseBoolOperand`, so `equals:true` no longer 400s on that route alone.

The PG-EAV `1.0`/`0.0` operand divergence named in the issue was already closed by #384 (`Truthy`); only the spelling changed here.

## Follow-ups (not in this PR)

- #565: pg-main pushdown still compares the raw column (`m.smallint_01 = ?` / `m.text_02 = ?`) against the converted operand rather than through the truthiness expression. Under the 0/1 storage contract this is equivalent; a stored off-contract value (e.g. `2` in a `bool_smallint` column) reads as true on the DuckDB/EAV legs and as not-equal to `1` on this route. The unflushed DuckDB hot leg inherits the pushdown. Option comparison on the issue (sargable range rewrite recommended over a boolean-expression equality, which would lose the hot-field index).
- #566: widen the write-path numeric type set (int8/uint*) across every funnel at once, so the bool funnel does not accept a width the numeric funnel rejects.

## Tests

- `internal/transform/bool_truth_table_test.go`: ~65-row truth table asserting both write funnels agree, including the unsupported widths and the `json.Number` boundary literals on both sides of 0 and 1; rejection messages name the image rule, the literal as sent, and the wrapped cause; EAV read by threshold; `bool_text` read rejects `"true"`, `"TRUE"`, `"t"`, `""`, `"banana"`.
- `internal/sqlgen/bool_truthiness_test.go`: `BoolTruthiness` rendering, PG EAV LHS spelling, `ConvertPgMainValue` bool operands for both encodings, `BuildPGSelectNoEAV` bool spelling, and the outer-select cast executed in DuckDB and scanned with the reader's own types.
- `internal/cdc/export_bool_truth_test.go`: export leg spellings match the hot leg; the export expression executed in DuckDB over NULL and every on/off-contract image preserves NULL.
- `internal/e2e_harness/production/bool_bound_parity_e2e_test.go`: an all-bound schema (`bool_smallint` + `bool_text`) read on the unflushed DuckDB hot leg (no-EAV projection), delta parquet after a real flush and base parquet after a real compaction — same row sets, verdict images for planted off-contract rows, unset row NULL on every tier.
- Existing SQL assertions updated from `<> 0` to `> 0.5`.

`make fmt-check`, `make lint`, the full unit suite (`scripts/test_with_container_runtime.sh`) and the full production e2e suite (`-tags=e2e`) pass.

### DIFF
--- a/docs/federated-query/design.md
+++ b/docs/federated-query/design.md
@@ -882,11 +882,18 @@ applied on both sides:
   `NUMERIC` while the DuckDB route raised a Conversion Error on
   `CAST('1e+30' AS BIGINT)`. Fractional `bigint` operands in range are
   unaffected.
-* **Bool**: both engines compare the `value_numeric <> 0` truthiness — the PG
-  EAV EXISTS predicate renders `(x.value_numeric <> 0) =/!= <bool>` — and
-  parse operands under one shared rule (`ParseBool` spellings, else any
-  integer with `>0` truthiness), so no spelling errors on exactly one route.
-  (The write-side bool truth table is #404.)
+* **Bool**: both engines compare the `value_numeric > 0.5` truthiness
+  (`sqlgen.BoolTruthiness`) — the PG EAV EXISTS predicate renders
+  `(x.value_numeric > 0.5) =/!= <bool>` — and parse operands under one shared
+  rule (`ParseBool` spellings, else any integer with `>0` truthiness), so no
+  spelling errors on exactly one route. The threshold is the #404 read-side
+  rule shared with the Go read path (`transform.float64ToBool`): a persisted
+  image is the nearest of 0/1, so float noise around either end does not flip
+  the answer. The write side is strict (`transform.boolFromAny`): a bool
+  input is a Go bool, a `ParseBool` string, or a number that is exactly 0 or
+  1; anything else is rejected rather than coerced. A `bool_text` main column
+  is read by its `"1"`/`"0"` contract on every leg (Go, DuckDB hot leg, CDC
+  export); other text is a storage consistency error on the Go read path.
 
 Parquet files written before this contract carry INT32/INT16 attribute columns
 and NULLs where a value exceeded the declared width; `union_by_name=true` scans
--- a/internal/cdc/duckdb_exporter.go
+++ b/internal/cdc/duckdb_exporter.go
@@ -294,9 +294,13 @@ func castMainValue(col string, meta forma.AttributeMetadata) string {
 	case forma.ValueTypeBool:
 		switch meta.ColumnBinding.Encoding {
 		case forma.MainColumnEncodingBoolInt:
-			return fmt.Sprintf("CASE WHEN %s <> 0 THEN TRUE ELSE FALSE END", col)
+			// The shared #404 read-side truthiness, as the hot-leg
+			// sqlgen.mainColBoolExpr derives it.
+			return fmt.Sprintf("CASE WHEN %s THEN TRUE ELSE FALSE END", sqlgen.BoolTruthiness(col))
 		case forma.MainColumnEncodingBoolText:
-			return fmt.Sprintf("CASE WHEN LOWER(%s) IN ('true','1') THEN TRUE ELSE FALSE END", col)
+			// bool_text stores "1"/"0"; read by that contract, as the hot leg
+			// does (#404) — not a wider spelling set only this leg accepts.
+			return fmt.Sprintf("CASE WHEN %s = '1' THEN TRUE ELSE FALSE END", col)
 		default:
 			return fmt.Sprintf("TRY_CAST(%s AS BOOLEAN)", col)
 		}
@@ -343,7 +347,7 @@ func castDateMainValue(col string, meta forma.AttributeMetadata) string {
 func castEAVValue(meta forma.AttributeMetadata) string {
 	switch meta.ValueType {
 	case forma.ValueTypeBool:
-		return "(value_numeric <> 0)"
+		return sqlgen.BoolTruthiness("value_numeric")
 	case forma.ValueTypeDate, forma.ValueTypeDateTime:
 		return "TRY_CAST(value_numeric AS BIGINT)"
 	case forma.ValueTypeSmallInt, forma.ValueTypeInteger:
--- a/internal/cdc/duckdb_exporter_type_lockstep_test.go
+++ b/internal/cdc/duckdb_exporter_type_lockstep_test.go
@@ -24,7 +24,7 @@ const eavExportFixture = "(SELECT CAST(1 AS DOUBLE) AS value_numeric, CAST('x' A
 // the UNION ALL and re-opens #205.
 //
 // The comparison is on typeof() rather than on the rendered SQL text: bool
-// renders as `(value_numeric <> 0)`, which carries no type literal to
+// renders as `(value_numeric > 0.5)`, which carries no type literal to
 // extract — only the engine knows the resulting type.
 func TestCastEAVValueMatchesNullScanTypeof(t *testing.T) {
 	db, err := sql.Open("duckdb", ":memory:")
--- a/internal/cdc/export_bool_truth_test.go
+++ b/internal/cdc/export_bool_truth_test.go
@@ -1,0 +1,33 @@
+package cdc
+
+import (
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/lychee-technology/forma/internal/sqlgen"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCastMainValue_BoolEncodings pins the #404 read-side rule on the export
+// leg: a bool_smallint column is read as the nearest of 0/1 (the `> 0.5`
+// truthiness every other reader spells), and a bool_text column by its
+// "1"/"0" contract — the same `= '1'` the DuckDB hot-leg pivot uses, not a
+// wider LOWER(...) IN ('true','1') set only this exporter accepted.
+func TestCastMainValue_BoolEncodings(t *testing.T) {
+	boolInt := &forma.MainColumnBinding{ColumnName: forma.MainColumnSmallint01, Encoding: forma.MainColumnEncodingBoolInt}
+	boolText := &forma.MainColumnBinding{ColumnName: forma.MainColumnText02, Encoding: forma.MainColumnEncodingBoolText}
+
+	require.Equal(t, "CASE WHEN (m.smallint_01 > 0.5) THEN TRUE ELSE FALSE END",
+		castMainValue("m.smallint_01", forma.AttributeMetadata{ValueType: forma.ValueTypeBool, ColumnBinding: boolInt}))
+	require.Equal(t, "CASE WHEN m.text_02 = '1' THEN TRUE ELSE FALSE END",
+		castMainValue("m.text_02", forma.AttributeMetadata{ValueType: forma.ValueTypeBool, ColumnBinding: boolText}))
+}
+
+// TestCastEAVValue_BoolMatchesHotLegSpelling: the export leg and the hot-leg
+// pivot must render the bool truthiness identically, not merely to the same
+// type (#404) — a stored 0.3 must read the same from parquet and from
+// Postgres.
+func TestCastEAVValue_BoolMatchesHotLegSpelling(t *testing.T) {
+	require.Equal(t, sqlgen.BoolTruthiness("value_numeric"),
+		castEAVValue(forma.AttributeMetadata{ValueType: forma.ValueTypeBool}))
+}
--- a/internal/cdc/export_sql_shared_test.go
+++ b/internal/cdc/export_sql_shared_test.go
@@ -217,7 +217,7 @@ func TestBuildSchemaDrivenProjection_MixedBoundAndUnboundAttributesKeepCastsAlia
 	require.Contains(t, projection.eavAgg[0], "TRY_CAST(value_numeric AS DOUBLE)")
 	require.Contains(t, projection.eavAgg[0], "AS employee_count")
 	require.Contains(t, projection.eavAgg[1], "attr_id = 22")
-	require.Contains(t, projection.eavAgg[1], "(value_numeric <> 0)")
+	require.Contains(t, projection.eavAgg[1], "(value_numeric > 0.5)")
 	require.Contains(t, projection.eavAgg[1], "AS is_active")
 	require.Equal(t, []string{"e.employee_count", "e.is_active"}, projection.eavSelect)
 	require.Equal(t, []int16{21, 22}, projection.eavAttrIDs)
--- a/internal/e2e_harness/production/numeric_width_parity_e2e_test.go
+++ b/internal/e2e_harness/production/numeric_width_parity_e2e_test.go
@@ -16,7 +16,7 @@ import (
 // Postgres and every DuckDB tier must answer the same filter identically —
 // projection and operand casts by storage width (DOUBLE: the write funnel
 // narrows everything through float64), and one bool truth rule
-// (value_numeric <> 0) with one operand parse rule on both routes.
+// (value_numeric > 0.5, #404) with one operand parse rule on both routes.
 //
 // Out-of-range values are planted by direct eav_data UPDATEs: the write
 // funnel rejects them since #384 (TestEAVIntegerWidthWriteRejection), so the
@@ -201,7 +201,7 @@ func TestEAVIntegerWidthWriteRejection(t *testing.T) {
 }
 
 // TestEAVBoolTruthinessParityBothDialects pins the #384 bool ruling: both
-// routes compare the value_numeric <> 0 truthiness. Pre-#384 the OLTP route
+// routes compare the value_numeric > 0.5 truthiness (#404). Pre-#384 the OLTP route
 // bound equality against exactly 1.0/0.0, so a stored 2 (planted; the write
 // funnels store only 1/0, #404) matched equals:1 on the DuckDB tiers and not
 // on the hot route.
--- a/internal/sqlgen/bool_truthiness_test.go
+++ b/internal/sqlgen/bool_truthiness_test.go
@@ -1,0 +1,64 @@
+package sqlgen
+
+import (
+	"testing"
+
+	"github.com/lychee-technology/forma"
+	"github.com/stretchr/testify/require"
+)
+
+// TestBoolTruthiness pins the one read-side spelling every SQL reader derives
+// a bool from (#404): the stored numeric image is the nearest of 0/1, so
+// float noise around either end does not flip the answer. This is the SQL
+// image of transform.float64ToBool's threshold; the two must stay in
+// lockstep.
+func TestBoolTruthiness(t *testing.T) {
+	require.Equal(t, "(value_numeric > 0.5)", BoolTruthiness("value_numeric"))
+	require.Equal(t, "(x.value_numeric > 0.5)", BoolTruthiness("x.value_numeric"))
+}
+
+// TestPgEavLeafPayload_ComparisonLHSUsesBoolTruthiness: the PG-EAV leg
+// compares the same truthiness the DuckDB tiers derive, in the same spelling.
+func TestPgEavLeafPayload_ComparisonLHSUsesBoolTruthiness(t *testing.T) {
+	p := PgEavLeafPayload{ValueColumn: "value_numeric", Truthy: true}
+	require.Equal(t, BoolTruthiness("x.value_numeric"), p.ComparisonLHS("x"))
+}
+
+// TestConvertPgMainValue_BoolAcceptsEveryOperandSpelling: the pg-main
+// main-table route parses a bool operand under the same rule as the EAV and
+// DuckDB routes (parseBoolOperand, #384 P2b). Before #404 it ran strconv.Atoi
+// alone, so `equals:true` was a 400 on this route and a match on the others.
+func TestConvertPgMainValue_BoolAcceptsEveryOperandSpelling(t *testing.T) {
+	boolInt := forma.AttributeMetadata{
+		ValueType:     forma.ValueTypeBool,
+		ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumnSmallint01, Encoding: forma.MainColumnEncodingBoolInt},
+	}
+	boolText := forma.AttributeMetadata{
+		ValueType:     forma.ValueTypeBool,
+		ColumnBinding: &forma.MainColumnBinding{ColumnName: forma.MainColumnText02, Encoding: forma.MainColumnEncodingBoolText},
+	}
+	for _, tc := range []struct {
+		operand  string
+		wantInt  any
+		wantText any
+	}{
+		{operand: "true", wantInt: int64(1), wantText: "1"},
+		{operand: "TRUE", wantInt: int64(1), wantText: "1"},
+		{operand: "t", wantInt: int64(1), wantText: "1"},
+		{operand: "false", wantInt: int64(0), wantText: "0"},
+		{operand: "1", wantInt: int64(1), wantText: "1"},
+		{operand: "0", wantInt: int64(0), wantText: "0"},
+		{operand: "2", wantInt: int64(1), wantText: "1"},
+	} {
+		t.Run(tc.operand, func(t *testing.T) {
+			got, err := ConvertPgMainValue(tc.operand, "flag", boolInt)
+			require.NoError(t, err)
+			require.Equal(t, tc.wantInt, got)
+			got, err = ConvertPgMainValue(tc.operand, "flag", boolText)
+			require.NoError(t, err)
+			require.Equal(t, tc.wantText, got)
+		})
+	}
+	_, err := ConvertPgMainValue("banana", "flag", boolInt)
+	require.ErrorIs(t, err, forma.ErrInvalidInput)
+}
--- a/internal/sqlgen/dualpath_sql_helpers.go
+++ b/internal/sqlgen/dualpath_sql_helpers.go
@@ -54,39 +54,27 @@ func ConvertPgMainValue(valStr string, attr string, meta forma.AttributeMetadata
 	}
 }
 
-// convertPgBoolValue converts a boolean string value respecting column encoding.
+// convertPgBoolValue converts a boolean string value respecting column
+// encoding. The operand parse is the engine-shared parseBoolOperand rule
+// (#384 P2b); this route used to run strconv.Atoi alone, so `equals:true`
+// was a 400 here and a match on the EAV and DuckDB routes (#404).
 func convertPgBoolValue(valStr string, attr string, meta forma.AttributeMetadata) (any, error) {
-	parsedInt, err := strconv.Atoi(valStr)
-	if err != nil {
+	parsed, ok := parseBoolOperand(valStr)
+	if !ok {
 		return nil, forma.InvalidInputf("invalid boolean value for '%s': %s", attr, valStr)
 	}
 
-	if meta.ColumnBinding == nil {
-		// default to text "1"/"0"
-		if parsedInt > 0 {
-			return "1", nil
-		}
-		return "0", nil
-	}
-
-	switch meta.ColumnBinding.Encoding {
-	case forma.MainColumnEncodingBoolInt:
-		if parsedInt > 0 {
+	if meta.ColumnBinding != nil && meta.ColumnBinding.Encoding == forma.MainColumnEncodingBoolInt {
+		if parsed {
 			return int64(1), nil
 		}
 		return int64(0), nil
-	case forma.MainColumnEncodingBoolText:
-		if parsedInt > 0 {
-			return "1", nil
-		}
-		return "0", nil
-	default:
-		// default to text "1"/"0"
-		if parsedInt > 0 {
-			return "1", nil
-		}
-		return "0", nil
 	}
+	// bool_text and the unbound default both bind the "1"/"0" text contract.
+	if parsed {
+		return "1", nil
+	}
+	return "0", nil
 }
 
 // detectValueType infers the forma.ValueType from a string literal when no metadata is available.
--- a/internal/sqlgen/duckdb_cold_scan_test.go
+++ b/internal/sqlgen/duckdb_cold_scan_test.go
@@ -156,7 +156,7 @@ const eavFixtureSubquery = "(SELECT CAST(1 AS DOUBLE) AS value_numeric, CAST('x'
 // pg_source leg and the NULL-augmented parquet leg and re-opens #205.
 //
 // The comparison is on typeof() rather than on the rendered SQL text: bool
-// renders as `(value_numeric <> 0)` and text as a bare `value_text`, so
+// renders as `(value_numeric > 0.5)` and text as a bare `value_text`, so
 // neither carries a type literal to extract — only the engine knows.
 func TestDuckDBNullScanTypeMatchesHotLegTypeof(t *testing.T) {
 	db, err := sql.Open("duckdb", ":memory:")
--- a/internal/sqlgen/duckdb_eav_pivot.go
+++ b/internal/sqlgen/duckdb_eav_pivot.go
@@ -36,8 +36,9 @@ func buildEAVPivotExpr(a attrProjectionInfo) string {
 			a.attrID, eavElementCastExpr(a.meta.EffectiveItemsType()), a.attrID)
 	}
 	if a.meta.ValueType == forma.ValueTypeBool {
-		// Wrap in <> 0 so the pivot column is BOOLEAN, not DOUBLE (#182).
-		return fmt.Sprintf("(MAX(CASE WHEN attr_id = %d THEN value_numeric END) <> 0)", a.attrID)
+		// Wrap in the truthiness comparison so the pivot column is BOOLEAN,
+		// not DOUBLE (#182), spelled by the shared #404 rule.
+		return BoolTruthiness(fmt.Sprintf("MAX(CASE WHEN attr_id = %d THEN value_numeric END)", a.attrID))
 	}
 	base := fmt.Sprintf("MAX(CASE WHEN attr_id = %d THEN %s END)",
 		a.attrID, eavValueColumn(a.meta.ValueType))
@@ -74,7 +75,7 @@ func buildEAVPivotExpr(a attrProjectionInfo) string {
 func eavElementCastExpr(vt forma.ValueType) string {
 	switch vt {
 	case forma.ValueTypeBool:
-		return "(value_numeric <> 0)"
+		return BoolTruthiness("value_numeric")
 	case forma.ValueTypeBigInt, forma.ValueTypeDate, forma.ValueTypeDateTime:
 		return "TRY_CAST(value_numeric AS BIGINT)"
 	case forma.ValueTypeInteger, forma.ValueTypeSmallInt:
@@ -103,12 +104,14 @@ func eavValueColumn(vt forma.ValueType) string {
 
 // mainColBoolExpr returns a DuckDB boolean expression that normalises a
 // column-bound bool main-table column to a BOOLEAN value.
-// bool_text encoding stores "1"/"0" as text; bool_smallint stores 0/1 as SMALLINT.
-// Both must produce a BOOLEAN so that COALESCE(hot_vals.<attr>, <expr>) is type-safe.
+// bool_text encoding stores "1"/"0" as text and is read by that contract;
+// bool_smallint stores 0/1 as SMALLINT and is read by the shared #404
+// truthiness. Both must produce a BOOLEAN so that
+// COALESCE(hot_vals.<attr>, <expr>) is type-safe.
 func mainColBoolExpr(colName string, enc forma.MainColumnEncoding) string {
 	if enc == forma.MainColumnEncodingBoolText {
 		return fmt.Sprintf("m.%s = '1'", colName)
 	}
 	// Default covers MainColumnEncodingBoolInt and any other numeric-like encoding.
-	return fmt.Sprintf("m.%s <> 0", colName)
+	return BoolTruthiness("m." + colName)
 }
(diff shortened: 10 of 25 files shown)


iceboundrock:
# Review: PR #564 — one bool truth table across write and read (#404)

Reviewed at `7cd2694` against the ruling recorded in #404 (comment `5652081093`), the issue body, and the second-comment extensions (`bool_text` read, read-side federated divergence). This is the single consolidated review comment (findings / spec / standards / assessment / observations / verification / summary).

## Findings

No blockers. All five pre-existing spellings identified in the ruling survey are folded into the two ruled behaviors, with the one explicitly scoped-out follow-up below.

1. **Write side is strict and shared — PASS.** `internal/transform/scalar_conversion.go:46` `boolFromAny` replaces both `toBool` and `toBoolForEAV`. Both funnels call it (`typed_value.go`, `attribute_converter.go:93`). Rule matches the ruling exactly: Go bool passthrough; `string`/`*string` via `strconv.ParseBool` (`boolFromString`); numerics/pointers/`json.Number` via `finiteBoolInput` then exactly `0`/`1` (`boolFromNumeric`), else `value %v is not a boolean image; must be 0 or 1`. Deleted helpers (`boolFromPositive`, `boolFromNonZero`, `isTrueStringForEAV`, `finiteFloat64ToBool`) cannot diverge again by construction.
2. **Read side is `> 0.5` everywhere in SQL — PASS.** New `internal/sqlgen/predicate_ir.go:BoolTruthiness` renders `(expr > 0.5)` and is used by all four SQL readers: PG-EAV `ComparisonLHS`, DuckDB EAV pivot (`duckdb_eav_pivot.go:buildEAVPivotExpr`, `eavElementCastExpr`), DuckDB hot `mainColBoolExpr` for `bool_smallint`, and CDC export (`internal/cdc/duckdb_exporter.go:castMainValue`, `castEAVValue`). Go `float64ToBool` threshold is unchanged (`bool_numeric.go`), and `extractValueFromEAVRecord` (`attribute_value_extraction.go`) now guards with `finiteBoolInput` then calls `float64ToBool` directly instead of routing through the write funnel — exactly what the ruling ordered.
3. **`bool_text` contract — PASS.** Go main-column read (`persistent_record.go:readWithEncoding`) now goes through `boolFromBoolText` (`bool_numeric.go`): `"1"`/`"0"` only, else plain storage-consistency error naming the column (`column %s: %w`). CDC exporter aligns from `LOWER(col) IN ('true','1')` to `= '1'`, matching the DuckDB hot pivot. SQL legs keep `= '1'` since SQL cannot reject — as ruled.
4. **Adjacent `convertPgBoolValue` — PASS.** `internal/sqlgen/dualpath_sql_helpers.go` now parses via engine-shared `parseBoolOperand` instead of `strconv.Atoi` alone, so `equals:true` no longer 400s on pg-main while matching on EAV/DuckDB. Nil-`ColumnBinding` handling is preserved (defaults to the `"1"`/`"0"` text bind) and is now branch-flat.
5. **Already-closed divergence correctly scoped.** The PG-EAV `1.0`/`0.0` operand divergence is, as the PR body states, already closed by #384 (`Truthy`); this PR only re-spells it through `BoolTruthiness`. No extra work needed there.

## Spec check (against #404 + #384/#322 interactions)

- Issue table rows all pinned by `internal/transform/bool_truth_table_test.go`: `float64(0.3)`/`(-1)` reject on both funnels (was true/false split), `"banana"`/`"NaN"`/`""` reject on both (was error vs silent-false split), `int(-1)` rejects on both (was true vs false split). `json.Number("1.0")` accepts as image `1` — correct under "exactly 0 or 1" since `1.0 == 1`.
- Read-side pins match the ruling: stored `0.3→false`, `0.7→true`, `2→true`, `-1→false`, `0.5→false`; `bool_text` `"true"/"TRUE"/"t"/""/"banana"` error naming `text_02` + `bool_text`, and `NotErrorIs(ErrInvalidInput)` (read-path consistency, not caller input).
- SQL spelling pins match on every leg (`bool_truthiness_test.go`, `export_bool_truth_test.go`, updated `<> 0` → `> 0.5` assertions). `NULL` semantics are unchanged (`NULL > 0.5` is `NULL`, same as `NULL <> 0`), so no result change for absent values.
- #322 non-finite guard preserved on both write (`boolFromNumeric` → `finiteBoolInput`) and read (`extractValueFromEAVRecord` guard). #384 `parseBoolOperand` integer `> 0` operand rule untouched, as ruled.
- `docs/federated-query/design.md` bool section updated to the new spelling with the threshold rationale and the strict-write rule — accurate, no stale `<> 0` left in the touched lines.

## Standards check (`coding-standard.md`, `docs/error-handling.md`)

- **File/function caps:** all touched non-test files stay under 500 lines (`persistent_record.go` 492, `duckdb_exporter.go` 376, `dualpath_sql_helpers.go` 229, `predicate_ir.go` 186, `scalar_conversion.go` 156); new helpers are short with early returns and flat structure (`boolFromAny`, `boolFromNumeric`, `derefNumericPointer`, `boolFromBoolText`, `BoolTruthiness`). Nesting never exceeds two levels.
- **Error semantics:** write path correctly builds `forma.InvalidInputf` carriers at the boundary (`populateTypedValue.handleConversionError`, `convertPgBoolValue`) with attribute/column naming and expected state (`active`, `0.3`, `must be 0 or 1` — pinned by test); `ToEAVRecord` stays plain with `attrID` context per its documented operator-visible contract. Read path (`boolFromBoolText`, `finiteBoolInput` on read) stays plain, never `InvalidInputf`. `TestNoBareSentinelWraps` passes (verified locally).
- **Wrapping:** every new error carries context (`cannot convert string %q`, `column %s: %w`, `convert to bool for attrID %d: %w`); no bare `return err` on a newly created error path except the intentional read-path passthrough of `finiteBoolInput`, which already carries its own prose (pre-existing shape).
- `gofmt` clean, `go vet` clean on `transform`/`sqlgen`/`cdc`.

## Assessment

Ready to merge. The change implements the ruling faithfully, deletes the divergent helpers rather than adding a third spelling, and pins the contract with a truth-table test that makes future divergence a test failure. The acknowledged pg-main follow-up (raw-column `m.smallint_01 = ?` vs truthiness expression) is correctly scoped out: equivalent under the now-enforced 0/1 storage contract, divergent only for off-contract legacy data.

Recommended follow-ups (not merge-blockers):

- File the pg-main `BoolTruthiness` follow-up as its own issue (searched open issues; no existing issue covers it) so the PR body's "worth a separate issue" does not live only in the PR text.
- Close #562 with this merge (or explicitly link it): `boolFromPositive`/`boolFromNonZero` and the pinned `int(-1)`/`int32(-1)` expectations it complains about are deleted/rewritten here; #562 is subsumed by the #404 ruling.
- Note `persistent_record.go` moves 488 → 492 lines, leaving 8 lines to the 500 cap flagged in #558 — the next edit there must move the read helpers out as #558 already plans.

## Observations (minor, non-blocking)

- `boolFromNumeric` inherits `numutil.Float64`'s type set, so `uint*`/`int8` still reject with `cannot convert %T to bool` — identical to the old funnels, so no regression, but the ruling's "any numeric width" in practice means "every width the codebase already funnels." HTTP callers (float64/`json.Number`/bool/string) are unaffected.
- `BoolTruthiness` lives in `predicate_ir.go`; functionally fine (no import cycle — `cdc` already imports `sqlgen`) and the file stays at 186 lines, but a dedicated home would be more discoverable if more shared SQL spellings follow.
- `castMainValue`'s nil-`ColumnBinding` dereference and the store path's `float64ToBool` threshold write (`persistent_record.go:storeWithEncoding`) are both pre-existing and harmless here (writes are now strictly 0/1, so threshold ≡ exact), left untouched appropriately.

## Verification

On the PR worktree (`7cd2694`), with `GOTOOLCHAIN=go1.26.0+auto`:

- `go test ./internal/transform/... ./internal/sqlgen/... ./internal/cdc/...` — all `ok`.
- Targeted: `TestBoolWriteFunnelsShareOneTruthTable` (~40 subtests), `TestBoolRejectionNamesTheImageRule`, `TestExtractValueFromEAVRecordReadsBoolByThreshold`, `TestBoolTextReadRejectsOffContractText`, `TestBoolTruthiness`, `TestPgEavLeafPayload_ComparisonLHSUsesBoolTruthiness`, `TestConvertPgMainValue_BoolAcceptsEveryOperandSpelling`, `TestCastMainValue_BoolEncodings`, `TestCastEAVValue_BoolMatchesHotLegSpelling` — all PASS.
- `go test . -run TestNoBareSentinelWraps` — PASS; `go test . ./factory/...` — ok; `go vet` on the three packages — clean; `gofmt -l` on all touched files — clean.
- Did not run container E2E (`numeric_width_parity_e2e_test.go` touch is comment-only) or `make lint`; CI covers both.

## Summary

This PR closes #404 as ruled: one strict write funnel (`boolFromAny`, exact 0/1) shared by both entry points, one tolerant read rule (`> 0.5` / `float64ToBool`) spelled identically on every SQL leg through `BoolTruthiness`, `bool_text` held to its `"1"`/`"0"` contract on every leg, and the pg-main operand parse unified with `parseBoolOperand`. Well-tested, standards-clean, no blockers — approve, with the pg-main truthiness follow-up and #562 linkage handled as separate issue actions.

iceboundrock:
Response to the review at `7cd2694`. Branch is now at `f5413fb`.

## Recommended follow-ups

**pg-main `BoolTruthiness` follow-up — filed as #565.** Confirmed the divergence is real for both the pg-main emitter (`dualpath_sql_generator.go`) and the hybrid builder's `IsMain` branch (`postgres_persistent_repository_query.go:emitMainLeaf`): both emit `m.smallint_01 = $1` with `int64(1)`, while the Go OLTP read of the same column goes `readWithEncoding` → `float64ToBool`, so a stored `2` reads as `true` and is not matched by `equals:true` on that one route. The issue compares three options: (A) leave the raw compare and document the contract; (B) rewrite `equals:true` to `m.col > 0.5` / `equals:false` to `m.col <= 0.5` at emit time — same rule as every other leg, still btree-range-sargable; (C) bind a Go bool against `(m.col > 0.5) = $1` like PG-EAV does — simplest spelling but a boolean-expression equality that a btree on the hot column cannot serve without an expression index. Recommendation in the issue is B: it closes the last route with its own bool table without regressing the plan the pushdown exists to produce. `bool_text` is unaffected on this route (`"1"`/`"0"` bind is already the contract every leg reads by). PR body updated to link it.

**#562 linkage — done.** PR body now carries `Closes #562`, and #562 has a comment mapping its four proposed-resolution points onto what this PR ships (strict `== 1` rule chosen and documented on `boolFromAny`; `boolFromPositive`/`boolFromNonZero` deleted; negative-int pins rewritten; `toBool` shares the helper).

**`persistent_record.go` 488 → 492 — acknowledged on #558.** Posted the new count there; the read-helper split proposed in #558 is the precondition for the next edit to that file. #565 is sqlgen-side and does not touch it.

## Observations

**`boolFromNumeric` type set (no `uint*`/`int8`) — no change.** Checked against the numeric write funnel: `toFloat64ForEAV` (`numeric_conversion.go`) accepts exactly the same set — `float32/64`, `int/16/32/64`, their pointers, `json.Number`, `string` — via the same `numutil.Float64` / `NumericScalar`. So the bool funnel accepts every width the package already accepts for numbers, and the ruling's "any numeric width" reads as "any width the write path knows". Widening one funnel alone would reintroduce a per-type-set asymmetry in the other direction; if `uint`/`int8` are ever wanted from embedded Go callers, that is a `numutil.NumericScalar` change covering both funnels, not a bool-only one.

**`BoolTruthiness` location — moved.** `f5413fb` gives it `internal/sqlgen/bool_truthiness.go`, matching the test file that already existed under that name. Pure move; no behaviour change. `go test ./internal/sqlgen/ ./internal/cdc/ ./internal/transform/`, `gofmt`, and `make lint` clean.

**`castMainValue` nil-binding deref and `storeWithEncoding` threshold write — no change.** The only caller of `castMainValue` (`export_sql_builder.go:130`) is inside `if meta.ColumnBinding != nil`, so the deref is guarded at the call site. `storeWithEncoding` receives `attr.ValueNumeric` that `boolFromAny` already reduced to exactly `0`/`1`, so `float64ToBool` there is a threshold over a two-valued domain — equivalent by construction, and it is the same `float64ToBool` the read side uses, so the pair stays in lockstep if the write rule ever changes.

iceboundrock:
## Findings

1. **P1: Preserve NULL when exporting optional, column-bound booleans.** [`castMainValue`](https://github.com/Lychee-Technology/forma/blob/f5413fbf2e47e25a36b55be50b3a7d3dc5093f2b/internal/cdc/duckdb_exporter.go#L292-L306) now renders `CASE WHEN <nullable bool expression> THEN TRUE ELSE FALSE END`. SQL three-valued logic maps both `NULL > 0.5` and `NULL = '1'` to the `ELSE`, so an unset optional `bool_smallint`/`bool_text` becomes persisted `false` at CDC flush. The hot projection retains `NULL`, thus `equals:false` and the returned entity can change solely when the row moves to Parquet. Export `sqlgen.BoolTruthiness(col)` or `col = '1'` directly; both yield nullable BOOLEAN. Add a CDC execution test for unset values, not only rendered SQL.

2. **P1: The no-EAV hot projection bypasses the new decoder, creating a flush-boundary regression for off-contract data.** [`BuildPGSelectNoEAV`](https://github.com/Lychee-Technology/forma/blob/f5413fbf2e47e25a36b55be50b3a7d3dc5093f2b/internal/sqlgen/duckdb_schema_projection.go#L285-L308) selects raw `m.<column>` instead of `mainColBoolExpr`. This optimization is live for a production schema whose attributes are all column-bound. For legacy/out-of-band `bool_text='true'`, the hot path casts raw text to BOOLEAN and returns `true`; this PR changes the CDC export to `false` because only `'1'` is true. `bool_smallint=-1` similarly reads true in the fast hot path but false after the new `> 0.5` export. Before this PR the old CDC spellings agreed with the raw casts, so this is a new flush-boundary behavior change. Preserve the attribute encoding in `SchemaProjection` and emit `mainColBoolExpr` for bool columns in the fast projection. Cover a no-EAV schema through hot and flushed reads for both encodings.

3. **P2: Implement the ruled numeric domain rather than the narrower `numutil.Float64` domain.** [`boolFromNumeric`](https://github.com/Lychee-Technology/forma/blob/f5413fbf2e47e25a36b55be50b3a7d3dc5093f2b/internal/transform/scalar_conversion.go#L72-L90) and [`derefNumericPointer`](https://github.com/Lychee-Technology/forma/blob/f5413fbf2e47e25a36b55be50b3a7d3dc5093f2b/internal/transform/scalar_conversion.go#L92-L116) reject valid `int8(0/1)`, every unsigned integer width, and their pointers because `numutil.Float64` and the dereferencer do not support them. That contradicts the #404 ruling and the PR description, both of which promise every numeric width and pointer-ness. Extend the accepted type set (or explicitly narrow the ruling and documentation), and add the missing scalar and pointer cases to the shared truth table.

4. **P2: Do not round `json.Number` before deciding whether its image is exact.** The same conversion turns `json.Number("1.0000000000000001")` into `float64(1)`, accepts it, and persists `true`, although the submitted numeric image is not exactly `0` or `1`. This defeats the strict-write rule for JSON-decoded programmatic inputs. Decide `json.Number` with exact decimal/integer parsing before a float64 conversion, and add boundary tests on both sides of 0 and 1.

## Specification

The #404 ruling requires strict writes (`bool`/`ParseBool` strings/exact numeric images 0 or 1), `> 0.5` reads for every numeric reader, and `'1'/'0'` for `bool_text`. The shared write helper, EAV extraction, PG-EAV predicate, normal projection, and CDC expressions move substantially toward that contract. Findings 1 and 2 mean the stated every-reader/flush-invariant is not yet satisfied; findings 3 and 4 mean the stated numeric acceptance domain is not yet implemented.

The PR body correctly records raw pg-main/hybrid-main comparison as the separately filed follow-up [#565](https://github.com/Lychee-Technology/forma/issues/565). It remains an off-contract-data divergence, but is not repeated here as a new finding.

## Standards

- File and function sizes are within the repository limits; the exported `BoolTruthiness` has suitable GoDoc; the extraction of the shared SQL spelling is cohesive.
- Error handling needs a small follow-up in `boolFromNumeric`: it discards the `numutil.Float64` error (`json.Number("abc")` becomes only `cannot convert json.Number to bool`). Wrap the cause and include the supplied value so the published invalid-input error remains actionable, as required by `coding-standard.md` and the write-error policy.
- New tests assert SQL strings and selected truth-table rows, but omit nullable CDC behavior, the no-EAV optimization, and the numeric shapes/boundaries above.

## Observations

- The `bool_text` Go read now correctly surfaces off-contract storage as a plain consistency error rather than silently rewriting it.
- `ConvertPgMainValue` now accepts the shared boolean operand spellings, resolving the prior `equals:true` route inconsistency.
- The documentation accurately explains the intended strict-write/tolerant-read split, but should be updated with the actual supported numeric domain only if the implementation is deliberately narrowed.

## Verification

- Reviewed both PR commits: `7cd2694` and `f5413fb`.
- `make fmt-check` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform ./internal/sqlgen ./internal/cdc` passed.
- GitHub checks are all green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

## Assessment

**Changes requested.** The consolidation is directionally strong, but the two P1 paths make boolean values change at a CDC flush boundary, and the P2 issues violate the explicit strict-input contract.

## Summary

This PR usefully centralizes the main write and read rules, documents the decision, and adds focused coverage. Preserve nullable booleans, normalize the no-EAV hot projection, and complete exact numeric handling before merge.

iceboundrock:
**Review response (PR #564, https://github.com/Lychee-Technology/forma/pull/564#issuecomment-5655186818)** — all four findings and the standards note addressed in `e847a11`; one pre-existing defect surfaced by the requested e2e and fixed in the same commit; one ruling narrowed and filed as #566.

### Finding 1 (P1) — CDC export collapses NULL to false

Confirmed. `castMainValue` wrapped the truthiness in `CASE WHEN … THEN TRUE ELSE FALSE END`; under three-valued logic `NULL > 0.5` is NULL and the wrapper turned it into a persisted `false`, so an unset optional bool matched `equals:false` only after flush.

Fixed: the export now emits the bare nullable expression the hot leg spells — `(m.col > 0.5)` for `bool_smallint`, `m.col = '1'` for `bool_text` — i.e. exactly `sqlgen.mainColBoolExpr`. `TestCastMainValue_BoolExportPreservesNull` executes the rendered expression in an in-memory DuckDB over `NULL`, `1`, `0`, `2`, `-1`, `'1'`, `'0'`, `'true'`, `''` and scans `sql.NullBool`; NULL stays NULL, the off-contract rows pin that the parquet image is the hot-leg verdict and not a wider one. The e2e below covers the unset row through a real flush and compaction.

### Finding 2 (P1) — `BuildPGSelectNoEAV` bypasses `mainColBoolExpr`

Confirmed, with one correction to the diagnosis: the reviewer's example (`bool_smallint = -1` reads *true* on the no-EAV route) is what the SQL would have said, but that route could not actually return any row for a bound `bool_smallint` — see "surfaced defect" below.

Fixed: `SchemaProjection` now carries `boundBoolEncodings`, and `BuildPGSelectNoEAV` emits `mainColBoolExpr(col, enc)` for bound bools, same spelling as the EAV-joined projection's COALESCE leg (`TestBuildPGSelectNoEAV_BoolColumnsUseMainColBoolExpr`).

E2E: `internal/e2e_harness/production/bool_bound_parity_e2e_test.go` (`TestBoundBoolParityNoEAVAllTiers`) uses an all-bound `e2e_simple` (`name`→`text_01`, `flagInt`→`smallint_01` `bool_smallint`, `flagText`→`text_02` `bool_text`), so the no-EAV shortcut is live, and reads the same rows on the unflushed DuckDB hot leg, delta parquet after `RunFlush`, and base parquet after `RunCompaction`. Rows: true, false, unset, plus planted `smallint_01 = -1`, `smallint_01 = 2`, `text_02 = 'true'` (direct `UPDATE entity_main`). Assertions: filter row sets on both encodings, the raw record image on every DuckDB tier (`2 → 1`, `-1 → 0`, `'true' → "0"`, unset → no value in either column).

One scoping note on the e2e: on the unflushed hot leg a filter *on the bool itself* is pushed down into the Postgres scan as the raw compare (`smallint_01 = 1`), which hides the off-contract rows from both sides of the filter before DuckDB sees them. That is #565's route (I have added a note there that the DuckDB hot leg inherits the pushdown, which its table did not say). So the hot-leg filter probes run over on-contract images and the off-contract rows are pinned on the hot leg through the projected image; the flushed tiers, which have no pushdown, pin the filter verdict on off-contract rows.

### Surfaced defect (pre-existing, fixed here) — bound bools unreadable through DuckDB

The e2e failed before any of the above mattered: `scan duckdb row: … converting driver.Value type bool ("true") to a int64: invalid syntax` on `smallint_01`. The outer select rendered `CAST(flagInt AS BOOLEAN) AS smallint_01` while `federated.duckDBScanBuffers` scans that column by its kind (`sql.NullInt64`); `database/sql` cannot assign a Go `bool` to an int64. For `bool_text` the same cast landed `"true"`/`"false"` in the `sql.NullString` slot the `"1"/"0"` funnel reads (silently `false` before this PR, a consistency error after it). This is independent of the no-EAV route — the EAV-joined hot leg and every parquet leg hand the outer select a BOOLEAN — so any DuckDB read of a schema with a bound `bool_smallint` attribute has failed since the outer select was written. No production fixture binds a bool, which is why nothing caught it; I confirmed the failure on this branch with the e2e and by reasoning that `duckDBAttrCast` is unchanged from `main`.

Fix (`duckDBMainColCast`): for a bound bool the outer select casts the verdict back to the physical column's shape — `CAST(attr AS SMALLINT)` (1/0) and `CAST(CAST(attr AS TINYINT) AS VARCHAR)` ('1'/'0'), NULL preserved. Non-bool attributes are unchanged. `TestBuildOuterSelect_BoundBoolCastsToPhysicalColumn` executes the cast in DuckDB and scans with the reader's own `sql.NullInt64`/`sql.NullString` for TRUE/FALSE/NULL. I kept it in this PR because the reviewer-requested e2e cannot pass without it and it is the same "one bool image on every route" contract; it is a separate paragraph in the commit message so it can be split out if you would rather track it on its own issue.

Consequence worth naming: a DuckDB read of an off-contract row returns the *verdict's* image (`-1` reads back as `0`), while the Postgres route returns the raw `-1`. Both feed `float64ToBool` to the same `false`, so the API answer agrees; only the raw `PersistentRecord` differs, and only for images the write funnel already rejects.

### Finding 3 (P2) — numeric type set narrower than the ruling

Confirmed that the ruling text ("any numeric width") and the code disagreed. I chose to **narrow the ruling** rather than widen `boolFromNumeric`:

- The bool funnel delegates to `numutil.Float64`, which is the write path's numeric type set (int, int16, int32, int64, float32, float64, json.Number, string) and is shared with `toFloat64ForEAV` for numeric attributes. Widening bool alone would make `int8(1)` a valid bool and `int8(1)` an invalid integer on the same request shape — the funnel asymmetry #562 just closed, reintroduced one type over.
- No producer sends these widths: httpapi decodes with `UseNumber`, so every HTTP number is a `json.Number`; the batch/relation paths hand through the same decoded values.
- Widening every funnel together is a `numutil`-level change with its own truth table (uint64 above 2^53, int8 range checks against declared widths) and is filed as #566.

Changes: `boolFromAny`'s doc now states the set precisely and names #566; the truth table adds reject rows for `int8`, `*int8`, `uint`, `uint8`, `*uint8`, `uint16`, `uint32`, `uint64`, `*uint64` so both funnels are pinned to the same verdict; `docs/federated-query/design.md` bool section carries the same wording; the #404 ruling comment is amended below the original text.

### Finding 4 (P2) — `json.Number("1.0000000000000001")` accepted

Confirmed. `boolFromNumberLiteral` now decides a `json.Number` on its decimal text: `strconv.ParseFloat` is the acceptance authority (so `"abc"`, `"1e400"` are rejected with the parse error), then `numutil.TryParseNumber` refines any literal denoting an integer exactly to `int64` — `"1"`, `"1.0"`, `"1e0"`, `"10e-1"`, `"-0"`, `"0.0"` are accepted; genuine fractions, integers past int64, `NaN`/`Inf` stay `float64` and are rejected. Boundary rows both sides of 0 and 1: `1.0000000000000001`, `0.99999999999999999`, `0.0000000000000000001`, `-0.0000000000000000001`, `1e-400`, `9007199254740993`, `99999999999999999999` — all rejected, with the message naming the literal the caller sent rather than its rounded image (`TestBoolRejectionNamesTheImageRule/json.Number rounding to an image`).

### Standards note — discarded `numutil.Float64` error

Fixed: `cannot convert %T value %v to bool: %w`. Pinned by `TestBoolRejectionNamesTheImageRule/unsupported width` (`uint8`, `value 1`, `cannot convert uint8 to float64` all present).

### Verification

- `make fmt-check`, `make lint` (root + infra): clean.
- `make test` (full unit suite via podman): green.
- `go test ./internal/e2e_harness/production/ -tags=e2e`: full production suite green, including the new `TestBoundBoolParityNoEAVAllTiers`.

### Follow-ups

- #565: added the DuckDB-hot-leg-inherits-the-pushdown note.
- #566: package-wide numeric width widening (int8/uint*), filed from finding 3.

iceboundrock:
## Review: PR #564 — one bool truth table across write and read

Reviewed final head `e847a11` against the amended #404 ruling, the PR description, `docs/federated-query/design.md`, `coding-standard.md`, and `docs/error-handling.md`. This is the single consolidated review comment covering findings, specification, standards, assessment, observations, verification, and summary.

## Findings

No new blockers or actionable findings. The four findings from the earlier review have been addressed in `e847a11`:

1. Bound optional bool CDC exports now retain NULL by emitting the bare nullable verdict rather than `CASE ... ELSE FALSE`.
2. `BuildPGSelectNoEAV` now uses `mainColBoolExpr` for bound bools, matching the EAV-joined hot projection and exported parquet verdict.
3. The numeric input domain is explicitly narrowed to the project-wide `numutil.Float64` funnel; unsupported `int8`/`uint*` inputs reject in both write funnels and the package-wide widening is tracked in #566.
4. `json.Number` is decided from its decimal text, rejecting literals that merely round to `0` or `1` in `float64`.

The final commit also correctly fixes the surfaced pre-existing bound-bool DuckDB scan defect by casting the logical verdict back to the physical main-column representation.

## Specification

- **Write rule: PASS.** `transform.boolFromAny` is the shared strict write funnel for `populateTypedValue` and `ToEAVRecord`: bools and `ParseBool` strings are accepted; supported finite numeric inputs must be exactly `0` or `1`; all other values reject.
- **Read rule: PASS.** Numeric persisted bool images are read as `> 0.5` in Go and every relevant SQL leg through `sqlgen.BoolTruthiness`.
- **`bool_text`: PASS.** Go accepts only stored `"1"`/`"0"` and raises a plain consistency error for other stored text. SQL correctly uses `= '1'`, because it cannot surface that consistency error.
- **Tier/nullability parity: PASS.** Bound bools retain NULL across hot, delta, and base. The no-EAV projection and the outer scan now preserve the physical `SMALLINT` or text image expected by the federated reader.
- **Predicate operand parity: PASS.** pg-main conversion now uses `parseBoolOperand`, so standard boolean operand spellings no longer differ by route.

The deliberately scoped raw pg-main comparison of off-contract legacy images remains tracked by #565. It is not a new finding here.

## Standards

- Touched production files and added helpers remain within the repository file/function limits; the new exported `BoolTruthiness` has appropriate GoDoc.
- Error classes follow the documented policy: write-boundary failures become `forma.InvalidInputf`; read-side storage inconsistencies remain plain errors. The `numutil.Float64` cause and supplied value are retained in conversion failures.
- The implementation is cohesive: divergent helpers were removed in favor of one write funnel and one shared SQL read expression.

## Observations

- The new production E2E covers an all-column-bound schema through unflushed DuckDB hot data, real CDC delta, and compaction base data, including NULL and planted off-contract images. This closes the previously missing path coverage.
- Existing off-contract data has intentionally asymmetric raw-record representation on the direct Postgres versus DuckDB route, but both produce the same logical API boolean verdict. This behavior is documented and #565 covers the remaining predicate-pushdown aspect.
- The PR is large but its added tests target the previously divergent behavior rather than duplicating unrelated coverage.

## Verification

Local review worktree at `e847a11`:

- `git diff --check origin/main...HEAD` — PASS
- `make fmt-check` — PASS
- `GOCACHE=/home/ruoshi/code/lychee/forma/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform -run 'TestBool' -count=1 -v` — PASS
- `GOCACHE=/home/ruoshi/code/lychee/forma/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen -count=1` — PASS
- `GOCACHE=/home/ruoshi/code/lychee/forma/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/cdc -count=1` — PASS
- `GOCACHE=/home/ruoshi/code/lychee/forma/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test . -run TestNoBareSentinelWraps -count=1` — PASS
- GitHub CI is green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

The full local multi-package command initially hit the `/tmp` filesystem quota while linking DuckDB; rerunning the affected packages with a cache on the root filesystem passed.

## Assessment

Ready to merge. The final head implements the amended #404 decision, preserves tier and NULL semantics, and has focused unit plus production E2E coverage.

## Summary

No new findings. PR #564 unifies strict boolean writes and tolerant persisted-image reads, applies the contract across Go, PostgreSQL, DuckDB, CDC, and Parquet tiers, and resolves the bound-bool projection/scan defects exposed during review.

Issue #565, opened 2026-09-13
sqlgen: pg-main/hybrid-main bool pushdown compares the raw smallint column, not the #404 truthiness image


iceboundrock:
Follow-up scoped out of #564 (the #404 fix); the PR review (https://github.com/Lychee-Technology/forma/pull/564#issuecomment-5654742475) asked for it to be filed rather than live only in the PR body.

## What diverges

After #564 every SQL reader derives a bool from a numeric image through one spelling, `sqlgen.BoolTruthiness(expr)` = `(expr > 0.5)`, and the Go read side uses the same threshold (`transform.float64ToBool`). The one route that still compares the **raw column** is the entity_main pushdown (pg-main, and the hybrid builder's `IsMain` branch — both go through `ConvertPgMainValue` → `convertPgBoolValue` in `internal/sqlgen/dualpath_sql_helpers.go`):

| Route | `bool_smallint` predicate for `equals:true` | stored `2` matches? |
|---|---|---|
| PG-EAV (`PgEavLeafPayload.Truthy`) | `(e.value_numeric > 0.5) = $1` | yes |
| DuckDB hot pivot (`mainColBoolExpr`) | `(m.smallint_01 > 0.5)` | yes |
| DuckDB warm/cold (CDC export `castMainValue`) | exported as `CASE WHEN (m.smallint_01 > 0.5) …` | yes |
| Go OLTP read (`readWithEncoding` → `extractValueFromEAVRecord`) | `float64ToBool(2)` | reads as `true` |
| **pg-main / hybrid-main** | `m.smallint_01 = $1` with `$1 = int64(1)` | **no** |

So a row whose `bool_smallint` column holds an off-contract image (anything other than `0`/`1`) is returned as `true` by every reader, yet `equals:true` on the pg-main route does not find it, and `not_equals:true` does. `bool_text` is not affected: the pg-main bind `"1"`/`"0"` is exactly the contract every other leg already reads by (`= '1'`).

## Why this is not a bug today

#564 made the write funnel strict — the only images `boolFromAny` accepts are exactly `0` and `1`, and `storeWithEncoding` writes `0`/`1` from that. An off-contract smallint can only arrive out of band (a direct SQL write, a legacy row from before the strict funnel, a schema re-bind of a column that previously held a different value type). Under the enforced contract the raw compare and the truthiness compare are equivalent, which is why it was scoped out of #564.

[...]

PR #567, opened 2026-09-13
test(sqlgen,e2e): pin bound-bool operand spellings to one verdict per route (#503)

### LINKED ISSUE #503: bound-bool predicate operand spellings still parse differently per engine (equals:true rejected only on the PG main-column route)
Follow-up from PR #500 fourth review observation (refs #384, #404). Pre-existing on `main`; non-blocking under #384's EAV-only bool scope.

#384/#500 unified the **EAV** bool operand parse across both engines (`sqlgen.parseBoolOperand`: ParseBool spellings, else any integer at the historical `>0` truthiness) and the EAV comparison itself (`value_numeric <> 0`). The **column-bound** bool path was left as-is: `ConvertPgMainValue` accepts only integer spellings for bool-int/bool-text encodings, so `equals:true` on a bound bool attribute is rejected by the Postgres main-column payload while the DuckDB route parses it fine — the same one-route-only failure shape #384 removed for EAV bools.

## Proposal

Route the bound-bool operand parse through the same `parseBoolOperand` rule inside `ConvertPgMainValue`, then encode per binding (bool-int → int64 1/0, bool-text → "1"/"0"), so one spelling contract covers every bool attribute. This naturally belongs with #404's bool unification work (the write-side truth tables), which owns the same bound-bool surface.

Acceptance: `equals:true` / `equals:2` / `equals:0` on a bound bool answer identically (rows or 400) on the PreferHot route and the DuckDB route, with characterization pins for both encodings.

iceboundrock:
Closes #503. Tests only; no production change.

## Why

The fix #503 asks for landed with #564 (#404): `convertPgBoolValue` parses the operand through the engine-shared `parseBoolOperand` rule and encodes per binding (bool-int → `int64` 1/0, bool-text → `"1"`/`"0"`). `TestConvertPgMainValue_BoolAcceptsEveryOperandSpelling` pins that converter alone. The issue's acceptance — one spelling on a column-bound bool answering identically (rows or 400) on the PreferHot route and the DuckDB route — was not held by any test: the three-emitter characterization matrix had bound-bool rows only for `equals:1` / `equals:0`, and the e2e probed only `equals:true` / `equals:false`.

## What

- **`internal/sqlgen/bool_truthiness_test.go`** — `TestToDualClauses_BoundBoolOperandSpellingParity` drives `ToDualClauses` over `active` (bool-int) and `verified` (bool-text) with `equals:true` / `equals:2` / `equals:0`, asserting the pg-main bind, the PG-EAV truthy bind and the DuckDB `CAST(? AS BOOLEAN)` bind carry one verdict; `equals:banana` is `forma.ErrInvalidInput` on every route. Placed here because `predicate_characterization_test.go` sits at the 500-line cap.
- **`internal/e2e_harness/production/bool_bound_parity_e2e_test.go`** — `boolSpellingProbes` expands each attribute's truthy/falsy row set over `true` / `2` / `false` / `0`, so `TestBoundBoolParityNoEAVAllTiers` now pins the integer spellings beside the words at every stage; `boolRejectionProbes` (`equals:banana` on both encodings) runs through the existing `runRejectionProbes` on hot-pg, hot-duck, warm-duck and cold-duck.

## Verification

- RED proof: both tests were run once against a local revert of `convertPgBoolValue` to the pre-#564 `strconv.Atoi` parse. Unit: `active/equals:true` and `verified/equals:true` failed with `pg main generation: invalid boolean value for 'active': true: invalid input`, every other subtest passed. E2E: `hot-pg/flagInt_true` failed with `postgres read failed: build hybrid conditions: invalid boolean value for 'flagInt': true: invalid input`. That is the one-route-only shape the issue describes. The revert was discarded.
- GREEN on `main`'s code: `make fmt-check`, `make lint` (root + infra), `make test` (full unit suite), and `go test -tags=e2e ./internal/e2e_harness/production/ -run 'TestBoundBoolParityNoEAVAllTiers|TestBigIntOperandRangeParityBothDialects|TestNumericWidth'` under rootless Podman.

## Out of scope

#565 (pg-main pushdown compares the raw smallint column rather than the truthiness image) is untouched; the e2e's existing on-contract/off-contract split already keeps this test clear of it.


iceboundrock:
## Findings
No blocking or non-blocking findings. The changes are test-only and the added assertions are consistent with the shared boolean parsing/truthiness contract.

## Spec
- Issue #503 acceptance is covered for both `bool_smallint` and `bool_text` bindings.
- `true`, `2`, `false`, and `0` are asserted to produce the same verdict across the generated Postgres main-table, Postgres EAV, and DuckDB clauses.
- Invalid `banana` operands are asserted as `forma.ErrInvalidInput` at the SQL generator level and through hot-pg, hot-duck, warm-duck, and cold-duck production paths.
- No production code or out-of-scope #565 behavior was changed.

## Standards
- Changed Go files remain below the repository 500-line file limit, and the added helpers/tests remain below the 100-line function limit.
- Formatting, root lint, and infra-module lint pass.
- The test structure reuses the existing probe helpers and keeps the route expectations explicit.

## Assessment
The implementation meets the stated acceptance criteria and is ready to merge.

## Observations
The unit test checks the exact clause and argument payloads for both physical boolean encodings, while the production E2E validates actual tier routing and row-set behavior. This gives both emitter-level and end-to-end coverage without introducing production changes.

## Verification
- `make fmt-check` passed.
- `DOCKER_HOST=unix:///run/user/1000/podman/podman.sock TESTCONTAINERS_RYUK_DISABLED=true make test` passed.
- `make lint` passed for root and infra modules.
- Targeted `TestToDualClauses_BoundBoolOperandSpellingParity` and converter tests passed.
- Targeted `TestBoundBoolParityNoEAVAllTiers` passed on Podman.
- GitHub checks passed: Benchmark Smoke, E2E Smoke Tests, Lint, Test, and k6 Smoke.

## Summary
Approved from a review perspective: no findings.

PR #568, opened 2026-09-14
Carve relation roots out of the write path's input-side required check

### LINKED ISSUE #389: Write-path required-attribute check has no relation carve-out: required_always beneath a relation root 400s every write
Spun out of #318. **Pre-existing, not a #318 regression.** The shipped schemas are safe today; this is the code gap.

## What

There are two required-attribute checks on the write path, and only one of them has a relation carve-out.

- `AttributeConverter.checkRequiredAttributes` (`internal/transform/attribute_converter.go:218`) **has** the carve-out — #315 added it:

  ```go
  if relationRoots.Covers(attrName) {
      continue
  }
  ```

- `validateRequiredAttributesFromInput` (`internal/transform/transformer.go:374`, called from `ToAttributes` at `transformer.go:79`) **has none**. Its loop walks the whole attribute cache and consults `EffectiveRequiredPolicy()` directly, with no `RelationRoots.Covers` guard and no access to the roots set at all.

## The failure

For a schema declaring `required_always` on an attribute beneath a relation root:

1. `StripComputedFields` removes the whole relation subtree from the payload (#318 — root plus every dotted descendant).
2. `validateRequiredAttributesFromInput` then looks for the value the strip has just removed and answers
   `forma.InvalidInputf("missing required attribute '<name>' (attrID=%d) in EAV records")`.

That is a `400` on every create and every update, unfixable by sending the field — sending it only means the strip has more to remove.

`required_if_parent_present` does not trip this: with the parent gone, the policy reports not-missing.

## Why the #318 startup guard cannot catch it

`internal.ValidateRelationSchemas` reads only `schema["required"]` from the JSON Schema document (`internal/relation_index.go:66-72` in `loadSchemaRelations`). Attribute required policies live in the separate `<name>_attributes.json` ledger, which the relation index never opens. The guard is structurally blind to this variant.

## Why the shipped schemas are safe

The only required policy beneath a relation root in the shipped set is `contactSnapshot.isAnonymous = required_if_parent_present`, in `cmd/server/schemas/visit_attributes.json` and `visit_full_attributes.json`. Every other `contactSnapshot.*` attribute in both ledgers has no policy. Verified by walking both ledger sets. So the hazard is latent, not live.

## Documentation

Documented on the #318 branch in `docs/error-handling.md`, in the "Relation subtrees are never caller-writable" section, as a known gap that is *not* caught at startup. Closing this issue should shorten that passage.

## Acceptance

Either the write-path check gains the same `RelationRoots.Covers` carve-out (which needs the roots lookup plumbed to `validateRequiredAttributesFromInput`, currently a free function with no receiver), or the ledger-side policy is rejected at startup the way the schema-side `required` now is. A red test declaring `required_always` beneath a relation root and asserting the create succeeds.

iceboundrock:
Closes #389.

## What

`ToAttributes` runs two required-policy checks on every create and update. The record-side one (`AttributeConverter.checkRequiredAttributes`) has carved relation roots out since #315; the input-side one (`validateRequiredAttributesFromInput`) had no carve-out, so a ledger declaring `required_always` beneath a relation root answered `400 missing required attribute` on every write — unfixably, because `StripComputedFields` (#318) removes the subtree before validation and sending the value only gives the strip more to remove. The startup guard cannot see this shape: it reads the JSON Schema document, never the attribute ledger.

The input check now takes the schema's `RelationRoots`, resolved through the converter's existing lookup, and skips the names `Covers` reports — the same boundary as #315: strictly beneath a root, the root's own policy still enforced, nil lookup leaves enforcement exactly as it was.

## Why the carve-out rather than startup rejection

Design record on the issue: https://github.com/Lychee-Technology/forma/issues/389#issuecomment-5657506124. In short: it makes the two checks on one write agree, avoids teaching the relation index to open attribute ledgers, and the plumbing already exists on the converter.

## Tests

- `internal/transform/relation_root_required_input_test.go` — carve-out at the `ToAttributes` seam, plus controls (outside-root still `ErrInvalidInput`, root itself still enforced, no lookup → unchanged).
- `internal/entity_write_relation_required_policy_test.go` — the acceptance test from the issue: `manager.Create` over a copy of the shipped schemas with `visit_attributes.json` patched to `required_always` on `contactSnapshot.email` succeeds end to end; control on `propertySnapshot.code` still 400s.

Red observed before the change (`missing required attribute 'contactSnapshot.email' (attrID=35)`), green after. `make lint`, `make fmt-check` and the full suite via `scripts/test_with_container_runtime.sh` pass.

## Docs

`docs/error-handling.md`'s "known gap, not caught at startup" passage is replaced by the resolved rule; stale comments in `attribute_required.go`, `relation_roots.go` and `entity_manager.go` updated.


iceboundrock:
## Review

### Findings

1. **P2: relation roots are resolved twice within one write conversion.** [`transformer.go:81-85`](https://github.com/Lychee-Technology/forma/blob/22344ad514b2f943b534022f4cdf2098f398e796/internal/transform/transformer.go#L81-L85) resolves the lookup for the input-side check, and [`attribute_required.go:32-35`](https://github.com/Lychee-Technology/forma/blob/22344ad514b2f943b534022f4cdf2098f398e796/internal/transform/attribute_required.go#L32-L35) resolves it again from `FromEAVRecords`. `SchemaRegistry` is a public extension point and does not promise stable repeated reads; implementations may read from a database or network. If `GetSchemaByID` changes schema name between calls, the two checks can disagree, and if the second read fails a payload that passed the first check becomes an operator-facing error. Please resolve the schema name/relation roots once and pass that snapshot through both checks (or provide an equivalent single-resolution path), and add a regression test for a registry that changes or fails on the second lookup.

2. **P2: the change exceeds the repository source-file limit.** [`internal/transform/transformer.go`](https://github.com/Lychee-Technology/forma/blob/22344ad514b2f943b534022f4cdf2098f398e796/internal/transform/transformer.go) is now 503 lines, up from 485 on `main`, while [`coding-standard.md#L5-L10`](https://github.com/Lychee-Technology/forma/blob/22344ad514b2f943b534022f4cdf2098f398e796/coding-standard.md#L5-L10) requires source files to stay at or below 500 lines. Please move the new required-policy helper or related code into an appropriate file before merge.

### Spec

- The intended #389 behavior is correct: required policies strictly beneath a relation root are inert because the subtree is stripped before validation.
- The relation root itself remains enforced, same-prefix siblings remain enforced, and a nil relation-root lookup preserves the previous behavior.
- The manager wiring reaches both the JSON transformer and the persistent-record converter.
- No additional functional spec violation was found beyond the duplicate-resolution issue above.

### Standards

- Error context on the new registry-resolution failure is present, and the direct return of the caller-input validation carrier is consistent with the repository error contract.
- New files and changed helper functions are within the function/file limits except for `transformer.go`, which is 3 lines over the explicit file limit.
- `gofmt` and lint checks are clean.

### Assessment

**Changes requested.** The core carve-out is narrowly scoped and the existing behavior is well documented, but the per-write double lookup can violate the registry consistency model and the changed source file violates a mandatory repository standard.

### Observations

- The added transform-level tests cover the carve-out, outside-root control, root-itself control, same-prefix behavior through existing tests, and no-lookup behavior.
- The end-to-end regression covers `Create`; an `Update` case would improve confidence because the same `ToAttributes` path is used after update merge.
- The proposed single-resolution fix should be tested through a persistent-record conversion as well as the direct transform seam.

### Verification

Passed:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform ./internal -run 'Test(ToAttributes|FromEAVRecords|CreateSucceedsWithRequiredAlwaysBeneathRelationRoot|CreateStillRejectsMissingRequiredAlwaysOutsideRelationRoot|NewEntityManager)' -count=1`
- `git diff --check main...HEAD`
- `make fmt-check`
- `make lint` (root and `infra/`)
- PR checks: Benchmark Smoke, E2E Smoke Tests, k6 Smoke, Lint, and Test all passed.

The local full `go test ./... -count=1` reached all packages but could not complete because `internal/e2e_harness/TestE2EHarnessMinimal` requires a container runtime and the environment has no usable `$XDG_RUNTIME_DIR`/Docker-compatible endpoint. The relevant remote E2E smoke check passed.

### Summary

The fix addresses the reported required-policy/relation-root mismatch with good targeted coverage and documentation. Please address the one-write relation-root snapshot consistency issue and split `transformer.go` below 500 lines; after that, the change should be in good shape for approval.

iceboundrock:
Addressed in c45fef4, responding to https://github.com/Lychee-Technology/forma/pull/568#issuecomment-5657592251 point by point.

### Finding 1 — relation roots resolved twice per write

Fixed. `ToAttributes` resolves the roots once and hands the set to a new unexported `AttributeConverter.fromEAVRecords(records, relationRoots)`; `checkRequiredAttributes` now takes the resolved set instead of looking it up. The exported `FromEAVRecords` keeps resolving on its own and delegates, because on the read path (`FromPersistentRecord`) it is the only resolution on the call and nothing upstream has a set to hand it.

Two regression tests, both red on the previous commit and green now:

- `TestToAttributesResolvesRelationRootsOnce` — the registry fails on its second `GetSchemaByID`; the write must still succeed and the call count must be 1.
- `TestToPersistentRecordResolvesRelationRootsOnce` — through the persistent-record conversion the manager actually calls, as you asked; the second read answers a different schema name, which would make the lookup return no roots and the record-side check enforce `contactSnapshot.tenantId` that the input-side check had just carved out.

One scoping note, stated rather than left implicit. The consistency-model argument applies more widely than the relation roots: one `ToPersistentRecord` already reads the attribute metadata from the registry three times (`ToAttributes`, `fromEAVRecords`, `ToPersistentRecord` itself), and that predates this PR. This PR makes the read it introduced single; making the metadata read single is a signature change along the whole conversion chain, on both write and read paths, and I have filed it as #569 rather than widen this change.

### Finding 2 — `transformer.go` over 500 lines

Fixed. `validateRequiredAttributesFromInput` and its three helpers moved to `internal/transform/input_required.go` (136 lines), next to `attribute_required.go`, which holds the record-side twin. `transformer.go` is 384 lines. The cross-references in `attribute_required.go`, `relation_roots.go`, `docs/error-handling.md` and the trailing pointer comment in `transformer.go` now name the new file.

### Observations

- Update case: added `TestUpdateSucceedsWithRequiredAlwaysBeneathRelationRoot` on the same patched-ledger harness. It seeds a row, updates `feedback` while also sending `contactSnapshot.email`, and asserts the merged document passes, the subtree still never reaches storage, and `feedback` lands.
- Persistent-record conversion coverage for the single-resolution fix: the second test above.

### Verification

- `make fmt-check`, `make lint` (root and `infra/`) clean.
- `./scripts/test_with_container_runtime.sh` — full suite green, including `internal/e2e_harness`.
- Largest file in `internal/transform` after the change: `persistent_record.go` at 492 lines (unchanged).

iceboundrock:
## Review

### Findings
No findings.

### Spec
The implementation satisfies #389 and its recorded design decision:
- The input-side required-policy check now skips only attributes strictly beneath a relation root, using the same `RelationRoots.Covers` boundary as the record-side check.
- A policy on the relation root itself remains enforced; policies outside relation roots remain enforced; an absent/empty lookup preserves prior enforcement.
- The create and update acceptance paths no longer reject a `required_always` ledger entry beneath `contactSnapshot`, while relation-subtree data remains stripped and unstored.
- The follow-up correctly resolves relation roots once per `ToAttributes` conversion and shares that result with both required checks, avoiding inconsistent repeated registry reads.

### Standards
- Changed production files remain within the repository 500-line file and 100-line function limits.
- New error propagation adds operation context and preserves the write-path `forma.InvalidInputf` carrier.
- The extraction to `input_required.go` keeps the write-only validation concern cohesive and avoids duplication.
- Exported and non-obvious behavior is documented; `docs/error-handling.md` was updated to remove the now-resolved gap.

### Assessment
Approved from a code-review perspective. The change is narrowly scoped, aligns the two required-policy checks on the write path, preserves the deliberately strict root boundary, and includes both seam-level and manager-level regression coverage.

### Observations
- The direct `FromEAVRecords` API retains its own relation-root resolution for read paths, where it is the sole required-policy check.
- Tests intentionally model an unstable `SchemaRegistry`; this pins the single-resolution contract introduced by the follow-up commit.
- No production schema ledger is changed. The acceptance tests patch a copied shipped ledger, which is appropriate because the reported configuration hazard is latent in shipped schemas.

### Verification
Passed:
- `make fmt-check`
- `make lint`
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform ./internal -count=1`
- Focused transform and manager acceptance/control tests for #389

Not run locally: container-runtime E2E suite (`scripts/test_with_container_runtime.sh`).

### Summary
No findings. The PR correctly closes #389 with the chosen write-path carve-out and documents the resulting inert-policy rule.

PR #570, opened 2026-09-14
fix(schemavalidate): classify out-of-range numeric literals as invalid input (#402)

### LINKED ISSUE #402: Numeric literal outside int64 and float64 range (e.g. 1e400) answers a redacted 500 for caller input

iceboundrock:
Closes #402.

## Part 1 — out-of-range numeric literal is caller input

A caller-supplied literal that fits neither `int64` nor `float64` — `{"score": 1e400}` — survives httpapi's `UseNumber` decode and `json.Marshal`'s verbatim re-emit, then failed inside `exactNumberInstance` as a plain error and answered a redacted 500 in both strict and report-only modes.

- `exactNumberInstance` now threads the dotted attribute path (`o.xs[1]`, the same spelling the #322 walker uses) down its recursion and fails with a `forma.InvalidInputf` carrier: `attribute "score" holds numeric literal "1e400" outside the representable range; a number within float64 range is required`. The `strconv` range text rides along as operator detail (the #453 toolchain-drift ruling), so the log keeps it and the body never sees it. A root-level literal gets the message without the attribute clause, mirroring the walker's root rule.
- `Validate` wraps that failure under its own plain text (`failed to rewrite numeric literals for schema %d`) so the genuinely internal redecode branch keeps `failed to decode payload` to itself — the split the issue asked for rather than a blanket classification of the shared wrap.
- Report-only mode absorbs the new carrier like any violation; transform's `json.Number.Float64` independently rejects the literal on numeric and bool attributes, which `TestReportOnlyUpdateStillRejectsOutOfRangeLiteral` pins alongside the #322 precedent.

## Part 2 — the deferred marshal-fallback question

`marshalRefusalError`'s walk-empty branch published every `json.Marshal` failure verbatim. It now classifies by the concrete error:

- `*json.UnsupportedValueError` (a non-finite in an un-walked shape, or a cycle): the library's text still publishes — prose about the value, and for a cycle the only truthful description.
- `*json.UnsupportedTypeError` / `*json.MarshalerError`: an owned message naming the Go type (`Go type chan int is not encodable` / `the MarshalJSON method of Go type X refused the value`); the library's text — for a `MarshalerError`, the embedder's own error prose — stays behind `forma.WithOperatorDetail`.

All branches are HTTP-unreachable (a decoded body holds only maps, slices, strings, bools, nils and grammar-valid `json.Number`s), but they still cross the public transport and are held to the same width rule as #453.

`docs/error-handling.md` records both rulings and drops its "known gap tracked in #402" passages.

## Tests

- `number_instance_test.go`: top-level, negative, nested-object, array-index and root literals → sentinel, published attribute path + literal, `strconv` text operator-only; rewrite wrap distinct from decode wrap.
- `entity_write_nonfinite_test.go`: strict create → carrier; report-only update → transform still rejects, stored value intact. Both verified failing without the validator change.
- `nonfinite_paths_test.go`: owned `UnsupportedTypeError` message; `MarshalerError` prose kept out of the body; cycle text still publishes.

`make fmt-check`, `make lint`, and `./scripts/test_with_container_runtime.sh` are green.


iceboundrock:
## Review — #402 out-of-range literal + marshal-fallback width (#570)

### Spec (vs #402)
- Part 1 matches the issue direction and the posted plan exactly: `exactNumberInstance`'s range failure is now a `forma.InvalidInputf` carrier naming attribute + literal (`internal/schemavalidate/number_instance.go:84-92`), the shared `"failed to decode payload"` wrap is split (`internal/schemavalidate/validator.go:296-306` keeps redecode plain, rewrite gets `"failed to rewrite numeric literals for schema %d"`), and report-only absorbability is made safe by the transform backstop.
- Part 2 matches the deferred ruling posted on #402: walk-empty fallback publishes library text only for `*json.UnsupportedValueError`; `*json.UnsupportedTypeError` / `*json.MarshalerError` publish an owned `Go type …` message with library/embedder prose behind `WithOperatorDetail` (`internal/schemavalidate/nonfinite_paths.go:90-103`). `docs/error-handling.md` drops both `#402 gap` passages and records the split.
- No scope creep beyond the two committed items.

### Standards
- File/function sizes: `number_instance.go` 93 lines, `nonfinite_paths.go` 266, `validator.go` 318; new helpers (`rewriteNumbers`, `outOfRangeLiteralError`, `unexplainedRefusalError`) each <35 lines. Under the 500/100 gates.
- Error semantics (`docs/error-handling.md`, #313/#453 rules): both new carriers use `InvalidInputf` + `WithOperatorDetail`, never a bare `%w` of a sentinel — `TestNoBareSentinelWraps` passes. Plain `fmt.Errorf` wraps are used only for operator context (`validator.go:306`), so prefixes stay out of the published body. Stdlib/embedder prose (`strconv` range text, `unsupported type`, `MarshalJSON` error) is asserted operator-only in tests.
- Structure: early returns, flat switch + loop, max nesting 3 (`rewriteNumbers` map/for/if). Path spelling (`o.xs[1]`, `joinAttributePath`) matches `marshalRefusalWalker`, so the two published families read alike. Comments cite the issue/toolchain rationale at each decision point.
- No >15-line duplication introduced; the recursion mirrors the walker's shape but serves a different pass (rewrite vs. refusal-explain).

### Assessment
No blockers. Correct classification on both halves, correct report-only interaction, honest docs, pinned tests. Two non-blocking nits below.

### Findings (all non-blocking)
1. `unexplainedRefusalError`'s final fallback (`nonfinite_paths.go:102`: `InvalidInputf("…: %v", marshalErr)`) publishes library text without operator detail. Unreachable today ("nothing else" per the comment) and correct for `UnsupportedValueError`/cycle, but a future `encoding/json` error kind with toolchain-dependent or embedder-authored wording would publish verbatim by default. Consider an owned message + `WithOperatorDetail` as the default, keeping `%v` publish only for `UnsupportedValueError`.
2. `TestValidateRewriteWrapIsDistinctFromDecodeWrap` (`number_instance_test.go:139-149`) pins the split negatively (`NotContains "failed to decode payload"`) but never asserts the positive wrap (`"failed to rewrite numeric literals"`). A positive `Contains` would lock the operator-distinguishable text the comment promises.

### Observations
- Root-array edge is consistent: `rewriteNumbers` renders a root `[]any` index as `"[0]"` (`number_instance.go:64`), same as the refusal walk, so a root-array `1e400` publishes `attribute "[0]"` rather than the root-literal form. Odd-looking but coherent across both families; no action.
- Report-only safety for `1e400` is pinned at manager level for the numeric funnel (`entity_write_nonfinite_test.go:167-183`); the bool funnel's `1e400` rejection is already pinned at unit level (`bool_truth_table_test.go:92,171-172`). Manager-level bool coverage would be redundant — noting only so a reader doesn't misread the single-funnel test as a gap.
- Docs simplification ("`json.Number.Float64` fails on an out-of-range literal") covers the numeric funnel; the bool funnel rejects on literal text (`boolFromNumberLiteral`), same verdict, different mechanism. Fine as prose.

### Verification
- Local (this branch, `GOTOOLCHAIN=go1.26.0+auto`): `internal/schemavalidate` targeted tests pass (`TestValidateClassifiesOutOfRangeLiteralAsInvalidInput` 4/4 subtests, `TestExactNumberInstanceRootLiteralHasNoAttribute`, `TestValidateRewriteWrapIsDistinctFromDecodeWrap`, `TestValidateOwnsUnsupportedTypeMessage`, `TestValidateKeepsMarshalerErrorOperatorOnly`, `TestValidateFallsBackOnCyclicPayload/BeyondDepthCap`); full `internal/schemavalidate` package green; `internal` report-only/out-of-range manager tests green with stored value intact; root `TestNoBareSentinelWraps` pass; `make fmt-check` clean.
- CI on the PR: `Lint` pass, `Test` pass; `Benchmark Smoke` / `E2E Smoke` / `k6 Smoke` pending at review time (validation-only change; no e2e-sensitive path touched beyond error classification).
- Did not re-run the full container suite here; PR body reports `./scripts/test_with_container_runtime.sh` green and CI `Test` corroborates.

### Summary
LGTM. Ships exactly what #402 asked for on both halves, with the publish-width discipline (#313/#453) applied consistently and the report-only backstop pinned. Nits above are optional follow-ups, not merge conditions.

iceboundrock:
## Response to review (e741f5a)

Both findings are addressed in this PR rather than deferred: each is a few lines inside code the PR already touches, and finding 1 turned out to be a code/doc mismatch, not only future-proofing.

### Finding 1 — fallback default publishes library text: fixed

Verified against the pinned toolchain before changing anything. Under the default build `json.Marshal` returns exactly the three kinds the switch names, so the `%v` default was unreachable — but `encoding/json`'s `GOEXPERIMENT=jsonv2` shim (`v2_inject.go:transformMarshalError`) already falls through to `transformSyntacticError(err)`, i.e. a `*json.SyntaxError` for a malformed `json.RawMessage`. That is a fourth kind one build flag away, with toolchain-authored wording. Separately, `docs/error-handling.md` already said the library text publishes *only* for `*json.UnsupportedValueError`; the code didn't enforce that.

Change (`internal/schemavalidate/nonfinite_paths.go`): `*json.UnsupportedValueError` is now matched explicitly and keeps publishing `%v` (cycle / beyond-cap non-finite; both existing tests still pass unchanged). Everything else takes `payload cannot be encoded as JSON; no offending attribute could be identified` with the library's text behind `forma.WithOperatorDetail`. The message names nothing on purpose: the walk found no attribute, and inventing one would be false. New `TestUnexplainedRefusalClosesByDefault` pins both branches through the package-private helper with a synthetic error kind (the only way to reach the default under the default toolchain). Docs sentence added recording the default.

### Finding 2 — positive wrap assertion: fixed

`TestValidateRewriteWrapIsDistinctFromDecodeWrap` now asserts `Contains "failed to rewrite numeric literals for schema 3"` alongside the existing `NotContains`. Also fixed the test's doc comment, which named a non-existent `TestValidateRedecodeFailureStaysPlain`.

### Observations — no action, agreed

- Root-array `"[0]"` spelling: kept. A root-level array body is not a shape httpapi produces (entity payloads are objects), and the two published families should keep one path grammar; diverging for an unreachable shape would be a special case with no caller.
- Bool funnel coverage: agreed that `bool_truth_table_test.go` already pins `1e400` at the unit level and a manager-level duplicate adds nothing.
- Docs prose on `json.Number.Float64`: agreed; the bool funnel's mechanism (`boolFromNumberLiteral` on literal text) is recorded next to that code, and the doc sentence describes the outcome, which is the same.

### Verification

`internal/schemavalidate` package green; root `TestNoBareSentinelWraps` green; `internal` out-of-range / non-finite manager tests green; `make fmt-check` and `make lint` clean. Trailer-free commit history confirmed.

No follow-up issue opened: nothing from the review is left outstanding.

iceboundrock:
## Review

### Findings
1. **P2 - Make the selected out-of-range literal deterministic.** [`internal/schemavalidate/number_instance.go:54-60`](https://github.com/Lychee-Technology/forma/blob/3a6b354cbc1105b37c34c8eaa2939824f9d40955/internal/schemavalidate/number_instance.go#L54-L60) iterates `map[string]any` directly and returns as soon as it finds an invalid `json.Number`. For a payload such as `{"z": 1e400, "a": 1e400}`, Go map iteration makes the published 400 body and the report-only warning name either `z` or `a` across identical requests. This newly public error surface should be stable; the adjacent marshal-refusal walker deliberately sorts map keys for exactly that reason. Collect and sort keys before recursing (and add a two-literal test that pins the selected path).

### Spec
- Part 1 matches #402: a `json.Number` outside both `int64` and `float64` now remains caller input, carries `forma.ErrInvalidInput`, publishes an owned path/literal message, and preserves the `strconv` detail for operators only. The redecode failure remains a plain internal error under its distinct wrap.
- Part 2 matches the issue ruling: only `*json.UnsupportedValueError` publishes `encoding/json` prose; unsupported types, failing marshalers, and unknown future kinds use an owned message and retain underlying detail only in the operator chain.
- Report-only behavior remains safe: numeric and boolean transform funnels reject `1e400`, while other writable value types cannot create an unreadable numeric image.

### Standards
- Error semantics follow `docs/error-handling.md`: new 4xx paths use `forma.InvalidInputf`, operator-only details use `forma.WithOperatorDetail`, and no bare sentinel wrap is introduced.
- Modified source files and helpers are below the 500-line / 100-line limits. The implementation is flat and contextual error wrapping is present.
- `docs/error-handling.md` accurately removes the former #402 gap and records the closed-default marshal-fallback rule.

### Assessment
**Changes requested** for the deterministic public-error finding above. I found no other correctness, disclosure, scope, or standards issues.

### Observations
- The prior two review nits are resolved in the current head: the unclassified marshal fallback closes by default, and the rewrite-wrap test now asserts the positive text.
- Root-array paths use `[0]`, consistent with the existing marshal-refusal walker; no action needed.

### Verification
- Passed: `make fmt-check`
- Passed: `make lint`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/schemavalidate ./internal -count=1`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/schemavalidate ./internal -count=1`
- Passed compile sweep: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./... -run "^$"`
- GitHub CI is green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Summary
The PR correctly closes #402 and hardens marshal-error disclosure. Sort the number-rewrite map traversal so identical invalid payloads produce one stable client and operator-facing error.

iceboundrock:
## Response to review (comment 5668301958)

### Finding 1 — deterministic out-of-range literal selection: fixed in e56799a

Verified: `rewriteNumbers` ranged the map directly and returned on the first failure, so `{"z": 1e400, "a": 1e400}` named `z` or `a` at random. A looped test (`TestValidateOutOfRangeLiteralSelectionIsDeterministic`, 64 iterations per case) reproduced it on run 1–3 before the change.

**Approaches weighed** (mid-execution ruling; recorded here rather than only in the commit):

| | Approach | Success-path cost | Verdict |
|---|---|---|---|
| A1 | Sort keys before recursing via `slices.Sorted(maps.Keys(t))` (the `walkField` idiom) | Measured **+14% allocs / +24% bytes** per `Validate` on a 21-object doc — the iterator form grows the slice through several allocations per object | Rejected: this walker runs on every successful write, unlike `marshalRefusalWalker`'s |
| A2 | Sort keys before recursing via pre-sized `make([]string, 0, len(t))` + `slices.Sort` (the `marshalRefusalWalker` idiom) | Measured **+1.7% allocs / +7% bytes** (1206 → 1227 allocs/op, 51.0 → 54.6 KB/op); ns/op indistinguishable from baseline noise | **Taken** |
| B | Iterate unsorted; on failure keep walking siblings and return the smallest key's error | Zero on the success path | Rejected: buys ~1% over A2 at the cost of an induction argument in a comment ("min key at each level ≡ sorted traversal") that the next reader has to re-derive; the package already has two sorted-traversal precedents and no min-key one |

Benchmark: temporary `Validate` benchmark on a `{"f0".."f19": {"k0".."k9": n}}` document, `-benchmem -count=5`, not committed. The doc comment on `rewriteNumbers` records the A1-vs-A2 choice so nobody "simplifies" it back to `slices.Sorted(maps.Keys(...))`.

**Test pins two cases**, both looped so a regression fails with probability ≈ 1 − 2⁻⁶⁴ rather than flaking:
- sibling literals `{"z": 1e400, "a": 1e400}` → `attribute "a"`
- a smaller key's nested subtree beats a larger key's direct literal: `{"z": 1e400, "m": {"deep": 1e400}, "n": [1e400]}` → `attribute "m.deep"`

### Spec / Standards / Observations sections

No action: these confirm the PR's rulings and the resolution of the two prior nits. Root-array `[0]` spelling stays as is.

### Follow-up issues

None needed; the finding is closed by this commit and nothing was deferred.

### Verification
- Passed: `make fmt-check`
- Passed: `make lint`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/schemavalidate ./internal -count=1` (and `-race` on `./internal/schemavalidate`)
- Passed: the new test at `-count=20` (60/60 subtests)

iceboundrock:
## Review

### Findings

No blocking, non-blocking, or test-coverage findings.

### Spec

- The implementation satisfies #402: an HTTP-originated `json.Number` outside `int64` and `float64` range now carries `forma.ErrInvalidInput`, publishes an owned actionable 4xx message, and preserves `strconv` detail only for operators.
- The validation rewrite and JSON-decode failures remain distinct, preserving the required caller-input versus internal-fault classification.
- The marshal-refusal fallback now publishes library text only for `*json.UnsupportedValueError`; unsupported Go types, failing custom marshalers, and unclassified future kinds use owned public messages with raw detail withheld. This matches the error-handling ruling and prevents toolchain or embedder prose from widening the public API.
- Report-only behavior remains safe for numeric and bool storage: the validator carrier may be reported and accepted, but the transform conversion rejects the unrepresentable literal before persistence.

### Standards

- Error chains use contextual wrapping and retain sentinel classification through `errors.Is`. Public/error-operator separation uses `forma.InvalidInputf` and `forma.WithOperatorDetail` correctly.
- New helpers are cohesive, use deterministic sorted map traversal, and are well below the 100-line function and 500-line source-file limits.
- Documentation in `docs/error-handling.md` reflects the implemented classification and disclosure contract.

### Assessment

The change is internally consistent, deterministic for multiple invalid literals, and appropriately scoped. No change request.

### Observations

- New tests cover positive and negative overflow, nested object and array paths, root behavior, deterministic offender selection, owned unsupported-type output, custom marshaler disclosure, default-closed fallback behavior, strict create, and report-only update protection.
- GitHub Actions run `34879596181` is fully green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Verification

- Passed: `git diff --check origin/main...HEAD`
- Passed: `make fmt-check`
- Passed: focused `go test` coverage for all changed schema-validation and manager-level scenarios.
- Local `make test` passed all regular packages and affected packages, but did not complete `internal/e2e_harness`: testcontainers panicked during Docker host discovery because `$XDG_RUNTIME_DIR` is unavailable in this review environment. This is environmental, not a PR failure; the PR E2E Smoke CI check passed.

### Summary

Ready to merge from this review.

PR #571, opened 2026-09-14
fix(transform): cap payload nesting depth so cyclic embedder payloads are rejected (#406)

### LINKED ISSUE #406: Cyclic embedder payloads crash the write path at normalization, before any guard can reject them
Found while resolving PR #403's round-2 review (issue #322). Pre-existing — not introduced or fixed by that PR.

## The gap

A Go embedder handing `EntityManager.Create`/`Update` a **cyclic** payload (`m := map[string]any{}; m["self"] = m`) crashes the process with unbounded recursion before any validation runs:

- `validateWritePayload` calls `transform.NormalizeDottedKeys` → `normalizeMap` → `normalizeValue`, which recurse through nested `map[string]any`/`[]any` with no cycle or depth protection (`internal/transform/normalize_keys.go`).
- Even with validation unconfigured, the write path's own flattening (`flattenToAttributes`, `internal/transform/transformer.go`) recurses over the same shapes.

By contrast the two layers downstream are safe: `json.Marshal` detects cycles and returns `json: unsupported value: encountered a cycle …`, and (since PR #403 round 2) the validator's non-finite path walker aborts at a depth cap and falls back to that text. The normalization/flattening layers sit **in front** of both, so today a cyclic payload never reaches either — it dies first, as fatal stack exhaustion (not a recoverable panic).

## Reachability

Go-embedder-only: HTTP JSON cannot express a cycle, and `encoding/json` bounds decode nesting, so request-body documents are always finite and acyclic. This is the same reachability class as #322's `math.NaN()` — an embedder handing in a Go value no JSON body could produce.

## Direction

A shared depth cap for the transform-layer recursions (normalize + flatten), aborting into a `forma.InvalidInputf` carrier ("payload nesting exceeds N levels / payload is cyclic") rather than plain, since a cyclic or absurdly deep document is caller input. PR #403's `maxNonFiniteWalkDepth = 1000` in `internal/schemavalidate/nonfinite_paths.go` is precedent for the cap value and the abort shape.

iceboundrock:
Closes #406.

## Problem

A Go embedder handing `EntityManager.Create`/`Update` a cyclic `map[string]any` or `[]any` (`m["self"] = m`) crashed the process with fatal stack exhaustion. The first recursions over caller-owned nesting are `transform.NormalizeDottedKeys` (validator configured) and `flattenToAttributes` (validator nil), and both sit in front of the two layers that could have refused the cycle (`json.Marshal`, and the validator's own capped walk from #403). HTTP JSON cannot express a cycle and `encoding/json` bounds decode nesting, so this is embedder-only, the same reachability class as #322.

## Change

- **`internal/transform/payload_depth.go`** (new): `maxPayloadNestingDepth = 1000`, a `payloadPosition{root, depth}` locator, `flattenPosition(path, indices)` deriving it from what the flattener already carries, and `checkPayloadDepth`, which past the cap builds a published `forma.InvalidInputf("payload nesting exceeds 1000 levels beneath attribute %q; the payload is cyclic or too deeply nested")`. Only the top-level key is named: at depth 1000 the full dotted path is a multi-kilobyte string.
- **`normalize_keys.go`**: the three walkers carry the position and return an error; `NormalizeDottedKeys` becomes `(map[string]any, error)`. Array elements count as an edge, so a cycle through `[]any`, where the dotted name never grows, is bounded too. The merge helpers need no cap: they recurse only over output the walk has already rebuilt.
- **`transformer.go`**: `flattenToAttributes` checks the cap at its map and slice cases. Its signature is unchanged; depth is `len(path)+len(indices)`.
- **`entity_write_validation.go`**: the normalize error is returned before `Validate` runs and regardless of `enforce`. Report-only mode forgives violations in stored data that an operator repairs later; a cyclic document is not one, and the write could never succeed anyway because the flattener refuses it with the same cap.
- **`docs/error-handling.md`**: new bullet under "Which errors are which class", and a note at the marshal-refusal passage that the validator's cycle fallback is now defence in depth.

The cap value matches `schemavalidate.maxMarshalRefusalWalkDepth` on purpose but stays a separate constant: that one degrades an error message on a payload already refused elsewhere, this one is what refuses the payload.

## Tests

- `payload_depth_test.go`: the exact 1000/1001 edge, cyclic map, and cyclic slice, each through `NormalizeDottedKeys` and through `ToAttributes` with no validator in front. The flatten edge test registers the 1001-segment attribute name so the only possible objection is the cap.
- `TestReportOnlyUpdateRejectsCyclicPayload`: an update with strict validation off is still rejected with the published carrier and nothing reaches storage.
- Existing normalize tests moved to a `mustNormalize` helper; behaviour unchanged. `TestNoBareSentinelWraps` still passes.

## Verification

`make fmt-check`, `make lint`, and the full unit suite via `scripts/test_with_container_runtime.sh` are green.

Design draft and implementation note are on #406.


iceboundrock:
## Review — fix(transform): cap payload nesting depth (#406 / PR #571)

### Spec (against #406 + design draft + implementation note)
- Issue asks for a shared depth cap over both transform recursions (`NormalizeDottedKeys` and `flattenToAttributes`), aborting into `forma.InvalidInputf`, returned regardless of `enforce`. All three hold here.
- Cap value 1000 matches `schemavalidate.maxMarshalRefusalWalkDepth` as precedent but stays a separate constant with the semantic difference documented in `payload_depth.go` ("degrades a message" vs "refuses the payload"). Matches the draft's explicit direction.
- Two documented deviations from the draft are both mechanical and justified:
  1. `payloadPosition{root, depth}` instead of bare `depth int` — deriving the top-level key from the dotted `name` would misreport a literal `"a.b"` key as `"a"`; threading `pos.into(key)` reports the key exactly as written. `flattenPosition(path, indices)` derives the same struct without changing `flattenToAttributes`'s signature, as planned.
  2. Cap checked at the map and slice cases in `flattenToAttributes`, not once before the `switch` — a single top check would also fire for scalar leaves one edge below their container and diverge from normalize at the exact 1000 edge. Both walks pin 1000-accept / 1001-reject, so edge parity is tested, not asserted.
- `entity_write_validation.go` returns the normalize error before `Validate` and regardless of `enforce`, with the doc comment updated to explain why (a cyclic document is not a repairable stored violation, and the flattener refuses it anyway). Correct placement.
- `docs/error-handling.md` records the cap under "Which errors are which class" and notes the validator's cycle fallback is now defence-in-depth. Both places that describe the marshal-refusal path are updated consistently.
- Completeness of the "first recursion" claim verified: `StripComputedFields` is a shallow top-level copy; `mergeMaps`/`copyMapDeep` recurse only while the finite stored side also has a map at the same key (depth bounded by stored data, update values otherwise assigned by reference); `validateRequiredAttributesFromInput` walks at most `len(schema-attr-segments)` steps over caller data (schema-bounded, and array expansion terminates on non-map contexts); `ToAttributes`'s `json.Marshal` default branch is cycle-safe. So normalize (validator configured) / flatten (validator nil, after the bounded required check) are indeed the first unbounded recursions, and both are now capped.

### Standards (`coding-standard.md`, `AGENTS.md`, `docs/error-handling.md`)
- File sizes: `payload_depth.go` 70, `normalize_keys.go` 359, `transformer.go` 390, `entity_write_validation.go` 264, `payload_depth_test.go` 127 — all ≤ 500.
- Function size: `flattenToAttributes` is exactly 98 lines — at the 100-line cap but not over. The implementation note already flags the next change there should extract the empty-list marker branch. No new violation.
- Error semantics: cap built via `forma.InvalidInputf` (published carrier naming logical state, offending top-level attribute, and expected state). `validateWritePayload` wraps it as `fmt.Errorf("failed to normalize payload for schema %d: %w", …)` — context-adding `%w`, not a bare sentinel wrap, so `errors.Is(ErrInvalidInput)` and `ResolvePublicMessage` (preorder unwrap to the inner carrier; outer prefix correctly stays out of the body) both hold. `flattenToAttributes` returns the carrier directly, consistent with its neighboring `InvalidInputf` returns. `TestNoBareSentinelWraps` passes (verified locally).
- Nesting/early-return style: `normalizeMap`/`normalizeSlice`/flatten additions are flat guard-clauses; no function exceeds three nesting levels; no unrelated refactoring.
- `gofmt` clean on all touched Go files (verified locally).

### Findings
No blocking findings. Two non-blocking notes (already deliberate in the PR, recorded here for the decision log):
1. The cap also rejects acyclic documents deeper than 1000, including HTTP JSON in the 1001–10000 band the decoder would otherwise accept. Documented as "no shipped schema comes near" — acceptable tradeoff of availability for stack safety, and consistent with the schemavalidate precedent.
2. Cycle handling is depth-bound, not visited-set-based: a cyclic payload burns ~1000 frames before aborting (bounded CPU, no fatal exhaustion) and a 1001-deep acyclic payload is indistinguishable from a cycle. Simpler and allocation-free; the message wording ("cyclic or too deeply nested") is the truthful statement of that.

### Observations (correctness details checked)
- Depth accounting is consistent across walks: map member and array element each count one edge (`pos.into(key)` / `pos.into("")` vs `len(path)+len(indices)`), so the `[]any` cycle where the dotted name never grows is bounded — pinned by `TestNormalizeRejectsCyclicSlice` / `TestFlattenRejectsCyclicSlice`.
- Root naming is consistent: normalize inherits the depth-0 key verbatim (dotted literals preserved); flatten uses `path[0]` as written. At depth 1000 only the root key is named, avoiding a multi-KB path — the right call.
- Off-by-one verified by reading `deepPayload` against the tests: root depth 0, innermost container at depth `levels`; 1000 accepted, 1001 refused on both walks. `deepAttributeRegistry(levels)` registers the `levels+1`-segment `"a…"` name so the accept-edge's only possible objection is the cap.
- Merge-helpers-need-no-cap argument holds: both operate on already-rebuilt output (new maps/slices, scalars shared), which is finite once the walk aborted on infinite input.
- Report-only boundary pinned both ways: violations still absorbed, depth-cap carrier still rejected, and the rejected update provably never reaches storage (`FromPersistentRecord` still reads `{"name":"open"}`).
- All `NormalizeDottedKeys` call sites migrated (`entity_write_validation.go` prod path + `entity_write_relation_diagnosis_test.go` + `mustNormalize` helper for the 29 fixture calls); `grep` shows no stale single-return callers.

### Verification
- CI on this PR is green: Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke (all `SUCCESS`).
- Local (branch `fix/406-payload-depth-cap`, `GOTOOLCHAIN=go1.26.0+auto` per Makefile): `go test ./internal/transform -run 'TestNormalizeRejects|TestFlattenRejects|TestNormalize'` PASS; `go test ./internal -run 'TestReportOnlyUpdateRejectsCyclicPayload|TestWriteValidationAttachesNoDiagnosis'` PASS; `go test . -run TestNoBareSentinelWraps` PASS; `go test ./internal/transform` PASS; `go test ./internal` PASS; `gofmt -l` on all touched files clean.

### Assessment / summary
Faithful execution of the #406 design with two well-reasoned mechanical deviations, both documented on the issue and in code. Error class, enforcement boundary, docs, and tests all line up; no standards violations; verification green locally and on CI. No changes requested — approve-worthy as is. The only follow-up worth tracking is the already-noted one: `flattenToAttributes` sits at 98/100 lines, so its next unrelated change should do the flagged extraction.

iceboundrock:
**Response to the review above** (assessment of the two non-blocking findings and the follow-up; no code change on this PR).

### Finding 1 — the cap also refuses acyclic documents deeper than 1000

Compared two options: keep 1000, or raise to 10000 so the cap matches `encoding/json`'s decode limit and can never fire on an HTTP payload.

Keeping 1000 is the right call, and for a stronger reason than "no shipped schema comes near". For an HTTP payload the cap cannot change the **verdict**, only the **message**: every scalar leaf at depth > 1000 needs a registered attribute whose dotted name has > 1000 segments, and without one `flattenToAttributes`'s default branch already refuses it as `attribute '…' is not defined for this schema` (also `InvalidInputf`). The only documents whose verdict flips are those whose content past depth 1000 consists solely of `null` values or empty lists under unregistered names — previously dropped silently by the `continue` / `return nil` skips, now refused. That is a negligible band and arguably an improvement. Raising to 10000 would multiply the frames burned on a cyclic payload by ten and buy nothing on the verdict side.

### Finding 2 — depth-bound instead of visited-set cycle detection

Compared the depth cap against a pointer-identity cycle detector. Three points decide it:

1. **A cycle detector cannot replace the cap, only supplement it.** An acyclic embedder payload nested 10^6 deep exhausts the stack exactly as fatally as a cycle. So the depth bound stays regardless; the question is only whether to *add* a detector.
2. **A correct detector is an ancestor-path set, not a visited set.** A shared subtree (`s := []any{1}; m["x"] = s; m["y"] = s`) is an acyclic DAG and a legal payload; a visited set would falsely reject it. The path set needs a push/pop per container on every write, and container identity for `map[string]any` / `[]any` needs `reflect.Value.UnsafePointer` or `unsafe.SliceData`, which is new plumbing in the hottest write-path function.
3. **What it would buy.** Distinguishing "cyclic" from "too deep" in the message, and aborting a cycle within its period rather than after 1000 frames. The 1000 frames cost microseconds; a cyclic payload is an embedder bug and the root key in the message already locates it.

Conclusion: no change. The "cyclic or too deeply nested" wording stays as the honest statement of what the cap knows.

### Follow-up

Opened #572 for the `flattenToAttributes` extraction (98/100 lines): move the empty-list marker branch into `appendEmptyListMarker`, pure refactor, before the next change lands there. Not folded into this PR: it is unrelated to the fix and the review approved the diff as is.

iceboundrock:
## Review

### Findings
No blocking findings. No changes requested.

Non-blocking observation: `flattenToAttributes` is now 98 lines, still within the 100-line limit in `coding-standard.md`. The already-tracked extraction follow-up in #572 is the appropriate place to address its next unrelated change.

### Spec
- #406 requires one shared bound over the caller-owned `NormalizeDottedKeys` and `flattenToAttributes` walks. `maxPayloadNestingDepth = 1000` and `checkPayloadDepth` meet that requirement.
- Map members and slice elements each count as a nesting edge. This correctly bounds a `[]any` cycle whose dotted attribute path does not grow.
- `payloadPosition` preserves the literal top-level key rather than deriving it from a dotted path, so diagnostics identify keys such as `a.b` as written by the caller.
- `validateWritePayload` returns the normalization error before schema validation and independently of report-only enforcement. This prevents an unwritable cyclic payload from being logged as an accepted legacy schema violation.
- The error-handling reference documents the carrier class, report-only boundary, depth trade-off, and the validator marshal fallback as defense in depth.

### Standards
- The new rejection is a published `forma.InvalidInputf` carrier. It names the nesting limit, top-level attribute, and expected state; callers retain `errors.Is(err, forma.ErrInvalidInput)` classification and the public body excludes outer wrapping context.
- `validateWritePayload` adds schema context with `%w`; this is not a bare sentinel wrap.
- All production callers of `NormalizeDottedKeys` use its new `(map[string]any, error)` contract. No stale single-result calls remain.
- Touched source files remain within the 500-line cap. New control flow uses early error returns and does not add excessive nesting.

### Observations
- The exact boundary is coherent across both walks: root is depth zero, depth 1000 is accepted, and a container at 1001 is rejected.
- Tests cover map and slice cycles through both normalization and flattening, including unknown schema keys, plus the report-only update path verifies that rejected input does not reach storage.
- The depth cap intentionally rejects both cycles and acyclic payloads beyond the limit. That is the required stack-safety trade-off, and the message accurately says `cyclic or too deeply nested` rather than claiming cycle detection.

### Verification
- `git diff --check 68b03f1^ 68b03f1`: PASS
- `make fmt-check`: PASS
- Targeted transform cycle/depth tests: PASS
- Targeted report-only and bare-sentinel tests: PASS
- `./scripts/test_with_container_runtime.sh`: PASS, including `internal/e2e_harness` via the detected Podman socket
- PR CI is green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

### Assessment / Summary
This is a faithful, minimal implementation of #406. The depth accounting, public error semantics, report-only behavior, documentation, and regression coverage align. The change is approve-worthy as written.

PR #573, opened 2026-09-16
fix: guard batch update with per-row locks and an in-transaction merge (#554)

### LINKED ISSUE #554: bug: batchUpdateAtomic is an unguarded read-modify-write, same lost-update shape as #457
Split out of #457, whose acceptance criteria named `entityCRUDService.Update`
only. PR #553 fixed that path; the batch path still has the identical bug.

## The problem

`entityBatchService.batchUpdateAtomic` (`internal/entity_batch_service.go:268`)
loops over the request's operations and, per row:

- reads the merge base with `s.repository.GetPersistentRecord` at
  `internal/entity_batch_service.go:296` — on the pool, outside any
  transaction, and before `BatchUpdatePersistentRecords` opens the write
  transaction at all;
- merges, validates, and collects the record;

then hands every collected record to
`DBPersistentRecordRepository.BatchUpdatePersistentRecords`
(`internal/postgres_persistent_repository_batch.go:66`), which opens its own
transaction and takes no `pg_advisory_xact_lock` per row and no version
predicate.

That is the same shape #457 described:

- two concurrent writers touching one row both merge onto the same pre-write
  snapshot, and because the EAV write is delete-all-then-reinsert, the second
  committer replaces the first committer's whole document — silently, both
  returning 200;
- a batch racing a single-row `Update` loses the same way, and now
  asymmetrically: since #553 the single-row path holds the per-row advisory
  lock across its read, but the batch path takes no such lock, so the batch
  can still clobber a concurrent `Update` that was itself correctly guarded.

`GetPersistentRecord` is no longer torn — #553 made it one statement — so the
torn-read half of #457 does not recur here. The lost update does.

## Why it was not fixed in #553

The single-row fix is `MergePersistentRecord`: lock, read inside the write
transaction, merge, write, commit. The batch analogue needs a multi-row merge
callback and a lock-ordering design, because a batch must take N per-row
advisory locks in one transaction. The existing precedent is that batch
writers acquire these locks in input order (see `lockRowVersion`'s doc
comment), which makes an order inversion between two overlapping batches a
deadlock PostgreSQL detects and errors rather than a correctness bug — but
that then needs a caller-facing story for the 40P01 it produces, which does
not exist yet either.

## Suggested shape

- A `BatchMergePersistentRecords(ctx, tables, keys, merge)` on the repository:
  one transaction, per-row advisory locks taken in a deterministic order
  (sort by `(schemaID, rowID)` rather than input order, so two overlapping
  batches cannot invert), each merge base read inside that transaction, each
  merge run there, all writes committed together.
- `batchUpdateAtomic` moves its per-operation read/merge/validate body into
  that callback, as `Update` did in #553.
- Decide what a deadlock abort answers to the caller — probably the same
  conflict protocol #457 deferred for the optimistic predicate.

## Regression coverage to add

- Two concurrent batches updating disjoint fields of one shared row: both
  fields survive.
- A batch and a single-row `Update` racing one row: neither loses the other's
  field.
- Two overlapping batches whose rows appear in opposite input order: no
  deadlock (the deterministic lock order makes this pass), or a defined error
  if the design keeps input order.

iceboundrock:
Closes #554.

## Problem

`batchUpdateAtomic` read every merge base on the pool, merged in memory, and handed finished records to `BatchUpdatePersistentRecords`, which wrote them under no advisory lock — the same lost-update shape #457 closed for the single-row path. Two concurrent batches (or a batch racing a single-row update) merged onto the same stale snapshot, and because the EAV write is delete-all-then-reinsert, the second committer replaced the first committer's whole document. The service also echoed its own pre-write merge, so the answered `ltbase_updated_at` was the caller's clock rather than the `GREATEST(now, prev + 1)` storage actually stored.

## Change (per the maintainer ruling on the issue)

- **One shared lock-order helper** (`postgres_persistent_repository_lock_order.go`): `sortedRowKeys` sorts a copy of the keys by ascending `(schema_id, row_id)` tuple — not the hashed lock key — and `lockRowKeys` takes `pg_advisory_xact_lock` for each in that order. No dedup (the lock is re-entrant); writes stay in input order.
- **All three batch writers lock up front**: `BatchInsertPersistentRecords`, `BatchDeletePersistentRecords`, and the new merge writer take every row's lock before their first write. Two batches over the same rows in opposite input order can no longer invert their acquisition, and the same sort orders the subsequent main-table row locks.
- **`BatchUpdatePersistentRecords` → `BatchMergePersistentRecords(ctx, tables, keys, merge)`** (`postgres_persistent_repository_batch_merge.go`): a READ COMMITTED transaction that locks the sorted keys, then per key in input order reads the base inside the transaction, hands it to the `PersistentRecordBatchMerge` callback, writes the result, and returns the stored records. The merge callback's error is wrapped with a plain `%w` so a published `NotFoundf` message survives to the caller (#313).
- **`batchUpdateAtomic`** moves read / merge / validate into that callback (`mergeBatchUpdateRecord`) and derives its response from what storage stored, so the answered version is the effective one.
- `lockRowVersion` godoc updated to describe the batch discipline.

## Verification

- `make fmt-check`, `make lint` (exit 0), `./scripts/test_with_container_runtime.sh` (full `make test`, exit 0).
- New coverage:
  - unit test on `sortedRowKeys` (schema-then-rowid-bytes order, input untouched, ignores hashed-key order);
  - pgxmock tests pinning the sorted lock pass ahead of input-order writes for insert, delete and merge, plus the missing-row path for merge (published message `operation[0]: entity not found: …` survives) and nil-merge / empty-keys guards;
  - service tests: merge base comes from the write transaction (`getCalls == 0`, a `beforeMerge` committer is seen), answered `version` equals the stored one, missing row → `ErrNotFound` naming the operation and nothing written;
  - real-Postgres integration tests: concurrent disjoint-field batches all survive, batch vs single-row merge neither loses, opposite-input-order batches both commit, batch merge vs batch delete in opposite order never deadlocks, same-millisecond double update answers stored `GREATEST(...)+1`.
- Sabotage checks against real Postgres: locking in input order instead of sorted order makes the opposite-order tests fail with SQLSTATE 40P01 on round 1 (and the two mock lock-order tests fail); removing the lock pass makes both lost-update tests fail with "attribute 1 was lost". Production code restored afterwards.


iceboundrock:
## Findings

No blocking findings. I found no correctness, data-loss, deadlock, error-semantics, or regression issue in the changed implementation.

## Spec

- The atomic batch update now performs the complete read/merge/write sequence inside one `READ COMMITTED` transaction.
- All batch writers acquire every per-row advisory lock up front in deterministic `(schema_id, row_id)` order, while preserving input order for writes and positional results.
- The merge base is read only after the lock pass, so overlapping batch updates and batch-vs-single-row updates cannot merge against a stale document.
- The repository returns the records after PostgreSQL applies the effective monotonic version, so the service response reflects the stored `ltbase_updated_at`.
- Callback failures abort the transaction, preserve the public error carrier/message, and leave the atomic batch unchanged.
- Existing validation, duplicate-row rejection, EAV schema-scope preservation, and batch atomicity behavior remain covered.

## Standards

- Changed production files remain within the repository limits: no source file exceeds 500 lines and no changed function exceeds 100 lines.
- Errors are wrapped with operation/row context and the public-vs-operator error handling convention is preserved.
- The implementation is formatted and passes the repository lint gate.

## Assessment

LGTM. No changes requested.

## Observations

- The new mock tests cover sorted lock acquisition, transaction-local merge reads, positional stored results, missing-row behavior, and rollback paths.
- The new real-PostgreSQL tests cover concurrent disjoint-field batches, batch versus single-row merge, opposite input order, batch merge versus batch delete, and same-millisecond version advancement.
- The local environment could not execute the real-PostgreSQL concurrency cases because `localhost:5432` was unavailable; those tests self-skipped. Remote CI reports the E2E smoke and other checks successful.

## Verification

- `make fmt-check`: passed.
- `make lint`: passed, including the infra module.
- Targeted `go test ./internal -run 'Test(EntityManager_BatchUpdate|BatchMerge|BatchInsertLocks|BatchDeleteLocks|SortedRowKeys)' -count=1 -v`: passed; PostgreSQL integration cases were skipped only because no local PostgreSQL was reachable.
- `make test`: all unit-test packages passed, but the command exited non-zero when the existing `internal/e2e_harness` test attempted to initialize testcontainers without a configured `$XDG_RUNTIME_DIR`/container runtime. This is an environment limitation, not a failure in the PR changes.
- PR CI: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke are successful.

## Summary

The PR addresses the reported lost-update and stale-version-response issues with a consistent lock/transaction discipline and has appropriate regression coverage. No blocking issues were identified.

PR #574, opened 2026-09-16
sqlgen: push bool_smallint equality down as an operand-independent BETWEEN range (#565)

### LINKED ISSUE #565: sqlgen: pg-main/hybrid-main bool pushdown compares the raw smallint column, not the #404 truthiness image

iceboundrock:
Closes #565.

## What was wrong

Every reader derives a column-bound bool through the read contract — `> 0.5` (`sqlgen.BoolTruthiness`) for `bool_smallint`, `= '1'` for `bool_text` (`mainColBoolExpr`, the CDC export, the PG-EAV `(x.value_numeric > 0.5) = $n` leaf, the Go read funnel). The two entity_main pushdown routes compared the raw column instead: the pg-main clause the DuckDB hot leg embeds in its `postgres_scan` (`PG_WHERE_CLAUSE`) and the hybrid builder's main branch on the Postgres/PreferHot route both emitted `m.smallint_01 = $1` with `1`/`0`, and `m.text_02 = $1` with `'1'`/`'0'`.

So a stored off-contract image — a `2` written before the strict funnel, or a `'true'` — reads as true/false on an unfiltered read and then matches neither `equals:true` nor `equals:false` on the hot routes, while matching correctly once flushed to parquet. The issue scoped this to `bool_smallint`; running the amended e2e acceptance showed `bool_text` has the same divergence for `equals:false` / `not_equals:false` (`= '0'` misses `'true'`, `!= '0'` returns it), so both encodings are fixed here (recorded on the issue: https://github.com/Lychee-Technology/forma/issues/565#issuecomment-5692413355).

## The fix

`sqlgen.BoolMainPredicate{Encoding, Truthy}` (`internal/sqlgen/bool_main_predicate.go`) is the pushdown form of a bool-encoded leaf under `equals` / `not_equals`, attached by the predicate normalizer to both `PgMainLeafPayload.Bool` and `HybridLeafPayload.MainBool`, and rendered by both emitters:

| encoding | clause text (every operand) | truthy binds | falsy binds |
|---|---|---|---|
| `bool_smallint` | `m.<col> BETWEEN ? AND ?` | `1, 32767` | `-32768, 0` |
| `bool_text` | `(m.<col> = '1') = ?` | `true` | `false` |

`not_equals:X` is `equals:!X`. On a SMALLINT the two ranges are exactly `> 0.5` / `<= 0.5`; the text form is the read contract verbatim. NULL stays NULL under every spelling, so an unset optional bool still matches neither side.

**Why not the issue's literal Option B (`> 0.5` / `<= 0.5`).** The federated plan cache keys on the query shape without operands (`queryplan.HashFederatedQueryShape`) and `duckdb_query_build.go` reuses the cached `PgMainClause` text on a hit, so `equals:true` and `equals:false` must produce the same clause text and differ only in binds. An operand-dependent operator rewrite would have served `equals:false` with the `equals:true` clause after the first hit. `TestPlanBindReusesBoolPredicateAcrossOperands` pins this: a plan compiled for `equals:true` is bound with `equals:false` and must yield the falsy binds under the same text.

**Sargability.** Integer bounds keep the smallint predicate on the `integer_ops` btree family (a `0.5` literal against an int2 column casts the column). For `bool_text` there is no range equivalent; the hot-column indexes are `(ltbase_schema_id, ltbase_row_id, <col>)`, so a `<col>` predicate is never a leading index condition on either encoding anyway.

**Placeholders.** A range leaf consumes two counter ticks on the pg-main emitter, so the sibling EAV `$n` numbering in the same `DualClauses` shifts (pinned in the characterization table, `span` 3 → 4). The hybrid builder allocates `Arity()` placeholders through a new `nextPlaceholder()` that the EAV leaf now also uses.

## Files

- `internal/sqlgen/bool_main_predicate.go` — new type + `pgMainBoolPredicate` classifier.
- `internal/sqlgen/predicate_ir.go`, `predicate_normalizer.go` — payload fields and normalizer hook (hybrid and pg-main).
- `internal/sqlgen/dualpath_sql_generator.go`, `dual_clause_plan.go` — pg-main emitter and re-bind.
- `internal/sqlgen/dualpath_sql_helpers.go` — `convertPgBoolValue` doc: equality leaves no longer reach it; the raw bind remains only for the hybrid route's non-equality operators (pre-existing, out of scope).
- `internal/postgres_persistent_repository_query.go` — hybrid `emitMainLeaf` → `mainPredicate` + `nextPlaceholder`.
- Tests: `TestPgMainAndHybridMain_BoolPredicateSpelling` (both encodings × both operators, next to `TestPgEavLeafPayload_ComparisonLHSUsesBoolTruthiness`), `TestBoolMainPredicate_MatchesReadContract` (executes pushdown vs. `mainColBoolExpr` on DuckDB over -32768…32767/NULL and `'1'`,`'0'`,`'true'`,`'2'`,`''`,NULL), `TestPlanBindReusesBoolPredicateAcrossOperands`, hybrid `TestHybrid_BoolIntEncoding` / `TestHybrid_BoolTextEncoding` (both anchors) / `TestHybrid_BoolIntRangeAdvancesArgCounter`; existing pins updated.
- `internal/e2e_harness/production/bool_bound_parity_e2e_test.go` — `TestBoundBoolParityNoEAVAllTiers` now runs the `parquetProbes` table on `hot-pg-planted` and `hot-duck-planted` right after planting, adds `not_equals` probes, drops the scoping paragraph (issue amendment).
- `docs/federated-query/design.md` — bool section names the pushdown as on-contract with the spelling and the plan-cache rationale.

## Verification

- `./scripts/test_with_container_runtime.sh` (rootless Podman): all packages ok.
- `go test -tags=e2e ./internal/e2e_harness/production/ -run TestBoundBoolParityNoEAVAllTiers -v`: PASS. Before the `bool_text` change the same run failed on `hot-pg-planted/flagText_false`, `flagText_0`, `flagText_ne_false` and the `hot-duck-planted` twins — the divergence the issue had scoped out.
- `make fmt-check`, `make lint` (golangci-lint v1.64.8, `--build-tags e2e`, root + infra): clean.


iceboundrock:
## Post-merge review of #574 (merge commit 96daa8a, closes #565)

Reviewed the squash-merged commit (`96daa8a`, 14 files, +440/-64) against the #565 spec. This is a single consolidated review: findings, spec, standards, assessment, observations, verification, summary.

### Spec

#565 required: (1) a sqlgen pin for the pg-main + hybrid-main bool predicate spelling, both encodings x both `equals`/`not_equals`, next to `TestPgEavLeafPayload_ComparisonLHSUsesBoolTruthiness`; (2) an integration assertion that a raw-SQL `smallint_01 = 2` row is returned by `equals:true` on the pg-main route, matching the DuckDB hot leg; (3) a `docs/federated-query/design.md` update naming the pg-main route as on-contract. The issue's amendment additionally required the hot-leg probes to become the shared `parquetProbes` table (dropping the scoping paragraph) plus `not_equals` probes.

Coverage, all met:

- (1) `TestPgMainAndHybridMain_BoolPredicateSpelling` (both encodings x `equals`/`not_equals`, plus `equals:2` truthiness-operand rows and the `equals:banana` rejection), hybrid `TestHybrid_BoolIntEncoding` / `TestHybrid_BoolTextEncoding` on both anchors, `TestHybrid_BoolIntRangeAdvancesArgCounter` for the placeholder shift, and the characterization-table updates (`span` 3 -> 4, EAV `$n` renumbering).
- (2) `TestBoundBoolParityNoEAVAllTiers` now runs `parquetProbes` on `hot-pg-planted` and `hot-duck-planted` immediately after planting (stored `2` and `'true'`), and `TestBoolMainPredicate_MatchesReadContract` executes pushdown vs `mainColBoolExpr` side by side on DuckDB over the full off-contract image set (`-32768..32767`/NULL, `'1'/'0'/'true'/'2'/''`/NULL).
- (3) `docs/federated-query/design.md` bool section rewritten to name the pushdown spelling, the plan-cache rationale, and the integer-bounds sargability note.
- Amendment: `not_equals` probes added to the spelling table; scoping paragraph replaced by the on-contract statement.

Two documented deviations from the issue text, both justified in the PR body, the #565 thread, code comments, and the design doc: (a) the operand-independent `BETWEEN ? AND ?` with integer bounds instead of the literal `> 0.5` / `<= 0.5` rewrite (the federated plan cache reuses `PgMainClause` text across operands, pinned by `TestPlanBindReusesBoolPredicateAcrossOperands`); (b) `bool_text` fixed in the same change (`(<col> = '1') = ?` with a bool bind) after the amended e2e acceptance showed `equals:false` / `not_equals:false` diverging on a stored `'true'`. Both deviations are the right call.

### Standards

- File/function size: new file `internal/sqlgen/bool_main_predicate.go` is 100 lines; all touched functions stay well under the 100-line limit; no file exceeds 500 lines (checked: `predicate_normalizer.go` 446, `postgres_persistent_repository_query.go` 391).
- Errors: `pgMainBoolPredicate` rejects bad bool operands with `forma.InvalidInputf` (user-facing 4xx carrier), consistent with `convertPgBoolValue`/`parsePgEavValue` and the `TestNoBareSentinelWraps` contract. No bare `%w` sentinel wraps introduced.
- Structure: early returns, flat classifier, max nesting 2. `gofmt -l` clean and `go vet ./internal/sqlgen/` clean on the merged tree.

### Findings

No blocking findings. Two minor non-blocking observations for a follow-up:

1. (Minor) `BoolMainPredicate.Render` indexes `placeholders[0]` / `[1]` without an arity guard. Both call sites allocate exactly `Arity()` placeholders (`pgMainTypedEmitter`, `hybridConditionBuilder.mainPredicate`) and `Bind` goes through `Args()`, so this is safe today, but a wrong-arity caller panics instead of erroring. Consider a length check returning an empty string or documenting the panic contract.
2. (Minor, pre-existing scope edge, now documented) Non-`=`/`!=` operators on a bool column still take the raw `ConvertPgMainValue` bind on the hybrid route (`m.<col> <op> $n` via `mainPredicate` fallthrough). The dual-path pg-main route rejects them (`classifyPredicate` allows only `equals`/`not_equals` for bool; EAV likewise in `normalizePgEavPayload`), so `gt`/`lt` on a bool is only expressible through hybrid, where it still compares the raw image. The helper comment says this explicitly ("left for the hybrid route's non-equality operators ... out of scope"), so this is recorded, not missed — a follow-up could either reject bool range operators on hybrid too or extend the predicate there.

Correctness notes I verified while reading: `not_equals:X == equals:!X` collapse (`parsed == (sqlOp == "=")`) is right on the two-valued domain; NULL stays NULL under all three spellings (`BETWEEN`, `(col='1')=bool`, `>0.5`); smallint bounds `(1,32767)/(-32768,0)` are exactly `>0.5`/`<=0.5` on SMALLINT; `bool_text` has no range equivalent so the verbatim-contract form is the only operand-independent option; `Arity()`-driven placeholder allocation keeps the sibling EAV `$n` numbering and `ParamSpan` consistent (plan-bind matrix + span test cover it).

### Observations

- The plan-cache constraint shaping the spelling (operand-independent text, binds carry the verdict) is the strongest part of this change: it rules out the issue's literal Option B text for a real reason, and the reuse test pins it rather than leaving it as prose.
- The `bool_text` scope expansion mid-execution was warranted and evidenced (the four `flagText` failures on both hot routes before the fix, quoted in the PR body); the issue's "not affected" assumption only held for `equals:true`.
- Sargability reasoning is sound in direction (integer bounds avoid casting the int2 column, unlike a `0.5` numeric literal). The hot-column indexes lead with `(schema_id, row_id)`, so this predicate was never a leading index condition anyway — the doc says so honestly instead of overselling the index story.

### Verification

- CI on the PR: all 5 checks SUCCESS (Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke); `mergeStateStatus` was CLEAN at merge.
- Local, on the merged tree (`96daa8a`): `go test ./internal/sqlgen -count=1` PASS; targeted `TestBoolMainPredicate_MatchesReadContract`, `TestPgMainAndHybridMain_BoolPredicateSpelling`, `TestPlanBindReusesBoolPredicateAcrossOperands`, `TestToDualClauses_BoundBoolOperandSpellingParity`, `TestHealth_BoolEncoding`, `TestPlanBindEquivalenceMatrix`, `TestPlanBindParamSpan` all PASS; `go test ./internal -run TestHybrid_Bool` PASS (`BoolIntEncoding`, `BoolIntRangeAdvancesArgCounter`, `BoolTextEncoding`). `gofmt`/`go vet` clean as above.
- Not re-run locally: the container-backed production e2e (`TestBoundBoolParityNoEAVAllTiers`) — relies on the author's reported PASS plus CI's E2E Smoke signal; the DuckDB-executed contract-equivalence unit test covers the same verdict table without containers.

### Assessment

Approve (post-merge, no action required). Spec fully met including the amendment, deviations from the issue text are justified and pinned by tests, standards clean, verification green. The two findings above are follow-up material at most.

### Summary

#574 closes the last route embodying its own bool table: both entity_main pushdown emitters now render `BoolMainPredicate` (range for `bool_smallint`, verbatim `= '1'` contract for `bool_text`) with operand-independent clause text, so off-contract images answer filters the way every reader already reads them. Well-scoped, well-pinned, well-documented.

iceboundrock:
Response to the post-merge review above (https://github.com/Lychee-Technology/forma/pull/574#issuecomment-5706977749). Both findings are addressed in #576 (tracked as #575, which carries the full option comparison).

**Finding 1 (`Render` arity guard):** verified — both call sites allocate `Arity()` placeholders, so the panic was unreachable, but only by convention. Of the two suggested resolutions, a length check returning `""` was rejected: an empty fragment spliced into a WHERE clause is a syntax error at best and a silently dropped predicate inside a composite at worst, which is worse than the panic. Documenting the panic contract would have left the latent bug class in place. #576 instead changes `Render` to take `next func() string` and call it once per bind, so the predicate allocates its own placeholders and there is no arity for a caller to get wrong; `Arity()` is removed. Both emitters already had an allocator to hand over.

**Finding 2 (bool range operators on hybrid):** verified as a real inconsistency, and slightly wider than described. Besides `gt`/`lt` rendering `m.<col> > $n` against the raw `1`/`0` bind, `starts_with`/`contains` on a bound bool rendered `m.<col> LIKE $n` against an `int64`, which Postgres rejects at execution (`smallint ~~ bigint`) as an opaque 500. Extending the predicate to range operators was rejected: there is no meaningful order on a two-valued domain. #576 rejects a bool attribute under any operator other than `=` / `!=` in the shared classifier `pgMainBoolPredicate`, with the EAV whitelist's message, so all three routes now answer the same 400. New `TestHybrid_BoolNonEqualityOperatorRejected` fails on `96daa8a` and passes with the fix. `docs/federated-query/design.md` names the all-route rejection.

**Everything else** (spec coverage, standards, correctness notes, sargability observation, verification) — no action; the reading matches mine, including the note that the hot-column indexes lead with `(schema_id, row_id)` so the bool predicate was never a leading index condition.

iceboundrock:
## Consolidated Post-Merge Review

Reviewed the merged commit `96daa8a` against the bool truthiness and pg-main/hybrid pushdown contract. The equality implementation is directionally correct, but the merged PR contains the following findings.

### Findings

1. **High: hybrid main-table bool predicates accept unsupported non-equality operators.**

   Location: [`internal/postgres_persistent_repository_query.go` (`mainPredicate` fallback)](https://github.com/Lychee-Technology/forma/blob/96daa8a0f19fd678faa019453363de9527f0f344/internal/postgres_persistent_repository_query.go#L343-L356) and [`internal/sqlgen/bool_main_predicate.go` (`pgMainBoolPredicate`)](https://github.com/Lychee-Technology/forma/blob/96daa8a0f19fd678faa019453363de9527f0f344/internal/sqlgen/bool_main_predicate.go#L77-L83).

   For a column-bound bool such as `gt:true`, `gte:1`, `lt:false`, `lte:0`, `starts_with:1`, or `contains:1`, `BoolMainPredicate` declines the operator and the hybrid emitter falls through to raw `m.<column> <op> $n` SQL. The pg-main and EAV routes reject the same input as invalid. This makes the route behavior inconsistent: range operators can return incorrect rows for off-contract encodings (for example, a stored `2` is greater than the raw bind `1`), while `LIKE` can fail at execution with a PostgreSQL type error instead of a 400.

   **Required action:** reject bool attributes under every operator other than `=` and `!=` in the hybrid classification path, or otherwise make all routes use one explicitly defined bool operator contract. #576 implements the rejection and adds coverage for both encodings and both anchors.

2. **Low: `BoolMainPredicate.Render` panics on an invalid placeholder arity.**

   Location: [`internal/sqlgen/bool_main_predicate.go` (`Render`)](https://github.com/Lychee-Technology/forma/blob/96daa8a0f19fd678faa019453363de9527f0f344/internal/sqlgen/bool_main_predicate.go#L47-L64).

   The variadic API indexes `placeholders[0]` and `[1]` without validating the supplied slice. The two production callers currently pass exactly `Arity()` placeholders, so this is not a currently observed request-path failure, but a caller mistake becomes a process-level panic. #576 removes the caller-sized arity contract by taking an emitter allocator and therefore closes this risk at the design level.

### Spec

- The `bool_smallint` equality pushdown uses integer `BETWEEN` bounds equivalent to the shared `> 0.5` truthiness rule, preserving NULL behavior and operand-independent SQL text for the plan cache.
- The `bool_text` form `(<column> = '1') = ?` correctly matches the read contract, including off-contract text images.
- Equality and `not_equals` coverage, plan-cache rebinding, `ParamSpan`, both hybrid anchors, and the design documentation are present.
- The merged PR did not fully satisfy the all-route operator consistency contract because hybrid non-equality bool operations remained on the raw fallback.

### Standards

- I found no file/function size, SQL identifier sanitization, or error-wrapping violation in the merged diff.
- The low-severity panic is an API robustness issue rather than a repository style violation; the follow-up removes the underlying invalid-arity state.

### Assessment

Post-merge assessment: the equality fix is sound, but #574 was incomplete without the follow-up. No rollback is indicated because the findings are narrowly addressed by #576, which should be treated as the required corrective change.

### Observations

- Expanding the fix to `bool_text` was justified: raw `= '0'` / `!= '0'` comparisons diverge from the `= '1'` read contract for stored values such as `'true'`.
- The operand-independent clause text is necessary because the federated plan cache reuses the compiled pg-main skeleton across operands; carrying the truth value in binds is the correct tradeoff.
- The merged tests cover the equality and plan-binding matrix well, but did not cover hybrid bool range/LIKE rejection or invalid `Render` arity.

### Verification

- PR CI for #574: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all passed.
- The merged-code paths reproduce the two findings: hybrid non-equality emits raw main-column SQL without an error, and a short `Render` placeholder list panics.
- The follow-up regression test in #576 fails against `96daa8a` and passes with the corrective change; the current targeted SQLgen/internal test run on #576 passed.

### Summary

#574 correctly repairs bool equality pushdown and truthiness parity, but its merged implementation left two narrowly scoped robustness/consistency gaps. Both are resolved by #576; no further action is requested on the already merged commit beyond retaining that follow-up.

PR #576, opened 2026-09-17
sqlgen: allocate bool predicate placeholders in Render and reject bool range operators on the hybrid route (#575)

### LINKED ISSUE #575: Bool main predicate: allocate placeholders in Render and reject bool range operators on the hybrid route (#574 follow-up)
Follow-up to the post-merge review of #574 (https://github.com/Lychee-Technology/forma/pull/574#issuecomment-5706977749), which closed #565. Both findings there were marked non-blocking; this issue records the analysis and the chosen resolution for each.

## Finding 1: `BoolMainPredicate.Render` indexes `placeholders[0]` / `[1]` without an arity guard

**What it means.** `Render(column, placeholders ...string)` trusts the caller to have allocated exactly `Arity()` placeholders. Both callers (`pgMainTypedEmitter.EmitTypedLeaf`, `hybridConditionBuilder.mainPredicate`) do, so it is safe today, but the contract is enforced only by convention: a third caller, or a future encoding with a different arity, can panic on an index out of range instead of erroring.

**Options compared.**

| Option | Verdict |
|---|---|
| (a) Document the panic contract, no code change | Leaves the latent class of bug in place; index-out-of-range is a poor error surface for an internal invariant. |
| (b) Length check returning `""` (the reviewer's first suggestion) | Worse than the panic: an empty fragment spliced into a WHERE clause is either a SQL syntax error at execution or, inside a composite, a silently dropped predicate. Rejected. |
| (c) Length check returning an error | Threads an error through a pure render function for a programmer error only; the two call sites already return errors so it is feasible, but it keeps the arity contract and just changes how it fails. |
| (d) **Make `Render` allocate its own placeholders** through a `next func() string` supplied by the emitter | Removes the contract by construction: the predicate calls `next()` once per bind, so there is no count for a caller to get wrong, and `Arity()` disappears. Both emitters already own a "next placeholder" notion (`?` + counter tick on pg-main, `$n` on hybrid), so the call sites get shorter. **Chosen.** |

## Finding 2: hybrid main branch still accepts range / LIKE operators on a bool column

**What it means.** `classifyPredicate` (pg-main pushdown) and `normalizePgEavPayload` (EAV) both refuse a bool attribute under anything but `equals` / `not_equals` with a 400. The hybrid builder's main branch (the Postgres-only `advanced_query` route for column-bound attributes) had no such whitelist: `gt:true` on a `bool_smallint` column rendered `m.<col> > $n` against the raw `1` bind (so a stored `2` was "greater than true"), and `starts_with:1` rendered `m.<col> LIKE $n` against an `int64`, which Postgres rejects at execution time as `operator does not exist: smallint ~~ bigint` — an opaque 500. The #574 helper comment recorded this as out of scope; the reviewer asks for either rejecting it on hybrid too or extending the predicate.

**Options compared.**

| Option | Verdict |
|---|---|
| Extend `BoolMainPredicate` to range operators | There is no meaningful order on a two-valued domain (`gt:true` has no rows, `gte:false` is every non-NULL row); it would invent semantics that no other route offers. Rejected. |
| **Reject on hybrid with the EAV whitelist's message** | Restores one behaviour across all three routes, turns the LIKE 500 into a 400, and costs one check in the shared classifier `pgMainBoolPredicate` (both normalizers already call it, so the rule lives in one place). **Chosen.** |

**Behaviour change.** A caller that sent `gt`/`gte`/`lt`/`lte`/`starts_with`/`contains` on a column-bound bool over the Postgres route previously received either a meaningless raw-image comparison or a 500; it now receives `operator '<op>' not supported for boolean attributes` (400), identical to what the same condition already returned on the federated route and on an unbound bool.

## Verification

- `TestHybrid_BoolNonEqualityOperatorRejected` (new) fails on `96daa8a` and passes with the fix.
- `TestBoolMainPredicate_MatchesReadContract`, `TestPgMainAndHybridMain_BoolPredicateSpelling`, `TestHybrid_BoolIntRangeAdvancesArgCounter` continue to pin the spelling and the placeholder count under the allocator-based `Render`.
- `make fmt-check`, `make lint`, `./scripts/test_with_container_runtime.sh` green.

`docs/federated-query/design.md` bool section names the all-route rejection.

iceboundrock:
Closes #575. Resolves both non-blocking findings from the post-merge review of #574 (https://github.com/Lychee-Technology/forma/pull/574#issuecomment-5706977749). The option comparison for each is on the issue.

## Finding 1: `Render` arity guard

`BoolMainPredicate.Render` now takes `next func() string` and calls it once per bind, instead of indexing a caller-allocated `placeholders` slice. The arity contract is gone by construction rather than guarded (a length check returning `""` would splice an empty fragment into the WHERE clause, which is worse than the panic), and `Arity()` is removed. Both emitters already owned an allocator: pg-main ticks `paramIndex` and returns `?`; hybrid passes `b.nextPlaceholder` directly.

## Finding 2: bool range operators on the hybrid route

`pgMainBoolPredicate` (the classifier both normalizers call) now rejects a bool attribute under any operator other than `=` / `!=` with the EAV whitelist's message, `operator '<op>' not supported for boolean attributes` (`forma.InvalidInputf`, 400). The pg-main and EAV routes already refused it; the hybrid main branch was the one place that accepted it, rendering `m.<col> > $n` against the raw `1`/`0` bind (a stored `2` was "greater than true") and `m.<col> LIKE $n` against an `int64`, which Postgres rejected at execution as `smallint ~~ bigint` (opaque 500). Extending the predicate to ranges was rejected: there is no meaningful order on a two-valued domain.

**Behaviour change:** `gt`/`gte`/`lt`/`lte`/`starts_with`/`contains` on a column-bound bool over the Postgres route is now a 400, matching the federated route and unbound bools.

## Tests

- New `TestHybrid_BoolNonEqualityOperatorRejected` (both encodings, both anchors, range and LIKE operators): fails on `96daa8a`, passes here.
- `TestBoolMainPredicate_MatchesReadContract` uses the allocator form; `TestPgMainAndHybridMain_BoolPredicateSpelling` and `TestHybrid_BoolIntRangeAdvancesArgCounter` still pin the spelling and the placeholder count.
- `make fmt-check`, `make lint`, `./scripts/test_with_container_runtime.sh` green.

`docs/federated-query/design.md` bool section names the all-route rejection.


iceboundrock:
## Consolidated Review

### Findings

No blocking findings. I found no correctness, regression, security, or standards issue in the #576 diff.

### Spec

- The `Render` arity contract is removed: both emitters now provide the placeholder allocator, and `BoolMainPredicate` consumes exactly one placeholder per bind.
- Hybrid main-table bool predicates now reject every operator other than `=` and `!=`, including ranges and `LIKE` operators, with the invalid-input error expected by the EAV contract.
- Coverage includes `bool_smallint` and `bool_text`, both hybrid anchors, equality and `not_equals`, operand parsing, placeholder advancement, and the relevant truthiness spellings.
- The federated-query design document records the all-route rejection behavior.

### Standards

- Error handling, identifier sanitization, and placeholder allocation follow the existing repository conventions.
- The touched files and functions remain within the repository size limits.
- No bare sentinel wrapping, unnecessary compatibility layer, or unrelated change was introduced.

### Assessment

Approve. This PR addresses both non-blocking follow-ups recorded against #574 and does so with the smaller, safer design: the predicate owns placeholder allocation instead of accepting a caller-sized variadic slice.

### Observations

- The pg-main path still uses its existing generic `unsupported operator` message, while the hybrid path uses `operator ... not supported for boolean attributes`; both preserve the required 400 behavior, so this is only an error-text consistency note, not a correctness issue.
- There is no dedicated hybrid test for a bool predicate followed by an EAV predicate, although the shared allocator and existing dual-path numbering tests cover the implementation pattern.

### Verification

- PR CI: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all passed.
- Local targeted verification on the #576 branch: `GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal -run 'Test(Bool|Hybrid|Plan|ToDualClauses|BuildHybrid)' -count=1` passed.
- The full container-backed suite was not rerun locally; the PR's E2E Smoke check was successful.

### Summary

#576 is a focused, well-tested follow-up: it closes the latent `Render` arity panic and makes bool non-equality operators consistently reject as invalid input across the hybrid, pg-main, and EAV routes. No changes are required before merge.

PR #577, opened 2026-09-17
sqlgen: normalise iso8601-bound date/datetime columns on the DuckDB hot leg and outer select (#555)

### LINKED ISSUE #555: bug: DuckDB hot-leg projection does not normalise iso8601-encoded date/datetime main columns

iceboundrock:
Closes #555.

## Problem

A date/datetime attribute bound to a text column with `encoding: iso8601` is admitted by `schemameta.ValidateColumnBinding` (#459) and round-trips through the write path, the Postgres read path and the CDC export, but the DuckDB reader broke at both ends:

1. **Hot leg.** `buildPGProjection` special-cased only bool and uuid, so the column fell to `COALESCE(ANY_VALUE(hot_vals.<attr>), m.text_NN)` — a `Binder Error: Cannot mix values of type BIGINT and VARCHAR in COALESCE`. `BuildPGSelectNoEAV` emitted the raw `m.text_NN AS <attr>`, which put a VARCHAR into the UNION ALL and failed on the first filter with `Could not convert string '2024-01-02T03:04:05Z' to INT64`.
2. **Outer select** (found while fixing 1). `duckDBMainColCast` rendered `CAST(<attr> AS BIGINT) AS text_NN`; the reader scans `text_NN` as a string, so `transform.readWithEncoding` received `"1704164645000"` and failed `time.Parse(RFC3339)`. This would have hit every tier, not only hot.

## Fix

- `boundBoolEncodings` (bool-only) becomes `boundMainExprs`: bound attr → normalised DuckDB expression, filled once by `boundMainExpr` and consumed by **both** hot projections, so they cannot drift. The iso8601 arm is the exporter's spelling `epoch_ms(TRY_CAST(m.text_NN AS TIMESTAMP))` (`cdc.castDateMainValue`, #219), so the hot leg and the parquet legs reach the same epoch-ms verdict.
- `duckDBMainColCast` re-derives the stored image for the pair, as it already does for bound bools (#404): `strftime(epoch_ms(CAST(attr AS BIGINT)), '%Y-%m-%dT%H:%M:%SZ') AS text_NN` — the write path's exact image (UTC, second precision, `Z`).
- `binding_compat.go`: the "DuckDB hot leg unverified; see #555" note is replaced.
- Benchmark projection: nothing to mirror (no iso8601-bound column there); the map-driven approach covers a future one.

Rulings (TIMESTAMP vs TIMESTAMPTZ, the second defect, test design) are recorded on the issue.

## Tests

- `internal/sqlgen/duckdb_iso8601_projection_test.go`: both projections' spelling; the hot-leg expression executed on DuckDB for `Z`, fractional, epoch, garbage and NULL images; the outer-select cast-back executed for positive/zero/negative/NULL verdicts, scanned as a string.
- `internal/e2e_harness/production/date_iso8601_bound_e2e_test.go` (`-tags=e2e`): two schema generations so both hot projections run (all-bound → no-EAV; plus an EAV-only attr → EAV-joined). Filters (`equals`/`gt`/`lt`/`gte`) and a `DESC` sort through hot-Postgres and the unflushed DuckDB hot leg, then the stored image pinned as the RFC3339 string **and** decoded through the real read funnel (`transform.FromPersistentRecord`) to epoch-ms, then flush → warm and compaction → cold, repeated. RED was observed with the fix stashed (both errors above); GREEN with it.
- `make fmt-check`, `make lint`, full unit suite via `scripts/test_with_container_runtime.sh`: green.


iceboundrock:
### PR Review

#### Scope & Intent

This PR closes #555: a `date`/`datetime` attribute bound to a `text` column with `encoding: iso8601` (admitted by `schemameta.ValidateColumnBinding`, #459) round-tripped through the write path, the Postgres read path and the CDC export, but the DuckDB federated reader broke at both ends:

1. **Hot leg in** — `buildPGProjection` and `BuildPGSelectNoEAV` projected raw `m.text_NN` (VARCHAR) against the EAV pivot / parquet legs' epoch-ms BIGINT (`Cannot mix BIGINT and VARCHAR in COALESCE`; `Could not convert string ... to INT64` in the UNION ALL).
2. **Outer select out** (found while fixing 1) — `duckDBMainColCast` emitted `CAST(attr AS BIGINT) AS text_NN`, so `streamDuckDBRows` scanned `"1704164645000"` into the text slot and `transform.readWithEncoding` failed `time.Parse(RFC3339)` on every tier.

The fix generalises `boundBoolEncodings` into `boundMainExprs` (attr → normalised DuckDB expression via `boundMainExpr`), consumed by both hot projections, with the iso8601 arm spelled exactly as the exporter (`epoch_ms(TRY_CAST(m.text_NN AS TIMESTAMP))`, `cdc.castDateMainValue` #219); and makes `duckDBMainColCast` re-derive the write-path image (`strftime(epoch_ms(CAST(attr AS BIGINT)), '%Y-%m-%dT%H:%M:%SZ')`) for date/datetime landing in a text column, mirroring the #404 bool pattern. `binding_compat.go` note is updated. I reviewed the change against the issue's approved design comment, the surrounding tier contracts, and the full diff (5 files, +497/−42, single commit `5f17406`).

Evaluated against, in order:

1. **Explicit spec** — issue #555 plus the approved design ruling on the same issue (two ends, `TIMESTAMP` not `TIMESTAMPTZ`, second-precision outer select, benchmark no-op, test design). The PR implements that design verbatim.
2. **Tier contract** — unified column for dates is epoch-ms BIGINT on every leg (#194/#200); CDC export spelling is authoritative (`castDateMainValue`, #219); write-path image is `unixMillisFloat64ToTimeUTC(...).Format(time.RFC3339)` (UTC, second precision, `Z`); read funnel `readWithEncoding` parses that slot with `time.Parse(time.RFC3339)`.
3. **Established patterns** — #404 bool normalisation (`mainColBoolExpr` + outer-select re-derivation), `TRY_CAST`-yields-NULL export parity (#205), `boundBoolEncodings` consumed by both hot projections (#564 review).
4. **Repo standards** — `coding-standard.md` (≤500-line files, ≤100-line funcs, wrapped errors), `docs/error-handling.md` (no new error carriers here), gofmt + `make lint` (pinned golangci-lint v1.64.8).
5. **DuckDB semantics** — verified by execution (see below), not assumed.

#### Findings

In particular, the candidates I checked and **rejected** as findings:

- `duckDBMainColCast` keys the iso8601 path off `(vt ∈ {date, datetime}, kind == Text)` rather than the binding encoding. This is safe because the validated matrix admits date/datetime→text *only* with `iso8601`; it is also the exact pattern the bool path already uses (kind, not encoding). An invalid legacy binding would already be rejected by registration / flagged by `validate-schema-consistency`.
- `boundMainExpr` dereferences `meta.ColumnBinding` without a nil guard. The sole call site (`BuildSchemaProjection`, `duckdb_schema_projection.go:105`) is guarded by `if meta.ColumnBinding != nil`, and the function is unexported. No reachable nil dereference.
- `BuildPGGroupByNoEAV` / `buildPGProjection` group by raw `m.<col>` while selecting `epoch_ms(...)` of it. This matches the pre-existing bool behaviour and DuckDB accepts an expression over a grouped column (verified by execution: `SELECT epoch_ms(TRY_CAST(m.text_02 AS TIMESTAMP)) ... GROUP BY m.text_02` succeeds). No new GROUP BY hazard.
- Outer-select second-precision truncation (`strftime` without `%f`) vs. a fractional stored string. The write funnel (`storeWithEncoding`, `persistent_record.go:231`) formats via `time.RFC3339`, so on-contract rows never carry fractions; off-contract history is consistently truncated on DuckDB while Postgres `readWithEncoding` would reject it — this is the issue's explicitly approved ruling, not a regression.

#### Observations

Non-blocking, for the author's awareness only:

- The PR description says "the map-driven approach covers a future [benchmark] one". Strictly, `BuildBenchmarkProjections` (`duckdb_benchmark_projection.go:233-237`) hardcodes `CAST(m.<col> AS VARCHAR)` and does not consult `boundMainExprs` (it also never adopted the bool map). True today because no benchmark schema binds bool/iso8601 (`tradeTime` is `bigint_02`), and consistent with how bool was left, but a future benchmark schema that binds either would need a manual mirror. No action required in this PR.
- Bound attributes never populate `OtherAttributes` (`ToPersistentRecord`, `persistent_record.go:103-113`: bound → `storeInMainColumn`, else → EAV). Hence for an iso8601-bound date the `COALESCE(ANY_VALUE(hot_vals...), mainExpr)` EAV arm is always NULL on real rows and the hot verdict is exactly the exporter's verdict — the design's parity claim holds, including for fractional inputs (both truncate at write).

**Unit** (`internal/sqlgen/duckdb_iso8601_projection_test.go`, 133 lines): pins both hot projections' spelling (EAV-joined `COALESCE` + `BuildPGSelectNoEAV`, plus `NotContains` guards that the raw `m.text_02/03 AS ...` never reaches the UNION ALL); executes the hot-leg expression on in-memory DuckDB over `Z` / fractional / epoch / garbage / NULL images; executes the outer-select cast-back over positive / zero / negative / NULL verdicts scanned as `sql.NullString` (the reader's actual scan type), plus a guard that date→bigint is untouched. Meaningful, behaviour-pinned, and would fail on the pre-fix code (spelling assertions + the RED errors recorded on the issue).

**E2E** (`internal/e2e_harness/production/date_iso8601_bound_e2e_test.go`, 281 lines, `-tags=e2e`): two schema generations so *both* hot projections run (all-bound → NoEAV; +EAV-only `note` → EAV-joined). Each walks hot-Postgres → unflushed DuckDB hot leg → flush/warm → compact/cold, probing `equals/gt/lt/gte` + a `DESC` sort with routing assertions (`UseDuckDB` pinned per leg), and pins *both* the physical image (`TextItems["text_02/03"]` is the RFC3339 string, byte-identical to Postgres) *and* the funnel decode (`FromPersistentRecord` → epoch-ms). The anchor-row trick (first parquet object so the federated read has an S3 leg to glob) and the unset-row / anchor-row non-interference (image map keyed by RowID, sort checks only late-vs-early order) are handled correctly. The date-typed `bornOn` seeded as bare `"2024-01-02"` and probed as midnight-RFC3339 correctly exercises the funnel's date parsing. No `lte`/`not_equals` probes, but the covered operators suffice for equality + range + ordering on an epoch-ms line.

**Executed here** (on the PR branch `fix/555-iso8601-duckdb-hot-leg`):

- `go test ./internal/sqlgen -run 'TestHotLegProjections_ISO8601|TestMainColISO8601Expr|TestBuildOuterSelect_ISO8601' -v` — PASS (all subtests).
- `go test ./internal/sqlgen` (full package) — PASS.
- `go test ./internal/sqlgen -run 'TestBool|TestBound|TestHotLeg|TestBuildOuter|TestBuildPG|TestSchemaProjection'` — PASS (bool parity intact).
- `go test ./internal/schemameta`, `go test ./internal/cdc -run 'TestCast|TestExport'` — PASS.
- `go vet ./internal/sqlgen ./internal/schemameta` and `go vet -tags=e2e ./internal/e2e_harness/production` — clean.
- `make fmt-check`, `make lint` — green.
- Ad-hoc DuckDB probes (`duckdb-go/v2`, in-memory): `typeof(epoch_ms(BIGINT))` is `TIMESTAMP` (naive), so the outer `strftime` is session-TZ independent (verified `UTC` / `America/New_York` / `Asia/Shanghai` render identically); `TIMESTAMP` hot-leg verdict is TZ-stable while `TIMESTAMPTZ` on an offset-less string is host-dependent — confirming the `TIMESTAMP` ruling; `TRY_CAST('...+08:00' AS TIMESTAMP)` drops the offset exactly as the issue notes (consistent with the exporter, off-contract only); `GROUP BY m.text_02` with `SELECT epoch_ms(...)` succeeds; negative-epoch and NULL round-trips match the unit expectations.

**Not executed**: the `-tags=e2e` production suite and full `make test` (container runtime out of scope for this pass). CI on the PR reports all 5 checks SUCCESS (Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke). The RED-with-fix-stashed observation was recorded on the issue and not re-reproduced here.

Standards: all touched files within the 500-line limit (452/144/133/281/107); no new func exceeds 100 lines; no new error values (no `errors.go` carrier concerns); comments name the logical type, column, and expected state where relevant.

#### Overall Assessment

Sound, minimal, well-pinned fix. The abstraction change (bool-only map → expression map consumed by both hot projections) removes the drift class rather than adding a third spelling; the exporter-verbatim hot-leg spelling and the bool-pattern outer-select re-derivation give all three tiers one verdict and the reader one image. The `TIMESTAMP`-vs-`TIMESTAMPTZ` and second-precision rulings are documented in code and tests and verified host-independent for on-contract (`Z`) data. Tests cover both projections, both ends (hot normalisation + outer cast-back with the reader's scan type), and all three tiers with routing assertions; verification above passes. Remaining risk is limited to off-contract history (fractional/offset strings planted behind the funnel), which is explicitly ruled out of contract and behaves consistently across DuckDB tiers.

#### Summary

- Findings: **0 blockers, 0 high, 0 medium, 0 low**.
- Major risks: none within contract; off-contract fractional/offset images diverge Postgres-vs-DuckDB by prior ruling, unreachable via the write funnel.
- Verification: targeted + package unit tests, vet (incl. `-tags=e2e` compile), `fmt-check`, `lint` all green; DuckDB TZ/GROUP BY semantics probed by execution; full e2e/container suite not re-run here (CI green).

PR #579, opened 2026-09-17
refactor(transform): extract the empty-list marker branch from flattenToAttributes (#572)

### LINKED ISSUE #572: Extract the empty-list marker branch from flattenToAttributes before it crosses the 100-line cap
## Context

`transformer.flattenToAttributes` (`internal/transform/transformer.go`) is 98 lines against the 100-line function cap in `coding-standard.md`. The #406 implementation note and the #571 review both flag it: the fix in #571 added two three-line depth-cap guards (map and slice cases) and consumed the remaining headroom. Any further change to the function, however small, will push it over the limit and force a refactor in the middle of an unrelated fix.

## Ask

Extract the empty-list marker branch of the `[]any` case into its own helper before the next change lands there. The branch is self-contained: given `schemaID`, `rowID`, `path`, `cache`, and `result`, it looks up the attribute, and if it is a registered `ValueTypeList` appends the marker row (`ArrayIndices: ""`, both value columns NULL) that makes an explicit `[]` round-trip instead of degrading to an absent attribute (#204). Something like

```go
// appendEmptyListMarker persists the explicit-empty-list marker row for a
// registered list attribute; an unregistered name is silently skipped, as
// before (#204).
func appendEmptyListMarker(schemaID int16, rowID uuid.UUID, path []string,
    cache forma.SchemaAttributeCache, result *[]taggedEAVRecord)
```

brings `flattenToAttributes` down to roughly 85 lines with a pure move and no behaviour change. The `#204` comment block goes with the branch.

## Constraints

- Pure refactor: no change to which rows are emitted, their `spelling`, or the silent skip for unregistered names. `TestFlatten*`, the #204 list tests, and `payload_depth_test.go` (1000-accept / 1001-reject on the flatten walk) must pass unchanged.
- Do not fold the depth-cap guards into the helper; they belong at the container cases so the flatten and normalize walks keep the same edge (#406 implementation note, deviation 2).

## Origin

- #406 implementation note: "`flattenToAttributes` ends at 98 lines against the 100-line cap; the next change to it should extract the empty-list marker branch."
- #571 review: "The only follow-up worth tracking is the already-noted one."

iceboundrock:
Closes #572.

## What

`flattenToAttributes` (`internal/transform/transformer.go`) sat at 98 lines against the 100-line function cap in `coding-standard.md`; the #571 depth-cap guards consumed the last of its headroom. This moves the explicit empty-list marker branch of the `[]any` case into a new `appendEmptyListMarker(schemaID, rowID, path, cache, result)` helper, exactly as the issue proposes. The `#204` comment block moves with it.

`flattenToAttributes` is now 83 lines; the file is 403 lines.

## Constraints from the issue

- **Pure refactor.** The helper keeps the same lookup condition (registered `ValueTypeList` attribute), emits the same marker row (`ArrayIndices: ""`, both value columns NULL) with the same `spelling`, and keeps the silent skip for an unregistered name. No test changed.
- **Depth-cap guards stay put.** Both `checkPayloadDepth` guards remain at the map and slice container cases so the flatten and normalize walks keep the same edge (#406 implementation note, deviation 2).

## Verification

- `go test ./internal/transform/...` — pass (25 `TestFlatten*` / list / depth / empty-list tests run verbosely, all pass; `payload_depth_test.go` 1000-accept / 1001-reject unchanged).
- `./scripts/test_with_container_runtime.sh` (full `make test`) — all packages pass.
- `make fmt-check` — clean.
- `make lint` (root and infra modules) — clean.


iceboundrock:
### PR Review

#### Scope & Intent

This PR extracts the explicit empty-list marker branch from `transformer.flattenToAttributes` into `appendEmptyListMarker`. The intended behavior is a pure refactor: registered list attributes still emit one marker EAV row with the same schema ID, row ID, attribute ID, empty `ArrayIndices`, nil value columns, and `spelling`; unregistered names remain silently skipped. The depth-cap checks remain at the map and slice container cases.

- Issue #572 explicitly requires a pure extraction with no behavior change.
- The #204 empty-list contract requires `[]` to round-trip as an explicit empty list and to clear list values under merge-update semantics.
- The flattening path must preserve spelling information used by `dedupeEAVRecords` to resolve dotted-key collisions.
- The #406/#571 depth-cap invariant requires both recursive container checks to remain in `flattenToAttributes` and `NormalizeDottedKeys` to prevent cyclic or excessively deep payloads from exhausting the stack.
- `coding-standard.md` caps source files at 500 lines and functions at 100 lines; this change reduces `flattenToAttributes` below that limit without altering adjacent behavior.

#### Findings

The extracted helper preserves the original lookup, `ValueTypeList` condition, marker construction, spelling calculation, and early-return behavior. Its call site is reached only after the existing slice depth check, and no other caller or contract is changed.

#### Observations

The existing tests already cover the relevant observable contract, including marker shape, empty-list round trips, dotted-key deduplication, and depth boundaries. No additional non-blocking observation requires action for this refactor.

Relevant existing coverage is present in `internal/transform/attribute_converter_list_test.go`, `internal/transform/eav_dedupe_test.go`, and `internal/transform/payload_depth_test.go`. These tests would detect changes to marker emission, spelling-based deduplication, round-trip behavior, and depth handling.

Executed checks:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform/...` - passed.
- `./scripts/test_with_container_runtime.sh` - passed; the full unit-test suite completed successfully using the configured rootless Podman socket.
- `make fmt-check` - passed.
- `make lint` - passed for both the root and infra modules.
- `gofmt -d internal/transform/transformer.go` - no differences.
- `git diff --check` - passed.

A direct `make test` invocation before runtime setup failed in the harness initialization because no Docker-compatible runtime endpoint was configured; rerunning through the repository setup script passed all packages. At review time, the GitHub `Test`, `Lint`, `Benchmark Smoke`, and `k6 Smoke` checks were passing; `E2E Smoke Tests` was still pending.

#### Overall Assessment

The implementation is a minimal, behavior-preserving extraction aligned with the linked issue and repository size constraints. The affected execution paths and existing regression tests support the claim that the empty-list marker contract, deduplication semantics, and depth guards remain unchanged.

#### Summary

- Findings: 0 Blocker, 0 High, 0 Medium, 0 Low.
- Major risks: none identified.
- Verification: targeted tests, full configured unit suite, formatting, and lint passed; one GitHub E2E check remained pending at review time.

PR #581, opened 2026-09-17
transform: place encoded main-column values by column type, in parity with checkBoundColumnFit (#559)

### LINKED ISSUE #559: transform: storeWithEncoding picks the slot from the encoding alone while checkBoundColumnFit keys on ColumnType()
Deferred from PR #556 (#459); the PR body lists it under "Deliberate exclusions / follow-ups" and the review records it as observation O6.

## Problem

`transform.storeWithEncoding` (`internal/transform/persistent_record.go`) picks the `PersistentRecord` slot from the encoding alone:

- `unix_ms` → `Int64Items`
- `bool_smallint` → `Int16Items`
- `bool_text` / `iso8601` → `TextItems`

It never consults `binding.ColumnType()`. `transform.checkBoundColumnFit` (`internal/transform/storage_fit.go`) does consult it: the width check for `unix_ms`/`bool_smallint` is keyed on the column's type, so a `date` with `unix_ms` bound to `integer_01` is refused as out of range for an integer column, while the store — had it been reached — would have written the value into `Int64Items["integer_01"]` and the main-table writer would have rejected the key loudly (`unsupported column`).

Today this is a slot-placement inconsistency, not a fidelity hole: the registration matrix (`schemameta.ValidateColumnBinding`) admits `unix_ms` only on bigint columns and `bool_smallint` only on smallint columns, and the writer refuses an unknown key. But the two dispatches — the check in `checkBoundColumnFit` and the store in `storeWithEncoding`/`storeWithDefaultEncoding` — are meant to mirror each other exactly (that mirror is what makes the "no value to store" error in `storeInMainColumn` unreachable), and one of them looks at the column type while the other does not.

## Expected

`storeWithEncoding` selects the slot from `(encoding, ColumnType())` the same way `checkBoundColumnFit` does, so an encoded value lands in the slot of the column it is bound to (or errors) rather than in the slot the encoding conventionally implies. A table test over every `(encoding, column type)` pair asserting that the check admits a value iff the store writes it into the column's slot would pin the parity.

## Notes

- `persistent_record.go` is at 488/500 lines; the split tracked in #558 should land first or together.

iceboundrock:
Closes #559. Plan and rationale: https://github.com/Lychee-Technology/forma/issues/559#issuecomment-5710679011.

## Problem

`storeWithEncoding` picked the `PersistentRecord` map from the encoding alone (`unix_ms` → `Int64Items`, `bool_smallint` → `Int16Items`, `bool_text`/`iso8601` → `TextItems`) while `checkBoundColumnFit` keyed on `ColumnType()`. The two dispatches are meant to be one rule written twice (that mirror is what makes the "no value to store" error unreachable), and they disagreed in both directions: the check admitted `iso8601`→`bigint_01` (nothing width-checks a text rendering) and `unix_ms`→`text_01` (the numeric slot is populated), and the store would have written `TextItems["bigint_01"]` / `Int64Items["text_01"]`, keys only the main-table writer catches as a redacted 500. Registration refuses every such pair, so no shipped schema was affected; the funnel is documented as not depending on registration.

## Rule

An explicit encoding names a rendering of the EAVRecord's numeric slot; the column type names the map it lands in:

| encoding | renders | column types admitted | slot |
|---|---|---|---|
| `unix_ms` | epoch millis (exact sidecar where the column keeps it) | smallint, integer, bigint, double | the column's numeric map, width-checked as before |
| `bool_smallint` | 0/1 | smallint, integer, bigint, double | same |
| `bool_text` | `"0"`/`"1"` | text | `TextItems` |
| `iso8601` | RFC3339 | text | `TextItems` |
| default | the slot the column consumes | unchanged | unchanged |

Any other pair is refused by `checkBoundColumnFit` as published invalid input (`datetime value cannot be stored in main column bigint_01 (bigint) with encoding iso8601, which renders a text value: the column binding's encoding and column type disagree`), and `storeWithEncoding` reports a funnel bypass naming the encoding and column rather than claiming an empty slot.

Admitting `double` for the integer renderings was the one judgment call: it keeps `unix_ms`/`bool_smallint` exactly parallel to the default-encoding numeric family (same column set, same width rule) and the value stores and reads back losslessly. Registration still refuses the pair, so nothing user-visible changes.

## Changes

- `internal/transform/persistent_record.go`: `storeNumericRendering` (one placement for the default numeric arms, `unix_ms` and `bool_smallint`; bigint prefers the exact sidecar as before) and `storeTextRendering`; `numericRenderingColumn` next to them so the check consults the store's own set. The empty-numeric-slot guard stays first in each arm, so `TestStoreInMainColumn_EmptySlotIsAnError` keeps its meaning. The default bigint arm now requires `ValueNumeric` like the check always did (every sidecar producer — `populateTypedValue`, `attribute_converter`, the read side — also fills the numeric slot).
- `internal/transform/storage_fit.go`: `checkBoundColumnFit` requires a text column for `bool_text`/`iso8601` and a numeric column for `unix_ms`/`bool_smallint`; `errEncodingMismatch`; comments restated around the rendering/map rule.
- No read-side change: `readWithEncoding` already falls back to the column's own slot, and the parity test proves the round-trip.
- No docs change: `docs/schema-consistency-migration.md` describes the registration matrix, which does not move. No API/type/schema change.

## Tests

- `internal/transform/bound_slot_parity_test.go`
  - `TestBoundColumnSlotParity`: every encoding × every column type × one sample per slot family (text, uuid, numeric, date, bool; values fit every integer width so only placement decides). Asserts check admits ⇔ store writes; when written, the key is in exactly the map `ColumnType()` names; and `readFromMainColumn` reads it back (against what the rendering preserves: bool renderings collapse to 0/1, RFC3339 keeps whole seconds). RED before the fix on 39 of the 150 triples.
  - `TestStoreWithEncoding_ColumnMismatchIsAnError`: the four mismatched shapes reaching the store error with the encoding named and write nothing.
- `storage_fit_test.go` (`TestCheckStorageFit_NonNumericFamilyColumnWidth`): `iso8601`→bigint, `bool_text`→smallint, `unix_ms`→text, `bool_smallint`→uuid refused with the published message; `bool_smallint`→integer and `unix_ms`→double admitted.

## Verification

Run on `4d3eead` (`GOTOOLCHAIN=go1.26.0+auto`, `GOTMPDIR` in `.gocache/tmp`):

- `make fmt-check` clean; `make lint` (root + infra) exit 0.
- `go test ./internal/transform ./internal/schemameta -count=1` ok.
- `./scripts/test_with_container_runtime.sh` (rootless Podman): all 34 packages `ok`, exit 0 — includes the `internal` entity-manager round-trip tests over the transform.
- Line budget: `persistent_record.go` 314, `storage_fit.go` 219, `storage_fit_test.go` 322, new test file 166; no function near 100.

## Observation, not in scope

The `iso8601` image is `Format(time.RFC3339)` — second precision, which #555 ruled the canonical image. A caller writing `2024-01-01T00:00:00.123Z` to a datetime bound that way has the millis dropped silently while an unbound or `unix_ms`-bound datetime keeps them. Filed as a follow-up question rather than changed here.


iceboundrock:
### PR Review

#### Scope & Intent

This PR fixes the parity gap between `checkBoundColumnFit` and `storeWithEncoding`: explicit encodings now render into the `PersistentRecord` map selected by `ColumnType()`, and incompatible encoding/column-family pairs are rejected rather than reaching the main-table writer with an invalid slot. The canonical registered bindings and read behavior are intended to remain unchanged; the PR also verifies the low-level funnel for bindings that bypass registration.

- The write funnel must validate the physical destination, preserve values within the destination width, and return published invalid-input errors for caller-supplied binding/value mismatches.
- `schemameta.ValidateColumnBinding` remains the registration contract for shipped schemas; this PR does not broaden that matrix.
- The main-table writer accepts each physical column only from its corresponding `PersistentRecord` map.
- `readFromMainColumn` must round-trip values from the column-type map, including the existing specialized-encoding path and its default-column-type fallback.
- Repository standards require contextual errors, flat helpers, and source/functions within the documented size limits.

#### Findings

#### Observations

- The low-level fit/store path now intentionally admits numeric renderings such as `bool_smallint` or `unix_ms` on additional numeric column types, while schema registration still rejects those noncanonical combinations. This matches the PR and issue rationale: the funnel remains safe for programmatic or already-deployed metadata without changing the shipped registration policy.
- The existing `iso8601` encoding formats with `time.RFC3339`, so sub-second precision is truncated. This is explicitly documented in the PR as follow-up issue #582 and is not introduced by the slot-parity change.

- The new `TestBoundColumnSlotParity` covers all five encodings, all six column families, and representative text, UUID, numeric, date, and bool slot values. It verifies admission parity, exact map placement, and read-back behavior.
- `TestStoreWithEncoding_ColumnMismatchIsAnError` verifies that funnel bypasses fail without writing to a wrong map.
- Existing width, exact-bigint, family-mismatch, persistent-record round-trip, and schema-binding tests passed.
- Passed locally: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/transform ./internal/schemameta -count=1 -v`; the targeted transformer race tests; `make fmt-check`; `make lint`; and `git diff --check`.
- PR CI passed all jobs: Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke. The CI Test job exercised the full race-enabled suite.
- A local plain `make test` run could not complete only because the environment lacked the container-runtime setup expected by `internal/e2e_harness` (`$XDG_RUNTIME_DIR`); this was independently covered successfully by the PR CI jobs.

#### Overall Assessment

The implementation is consistent with the stated issue contract. The shared numeric placement helper, encoding-family validation, preserved exact-bigint handling, and parity tests address the original failure mode without changing the registration matrix or canonical production paths. No merge-blocking concerns remain from this review.

#### Summary

- Findings: 0 Blocker, 0 High, 0 Medium, 0 Low
- Major risks: none identified
- Verification: focused local tests, race-targeted tests, formatting, lint, and all PR CI jobs passed; only the local container-dependent `make test` invocation was environment-limited

Issue #549, opened 2026-09-08
schemameta: a retired ledger entry whose folded column is reserved hard-blocks registry construction with an unactionable remedy


iceboundrock:
Raised as F3 in the review of #548 (fix for #532). Not a defect in that PR — the behavior is pre-existing design for exact-case names and #548 only widens the trigger set — but the widening makes it worth its own issue.

## What happens

`validateSchemaAttributeCache` deliberately runs `sqlgen.ValidateParquetAttrColumns` over the **full** attribute cache, before `activeAttributeCache` strips retired entries (`internal/schemameta/validation.go:47-57`; the #342 contract is documented in the comment there). Retired entries are in scope on purpose: their folded columns still exist in flushed parquet files.

Two consequences follow, neither of which the current diagnostics acknowledge.

### 1. One offending schema fails the whole registry

Every caller returns on first error inside the per-schema load loop — `internal/schemameta/loader.go:81`, `loader.go:360`, `internal/schemameta/file_registry.go:192`, `file_registry.go:389`. A single legacy schema carrying a reserved-folding attribute therefore fails registry construction for the entire directory or registry table at startup, not just its own entity. Fail-closed at boot is defensible; the operator seeing it has no signal that the other N-1 schemas are fine.

### 2. For a *retired* entry the remedy is unactionable

"rename the attribute" is the remedy both messages give. For a retired ledger entry it is not a real option:

- the entry exists precisely to preserve the historical name — it is the attributeID ledger #342 introduced, and #341 orphan detection reads it to tell a #294-preserved EAV row from a genuinely orphaned one;
- the name matches a column already sitting in flushed parquet, so renaming it desynchronizes the ledger from the data it describes.

A retired `Created_At` is also the one sub-case that is not "already broken today" in the sense #532 used: `activeAttributeCache` removes it before any consumer sees it, so it produces no wrong answers on any read path — yet after #548 it hard-blocks boot.

## Why it is not just a doc fix

The guard is right to see retired entries (an intra-schema collision between a retired and an active attribute is real). What is missing is that a reserved-column hit on a *retired* entry has a different remedy from one on an active entry, and the error says nothing about which case the operator is in.

## Options

1. **Diagnostics only** — when the offending entry is `Retired`, say so and give the real remedy (the entry is a ledger record; the fix is an attributeID/name migration, not a rename). Cheapest, keeps fail-closed. `attrIDCollisionError` already has precedent for naming the retired side of a collision (`validation.go:96-106`).
[...]

PR #583, opened 2026-09-17
sqlgen: compose reservedParquetColumns from the dedup column set (#552)

### LINKED ISSUE #552: sqlgen: reservedParquetColumns restates the dedup column names a third time, leaving the registration guard off the single source
## Summary

#531 / #551 single-sourced the dedup column set across `internal/sqlgen` and
`internal/federated`: `federatedDedupColumns` is now the one definition, and
`federated.keysetRejectedColumns` derives from `sqlgen.FederatedDedupColumns()`.

A **third** copy of the same two names survives, in the same file:
`reservedParquetColumns` (`internal/sqlgen/parquet_attr_column.go:43`) restates
`"rn"` and `"source_tier_priority"` as literals, with a doc comment that names
them as "the dedup machinery".

```go
var reservedParquetColumns = map[string]struct{}{
	...
	"source_tier_priority": {},
	"rn":                   {},
	...
}

var federatedDedupColumns = map[string]struct{}{
	"rn":                   {},
	"source_tier_priority": {},
}
```

Nothing keeps the two in sync — the same hazard #531 was opened for, one layer
down.

## Why it matters

`reservedParquetColumns` is not a redundant restatement. It is the **only**
guard covering the registered-attribute path:

- `ValidateParquetAttrColumns` (`parquet_attr_column.go:190`, called from
  `internal/schemameta/validation.go:53` at registry construction,
  `internal/cdc/export_sql_builder.go:124`, and
  `internal/sqlgen/duckdb_schema_projection.go:119`) checks folded attribute
  names against `reservedParquetColumns` only. It never consults
  `federatedDedupColumns`.
- `ValidateUnregisteredParquetAttrColumn` consults both, but by construction
  only fires for attributes the schema cache does *not* know —
  `normalizeDuckPayload` gates it behind `if !hasMeta`, and its comment says so
  outright: "Registered names were already checked at registration by
  `ValidateParquetAttrColumns`."

So today the registered path is safe purely because `reservedParquetColumns`
happens to repeat the dedup names.

## Failure mode

Add a third dedup column — say `tier_rank` — to `federatedDedupColumns` without
also editing `reservedParquetColumns`:

1. The filter guard refuses it (derives from `federatedDedupColumns`). ✅
2. The keyset cursor guard refuses it (derives via `FederatedDedupColumns()`,
   #551). ✅
3. **Registration accepts an attribute that folds onto it** (`"tier rank"`,
   `"tier.rank"`, `"tier_rank"`), because `ValidateParquetAttrColumns` reads
   only `reservedParquetColumns`. ❌

That attribute then has metadata, so `normalizeDuckPayload` takes the
`hasMeta` branch and never reaches the dedup check. The generated DuckDB clause
binds against the dedup column and filters on the rank instead of the
attribute — a silent wrong answer, the #354 family, on the exact path #531 set
out to close.

The window is a future dedup column, not the current tree: `rn` and
`source_tier_priority` are present in both sets today. This is a drift hazard,
same class and same severity argument as #531.

## Proposed fix

Compose `reservedParquetColumns` from its parts at package init instead of
listing the dedup names a second time, so a column added to
`federatedDedupColumns` reaches the registration guard too. Roughly:

```go
var reservedParquetColumns = buildReservedParquetColumns()

func buildReservedParquetColumns() map[string]struct{} {
	cols := map[string]struct{}{
		"row_id": {}, "schema_id": {}, "changed_at": {}, "deleted_at": {},
		"created_at": {}, "ver_ts": {}, "deleted_ts": {},
		"attributes_json": {}, "total_records": {}, "total_pages": {}, "current_page": {},
		"ltbase_row_id": {}, "ltbase_schema_id": {},
		"ltbase_created_at": {}, "ltbase_updated_at": {}, "ltbase_deleted_at": {},
	}
	for col := range federatedDedupColumns {
		cols[col] = struct{}{}
	}
	return cols
}
```

Declaration order matters (`federatedDedupColumns` must be initialised first);
Go's package-level dependency ordering handles it, but the test below should
pin the outcome rather than trusting it.

Constraints to respect:

- The **contents** of `reservedParquetColumns` must not change in this tree.
  It gates schema registration, so adding or dropping an entry can reject
  schemas that register today. The refactor must be a pure identity on the
  current set — worth asserting explicitly against a literal expected map.
- Entries stay lower-cased: `ValidateParquetAttrColumns` looks up
  `duckdbFoldIdentifier` keys, `ValidateUnregisteredParquetAttrColumn` looks up
  `strings.ToLower` keys (that divergence is #550's subject, not this one).

## Tests

- The composed set equals the current literal set, byte for byte.
- A dedup column added to `federatedDedupColumns` is refused at registration —
  the same drive-the-guard-from-the-source shape #551 used for the cursor
  guard, so the test bites on drift rather than restating the list.

## Scope

Deliberately excluded from #551, which was reviewed and approved as a
no-behaviour-change cross-package refactor; this touches a set that gates
schema registration and deserves its own review. Independent of #550 (fold
width) and #549 (retired-ledger reserved collisions), though all three live in
the same file.

iceboundrock:
Closes #552.

## Problem

#531 / #551 made `federatedDedupColumns` the single definition of the dedup column set and derived the filter and cursor guards from it. A third copy survived in the same file: `reservedParquetColumns` restated `rn` and `source_tier_priority` as literals.

That copy is not redundant. `ValidateParquetAttrColumns` — the only guard on the registered-attribute path (called at registry construction, from the CDC export builder, and from the DuckDB schema projection) — consults `reservedParquetColumns` alone. `ValidateUnregisteredParquetAttrColumn` reads both sets, but only fires for attributes the schema cache does not know. So a column added to `federatedDedupColumns` without the matching edit here would be refused by the filter and cursor guards but accepted at registration, and the resulting attribute would take the `hasMeta` branch of `normalizeDuckPayload` and bind against the dedup rank: the #354 silent-wrong-answer class, on the path #531 set out to close.

## Change

`internal/sqlgen/parquet_attr_column.go`

- `var reservedParquetColumns = buildReservedParquetColumns(federatedDedupColumns)`: the system columns are listed once inside the builder, and the dedup set is unioned in. Go's initialisation dependency ordering places `federatedDedupColumns` first; the identity test below pins the outcome rather than relying on it.
- One deviation from the issue's sketch: the builder takes the dedup set as a **parameter** rather than closing over the package variable, so a test can prove the composition with a synthetic third column without mutating the guard every other test runs against.
- Doc comments on both sets updated to describe the composition.

No behaviour change. The contents of `reservedParquetColumns` are byte-identical to the previous literal (18 entries, lower-cased), and neither validator's lookup key changes. #550 (fold width) and #549 (retired-ledger collisions) are untouched.

## Tests (`internal/sqlgen`)

- `TestReservedParquetColumnsIsUnchanged`: the composed set equals the previous literal map — the pure-identity constraint from the issue, since this set gates schema registration.
- `TestBuildReservedParquetColumnsFollowsDedupSet`: the builder, given the dedup set plus a synthetic `tier_rank`, returns every existing reserved column plus `tier_rank`. Fails before this change (no builder).
- `TestValidateParquetAttrColumnsRejectsEveryDedupColumn`: drives the registration guard from `federatedDedupColumns` under every spelling the fold and DuckDB's case-insensitive resolution reach a column by (`col`, `COL`, `[col]`, `[COL]`, `a.b`, `a b`) — the drive-the-guard-from-the-source shape #551 used for the cursor guard.

The drift guard was verified to bite: with `tier_rank` added to `federatedDedupColumns` and the reserved set regressed to the pre-PR restated literal, the last test fails with

```
attribute "tier_rank" folds onto dedup column "tier_rank" and must be refused at registration
```

and passes with the composition in place.

## Verification

- `go test ./internal/sqlgen ./internal/federated ./internal/schemameta ./internal/cdc ./cdc`: ok
- `./scripts/test_with_container_runtime.sh` (full `make test`): 34 packages ok
- `make fmt-check`, `make lint`: clean


iceboundrock:
### PR Review

#### Scope & Intent

Reviewed PR #583, commit `ad431fd23d72fa723b1191cbf9de6978f205f2e2`, including the complete two-file diff, its parent state, the linked issue #552, affected callers, related dedup/cursor changes from #531/#551, tests, documentation, repository standards, and CI configuration.

The PR replaces the duplicated `rn` / `source_tier_priority` entries in `reservedParquetColumns` with a composition from `federatedDedupColumns`. The intended behavior is to preserve the current 18-entry reserved set exactly while ensuring future dedup columns are rejected on the registered-attribute path as well as filter and cursor paths.

- Explicit issue requirement: the current `reservedParquetColumns` contents must remain unchanged; entries are lower-cased because validator lookup keys are normalized.
- Explicit issue requirement: a future entry added to `federatedDedupColumns` must reach registration validation without a second literal list.
- Existing contract: `ValidateParquetAttrColumns` is used at schema registration, CDC export projection, and DuckDB schema projection; it must reject folded collisions with emitted system/dedup columns.
- Existing architecture: `sqlgen` owns the shared definition; `federated` may depend on `sqlgen`, not the reverse.
- Repository standards: touched Go files remain below the 500-line limit, functions remain below the 100-line limit, errors and error classes are unchanged, and formatting/lint gates must pass.

#### Findings

The initializer dependency in `var reservedParquetColumns = buildReservedParquetColumns(federatedDedupColumns)` is valid Go package initialization ordering: the referenced package variable is initialized before the dependent variable. The helper constructs a new map rather than aliasing or mutating the source map. The exact-set identity test confirms the current 18 entries are unchanged, and the synthetic-column test verifies the intended composition behavior.

#### Observations

No additional non-blocking observations materially requiring author action.

Relevant coverage is present for exact reserved-set identity, synthetic dedup composition, registration rejection of every current dedup column under bare/case/punctuation/folded spellings, existing filter/cursor behavior, and package initialization outcomes. The registration call sites remain covered through the existing package tests and full suite.

Executed locally:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/federated ./internal/schemameta ./internal/cdc ./cdc -count=1`: passed.
- The same targeted package set with `go test -race`: passed.
- Focused new tests in `internal/sqlgen`: passed.
- `go vet ./internal/sqlgen ./internal/federated`: passed.
- `make fmt-check`: passed.
- `make lint`: passed, including the infra module lint run.
- `make check-infra`: passed.
- `./scripts/test_with_container_runtime.sh`: passed; it configured the rootless Podman socket and the full `make test` completed successfully.
- `git diff --check origin/main...HEAD`: passed.

A bare `make test` run before the wrapper failed only at the Testcontainers smoke test because no Docker/Podman endpoint was configured. This was an environment setup failure, not a code failure; rerunning through the repository-prescribed wrapper passed all packages.

PR CI run `35200442015` completed successfully for Test, Lint, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

#### Overall Assessment

The implementation is correctly scoped, preserves current behavior, closes the identified registration-path drift hazard, respects package layering, and adds meaningful regression protection. No merge-blocking or follow-up correctness concerns remain in the reviewed scope.

#### Summary

Findings: 0 Blocker, 0 High, 0 Medium, 0 Low. Major risks: none identified. Verification: local targeted and race tests, full container-runtime test suite, formatting, lint, vet, infra checks, and all PR CI jobs passed.

Issue #598, opened 2026-09-19
QueryConfig.MaxRows is declared and validated but never enforced


iceboundrock:
Split out of #465, which wired every other declared limit (`Entity.MaxEntitySize`, `Performance.MaxBatchSize`, `Query.DefaultTimeout`, `Transaction.DefaultTimeout`, `DuckDB.QueryTimeout`) into an actual bound.

## Problem

`forma.QueryConfig.MaxRows` (default 10000) is validated by `Config.Validate` and never read. `Query.MaxPageSize` caps a single page, but nothing caps how deep a caller can page: `page=1000000&items_per_page=1000` addresses `OFFSET 999999000`, which Postgres and DuckDB both execute by scanning and discarding every row before it. The page-overflow guard added in #465 only refuses an offset that does not fit in an `int`; it does not bound the work a valid offset costs.

## Why it was left out of #465

Enforcing `MaxRows` is a behaviour change with a design question attached, and #465's acceptance criteria did not name it:

- Refusing `offset + items_per_page > MaxRows` with a 400 makes deep offset pagination fail loudly, which is arguably the point, but breaks any client that walks a large table by page number today.
- Clamping the offset silently repeats page N for every deeper page, which is worse.
- The keyset-paginated federated path (`internal/federated`) has no offset at all, so the limit only makes sense on the offset-based OLTP path, and the two paths would then answer the same request differently.

## Proposed direction

Refuse, do not clamp: `Query`/`CrossSchemaSearch` return `forma.InvalidInputf` when `(page-1)*items_per_page + items_per_page > MaxRows`, with a published message that names the limit and points at cursor pagination for deeper walks. `pageOffset` in `internal/entity_request_limits.go` is the single place to add the check. The default of 10000 should be revisited at the same time (the bun E2E suite and `cmd/sample` should be checked for deep-page assumptions), and `MAX_QUERY_ROWS` would join the `ApplyLimitsFromEnv` overlay.

## Acceptance criteria

- A query whose page window ends past `Query.MaxRows` is refused with a 400 whose body names the limit, before any repository call.
- `Query.MaxRows <= 0` (library embedders with a zero-value config) leaves pagination unbounded, matching the zero-means-unbounded convention #465 set for the other limits.
- Regression tests for both, and the README/`docs/error-handling.md` request-limits section updated.

iceboundrock:
[...]

PR #599, opened 2026-09-19
Enforce request-body, batch, timeout and page-offset limits (#465)

### LINKED ISSUE #465: bug(httpapi): no request-body/batch/timeout limits and four declared config limits are inert — memory-exhaustion and head-of-line-blocking DoS
## Problem

The HTTP surface has no request-body size cap, no batch-size cap, and no server timeouts, and four limits that *are* declared, defaulted, and validated in config are never read by the code that would enforce them. A single request can exhaust memory or head-of-line-block the whole federated route.

## Current behavior

- `readJSONBody` (`internal/httpapi/server.go`) wraps `r.Body` in `json.NewDecoder` with no `http.MaxBytesReader` (zero occurrences repo-wide).
- `cmd/server/main.go` constructs `&http.Server{Addr, Handler}` only — no `ReadTimeout`, `ReadHeaderTimeout`, `WriteTimeout`, `IdleTimeout`, `MaxHeaderBytes` (zero occurrences repo-wide) → slowloris exposure.
- `validateBatchOperation` (`internal/entity_batch_service.go`) is a nil-check with no `len(Operations)` cap; `handleCreate`/`handleDelete` build one `Atomic:true` batch from the whole array.
- Inert config (declared + defaulted + validated in `config.go`, read nowhere else — grep-confirmed): `EntityConfig.MaxEntitySize` (1 MB), `PerformanceConfig.MaxBatchSize` (1000), `QueryConfig.MaxRows` (10000), `QueryConfig.DefaultTimeout` / `TransactionConfig`.
- No Postgres `statement_timeout` (`internal/bootstrap/postgres.go` sets only `ConnConfig.ConnectTimeout`).
- `DuckDBConfig.QueryTimeout` (30 s) is never enforced: copied into `RoutingDecision.QueryTimeout` (`internal/federated/routing.go`) and read by nobody; `DuckDBClientQueryExecutor.Query` passes the request ctx straight to `QueryContext`. Only the CDC exporter honours its own `QueryTimeout`. Because DuckDB runs `SetMaxOpenConns(1)` (`internal/federated/duckdb_conn.go`), one unbounded federated scan head-of-line-blocks *every* federated reader.

Minor, same area: `Offset: (req.Page - 1) * req.ItemsPerPage` (`internal/entity_query_service.go`) clamps `Page` from below only; a large `page` in the `advanced_query` body overflows `int` to a negative offset → opaque 500.

## Reproduction

`POST /api/v1/{schema}` with a multi-GB JSON array (whole body + decoded `[]map[string]any` + N `EntityOperation`s resident before any check); a slowloris connection (no read timeout); a heavy federated scan (no query timeout, single DuckDB conn → all federated readers stall).

## Root cause

The limits were designed into the config surface but never wired to the paths that need them.

## Impact

Memory-exhaustion DoS; one giant atomic transaction pinning a pgx connection and holding Postgres locks; slowloris connection exhaustion; unbounded federated scan blocking all federated reads with no server-side ceiling. All from a single caller with no privilege.

- Same "declared-but-inert config" pattern as #423 (telemetry emitters never registered), different subsystem — INDEPENDENT. Searched `MaxBytesReader`, `ReadTimeout`, `batch size`, `MaxEntitySize`, `QueryTimeout`, `DoS`, `slowloris`, `inert config`. NEW.

## Acceptance criteria

- [ ] Request bodies are bounded by `MaxEntitySize` via `http.MaxBytesReader` (or equivalent) before full decode.
- [ ] Batch operations are bounded by `MaxBatchSize`.
- [ ] `http.Server` sets read/write/idle/header timeouts.
- [ ] `DuckDBConfig.QueryTimeout` (and a Postgres `statement_timeout`/`QueryConfig.DefaultTimeout`) are actually enforced on the query context.
- [ ] `page`/offset arithmetic cannot overflow to a negative offset.

- [ ] Oversized body / oversized batch rejected with a 4xx before full materialization.
- [ ] A query exceeding `QueryTimeout` is cancelled.

iceboundrock:
Closes #465. Follow-up for the one limit left unenforced: #598 (`QueryConfig.MaxRows`).

## Problem

Every request limit `forma.Config` declares was validated and never read, and the HTTP surface accepted unbounded work: `readJSONBody` decoded whole bodies, batch endpoints forwarded any number of operations, `cmd/server` built an `http.Server` with no timeouts, and `page * items_per_page` could overflow to a negative `OFFSET`.

## Change

| Limit | Where enforced | Answer when exceeded |
|---|---|---|
| `Entity.MaxEntitySize` (1 MiB) as HTTP body cap | `httpapi.readJSONBody` via `http.MaxBytesReader` under `Options.MaxBodyBytes`; the capped stream is always read to its terminal result before answering, after a successful decode and after a refused one alike | 413 `request body too large: request body exceeds N bytes`, before the decoder materializes the body; a malformed prefix under the cap followed by bytes past it is still 413 |
| body framing | `httpapi.bodyReader` tags every body read failure; transport failures (`net.Error`, `io.ErrUnexpectedEOF`) are told apart from net/http refusing to frame the body | malformed chunked encoding answers a disclosed 400 `read request body: malformed request body`, net/http's prose stays on the log line; a read timeout stays 408, a disconnect stays a redacted 500 |
| `Performance.MaxBatchSize` (1000) | `validateBatchOperation` at the manager, all three batch entry points, atomic and best-effort | 400 `batch of N operations exceeds the maximum batch size of M`, before any repository call |
| `Query.DefaultTimeout` (30s) | `Get`, `Query`, `CrossSchemaSearch` run under `context.WithTimeout` | 504 `request timed out` (redacted `timeout` error class) |
| `Transaction.DefaultTimeout` (30s) | `Create`, `Update`, `Delete`, and each atomic batch; a best-effort batch is bounded per operation through the CRUD service | 504 as above |
| `DuckDB.QueryTimeout` (30s) | armed once per request in the engine's `Query` (`queryDuckDBRouted`): the page pass, a corrupt-parquet retry, the deep-page recount and a degraded Postgres fallback share one deadline; `StreamDuckDBFederatedQuery` arms the same budget per pass for direct callers, nested under the request's | 504; an expired budget is not degraded to Postgres (repeated timeouts still open the breaker, and *that* rejection degrades) |
| page offset overflow | `pageOffset` in `Query`, `CrossSchemaSearch` validation and `fetchCrossSchemaResults` | 400 `page N with M items per page addresses an offset beyond the supported range` |
| `http.Server` phases | `bootstrap.NewHTTPServer`: read-header 10s, read 30s, write 60s, idle 120s, max header 1 MiB, `HTTP_*_SECONDS` / `HTTP_MAX_HEADER_BYTES` overrides | connection closed by net/http |

Design points, also recorded on #465:

- The budgets are real context deadlines, not a Postgres `statement_timeout`: pgx cancels the running statement on expiry, duckdb-go v2 interrupts the running query, and the same deadline covers the Go-side work around the statements. The pool constructor is shared with the CDC/compaction tools, so a session-level setting there would have leaked into them.
- Zero means unbounded for every timeout and for the batch cap, so a library embedder with a zero-value config keeps the pre-#465 behaviour instead of getting an instant timeout or an empty batch.
- `context.DeadlineExceeded` maps to 504 because handler contexts carry no deadline of their own, so a deadline in the chain is always one the server set. `context.Canceled` (client gone) stays 500; no one reads that status.
- `bootstrap.ApplyLimitsFromEnv` overlays `MAX_ENTITY_SIZE_BYTES`, `MAX_BATCH_SIZE`, `QUERY_TIMEOUT_SECONDS`, `TRANSACTION_TIMEOUT_SECONDS`, `DUCKDB_QUERY_TIMEOUT_SECONDS`; `cmd/server` and `cmd/lambda` both apply it, and `config.Validate` still runs afterwards. Lambda has no `http.Server`, API Gateway bounds the connection.

## Tests (all red before the fix)

- `internal/httpapi/body_limit_test.go`: oversized body on create/update/delete/advanced_query answers 413 with a parse-probe manager that records no call; a malformed prefix under the cap followed by bytes past it is 413 while the same malformed body under the cap keeps its 400; a chunked POST with a non-hexadecimal chunk-size line over a real listener answers 400 without reaching the manager, and the transport-failure table covers timeout (408), reset and truncated body (500) and framing (400); a body under the limit reaches the manager; the default cap is 1 MiB; `classifyManagerError(context.DeadlineExceeded)` is 504 and the HTTP body is the redacted `request timed out` with `error_class: timeout` and an `error_id`, `context.Canceled` stays 500.
- `internal/entity_request_limits_test.go`: 3 operations against `MaxBatchSize: 2` are refused on Create/Update/Delete, atomic and best-effort, with every repository counter at zero, while 2 succeed; zero cap is unbounded; `pageOffset` overflow cases; `Query` and `CrossSchemaSearch` with `Page: math.MaxInt` are refused before the repository sees a query; a repository that blocks on `ctx.Done` makes `Query` return `context.DeadlineExceeded` within `Query.DefaultTimeout`; zero budgets leave the context unbounded, defaults are 30s/30s/1000.
- `internal/federated/query_timeout_test.go`: a DuckDB executor that blocks on `ctx.Done` makes `StreamDuckDBFederatedQuery` return `context.DeadlineExceeded` within `QueryTimeout`, observed a deadline, and was called once (no degrade retry); a zero timeout leaves the caller's context alone; a deep page whose empty scan triggers the recount runs the recount under the very deadline the page pass ran under (one budget per request, not per pass).
- `internal/bootstrap/limits_test.go`: server defaults bound every phase and `WriteTimeout` exceeds the default query/transaction budgets; `HTTPServerConfigFromEnv` and `ApplyLimitsFromEnv` overrides, unset and unparsable values.

## Verification

- `./scripts/test_with_container_runtime.sh` (unit + e2e smoke): exit 0.
- `make fmt-check`, `make lint`, `go build ./...`: clean.
- Existing callers checked against the new caps: the bun E2E generator posts 500-record batches of a few hundred bytes each (under 1000 ops and 1 MiB); the Go harness seeds create rows one at a time; `cmd/sample` uses batches of 100.

## Docs

README env-var table gains the ten new variables; `docs/error-handling.md` gains the 504 mapping and a "Request limits (#465)" section covering 413, the batch and page refusals, the budgets, and the still-unenforced `MaxRows`.


iceboundrock:
## Scope & Intent

Reviewed PR #599 at head `48fe45ec9480023a8740951463dd7a5609de9909` against `main` at `96455804c36cd12dcb004ecb2116f9b7466dbdb7`, with issue #465 as the governing scope. The change adds HTTP body and batch caps, page-offset overflow protection, query/transaction/DuckDB budgets, connection-level HTTP bounds, server/Lambda environment overlays, and timeout response semantics.

## Specification / Standards

- Checked the repository `AGENTS.md`, `coding-standard.md`, `docs/error-handling.md`, the issue contract, the complete source and test diff, and all affected call paths.
- The relevant error contract requires `errors.Is`/`errors.As`, published `forma.InvalidInputf` carriers for caller-actionable write-path failures, redacted read-path failures, and `504` timeout handling without leaking operator details.
- The implementation is generally focused and preserves the existing manager-level enforcement model for library embedders, not only HTTP callers.

## Findings

### 1. [Medium] Enforce the body cap across the entire request body

**Location:** `internal/httpapi/server.go:201-205`, `Server.readJSONBody`; all four JSON handlers in `internal/httpapi/handlers.go` use this helper.

`readJSONBody` wraps the request in `http.MaxBytesReader`, but calls `json.Decoder.Decode` only once. `Decode` is allowed to return after the first complete JSON value, so it does not necessarily read the rest of the request body. Consequently, a request containing a valid first JSON value followed by oversized trailing whitespace or another value can exceed `MaxBodyBytes` while returning no decode error and proceeding to the handler.

This violates the documented contract that the cap applies to every HTTP request body. I reproduced it with a 64-byte cap and valid JSON followed by 1,000 trailing spaces: `readJSONBody` returned `nil`. The existing oversized-body coverage does not exercise a body whose first JSON value completes before the cap is crossed.

The helper should consume the capped stream through EOF after the first decode, reject a second JSON value as invalid input, and preserve `*http.MaxBytesError` as the `413` path. Add a regression test for trailing whitespace/data beyond the cap.

### 2. [Medium] Reject invalid negative limit and timeout configuration before startup

**Location:** `internal/bootstrap/limits.go:21-29`, `internal/bootstrap/httpserver.go:44-51`, `config.go:488-514`, with startup wiring in `cmd/server/main.go:214-217` and `cmd/lambda/main.go:154-157`.

`ApplyLimitsFromEnv` and `HTTPServerConfigFromEnv` use permissive integer parsing, but the claimed post-overlay validation does not occur in the production entry points. `Config.Validate` does not validate `Entity.MaxEntitySize`, the three configured budgets, or HTTP server durations, and `factory.newEntityManagerWithConfigContext` does not call `Config.Validate`.

As a result, invalid negative values are silently accepted. For example, `MAX_ENTITY_SIZE_BYTES=-1` is accepted and `bodyLimit` silently falls back to the default; `MAX_BATCH_SIZE=-1` is accepted and the service treats the batch cap as unlimited; negative query, transaction, or DuckDB budgets are treated by `withBudget` as unlimited; and negative `HTTP_*_TIMEOUT_SECONDS` values are passed to `net/http`, where they disable the corresponding timeout. This can silently remove protections while contradicting the documented fail-at-boot behavior.

Validate all overlaid values before opening the database or starting the server, with explicit rules for positive size/batch caps and nonnegative durations. Apply the same validation in both server and Lambda startup paths, and add regression tests for invalid environment values and the intended zero-value semantics.

## Observations

- The default `WriteTimeout` is 60 seconds while the default query and transaction budgets are 30 seconds. Custom budgets must remain below the HTTP write timeout; the README documents this relationship, but it is not cross-validated.
- Best-effort batch operations apply the write budget per operation and convert operation failures into `OperationError` entries rather than necessarily returning a top-level `504`. That is consistent with the implementation comments, but the distinction should remain explicit in the contract and tests.
- `QueryConfig.MaxRows` remains declared but unenforced, and the PR explicitly tracks it separately from #465. I am not treating that as a finding against this PR.

## Test / Verification Assessment

Passed:

- `go test ./internal/httpapi ./internal/bootstrap ./internal/federated ./internal -count=1`
- DuckDB driver interrupt test
- `make fmt-check`
- `make lint`
- `git diff --check`
- PR checks: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke
- Temporary focused tests reproduced both findings and were removed afterward; the worktree is clean.

Environment-limited:

- `go test ./internal/... -count=1` reached the `internal/e2e_harness` package but failed during `gcc` linking.
- `go build ./...` failed while linking `cmd/lambda` through `gcc`.

These linker failures prevented full local completion, but the affected unit packages and all PR CI checks passed.

## Overall Assessment

Request changes. The implementation is directionally aligned with #465 and the core manager timeout/batch/page protections are well covered, but the body cap currently protects only bytes consumed by the first JSON decode, and invalid environment values can silently disable or replace configured protections instead of failing at startup. Both issues need fixes and regression tests before approval.

## Summary

- Fix complete-body enforcement for `MaxBodyBytes`.
- Validate all limit, budget, and HTTP timeout environment overlays before startup.
- Add regressions for trailing body data and invalid negative configuration values.

iceboundrock:
Addressed the review at f51ef3d. Both findings were confirmed against the code before changing anything; the observation on `WriteTimeout` is now enforced too.

## Finding 1: body cap covered only the first JSON value — fixed

**Root cause.** `readJSONBody` called `json.Decoder.Decode` once and returned. `Decode` stops at the end of the first value, so nothing after it was ever read through the `http.MaxBytesReader`, and a valid value followed by trailing bytes past the cap (or a second value) reached the handler.

**Fix** (`internal/httpapi/server.go`). After the decode the helper drains the capped stream with `dec.Token()`:

- `io.EOF` → the body is exactly one value plus optional whitespace: accepted.
- `*http.MaxBytesError` → trailing bytes crossed the cap: `413`, same published message as before.
- a token → a second JSON value under the cap: `400 invalid json body: unexpected data after the JSON body`. This was previously silently discarded, which is the same class of bug (the cap and the parse both claimed to cover the whole body).
- a syntax error (`{}x`) → `400` with encoding/json's prose, as before.

`errTrailingBodyValue` is a plain error so `respondBodyError` publishes it exactly as it publishes encoding/json's own messages; the #360 contract for the helper is unchanged.

**Tests** (`internal/httpapi/body_limit_test.go`), red on the previous helper: `TestBodyCapCoversBytesAfterTheFirstValue` (64-byte cap, small valid value plus 1000 trailing spaces, with and without a second value → `413`, manager never called); `TestSecondJSONValueUnderTheCapIsInvalidInput` (second object, second scalar, trailing garbage → `400`; trailing newline → `200` and the manager receives the request).

## Finding 2: negative limits silently accepted — fixed

**Root cause.** The `ApplyLimitsFromEnv` doc comment claimed `config.Validate` ran after it in every entry point. It did not: neither `cmd/server`, `cmd/lambda`, nor the factory calls `Validate`, and `Validate` had no rule for `Entity.MaxEntitySize` or the query/transaction budgets. So a negative overlay flowed through `bodyLimit`'s default fallback or `withBudget`'s "non-positive means unbounded" branch, and negative `HTTP_*` values reached `net/http`.

**Where the fix lives.** I put the range rules in `Config.Validate` rather than inside the overlay, because `Validate` is already the authority for config invariants (it already covers `duckdb.queryTimeout >= 0` and `maxBatchSize >= batchSize > 0`, which is why `MAX_BATCH_SIZE=-1` and `=0` need no new rule), and both entry points needed the whole-config check anyway.

- `config_validate.go` (moved out of `config.go`, which was already over the 500-line cap on main): `Validate` gains `validateRequestLimits`: `entity.maxEntitySize > 0`, `query.defaultTimeout >= 0`, `transaction.defaultTimeout >= 0`. Zero budgets stay legal and mean unbounded, as documented.
- `internal/bootstrap/httpserver.go`: `HTTPServerConfig.Validate(budgets)` refuses a negative phase timeout or `MaxHeaderBytes` (net/http reads both as "no bound"/"default") and, when `WriteTimeout` is bounded, a `WriteTimeout` shorter than the query or transaction budget. This is the cross-validation from the observation; an unbounded budget is left to the write timeout and needs no check.
- `cmd/server/main.go`: `serverConfigFromEnv` assembles the entire `forma.Config` and the HTTP bounds from the environment, runs both validations, and does no I/O; `bootstrapServer` calls it before `NewPostgresPoolFromConfigContext`, extending the existing "validated before any I/O" invariant that previously covered only the DuckDB manifest surface (those two explicit DuckDB checks are subsumed by `Validate`, which already calls them). The registry is attached once the pool exists.
- `cmd/lambda/main.go`: `lambdaFormaConfig` now builds, validates and returns an error before the pool is opened, same shape.
- `ApplyLimitsFromEnv`'s doc now states the real contract: it only parses; the range rules are `Validate`'s, and both entry points run it before the pool.

**Tests.** `config_test.go`: `TestConfigValidationRequestLimits` covers zero and negative entity size, negative query/transaction/DuckDB budgets and negative batch cap (each naming its field), and that all-zero budgets pass. `internal/bootstrap/limits_test.go`: `TestHTTPServerConfigValidate` (each negative phase, negative header cap, all-zero config passes, short `WriteTimeout` vs query budget, vs transaction budget alone, unbounded budgets or unbounded write timeout pass, nil budgets skip the cross-check); `TestNegativeLimitOverlayFailsValidation` (each of the five `*_SECONDS`/`MAX_*` variables set to `-1` is refused by `Validate` on the overlaid config, zero budgets pass).

**Binary check.** Built `cmd/server` and started it with `DB_PORT=1` (no database):

| env | result |
|---|---|
| `MAX_BATCH_SIZE=-1` | `invalid configuration: ... 'performance.maxBatchSize': must be greater than or equal to batchSize` |
| `MAX_ENTITY_SIZE_BYTES=0` | `... 'entity.maxEntitySize': must be greater than 0` |
| `QUERY_TIMEOUT_SECONDS=-5` | `... 'query.defaultTimeout': must be greater than or equal to 0` |
| `HTTP_WRITE_TIMEOUT_SECONDS=10` | `invalid http server configuration: ... 'http.writeTimeout': 10s is shorter than the query budget 30s` |
| `HTTP_IDLE_TIMEOUT_SECONDS=-1` | `... 'http.idleTimeout': must be greater than or equal to 0` |
| `QUERY_TIMEOUT_SECONDS=0` | passes validation, fails at `failed to create database pool` (proves the ordering) |

## Observations

- **`WriteTimeout` vs budgets**: now cross-validated at boot, see above. README line for `HTTP_WRITE_TIMEOUT_SECONDS` updated.
- **Best-effort batch and `504`**: unchanged and still explicit in `docs/error-handling.md` ("a best-effort batch is bounded per operation") and in the `requestBudgets.write` comment; per-operation failures stay `OperationError` entries.
- **`MaxRows`**: still #598.

## Not changed, and why

The finding mentions "permissive integer parsing". I kept the set-but-unparsable case on the package convention (`QUERY_TIMEOUT_SECONDS=30s` still keeps the default): every `EnvInt` caller in `internal/bootstrap` behaves that way, `TestHTTPServerConfigFromEnv` pins it, and changing it only for these five variables would leave the package inconsistent. Filed #600 for making a set-but-unparsable integer fail boot across all overlays.

## Docs

README: body cap covers trailing bytes, write timeout is checked at boot, new paragraph on boot-time validation. `docs/error-handling.md`: body-cap entry describes the drain and the second-value `400`; new "Boot-time validation" bullet.

## Validation run

- `./scripts/test_with_container_runtime.sh` (unit + e2e smoke, rootless Podman): exit 0.
- `make lint`: clean (after splitting the extended `TestConfigValidationDetailed`, which funlen flagged at 128 lines).
- `make fmt-check`, `go build ./...`, `go vet ./cmd/... .`: clean.
- Not run: the full `-tags=e2e` federated suite, `test-e2e-production`, and the Bun suite. None of the Go harnesses launch the `cmd/server` binary, so the binary check above is the end-to-end coverage of the restructured boot path; the Bun suite starts the server via `scripts/local_server.sh` with default limits and is exercised by CI's k6 smoke.

iceboundrock:
### PR Review

#### Scope & Intent

PR #599 enforces the request limits identified in issue #465: JSON body and batch caps, page-offset overflow protection, manager query/transaction deadlines, DuckDB query deadlines, HTTP connection bounds, environment overlays, startup validation, and `504` timeout responses. The intended behavior is to bound untrusted work without changing the zero-value behavior of library embedders, while preserving Forma's existing published-versus-redacted error contract.

The review covers the final PR head `f51ef3d`, both commits, the complete diff, affected manager/repository/federated paths, HTTP handlers, startup wiring, tests, documentation, and the linked issue/specification.

- Issue #465 acceptance criteria require body and batch limits before unbounded materialization/work, HTTP server timeouts, enforced query/DuckDB deadlines, and safe page-offset arithmetic.
- The repository error contract requires `errors.Is`/`errors.As`, `forma.InvalidInputf` carriers for caller-actionable 4xx errors, redaction of operator detail, and the documented `504` timeout shape.
- The new body documentation promises that the cap covers the complete body and that an over-limit body answers `413`.
- `net/http` semantics are load-bearing here: `ReadTimeout` covers the entire request including the body, while `WriteTimeout` is installed after request headers are read and before the handler reads the body.
- Repository standards also require focused tests for observable behavior and keep the changed code within the documented file/function-size limits.

#### Findings

**[Medium] Continue enforcing the body cap after detecting a second JSON token**

- **Location:** `internal/httpapi/server.go / Server.readJSONBody`, especially lines 214-221
- **Issue:** After the first decode, the helper calls `dec.Token()` once. If that sees a second JSON value, the helper immediately returns `errTrailingBodyValue` and does not consume the rest of the capped stream.
- **Impact:** A body whose second value begins within the limit but whose later bytes cross the limit is answered as `400 invalid json body` instead of the documented `413`. For example, a valid first value followed by ` {` and then 1,000 spaces with a cap ending at the `{` returns `400`; the remaining bytes are never used to select the over-limit response.
- **Evidence:** A focused repro against the final code returned `400` with `unexpected data after the JSON body` for that input. The implementation's `dec.Token()` call can observe the second token without reaching the later `MaxBytesError`, despite the documentation stating that the reader is drained to EOF.
- **Recommended direction:** Consume all remaining bytes through the same capped reader before choosing the final error, retaining the second-value diagnosis but prioritizing `*http.MaxBytesError` when the cap is crossed. Add a regression for a second token followed by over-limit data.

**[High] Account for body-read time in the HTTP write deadline**

- **Location:** `internal/bootstrap/httpserver.go / HTTPServerConfig.Validate` and `DefaultHTTPServerConfig`; `internal/httpapi/server.go / Server.readJSONBody`
- **Issue:** Validation only compares `WriteTimeout` with the manager query and transaction budgets, and permits equality. In `net/http`, the write deadline starts after the request headers are read, before the handler consumes the body. It is therefore also running while `readJSONBody` reads a slow but otherwise legal upload.
- **Impact:** With the shipped defaults, a request can spend almost the 30-second `ReadTimeout` receiving its body and then almost the 30-second query/transaction budget doing work, leaving no time to serialize and write the promised response before the 60-second `WriteTimeout`. The client can receive a closed connection instead of the intended `504` response. The same issue exists for custom values that satisfy the current `WriteTimeout >= budget` check.
- **Evidence:** A real TCP-listener repro using `NewHTTPServer` delayed the body until after a 50ms `WriteTimeout`; the handler completed body reading but the response write produced no response. The existing tests use `httptest.ResponseRecorder`, which does not enforce socket deadlines, so they cannot detect this interaction.
- **Recommended direction:** Define and enforce a deadline invariant that includes the request-body/read phase and response-writing margin, or use a timeout arrangement that does not spend the response deadline during body consumption. Add a real-listener regression for a slow body followed by a budget-consuming handler; do not treat equality as sufficient for a response deadline.

**[Medium] Include an active DuckDB budget in write-timeout validation**

- **Location:** `internal/bootstrap/httpserver.go / HTTPServerConfig.Validate`, lines 85-95; `internal/bootstrap/limits.go / ApplyLimitsFromEnv`, line 30
- **Issue:** `HTTPServerConfig.Validate` checks `Query.DefaultTimeout` and `Transaction.DefaultTimeout`, but not `DuckDB.QueryTimeout`. The documented relationship that the DuckDB budget is “inside the query budget” is not true when the general query budget is zero.
- **Impact:** `QUERY_TIMEOUT_SECONDS=0`, `TRANSACTION_TIMEOUT_SECONDS=0`, `DUCKDB_QUERY_TIMEOUT_SECONDS=120`, and `HTTP_WRITE_TIMEOUT_SECONDS=60` pass startup validation. A federated pass can then run until its 120-second DuckDB deadline, but the HTTP connection's response deadline expires at 60 seconds, so the caller cannot receive the documented timeout response.
- **Evidence:** A focused validation repro with those values was accepted by the final `HTTPServerConfig.Validate` implementation. The existing test explicitly allows a short write timeout when the query and transaction budgets are zero, while the default DuckDB budget remains active.
- **Recommended direction:** Validate the effective federated deadline as part of the HTTP/write-timeout invariant, or require the DuckDB budget to be subordinate to a finite query budget. Add coverage for an unbounded general query budget with a longer active DuckDB budget.

**[Low] Do not publish read-transport timeouts as malformed JSON**

- **Location:** `internal/httpapi/body_limit.go / respondBodyError`, lines 29-37; all body-carrying handlers in `internal/httpapi/handlers.go`
- **Issue:** Every body-read error other than `*http.MaxBytesError` is wrapped in `forma.InvalidInputf` and returned as a disclosed `400 invalid json body`. That includes the `net.Error` produced when the new `http.Server.ReadTimeout` expires while the body is being uploaded.
- **Impact:** A transport timeout is classified as caller-supplied malformed JSON, causing clients to suppress retry or report a bad request. The disclosed message can also contain the network read error rather than JSON parse feedback.
- **Evidence:** A real listener with a 50ms `ReadTimeout` and a partial JSON body reached `respondBodyError` and returned an HTTP `400`. This path is not exercised by the `httptest` body cases.
- **Recommended direction:** Distinguish size errors, JSON syntax/type errors, and transport cancellation/timeouts. Let the server close the connection or map a read timeout to an appropriate transport status without publishing the network error as caller JSON feedback.

#### Observations

- `QueryConfig.MaxRows` remains intentionally unenforced and is tracked separately in issue #598; I am not treating that as a finding against this PR.
- The manager tests cover the query repository deadline and the federated tests cover one DuckDB scan deadline, but there is no equivalent behavioral test for `Get`, `Create`, `Update`, `Delete`, atomic batches, or best-effort batch timeout semantics.
- The environment overlay intentionally preserves the package-wide behavior that unparsable integer values keep defaults; that is documented and not treated as a defect here.

Relevant PR coverage is meaningful: body limits cover all four body endpoints, batch admission covers all three batch entry points and both modes, page overflow is tested through `Query` and `CrossSchemaSearch`, query and DuckDB cancellation are tested with blocking fakes, timeout classification is tested, and startup/env validation has unit coverage.

Additional verification performed:

- `git diff --check origin/main...HEAD` passed.
- `make fmt-check` passed.
- `go vet ./cmd/... .` passed.
- Focused tests for `internal/bootstrap`, `internal/federated`, and `internal` request-limit paths passed.
- Focused real-listener and body-cap repros confirmed all four findings above; temporary repro tests were removed and the worktree is clean.
- `make test` could not complete locally because the environment hit disk quota while linking CGO/DuckDB packages, and the container-based E2E harness lacked a usable container runtime environment.
- `go build ./...` was likewise blocked by the local disk quota during CGO linking.
- GitHub PR checks are green: Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

The existing tests would not fail for the second-token body-tail case, the real HTTP deadline interactions, or the read-timeout classification.

#### Overall Assessment

Request changes. The core manager-level batch, page, and context cancellation work is directionally sound and has useful focused coverage. The final implementation still has an incomplete whole-body cap path, HTTP deadline composition that can prevent legal requests from returning their promised timeout response, an uncovered DuckDB-to-HTTP deadline combination, and an incorrect transport-error classification. These should be corrected before merge, with real-connection regressions for the HTTP phase behavior.

#### Summary

- 1 High finding: `WriteTimeout` can expire during body consumption and suppress the response.
- 2 Medium findings: incomplete body-cap enforcement after a second token; DuckDB budget omitted from write-timeout validation.
- 1 Low finding: read transport timeouts are disclosed as malformed JSON.
- Major risk: the new timeout protections are not composable across HTTP body-read, manager, DuckDB, and response-write phases.
- Verification: focused package tests, formatting, vet, and all GitHub PR checks passed; full local test/build verification was limited by disk quota and unavailable container runtime.

iceboundrock:
### Response to the second review (comment 5744449243)

This is the disposition record for the four findings and the observations. All four findings are addressed in commit `66fe1ea`; nothing was deferred to a follow-up issue. There are no inline review threads on this PR, so every disposition lives here.

#### Dispositions

| Finding | Disposition |
|---|---|
| [Medium] Continue enforcing the body cap after detecting a second JSON token | `fixed` |
| [High] Account for body-read time in the HTTP write deadline | `fixed` |
| [Medium] Include an active DuckDB budget in write-timeout validation | `fixed` |
| [Low] Do not publish read-transport timeouts as malformed JSON | `fixed` |
| Observation: no behavioral timeout test for the write paths | `fixed` for `Create` and atomic `BatchCreate`; `Get`/`Update`/`Delete`/best-effort batches are still untested at the behavioral level (see "Remaining risks") |
| Observation: `MaxRows` unenforced (#598), unparsable ints keep defaults (#600) | `replied` here: both already tracked, no change |

#### Root causes

1. **Body cap after a second token.** `readJSONBody` decided the verdict at the first non-whitespace token after the value (`dec.Token()`), so the capped reader was never driven to EOF and `*http.MaxBytesError` could not fire once a second value had started under the cap. The documented contract ("the cap covers the whole body") was implemented only for the whitespace-tail case.
2. **Write deadline vs body read.** `Validate` compared `WriteTimeout` with the budgets alone and accepted equality. net/http arms the write deadline in `readRequest` (after the header read, before the handler runs), so the body upload phase is already spending it. The invariant was missing the body phase and the response margin.
3. **DuckDB budget.** `LargestRequestBudget` did not exist; `Validate` looked at `Query.DefaultTimeout` and `Transaction.DefaultTimeout` only. "DuckDB runs inside the query budget" holds when the query budget is bounded (nested `context.WithTimeout` keeps the earlier deadline) and is false when it is `0`.
4. **Transport errors as JSON feedback.** `respondBodyError` had two branches: `*http.MaxBytesError` and "everything else is `forma.InvalidInputf("%v", err)`". `json.Decoder` passes reader errors through unchanged, so a `net.Error` timeout from `ReadTimeout` took the published-400 branch.

#### Design decisions

- **One composed deadline invariant, checked at boot, instead of runtime deadline juggling.** `HTTPServerConfig.Validate(budgets)` now requires `WriteTimeout > ReadTimeout + LargestRequestBudget` (strict, so the difference is the response margin) and refuses a bounded `WriteTimeout` paired with an unbounded `ReadTimeout` (an unbounded upload can consume any write deadline). The alternative, calling `http.ResponseController.SetWriteDeadline` from the handler after the body read, would have moved the guarantee into every handler and made it depend on the body being read at all; the boot-time check matches how the rest of #465 is enforced and needs no per-request code.
- **`LargestRequestBudget` is the single definition of "how long a handler may legitimately run".** Read side: `Query.DefaultTimeout` if bounded, else `DuckDB.QueryTimeout`; then max with `Transaction.DefaultTimeout`. It is exported so a future check (cmd/lambda, an operator tool) uses the same rule.
- **Default `WriteTimeout` moves from 60s to 90s.** With the shipped 30s read timeout and 30s budgets, 60s left zero margin. 90s = 30s upload + 30s budget + 30s margin. `ReadTimeout` stays 30s.
- **Drain to EOF before choosing the verdict, cap wins.** `drainBody` reads the tail through the same capped reader, tracks whether any non-whitespace was seen, and returns the reader error first. So "second value, then over the cap" is `413`, "second value under the cap" is `400 unexpected data after the JSON body`, whitespace tail is legal.
- **Provenance-tagged transport errors.** A `bodyReader` under `http.MaxBytesReader` wraps every non-EOF read error as `*bodyReadError`. Classification is by provenance, not by sniffing error types, so a `net.Error` timeout that surfaces from inside the manager (a Postgres socket) is deliberately *not* treated as a body timeout and stays a 500 of the internal class. A body read timeout answers `408 Request Timeout` (RFC 9110 §15.5.9) on the redacted branch, `error_class: "timeout"`, fixed message `request timed out`, `Connection: close`; any other transport failure (reset, EOF mid-body) is a redacted 500, the same treatment `context.Canceled` gets.
- **Not done: mapping read timeouts via `classifyManagerError`.** Kept apart on purpose (documented on `isBodyReadTimeout`).

#### Code changes

- `internal/httpapi/body_limit.go`: `bodyReadError` + `bodyReader`, `drainBody`, `isJSONWhitespace`, `errTrailingBodyValue` moved here; `respondBodyError` grows the 408/500 branch.
- `internal/httpapi/server.go`: `readJSONBody` wraps the body in `bodyReader` inside `http.MaxBytesReader`, decodes, then `drainBody(io.MultiReader(dec.Buffered(), body))`.
- `internal/httpapi/error_response.go`: `errorClass` maps `isBodyReadTimeout` to the timeout class.
- `internal/bootstrap/httpserver.go`: type comment on phase composition; default `WriteTimeout` 90s; `Validate` composed invariant; new exported `LargestRequestBudget`.
- `internal/entity_manager_mocks_test.go`: `insertFunc` hook on the mock repository.
- `cmd/server/main.go`, `README.md`, `docs/error-handling.md`: comments and docs updated to the new rule, the 408, and the 90s default.

#### Tests added or modified

- `internal/httpapi/body_limit_test.go`
  - `TestBodyCapCoversBytesAfterTheFirstValue`: new case "second value then over the cap" → 413 (the reviewer's repro input).
  - `TestSecondJSONValueUnderTheCapIsInvalidInput`: trailing garbage and a form feed → 400 `unexpected data after the JSON body`.
  - `TestBodyReadFailureIsNotMalformedJSON`: a `net.OpError` deadline error from the body → 408, timeout class, `Connection: close`, no network text in the body, manager not reached; a connection reset → redacted 500.
  - `TestReadTimeoutWhileUploadingAnswers408`: real `net.Listen` + `http.Server{ReadTimeout: 300ms}`, raw TCP POST whose `Content-Length` exceeds what is sent → 408 with `Connection: close`.
- `internal/bootstrap/httpserver_deadline_test.go` (new): `TestWriteDeadlineStartsBeforeTheBodyIsRead`, a real listener through `NewHTTPServer`, client delays the body 400ms, handler sleeps 400ms after reading it; `WriteTimeout` 800ms yields no response, 5s yields 200. This pins the net/http semantic the invariant depends on, so a future Go release changing it fails the test rather than the invariant silently going stale.
- `internal/bootstrap/limits_test.go`
  - `TestHTTPServerConfigDefaultsBoundEveryPhase` now asserts the defaults satisfy `Validate(DefaultConfig)` and the strict inequality.
  - `TestWriteTimeoutMustCoverBodyReadAndBudget`: equality refused, +1s accepted, `ReadTimeout=0` refused naming `http.readTimeout`, transaction-only budget, unbounded budget/write/nil config accepted.
  - `TestLargestRequestBudgetIncludesDuckDB`: the reviewer's `0/0/120/60` combination is refused naming `duckdb query`; a bounded query budget hides DuckDB; the transaction budget wins when longest.
- `internal/entity_request_limits_test.go`: `TestTransactionTimeoutCancelsABlockingWrite` covers `Create` and atomic `BatchCreate` against a blocking insert with a 50ms transaction budget: the repository context carries a deadline and the call fails with `context.DeadlineExceeded`.

#### Follow-up issues

None created. `MaxRows` is #598 and unparsable integer overlays are #600, both pre-existing.

#### Validation actually run

All commands below were run on the final tree at `66fe1ea` with `GOTOOLCHAIN=go1.26.0+auto` and the Makefile's cache/tmp settings.

| Command | Result |
|---|---|
| `go test ./internal/httpapi -count=3 -race` | pass (includes the four new/extended body tests) |
| `go test ./internal/bootstrap -count=2 -race` | pass (includes the real-listener deadline test) |
| `go test ./internal -run 'TestTransactionTimeoutCancelsABlockingWrite\|TestQueryTimeoutCancelsABlockingRepository\|TestBatchOverMaxBatchSize' -race` | pass |
| `./scripts/test_with_container_runtime.sh` (rootless Podman; runs `make test`, unit suite + e2e smoke harness) | pass, 34 packages ok, exit 0 |
| `make fmt-check` | pass |
| `go vet ./cmd/... . ./internal/httpapi ./internal/bootstrap ./internal` | pass |
| `make lint` (golangci-lint v1.64.8, root + infra modules) | pass |
| `git diff --check` | clean |

Not run: the `-tags=e2e` federated suite, the production harness, and the benchmarks. None of the changed code is on those paths beyond what the smoke harness already exercises; CI will run the `-short` federated suite and the production harness on this push.

#### Remaining risks and limitations

- `Get`, `Update`, `Delete` and best-effort batches apply the same `withBudget` call but have no behavioral test; the mock repository exposes hooks only for query and insert. Adding hooks for merge/delete/get would be a small test-only change if wanted.
- The invariant assumes the handler's work is bounded by `LargestRequestBudget`. Work outside any budget (schema registry lookups, JSON encoding of a large page) is not counted; the 30s default margin covers it in practice.
- `408` is only answered when the handler is already reading the body. A `ReadTimeout` that expires before the handler is invoked is handled by net/http itself (connection closed, no response), which is unchanged.
- cmd/lambda does not build an `http.Server`, so the composed invariant is not checked there; its budgets are still validated by `forma.Config.Validate`.

iceboundrock:
### PR Review

#### Scope & Intent

This PR implements #465 across the HTTP and library entry points: it caps JSON request bodies, limits batch operations at the manager boundary, applies query/write/DuckDB context budgets, protects page-offset arithmetic, configures `http.Server` connection limits, exposes the corresponding environment overlays, and documents the resulting 400/408/413/504 behavior. It intentionally leaves `QueryConfig.MaxRows` unenforced, as recorded in the PR and #598.

I reviewed the complete three-commit diff, #465 and its decision record, affected handlers and manager services, federated retry/degradation flow, configuration/factory/Lambda wiring, public error contract, changed tests, README/error-handling documentation, repository coding standards, and CI configuration.

- **Explicit requirements:** #465 requires pre-decode body bounds, manager-level batch admission, HTTP connection timeouts, context-enforced query/DuckDB limits, and non-overflowing page offsets.
- **Repository contracts:** public caller faults use `forma.InvalidInputf` and published 4xx messages; read/infra failures remain redacted; context cancellation must reach database/federated work; library embedders retain documented zero-value semantics.
- **Established conventions:** `cmd/server` and `cmd/lambda` assemble environment-backed configuration before startup; DuckDB degradation must not return an answer that violates a query consistency or latency invariant; source files/functions remain within the project limits.
- **Language/runtime semantics:** `http.MaxBytesReader` must remain in the decode path; `http.Server.WriteTimeout` starts after headers and also covers the body-read/handler interval; `context.WithTimeout` preserves an earlier caller deadline and is observable by pgx and DuckDB.

#### Findings

**[Low] Real-listener deadline test has no reliable timing boundary**

- **Location:** `internal/bootstrap/httpserver_deadline_test.go:20-82`, `TestWriteDeadlineStartsBeforeTheBodyIsRead`
- **Issue:** The failing case sets `WriteTimeout` exactly equal to the client-side body delay plus handler sleep (400 ms + 400 ms) and requires the response to be lost. It does not synchronize the client delay with the server having read the headers and armed the write deadline.
- **Impact:** This creates a flaky regression test rather than a deterministic proof of the deadline composition rule. The test can receive a 200 when the server reads the headers after the client starts its 400 ms delay, leaving enough of the 800 ms server deadline for the handler to write.
- **Evidence:** A focused `go test ./internal/bootstrap -count=1` run failed at this assertion with `expected the write deadline to swallow the response, got 200`; the same test passed over ten subsequent runs. The timing construction itself permits both outcomes.
- **Recommended direction:** Coordinate the client delay with handler entry, or use a material timing margin, so the server has definitely armed its write deadline before the body delay begins and the failure case cannot legitimately complete.

#### Observations

- The body reader correctly drains after decoding the first JSON value, preserving the cap for trailing data and distinguishing an oversized body (413) from trailing non-whitespace under the cap (400).
- Batch admission occurs before repository work on all three batch APIs and both execution modes; the body cap still bounds HTTP-side decode allocation before manager admission.
- Read/write budgets are placed on the relevant repository/federated contexts, and DuckDB deadline expiry is deliberately non-degradable, preserving the stated latency invariant.
- Server and Lambda startup both apply and validate the limit overlays before opening their pools; the server-only transport deadlines are correctly excluded from Lambda.

The added tests cover the important observable behaviors: each body endpoint and trailing-body handling, transport read errors, batch rejection before repository calls, page overflow rejection, manager budget cancellation, DuckDB budget cancellation/non-degradation, environment overlays, configuration validation, and HTTP deadline composition. The deadline-composition test needs the reliability correction described above.

Commands executed:

- `git diff --check origin/main...HEAD` - passed.
- `make fmt-check` - passed.
- `make lint` - passed.
- `go test ./internal/httpapi -count=1` - passed.
- `go test ./internal/bootstrap -run TestWriteDeadlineStartsBeforeTheBodyIsRead -count=10` - passed after the one focused full-package failure described in the finding.
- `go test ./internal -run 'Test(BatchOverMaxBatchSizeIsRefusedBeforeTheRepository|ZeroMaxBatchSizeLeavesBatchesUnbounded|PageOffsetRefusesOverflow|QueryWithOverflowingPageIsRefusedBeforeTheRepository|QueryTimeoutCancelsABlockingRepository|TransactionTimeoutCancelsABlockingWrite)$' -count=1` - passed.
- `go test ./internal/federated -run 'Test(QueryTimeoutCancelsTheDuckDBPass|ZeroQueryTimeoutLeavesTheCallerContextAlone)$' -count=1` - passed.
- `go test ./cmd/server ./cmd/lambda -count=1` - passed.

The PR CI is also green: lint, race-enabled unit tests, E2E smoke, k6 smoke, and benchmark smoke all passed. I did not rerun the repository-wide container/E2E suite locally.

#### Overall Assessment

The implementation is coherent across the HTTP, manager, federated, configuration, and documentation layers. It satisfies the #465 acceptance criteria without bypassing the established error-disclosure or federated fallback contracts. The only issue identified is a low-severity timing race in a newly added test; it does not indicate a production correctness defect, but it can create intermittent CI noise and weaken protection for the deadline invariant.

#### Summary

Findings: **0 Blocker, 0 High, 0 Medium, 1 Low.**

Major implementation risks were not identified. Focused verification and all reported PR CI checks are successful, with the one locally observed flaky deadline-test result noted above.

iceboundrock:
### Response to the third review (comment 5744678797)

This is the disposition of the one finding, addressed at f4175e8. The review left no inline threads, so there is nothing to resolve beyond this comment.

#### [Low] Real-listener deadline test has no reliable timing boundary — fixed

**Concern as understood.** `TestWriteDeadlineStartsBeforeTheBodyIsRead` proved the deadline-composition rule with a race: the client started its 400 ms body delay when its header write returned, but net/http arms the write deadline only once it has parsed those headers (the deferred `SetWriteDeadline` in `conn.readRequest`). Under scheduling pressure the server's 800 ms budget could start later than the client's clock and leave room for the response, so the "response is lost" assertion could legitimately fail. The reviewer observed that once in a full-package run.

**Root cause.** Two independent clocks (client sleep, server deadline) with no ordering between them, and a failing case that sat exactly on the boundary, so any skew in the server's favour let the response through.

**Fix.** Both halves of the race are removed:

1. **Ordering.** The handler signals the client on entry over a channel, and the client only starts its body delay after that signal. net/http invokes the handler after the headers are parsed and the write deadline is set, so the delay can never begin before the deadline does. The wait is bounded by a 5 s guard that fails the test with a clear message if the server never reaches the handler.
2. **Margin.** The failing `WriteTimeout` is now `bodyDelay + handlerWork/2` (600 ms) instead of the exact sum (800 ms). It sits strictly between the handler's work alone (400 ms, which would fit) and the body delay plus that work (800 ms, which cannot), so the body delay counting against the deadline remains the only explanation for a lost response, and there is a 200 ms margin before the boundary. The passing case keeps a 5 s `WriteTimeout`.

With the ordering in place the handler's write happens at least `bodyDelay + handlerWork` after the deadline was armed, which exceeds the 600 ms budget by at least 200 ms regardless of scheduling, so the outcome no longer depends on timing luck. The test's comment was rewritten to state this construction.

**Verification actually run** (all with the Makefile's `GOTOOLCHAIN=go1.26.0+auto`):

- Before the fix, `go test ./internal/bootstrap -run TestWriteDeadlineStartsBeforeTheBodyIsRead -count=40` and `go test ./internal/bootstrap -count=6 -cpu=1`: 46 passes, no local reproduction of the flake. The race is established by the timing analysis above and the reviewer's observed failure, not by a local repro.
- After the fix, `go test ./internal/bootstrap -run TestWriteDeadlineStartsBeforeTheBodyIsRead -count=25`: pass.
- `go test ./internal/bootstrap -race -cpu=1 -count=4`: pass.

No production code changed, so the container/E2E suite was not rerun.

iceboundrock:
### PR Review

#### Scope & Intent

Reviewed final head `f4175e8` for #465. The PR enforces HTTP body and batch caps, offset-overflow rejection, query/write/DuckDB context budgets, HTTP connection bounds, environment overlays and startup validation, plus documented 408/413/504 error behavior. `QueryConfig.MaxRows` remains intentionally out of scope under #598.

- #465 and the PR description require request work to be bounded before materialization or execution.
- `docs/error-handling.md` and the README explicitly promise a complete-body cap with `413` taking precedence for an over-limit body, and an HTTP write deadline that leaves room to return a response after bounded work.
- Forma's error contract publishes caller-addressable 4xx failures through public carriers and redacts infrastructure failures.
- `net/http` starts `WriteTimeout` after headers and before the handler reads the body; each `context.WithTimeout` establishes its own deadline unless its parent is already earlier.

#### Findings

**[Medium] Complete-body cap is skipped after an early JSON parse failure**

- **Location:** `internal/httpapi/server.go:213-216`, `Server.readJSONBody`
- **Issue:** `readJSONBody` only drains the `MaxBytesReader` after a successful first decode. A JSON syntax error returns immediately, so later bytes never reach the cap.
- **Impact:** A request whose malformed prefix fits under the cap but whose remainder exceeds it returns `400` instead of the documented `413`. This makes the stated whole-body/cap-precedence contract dependent on JSON validity.
- **Evidence:** A focused repro with `MaxBodyBytes=64` and body ``{"schema_name" x`` followed by 1,000 spaces returned `400 invalid json body: invalid character 'x' after object key`. The cap was not observed because the return at line 214 bypasses `drainBody`.
- **Recommended direction:** Preserve the parse error for a body that fits, but consume the capped reader to its terminal result before selecting the response. If it crosses the limit, retain `*http.MaxBytesError` and return the established `413`. Add the malformed-prefix/over-limit-tail regression case.

**[Medium] HTTP write-timeout validation underestimates an unbounded-query federated request**

- **Location:** `internal/bootstrap/httpserver.go:121-137`, `LargestRequestBudget`; `internal/federated/duckdb_query.go:69-86, 217-226`; `internal/federated/engine.go:247-263`
- **Issue:** When `Query.DefaultTimeout` is zero, `LargestRequestBudget` counts exactly one `DuckDB.QueryTimeout`. A federated request can run several sequential DuckDB passes, each with a fresh timeout: `ExecuteDuckDBFederatedQuery` retries a corrupt-parquet pass once, and an empty deep page invokes `computeFederatedCount`, which calls `ExecuteDuckDBFederatedQuery` again.
- **Impact:** Valid startup configuration can still let `WriteTimeout` expire before a legal request can write its result or timeout response. For example, query and transaction budgets disabled, DuckDB budget 30s, read timeout 30s, and write timeout 61s pass `Validate`; a slow upload plus a nearly-30s deep-page scan and a nearly-30s recount exceed the 61s write deadline.
- **Evidence:** Each `StreamDuckDBFederatedQuery` creates its own `withQueryTimeout` context. The retry comment expressly says it gets its own budget, and the deep-page recount is a separate call path. The current validator and its tests model only one DuckDB budget when the outer query budget is disabled.
- **Recommended direction:** Make the write-deadline invariant account for the maximum allowed request-level federated work, or apply one request-scoped deadline across all DuckDB passes when the outer query budget is disabled. Add coverage for an unbounded query budget with a deep-page recount and retry-capable DuckDB path.

**[Low] Malformed chunk framing is reported as an internal server error**

- **Location:** `internal/httpapi/body_limit.go:55-60, 132-140`
- **Issue:** `bodyReader` labels every non-EOF body-read error as a transport failure, and `respondBodyError` maps every non-timeout instance to a redacted `500`. This includes malformed HTTP transfer framing, not only a reset/disconnected client.
- **Impact:** A client request with invalid chunked encoding is treated as a server failure, adding false 5xx signal and misclassifying caller input.
- **Evidence:** A raw chunked POST with an invalid chunk-size line (`Z\r\n`) reaches the handler and returns `500`; `net/http` produces the framing error during `r.Body.Read`, which is then wrapped as `bodyReadError`.
- **Recommended direction:** Distinguish HTTP body-framing/protocol errors from transport disconnects and return a redacted `400` for malformed framing. Keep actual read timeouts as the documented `408` and connection failures redacted.

#### Observations

- The latest synchronization change in `TestWriteDeadlineStartsBeforeTheBodyIsRead` removes the previously reported timing race. The handler-entry signal establishes the deadline ordering, and the 200ms failing-case margin is sufficient.
- Batch admission, offset checking, error redaction, startup validation, and the normal body-tail drain behavior are coherent with the stated contracts.

Relevant coverage already exercises all body endpoints, normal trailing-body behavior, manager batch admission, page arithmetic, configuration ranges, real-listener body/deadline behavior, and per-pass DuckDB cancellation. It does not cover the malformed-prefix body-cap case, malformed HTTP chunk framing, or the multi-pass DuckDB/write-deadline combination identified above.

Commands executed:

- `git diff --check origin/main...HEAD` passed.
- `make fmt-check` passed.
- `go test ./internal/httpapi -count=1 -race` passed.
- `go test ./internal -run 'Test(BatchOverMaxBatchSizeIsRefusedBeforeTheRepository|ZeroMaxBatchSizeLeavesBatchesUnbounded|PageOffsetRefusesOverflow|QueryWithOverflowingPageIsRefusedBeforeTheRepository|QueryTimeoutCancelsABlockingRepository|TransactionTimeoutCancelsABlockingWrite)$' -count=1` passed.
- `go test ./internal/bootstrap -run TestWriteDeadlineStartsBeforeTheBodyIsRead -count=10` passed.
- Focused temporary repros confirmed the first and third findings; they were removed and the review worktree is clean.

Local limitations:

- Race-enabled bootstrap and federated test binaries could not link because the shared `/tmp` filesystem exhausted its quota. The full federated targeted test and local lint were therefore not rerun.
- GitHub CI at the reviewed head reports `Lint`, `Test`, `k6 Smoke`, and `Benchmark Smoke` successful; `E2E Smoke Tests` was still in progress at review time.

#### Overall Assessment

Request changes. The PR is well integrated across HTTP, manager, federated, configuration, and documentation layers, and the latest deadline-test repair is sound. However, the remaining cap-precedence gap and underestimated multi-pass DuckDB deadline violate the PR's advertised bounded-work and response-delivery guarantees. The malformed-framing classification should also be corrected to avoid false 5xx responses.

#### Summary

- Findings: **0 Blocker, 0 High, 2 Medium, 1 Low**.
- Major risks: body-cap response contract can be bypassed by malformed input; valid multi-pass federated requests can outlive a configuration that passed HTTP deadline validation.
- Verification: targeted HTTP, manager, and deadline tests passed; full local federated/race verification was limited by `/tmp` disk quota; most CI checks are successful and E2E smoke remains in progress.

iceboundrock:
### Review response (head `b42649a`)

All three findings are addressed in code. Dispositions, the root cause behind each, and the design decisions taken while resolving them:

**[Medium] Complete-body cap is skipped after an early JSON parse failure — fixed** (`5ab4e92`)

- Root cause: `readJSONBody` treated the decoder's verdict as final. The cap is a verdict on the whole body, so the capped stream has to reach its terminal result before any answer is chosen, regardless of what the decoder said.
- Change: `settleDecodeError` reads the capped stream to its end after a refused decode and lets the cap (`*http.MaxBytesError`) or a read failure met on the way outrank the parse error. Only when the rest arrives whole and under the cap does the parse error stand. `drainBody` and the new path share one read loop (`consumeBody`), so the success and failure paths follow the same rule rather than two similar ones.
- Tests: `TestMalformedPrefixOverTheCapIsStill413` (the review's repro: `{"schema_name" x` + 1,000 spaces under `MaxBodyBytes=64` is 413; the same malformed body under the cap keeps `invalid json body: invalid character 'x' after object key`). Red before the fix, green after.

**[Medium] HTTP write-timeout validation underestimates an unbounded-query federated request — fixed** (`3e1a17b`)

- Root cause: `DuckDBConfig.QueryTimeout` was armed per DuckDB pass while `LargestRequestBudget` counted it once. A request can take up to four passes (page, corrupt-parquet retry, deep-page recount, its retry).
- Ruling between the two directions the review offered: I chose one request-scoped deadline over teaching the validator a pass count. A pass-count constant in `bootstrap` would have to track the engine's structure by hand and would still leave the request's latency unbounded by the DuckDB budget, which is the opposite of what #465 asks for. Arming once at the request's DuckDB-routed entry makes the validator's single-budget formula exact and bounds the request. It changes nothing under the default configuration: with `Query.DefaultTimeout` bounded (30s), the outer query deadline already made the retry and the recount share a budget, so the "retry gets its own budget" claim only ever held when the query budget was disabled. This makes the opted-out case behave the same way.
- Change: `Query` hands the DuckDB branch to `queryDuckDBRouted`, which arms `QueryTimeout` once for the page pass, the recount and a degraded Postgres fallback together; later passes get the remainder. The per-pass arming in `StreamDuckDBFederatedQuery` stays as the bound for direct callers (`ExecuteFederatedPaginatedQuery`, the benchmark harness) and nests under the request's deadline, which can only be earlier or equal. Extracting the branch also brings `Query` (previously 110 lines) back under the 100-line limit. The `LargestRequestBudget` doc, `config_duckdb.go`, the README row and `docs/error-handling.md` now describe the budget as per request, and the PR description's DuckDB row is updated.
- Note on the degraded fallback: with the query budget disabled, the Postgres-only fallback now runs under the remainder of the DuckDB budget instead of unbounded. That is the same shape the query budget already imposes on it when bounded, and it is what makes the validated write deadline actually cover the request.
- Tests: `TestQueryTimeoutIsOneBudgetForTheWholeRequest`: a deep page whose empty scan triggers the recount runs the recount under the very deadline the page pass ran under (deadline equality, deterministic). Red before the fix, green after. The validator itself is unchanged; its test comment now points at this test for the other half of the invariant.

**[Low] Malformed chunk framing is reported as an internal server error — fixed** (`5ab4e92`)

- Root cause: `bodyReader` tagged every non-EOF body error as a transport failure, and the response mapped every non-timeout one to a redacted 500.
- Change: transport failures are recognised positively (`net.Error`, which covers deadlines, resets and closed sockets, or `io.ErrUnexpectedEOF`, which net/http raises for a body that ends before its declared length). Everything else `r.Body.Read` returns is net/http refusing to frame what the client sent, and answers a disclosed 400 `read request body: malformed request body` built with `forma.InvalidInputf`, with net/http's own prose attached as operator detail (Warnw line with an `error_id`, as #361 prescribes for a withheld detail). Positive recognition is the only option because net/http's framing errors are untyped; an unrecognised failure therefore lands on the caller's side, where the client is at least still connected to read the answer. 408 for a read timeout and the redacted 500 for a disconnect are unchanged.
- Tests: `TestMalformedChunkFramingAnswers400` (real listener, chunk-size line `Z\r\n`, 400, manager never reached, no "chunk" prose in the body) and the framing and truncated-body rows in `TestBodyReadFailureIsNotMalformedJSON`. Red before the fix, green after.

**Verification run locally at `b42649a`**

- `make fmt-check`: passed.
- `make lint` (root and infra modules): passed.
- `./scripts/test_with_container_runtime.sh` (full `make test` under Podman): exit 0.
- `go test -race -count=1 ./internal/httpapi ./internal/federated ./internal/bootstrap`: passed (the race run the review could not link).
- Each new test was also run against the pre-fix production files (stashed) and failed there.

No follow-up issues were needed; nothing in the three findings is out of the PR's scope.

iceboundrock:
### PR Review

#### Scope & Intent

Reviewed final head `b42649a` for #465. The PR enforces request-body, batch, query/write/DuckDB timeout, page-offset, and HTTP connection limits; wires their environment configuration through the server and Lambda entry points; and documents the resulting `400`/`408`/`413`/`504` behavior. `QueryConfig.MaxRows` is explicitly left out of scope and tracked by #598.

The review covered the PR title and description, all commits and the complete final diff, #465 and the PR decision record, affected HTTP handlers, manager services, federated routing/degradation paths, config and startup wiring, public error semantics, changed tests and docs, repository standards, and CI configuration.

- **Explicit requirements:** #465 and the PR require work to be bounded before untrusted request material is decoded or executed, context deadlines for manager and DuckDB work, safe page-offset arithmetic, and bounded HTTP server phases.
- **Public/error contracts:** caller-addressable validation uses `forma.InvalidInputf`; infrastructure and timeout errors follow the established redacted response contract; context expiry must propagate through repository and federated work.
- **Architecture conventions:** manager-level enforcement also protects library embedders; the HTTP server and Lambda assemble and validate environment-backed configuration before opening pools; a DuckDB timeout must not silently degrade to a partial Postgres result.
- **Runtime semantics:** `http.MaxBytesReader` must cover the complete decoded body, `WriteTimeout` starts after headers and includes body reading plus handler work, and nested `context.WithTimeout` contexts preserve the earlier deadline.
- **Code standards:** changed production files and functions remain within the documented size limits; errors retain contextual wrapping and public/error-carrier semantics.

#### Findings

#### Observations

- The body reader now settles both successful and failed JSON decoding against the terminal capped stream. This correctly gives `413` precedence for an oversized tail, rejects non-whitespace trailing data under the cap, and separates transport failures from malformed framing.
- Batch admission is at the manager boundary for all three batch APIs and both execution modes. The server routes only atomic create/delete batches, while direct best-effort callers retain the explicitly documented per-operation transaction budget.
- DuckDB timeout is armed once for the complete routed request, so page scans, retries, deep-page recounts, and any degraded fallback share one budget. Excluding `context.DeadlineExceeded` from degradation preserves the latency bound rather than returning a potentially late partial result.
- `MAX_BATCH_SIZE=0` is intentionally only an unbounded runtime setting for embedders that construct an unvalidated config by hand. Server/Lambda startup validates it against `Performance.BatchSize`, as documented in the README; this is consistent with the existing `Config.Validate` invariant.

The added tests exercise observable behavior rather than only helpers: all body-carrying endpoints, whole-body cap precedence, malformed framing and transport read failures, manager batch admission, page overflow before repository access, query/write cancellation, DuckDB cancellation and shared multi-pass deadlines, environment overlays, startup validation, and real-listener HTTP deadline behavior. They cover the risks introduced by the PR.

Commands executed:

- `git diff --check main...HEAD` - passed.
- `make fmt-check` - passed.
- `go test . -run TestConfigValidationRequestLimits -count=1` - passed.
- `go test ./internal -run Test(BatchOverMaxBatchSizeIsRefusedBeforeTheRepository|ZeroMaxBatchSizeLeavesBatchesUnbounded|PageOffsetRefusesOverflow|QueryWithOverflowingPageIsRefusedBeforeTheRepository|QueryTimeoutCancelsABlockingRepository|TransactionTimeoutCancelsABlockingWrite)$ -count=1` - passed.
- `go test ./internal/httpapi -count=1` - passed.

A combined local run of `./internal/httpapi`, `./internal/bootstrap`, and `./internal/federated` was attempted. `internal/httpapi` passed, but bootstrap and federated test linking could not complete because the local filesystem hit a disk quota during CGO linking. This is an environment limitation, not a test failure.

GitHub CI was also inspected at the reviewed head: `Lint`, race-enabled `Test`, `E2E Smoke Tests`, `k6 Smoke`, and `Benchmark Smoke` are all successful.

#### Overall Assessment

The implementation is coherent across the HTTP, manager, federated, configuration, documentation, and error-boundary layers. The final design preserves whole-body enforcement, response-deadline composition, context cancellation, and federated consistency/latency invariants while retaining the documented direct-embedder zero-value behavior.

#### Summary

- Findings: **0 Blocker, 0 High, 0 Medium, 0 Low.**
- Major risks: none identified in the reviewed scope.
- Verification: focused local checks passed; two CGO-linked package runs were locally blocked by disk quota; all PR CI checks are successful.

PR #601, opened 2026-09-19
Release the single-flight plan/projection slot on panic and honour ctx while waiting (#467)

### LINKED ISSUE #467: bug: a panic inside a single-flight plan/projection build permanently deadlocks every later caller on that key
## Problem

The single-flight caches guarding plan/projection building do not release their in-flight slot on panic: `wg.Done()` and `delete(inflight, key)` are plain statements, not deferred. A panic in the leader's `build()` unwinds past both, leaving a `WaitGroup` that is never `Done`. Every later caller for that key blocks forever at `wg.Wait()` (no context, no timeout), leaking goroutines and their pgx/DuckDB holds.

## Current behavior

`internal/queryplan/cache.go` `Cache.GetOrBuild` and `internal/sqlgen/plan_cache.go` `ProjectionCache.GetOrBuild`:

```go
fc := &flightCall{}; fc.wg.Add(1); c.inflight[key] = fc; c.flightMu.Unlock()
v, err := build()          // no defer around the cleanup below
...; delete(c.inflight, key); fc.wg.Done()
```

`net/http` recovers a per-request panic so the process survives, but `c.inflight[key]` retains an un-`Done` `WaitGroup`. Followers block at `fc.wg.Wait()` with no `ctx` escape hatch — permanent unavailability for that query shape plus an unbounded goroutine leak. Even without a panic, `fc.wg.Wait()` is uncancellable, so a slow leader blocks all followers with no ctx timeout. Secondary: `internal/postgres_persistent_repository_query.go` does an unchecked `queryAny.(string)` on the result.

## Reproduction

Any panic from `build()` — `renderTemplate`/`BuildSchemaProjection` on malformed metadata, or a nil-map/index panic in a future builder — then a second request for the same key hangs forever.

## Root cause

Cleanup of the single-flight slot is not deferred, and the follower wait is uncancellable.

## Impact

One recoverable request panic converts into permanent unavailability for that query shape and a growing goroutine leak until the process is killed. Confidence: High on mechanism/deadlock; Medium that a panic is reachable from today's builders (so this is defense-in-depth for a real latent trap).

- #142 (closed) introduced the caches. #406 (open) is a *different* panic (write-path normalization). Searched `singleflight`, `deadlock`, `recover`, `plan cache panic`. NEW.

## Acceptance criteria

- [ ] The in-flight slot is released (Done + delete) via `defer`/`recover`, so a leader panic frees followers with an error rather than deadlocking them.
- [ ] `wg.Wait()` honours the caller's context.
- [ ] The `queryAny.(string)` assertion is checked.

- [ ] A panicking `build()` leaves the cache usable and does not hang the next caller for that key.

iceboundrock:
Fixes #467.

## Problem

`queryplan.Cache.GetOrBuild` and `sqlgen.ProjectionCache.GetOrBuild` released their single-flight slot with plain `delete` + `wg.Done()` statements after `build()`. A panic in the leader (recovered per request by `net/http`) skipped both, leaving an un-`Done` `WaitGroup` in `inflight`; every later caller for that key blocked forever at `wg.Wait()`, which also had no context escape hatch.

## Change

- **Deferred, panic-safe release.** The leader defers `releaseFlight`, which `recover()`s a panic, records it as the flight's error (so waiters get `… build for schema N panicked: <value>` instead of hanging), deletes the slot, closes the flight's `done` channel, and then re-raises the panic so the leader's own failure mode is unchanged.
- **Context-aware wait.** `GetOrBuild` takes `ctx` as its first parameter on both caches. Followers `select` on the flight's `done` channel vs `ctx.Done()`; a cancelled wait returns the wrapped context error and touches no hit/miss counter. The leader's build is never cancelled and its result is still cached, so a follower timing out cannot poison the key.
- **Context threading.** `renderOptimizedQuerySQL`, `buildDuckDBQueryWithPlan`, `serveFromPlanCache`, `injectSchemaProjections` and `schemaProjection` now take the request context.
- **Checked assertions.** `renderOptimizedQuerySQL` returns an error naming the schema, the actual artifact type and the expected `string` instead of panicking on `queryAny.(string)`. The identical `entryAny.(*duckCompiledEntry)` in `serveFromPlanCache` now falls through to the direct builder (`ok=false`, the existing contract for every other failure there).

Rejected `golang.org/x/sync/singleflight`: it re-panics in every follower, the opposite of the acceptance criterion, and would change the hit/miss accounting the benchmark counters rely on. Not done here: folding the two duplicated `GetOrBuild` bodies into one generic store. That duplication predates this issue and is a separate refactor.

## Tests

- `internal/queryplan/cache_test.go`, `internal/sqlgen/plan_cache_test.go`: `…LeaderPanicFreesFollowers` (leader panics while a follower waits: follower returns an error promptly, leader re-panics with the original value, `inflight` is empty, next caller rebuilds) and `…FollowerWaitHonoursContext` (follower's ctx cancelled while the leader is blocked: follower returns `context.Canceled`, leader's result is cached, no spurious miss). Both fail against the old bodies with "follower deadlocked" / "follower ignored context cancellation".
- `internal/postgres_persistent_repository_test.go`: `TestStreamOptimizedQueryRenderCacheRejectsForeignArtifact`, a pre-seeded `int` under the render key yields the typed error and no query reaches the database.
- Existing tests updated for the new signature.

## Verification

- `go test -race` on `internal/queryplan`, `internal/sqlgen`, `internal/federated`, `internal`: pass.
- New concurrency tests repeated 30× under `-race`: no flakes.
- `./scripts/test_with_container_runtime.sh` (full `make test`): 34 packages ok, exit 0.


iceboundrock:
### PR Review

#### Scope & Intent

This PR fixes both single-flight caches so a leader panic releases the in-flight slot, wakes followers with an error, and re-panics on the leader path. It also adds context-aware follower waits, threads request contexts through the PostgreSQL and federated planning paths, and replaces two unchecked cache-artifact assertions with checked handling.

The review basis is that successful builds remain single-flighted and cached, the leader build remains uncancelled, cache hit/miss accounting remains meaningful, the existing federated direct-builder fallback remains valid for non-cacheable or non-context cache failures, and request cancellation should not be bypassed by an uncancellable planning path.

- Explicit issue #467 requirements: release both in-flight slots on panic; make follower waits honor context; check the optimized-query cache assertion; and add regression coverage proving a panicking build does not permanently deadlock the key.
- Existing cache contracts in `internal/queryplan` and `internal/sqlgen`: one build per missing key, successful waiters count as hits, errors and nil artifacts are not cached, and the leader's successful result remains available after a follower cancels.
- Existing federated contract: the compiled-plan path may fall back to direct rendering for ordinary cacheability, compile, or bind failures, but request context must still bound request work.
- Repository standards: preserve error wrapping and context propagation, maintain race-safe concurrent behavior, and satisfy the Go 1.26 CI formatting, lint, unit, and E2E gates.

#### Findings

**[Medium] Federated plan-cache cancellation falls through to uncancellable rendering**

- **Location:** `internal/federated/duckdb_query_build.go:305-307` (`serveFromPlanCache`), with the direct fallback at `buildDuckDBQueryWithPlan:98-104`
- **Issue:** `queryplan.Cache.GetOrBuild(ctx, ...)` now correctly returns a wrapped `context.Canceled` or `context.DeadlineExceeded` when a follower stops waiting, but `serveFromPlanCache` discards every returned error and reports only `ok=false`. Its caller interprets that as an ordinary cache miss and invokes the direct DuckDB builder.
- **Impact:** If a federated request's context expires while another request is compiling the same plan, the canceled request still performs direct template rendering, which has no context parameter or cancellation check. Multiple canceled followers can duplicate this work, and cancellation is not observed at the planning boundary until the later DuckDB call. This also makes the federated plan-cache path inconsistent with the projection-cache and PostgreSQL callers, which propagate the context error.
- **Evidence:** `internal/queryplan/cache.go:118-128` and `internal/sqlgen/plan_cache.go:107-119` return promptly on `ctx.Done()`. The federated caller handles `err != nil` identically to a cache miss, and the direct builder does not receive `ctx`. No existing check prevents the fallback after cancellation. The new tests verify the cache methods in isolation but do not exercise a canceled federated waiter through `serveFromPlanCache`.
- **Recommended direction:** Preserve the current direct fallback for non-context cache/build/bind failures, but propagate context cancellation/deadline errors through `serveFromPlanCache` and `buildDuckDBQueryWithPlan` instead of converting them to `ok=false`. Add a federated regression test with a blocked leader that cancels a follower and verifies that direct rendering is not invoked and the returned error still matches the caller's context error.

#### Observations

- The deferred `releaseFlight` implementation is sound: it records a waiter-visible panic error, deletes the slot, closes the completion channel, and re-raises the original panic for the leader.
- All repository call sites found by the review were updated to the new cache signatures; no stale `GetOrBuild`, projection, render, or DuckDB-builder call sites remain.
- The checked PostgreSQL artifact assertion preserves the existing error-return behavior. The federated artifact assertion preserves its documented direct-builder fallback, subject to the cancellation issue above.

- New tests cover leader panics and follower cancellation for both cache implementations, including slot cleanup, leader re-panic behavior, cache reuse after cancellation, and hit/miss accounting.
- The new PostgreSQL foreign-artifact test verifies a non-string render artifact returns a typed error without reaching the database.
- Missing behavioral coverage: the federated production path needs a test for cancellation while waiting on `serveFromPlanCache`; this is the scenario described in the finding.
- Passed locally: `make fmt-check`.
- Passed locally: `make lint`.
- Passed locally: `GOTMPDIR=$PWD/.gotmp ./scripts/test_with_container_runtime.sh` with rootless Podman; all unit and container-backed tests passed.
- Passed locally under race detection: `go test -race ./internal/queryplan`, `./internal/sqlgen`, and `./internal/federated` (run package-by-package with a workspace temporary directory).
- Passed locally 30 times: the new panic and cancellation tests in both cache packages.
- GitHub PR checks are all green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.
- An initial bare `make test` could not initialize testcontainers because no Docker socket was configured; the documented Podman wrapper was then run successfully. An initial combined CGO race invocation also hit `/tmp` space during parallel linking; the affected packages passed when rerun sequentially with `GOTMPDIR` on the workspace filesystem.

#### Overall Assessment

The core single-flight fix is technically well-structured and addresses the deadlock described by #467 without changing the leader's panic semantics. The remaining issue is a real context-propagation gap in the federated plan-cache caller: the cache honors cancellation, but the surrounding production path immediately converts that result into uncancellable direct work. I recommend fixing that before merge; no blocker or high-severity issues were identified.

#### Summary

- Findings: 0 Blocker, 0 High, 1 Medium, 0 Low.
- Major risk: canceled federated plan-cache waiters can duplicate uncancellable SQL rendering and delay cancellation.
- Verification: repository gates and GitHub CI passed; the targeted race and repeated concurrency checks passed; the one medium finding remains unaddressed.

iceboundrock:
### Review response

**[Medium] Federated plan-cache cancellation falls through to uncancellable rendering** — fixed in 629ea1e.

**Root cause.** `serveFromPlanCache` folded every `GetOrBuild` error into `ok=false`, so the new follower-side `context.Canceled`/`DeadlineExceeded` return was indistinguishable from a compile failure and took the direct-builder fallback, which has no context parameter.

**Change.**
- `serveFromPlanCache` now returns `(sql, args, ok, err)`. `err` is set only when the returned cache error is the caller's own context error (`ctx.Err() != nil && errors.Is(err, ctx.Err())`); compile errors, leader panics and foreign artifacts keep the documented `ok=false` fall-through, so the direct builder still owns error reporting for those. `buildDuckDBQueryWithPlan` propagates the error as `build duckdb query: wait for in-flight duckdb_federated plan build for schema N: context canceled`, which `errors.Is` matches against the caller's context error.
- Key derivation is extracted to `duckPlanCacheKey` (no behavioural change) so a test can occupy the request's own key; `sqlgen.CompileDuckDBQuery` offers nothing to block on. The unused `cache` parameter of `serveFromPlanCache` was dropped in passing.

**Test.** `TestBuildDuckDBQueryCancelledPlanWaiterDoesNotRenderDirectly` (`internal/federated/plan_cache_cancel_test.go`): a synthetic leader holds the request's key in flight, a follower with a cancelled context calls `buildDuckDBQueryWithPlan` and must get `context.Canceled` with no SQL; after release the same inputs render with a live context, which is what proves the cancelled call would otherwise have produced SQL via the direct builder. Against the previous body the test fails with "Expected error with "context canceled" in chain but got nil" (a rendered query came back).

**Validation run locally.**
- `go test -race ./internal/federated ./internal/queryplan ./internal/sqlgen`: pass.
- New test 30× under `-race`: no flakes.
- `GOTMPDIR=$PWD/.gocache/tmp ./scripts/test_with_container_runtime.sh` (full `make test`, rootless Podman): 34 packages ok, exit 0.

Not changed: the direct builder itself still takes no context. That is pre-existing and a leader (whose build is deliberately never cancelled) is followed immediately by the context-aware DuckDB execute, so the remaining exposure is one render for a request that ends between the two, not the duplicated fan-out the finding described.

iceboundrock:
### PR Review

#### Scope & Intent

This PR updates both single-flight caches (`internal/queryplan.Cache` and `internal/sqlgen.ProjectionCache`) so leader cleanup is panic-safe and follower waits can honor request context cancellation. It threads context through the PostgreSQL optimized-query render path and the federated DuckDB planning path, preserves leader re-panics and successful-result caching, and replaces two unchecked cache-artifact assertions with checked handling.

The review basis was the complete two-commit diff, issue #467, all affected call sites, cache and query contracts, federated query design documentation, repository coding standards, and CI configuration.

- Issue #467 explicitly requires both in-flight slots to be released on leader panic, follower waits to honor context, the optimized-query assertion to be checked, and regression coverage proving that a panicking build does not permanently deadlock a key.
- Existing cache contracts require one build per missing key, successful followers to count as hits, failed or nil builds not to be cached, and a leader's successful result to remain cacheable when a follower cancels.
- The federated plan-cache contract allows ordinary compile, bind, and artifact failures to fall back to the direct builder, but a caller whose own context ended while waiting must not continue with uncancellable direct rendering.
- Repository standards require context propagation through request work, wrapped errors with useful context, race-safe concurrency, gofmt cleanliness, and files/functions within the documented size limits.
- CI validates formatting, linting, unit tests under `-race`, federated E2E smoke tests, k6 smoke tests, and benchmark smoke tests.

#### Findings

#### Observations

- The two leader-panic regression tests use a fixed `50ms` sleep to try to ensure the follower has reached the wait (`internal/queryplan/cache_test.go:181-192` and `internal/sqlgen/plan_cache_test.go:802-813`). This is not a production defect, and the repeated race runs passed, but a deterministic waiting barrier would remove the remaining scheduler-dependent test flake risk.

- New cache tests cover panic release, follower-visible panic errors, leader re-panics, slot reuse, context cancellation, successful-result caching after follower cancellation, and hit/miss accounting for both cache implementations.
- The PostgreSQL regression test verifies that a foreign render artifact returns a typed error and does not reach the database.
- The federated regression test verifies that a canceled plan-cache waiter receives the context error instead of falling through to direct SQL rendering, and that live callers can still render afterward.
- `make fmt-check`: passed.
- `make lint`: passed, including the infra module lint run.
- `git diff --check 40964bd..629ea1e`: passed.
- `go vet ./internal/queryplan ./internal/sqlgen ./internal/federated ./internal`: passed.
- `go test -race ./internal/queryplan ./internal/sqlgen ./internal/federated ./internal`: passed when run sequentially with workspace temporary storage.
- The new cache panic/cancellation tests and federated cancellation test each passed 100 times under `-race`.
- `gh pr checks 601`: all reported checks passed: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.
- A local `GOTMPDIR=$PWD/.gotmp make test` attempt was not a complete local-suite result: the compaction test binary hit the environment's disk quota during linking, and the container-backed E2E harness could not find a usable Docker-compatible runtime socket. The affected changed packages passed independently, and GitHub's full PR checks passed.

#### Overall Assessment

The implementation correctly closes the deadlock and head-of-line blocking gaps described by issue #467. The deferred release publishes waiter-visible results before closing the completion channel, removes the flight slot on both normal and panic paths, and re-raises the leader panic. Context cancellation is propagated through the federated caller rather than being converted into a direct-builder miss, while ordinary cache failures retain their existing fallback behavior. The checked artifact assertions improve failure handling without changing the intended query path.

The remaining test-synchronization observation is low risk and non-blocking. No production behavior requires attention before merge.

#### Summary

- Findings: 0 Blocker, 0 High, 0 Medium, 0 Low.
- Major risks: none identified in the reviewed scope.
- Verification: targeted local race tests, vet, formatting, lint, repeated concurrency tests, and all reported GitHub PR checks passed; the complete local `make test` was environment-limited as described above.

Issue #604, opened 2026-09-21
schemameta: report every failing schema at registry construction instead of the first


iceboundrock:
Split out of #549 (option 2 there), which fixed the per-entry diagnostic but not the blast-radius problem.

## What happens

Every registry construction path returns on the first error inside its per-schema load loop — `internal/schemameta/loader.go` (both the DB and file loaders) and `internal/schemameta/file_registry.go` (directory and map constructors). One schema failing `validateSchemaAttributeCache` therefore fails the whole directory or registry table at startup, and the operator sees exactly one diagnosis. If three legacy schemas each carry a retired `Created_At` (#549) or a case-variant reserved name (#532), the operator fixes one, reboots, and discovers the next.

## What should happen

Keep fail-closed: a registry with any invalid schema must still refuse to construct. But the error returned should aggregate every failing schema's diagnosis (`errors.Join` or an equivalent that keeps each "schema <name>: ..." line intact), so one boot attempt shows the full set of schemas that need attention.

## Notes

- `validateSchemaAttributeCacheOpts(..., checkBindings=false)` already has a precedent for "report everything" on the validate-schema-consistency path (`ValidateColumnBinding`); this issue is about the per-schema loop one level up, on the runtime load paths.
- Per-schema ordering of the aggregated report should be deterministic (sorted by schema name) so the boot log is stable.
- Tests: a directory with two failing schemas and one valid one must produce an error naming both failing schemas and must not register the valid one.

PR #605, opened 2026-09-21
sqlgen: give a retired ledger entry the migration remedy on a reserved or colliding parquet column (#549)

### LINKED ISSUE #549: schemameta: a retired ledger entry whose folded column is reserved hard-blocks registry construction with an unactionable remedy

iceboundrock:
Closes #549.

## Problem

`validateSchemaAttributeCache` runs `sqlgen.ValidateParquetAttrColumns` over the full attribute cache, retired entries included, because a retired attribute's folded column still exists in flushed parquet files (#342). Since #548 widened the reserved-column check to case variants, a retired `Created_At` fails registry construction, and the rejection said "rename the attribute", which a retired entry cannot do: it is the attributeID ledger for the flushed values, and a rename in place desynchronizes the ledger from the data. The intra-schema fold collision had the same gap.

## Change

The two halves of the guard now treat a retired entry differently, because the two hazards live in different places (option 3 of #549, narrowed to what the engine supports; ruling and evidence in [this comment](https://github.com/Lychee-Technology/forma/pull/605#issuecomment-5755648029)):

- **Reserved-column half: retired entries are exempt.** No flushed file holds two spellings of a system column — the exporter emits `ltbase_created_at`, never `created_at`, and DuckDB deduplicates a case-variant of a physical export column (`Row_Id` beside `row_id`) to `Row_Id_1` at COPY time. The ambiguity arises only when the attribute is projected beside the system column, and every projection (exporter select, `s3_source`, `pg_source`, outer select) is built from the active cache. A retired column sits unreferenced in the raw `SELECT *` scans of the cold-scan source and the compaction merge and binds nothing. A retired `Created_At` therefore loads again; the same attribute active is still refused with the #532 wording, byte for byte.
- **Collision half: retired entries stay in scope, with the ledger remedy.** `read_parquet(..., union_by_name=true)` resolves same-folding columns from different files onto one column, so a retired attribute's flushed column interferes with an active or another retired column regardless of projection. A retired/active pair names the retired side as `retired attribute "X" (id N, valueType T)` and directs the rename at the active attribute (or the migration: move the flushed column and the ledger entry together, keeping the id), mirroring `attrIDCollisionError` / `bindingCollisionError` in `schemameta`. A retired/retired pair gets the migration remedy for both.

Only `schemameta` passes retired entries through the guard; the CDC exporter and the DuckDB projection pass active caches and see no change.

## Design decision recorded

`docs/federated-query/design.md` §4.4 gains a paragraph next to the #532 case-insensitivity text stating the split and why, naming the engine tests that pin each claim.

## Follow-up

Option 2 (aggregate every failing schema at startup instead of returning on the first) is split out to #604.

## Tests

- `internal/sqlgen/parquet_attr_column_retired_test.go`: retired reserved names (exact, case-variant, physical, dedup) register while the same names active are refused; a retired reserved entry still collides with a same-folding sibling; retired/active collision in both sort orders; retired/retired collision; the active-message pin.
- `internal/sqlgen/parquet_attr_column_retired_scan_test.go` (engine-backed): `TestParquetScan_RetiredReservedColumnIsInert` runs the exact `s3_source` projection over the exact cold-scan source with `Created_At` and `Row_Id_1` in the file and asserts every system column binds to the system value; `TestParquetScan_FoldCollisionMergesAcrossFiles` shows `union_by_name` reading the retired `Foo` through the active `foo`.
- `internal/compaction/merge_test.go`: `TestDuckMerger_CarriesRetiredReservedNameColumnsThrough` runs the real merge over such a file.
- `internal/schemameta/retired_guard_test.go`: the #342 strip-after-validate ordering pin now uses a retired/active fold collision on the directory load path; a retired `Created_At` loads, is stripped from the active cache, and the active variant still fails with "rename the attribute"; plus the retired/active collision through `validateSchemaAttributeCache`.

All new "loads" and "refuses" tests were watched failing against the previous guard before it was changed.

## Verification

- `make fmt-check`, `make lint` (root + infra), `git diff --check`: clean
- `./scripts/test_with_container_runtime.sh` (full `make test`): all packages ok


iceboundrock:
### PR Review

#### Scope & Intent

This PR changes `sqlgen.ValidateParquetAttrColumns` diagnostics so retired metadata entries are identified as attributeID ledger records, with migration guidance for reserved-column and folded-column collisions. It keeps retired entries in full-cache validation, preserves the existing active-only messages, adds registry-path and validator tests, and records the decision in `docs/federated-query/design.md` §4.4. The review basis was the #549 acceptance criteria, the #342 retired-ledger contract, the active-cache stripping order, and the actual CDC, compaction, and federated projection SQL.

- Retired entries are validation-only ledger records. They remain available for ID/binding invariants, then are stripped before active read/write/projection consumers.
- Retired attribute IDs and historical physical names must not be rebound or changed in place when doing so would desynchronize preserved EAV or flushed lakehouse data.
- Folded-column collisions between a retired and active attribute remain meaningful because the active projection can bind an ambiguous identifier; active-only diagnostics must remain stable.
- The repository requires deterministic operator-facing errors, context-rich plain errors on consistency failures, and tests for observable behavior rather than only implementation structure.

#### Findings

**[Medium] Re-evaluate the retired-only reserved-column boot failure using the actual scan shapes**

**Location:** `docs/federated-query/design.md:218-224`; enforced by `internal/schemameta/validation.go:66-74` and exercised by `internal/schemameta/retired_guard_test.go:132-152`

**Issue:** The new design paragraph says a retired `Created_At` is beside a `created_at` system column in flushed Parquet and that both compaction and the federated `s3_source` scan those columns with `SELECT *`, making the reserved-name collision a real read-side ambiguity. The current implementation does not support that description. CDC base and delta writers emit `ltbase_created_at`, not `created_at` (`internal/cdc/duckdb_exporter.go:186-195`). Compaction reads and writes the raw Parquet schema with `SELECT *` (`internal/compaction/merge_sql.go:75-85`), so the relevant system name is also `ltbase_created_at`. The federated `s3_source` is an explicit `SELECT {{.S3SourceSelect}}`, not `SELECT *` (`internal/sqlgen/advanced_query_template_duckdb.go:46-60`); `S3SourceSelect` is built from the already-stripped active cache (`internal/schemameta/loader.go:373-382`, `internal/sqlgen/duckdb_schema_projection.go:162-174`), so it never selects a retired `Created_At` field.

**Impact:** A schema containing only a retired case-variant reserved attribute, such as `Created_At`, still fails entire registry construction before serving any schema, even though the active projection does not reference that retired field and the raw writer/compaction names do not collide. This is a narrow legacy case and the failure predates this PR, but this PR explicitly institutionalizes it as the fail-closed design and bases the migration remedy on the claimed ambiguity. The proposed migration can therefore impose an unnecessary rewrite and ledger change, while the actual read-side invariant that justifies the boot failure is not demonstrated.

**Evidence:** I traced the loader, writer, compaction, and federated projection paths above and ran a review-only DuckDB smoke test with the exact relevant names and shapes: raw `ltbase_created_at` plus `Created_At`, the `SELECT * REPLACE` scan wrapper, and the explicit `ltbase_created_at AS created_at` projection all bound successfully when the retired field was not selected. The retired/active folded-attribute collision remains a separate case and is not covered by this objection. No current producer in the reviewed history emits a Parquet `created_at` system field.

**Recommended direction:** Revisit the option-3 decision before merging. The invariant should retain retired entries for retired/active folded-column collisions, but a retired-only reserved hit should either be exempted if it cannot affect any active projection/read binding, or be supported by an actual legacy Parquet format that contains `created_at` and an integration test binding that exact scan. In either case, correct §4.4 and the registry test so the documented reason matches the real execution path.

#### Observations

The diagnostic implementation itself is deterministic and narrowly scoped: it carries the retired ID/value type, orders a retired side first regardless of map iteration, gives retired/active collisions an active-side remedy, and preserves the pre-existing active-only messages byte for byte. Splitting startup error aggregation into #604 is consistent with the stated scope.

The new tests cover exact and case-variant reserved hits, retired/active collisions in both sort orders, retired/retired collisions, the registry load path, and active-message compatibility. They do not test the actual compaction/federated scan shape used to justify the retained retired-only reserved rejection; that is the behavioral gap behind the finding.

Executed locally:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/schemameta` passed.
- `make fmt-check` passed.
- `git diff --check` passed.
- A temporary review-only DuckDB scan-shape smoke test passed.
- `make test` was attempted. The packages that completed passed, but multiple CGo/DuckDB test binaries could not link because the environment hit disk/quota limits (`disk quota exceeded` / `No space left on device`); this is not attributable to the PR.
- The PR status shown on GitHub was green for Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

#### Overall Assessment

The changed diagnostic logic is well targeted and the requested retired-ledger message coverage is strong. One medium design/correctness issue remains: the documented and tested justification for rejecting a retired-only reserved name does not match the current storage and query projections. That decision should be corrected or substantiated before merge; the rest of the reviewed implementation appears sound.

#### Summary

Findings: 1 Medium, 0 Blocker, 0 High, 0 Low.

Primary risk: unnecessary registry-wide startup failure and an unsupported migration rationale for retired-only reserved-column names.

Verification: targeted tests, formatting, lint, diff checks, and a DuckDB shape smoke test passed; the full local suite was limited by environment disk/quota failures, while the PR GitHub checks were green.

iceboundrock:
**Artifact: mid-execution ruling, revising the #549 option-3 decision** (working copy: none on disk; this comment is the record). Addresses the [Medium] finding in https://github.com/Lychee-Technology/forma/pull/605#issuecomment-5755501343.

## Understanding of the concern

The §4.4 paragraph and the registry test justified failing boot on a retired `Created_At` by claiming the column sits beside a `created_at` system column in flushed parquet and that compaction and `s3_source` scan both with `SELECT *`, making the reference ambiguous. The finding says the writers, the merge and the projection do not have that shape. I traced the same paths and ran the shapes against the pinned DuckDB rather than reasoning about them.

## What the engine shows

Verified with throwaway DuckDB tests (since turned into the committed tests below):

1. **No flushed file holds two spellings of a system column.** The exporter's base select emits `ltbase_created_at`, never `created_at` (`cdc.exportModeSpec.baseSelectColumns`). For a physical export column, DuckDB deduplicates the case variant at COPY time: `SELECT ... row_id, ... AS Row_Id` writes a parquet whose `parquet_schema` holds `row_id` and `Row_Id_1`. So even for `row_id`/`changed_at`/`ltbase_*`, the "both spellings in one file" premise is false.
2. **A retired column is inert on every read path.** The exact `s3_source` select (`BuildSchemaProjection` over the active cache) over the exact cold-scan source (`BuildParquetScanSource`) binds `row_id`, `created_at`, `ver_ts`, `deleted_ts` to the system values with `Created_At` and `Row_Id_1` present in the file. `DuckMerger.MergeToTmp` over such a file merges and carries the retired columns through verbatim.
3. **The active hazard is silent-wrong, not a loud bind error.** With `Created_At` projected as an active attribute beside `ltbase_created_at AS created_at`, `SELECT created_at FROM ranked` binds without error to whichever came first in the select list. That is what the reserved check protects active entries from, and it can only arise when the attribute is projected — which every projection (exporter select, `s3_source`, `pg_source`, outer select) does from the active cache.
4. **The fold-collision hazard is physical.** `read_parquet([older, newer], union_by_name=true)` resolves `Foo` (older file) and `foo` (newer file) onto one column: `SELECT foo` returns the retired `Foo` values too. Same-file `Foo`/`foo` is deduplicated to `foo_1` at write. So a retired entry's column interferes with a same-folding column regardless of projection, and the collision half must keep retired entries in scope.

## Ruling

Option 3 of #549, narrowed to what (1)–(4) support: **a retired entry is exempt from the reserved-column half and stays in scope for the collision half.** Failing boot on a retired reserved name imposed a migration that fixes nothing, and the issue text already anticipated this case ("produces no wrong answers on any read path — yet after #548 it hard-blocks boot"); the option-3 gate was "shown harmless first", which the engine tests now do. The alternative — keep the rejection and find a legacy format that contains `created_at` — has no such format in the reviewed history and would keep rejecting a name that no scan binds.

The earlier ruling I posted on #549 (the "SELECT * ambiguity" rationale) is superseded by this comment; I am posting a pointer there.

## Changes (commit ccc420b)

- `internal/sqlgen/parquet_attr_column.go`: the reserved lookup skips `Retired` entries; `reservedParquetColumnError` returns to its #532 signature and wording (the retired branch was dead under the exemption). The collision half is unchanged from the previous revision: retired/active directs the rename at the active side, retired/retired gets the migration remedy, active-only wording is byte-stable.
- `internal/sqlgen/parquet_attr_column_retired_test.go`: the two reserved-rejection tests become `..._RetiredReservedSystemColumnExempt` (exact, case-variant, physical and dedup names all register when retired, all refused when active — the exemption keys on `Retired`, not the name) and `..._RetiredReservedStillCollides`.
- `internal/sqlgen/parquet_attr_column_retired_scan_test.go` (new, engine-backed): `TestParquetScan_RetiredReservedColumnIsInert` pins (1) and (2) on the federated side; `TestParquetScan_FoldCollisionMergesAcrossFiles` pins (4).
- `internal/compaction/merge_test.go`: `TestDuckMerger_CarriesRetiredReservedNameColumnsThrough` pins (2) on the real merge.
- `internal/schemameta/retired_guard_test.go`: the #342 ordering-invariant pin (`strip runs after validate`) moves from a retired `row.id` to a retired/active fold collision on the directory load path; `TestDirectoryRegistry_RetiredReservedFoldedColumnLoads` loads a retired `Created_At`, asserts it is still stripped from the active cache, and asserts the same attribute active is still refused with "rename the attribute".
- `internal/schemameta/validation.go` comment and `docs/federated-query/design.md` §4.4 rewritten to state the real reason and name the pinning tests.

Both new "loads" tests were watched failing against the previous guard before the condition was changed.

## Verification

- `make fmt-check`, `make lint` (root + infra), `git diff --check`: clean
- `./scripts/test_with_container_runtime.sh` (full `make test`): every package ok

iceboundrock:
### PR Review

#### Scope & Intent

This review covered the PR description, both commits, the complete seven-file diff, the affected schema-registry, CDC, federated projection, cold-scan, and compaction paths, the relevant tests and documentation, issue #549 and its acceptance discussion, and the CI configuration.

The PR changes `sqlgen.ValidateParquetAttrColumns` so retired metadata entries remain in scope for folded-column collisions, but are exempt from reserved-column projection checks. It adds retired-ledger-aware collision diagnostics, preserves active-only diagnostics, documents the revised ruling in the federated-query design, and adds registry-path and DuckDB-backed regression coverage. The intended invariant remains that retired entries are validated before being stripped from active caches, while active CDC and federated projections never expose retired attributes.

- Issue #549 requires retired entries to retain their attribute ID and historical physical-name ledger semantics; renaming a retired entry in place must not be presented as a valid repair.
- The #342 contract requires full-cache validation before `activeAttributeCache` removes retired entries, while all production read, write, flush, and projection consumers receive active-only caches.
- Active attributes that fold onto reserved system or dedup columns must fail closed before writes or federated reads; folded-column collisions must remain rejected because DuckDB `union_by_name=true` merges case-insensitive identifiers across files.
- DuckDB unquoted identifier matching is ASCII case-insensitive, so the guard must match the engine while preserving byte-stable physical names and deterministic diagnostics.
- Repository standards require context-rich operator errors, stable public behavior where compatibility is specified, and tests of observable storage and query behavior rather than only helper implementation details.

#### Findings

#### Observations

- The revised reserved-versus-collision split is consistent with the actual paths: active-cache projections can create reserved-column ambiguity, while retired columns are unreferenced by those projections; folded collisions remain physical hazards in multi-file `union_by_name` scans and compaction.
- The implementation keeps retired entries ordered deterministically in diagnostics, includes the ledger ID and value type, directs retired/active repairs at the active attribute, and preserves the pre-existing active-only messages byte for byte.
- Startup aggregation of multiple failing schemas remains tracked separately as issue #604 and is not a gap introduced by this PR.

The added tests cover retired exact, case-variant, physical, and dedup-name reserved entries; retired/active collisions in both map-order cases; retired/retired collisions; active-message compatibility; registry loading and active-cache stripping; the exact federated cold-scan projection; `union_by_name` cross-file behavior; and the real compaction merge.

Executed locally:

- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/sqlgen ./internal/schemameta ./internal/compaction` passed.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/federated` passed.
- `make fmt-check` passed.
- `make lint` passed for the root and `infra/` modules.
- `git diff --check` passed.
- `./scripts/test_with_container_runtime.sh` passed the complete `make test` suite, including the container-dependent E2E harness. A bare `make test` attempt first failed only because this shell lacked `$XDG_RUNTIME_DIR`; the repository wrapper configured the available Podman socket and reran the full suite successfully.
- The PR CI checks observed at review completion all passed: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

No material behavioral test gap was found for the changed paths.

#### Overall Assessment

The PR appears technically sound. It implements the revised #549 ruling narrowly, preserves active-schema safety checks, avoids changing the parquet naming contract, and validates the important engine behavior directly. No changes are required based on this review.

#### Summary

Findings: 0 Blocker, 0 High, 0 Medium, 0 Low.

Major risks: none identified in the reviewed scope.

Verification: targeted tests, full local wrapper-based test suite, formatting, lint, diff validation, and all observed PR CI checks passed.

Issue #396, opened 2026-08-11
Best-effort batch failure log records the caller's payload


iceboundrock:
Follow-up from #318 / PR #392. **Re-scoped 2026-08-11:** the two disclosure items originally filed here were fixed in PR #392 after a later review round reframed them as new-code defects rather than pre-existing ones. What remains is the logging item.

## The gap

`executeBestEffortBatch` (`internal/entity_batch_service.go`) logs a failed operation like this:

```go
zap.S().Warnw(operationName+" operation failed", "operation", op, "error", err)
```

`op` is a `forma.EntityOperation`, so `Data`/`Updates` — arbitrary caller-supplied document content — land in the log at Warn level.

The repository is explicit that this is not acceptable elsewhere. `internal/entity_write_validation.go` deliberately withholds the payload from its report-only warning, saying entity data is caller content and may be sensitive. The batch path does the opposite for every failed operation.

Log the schema name, the row id and the error. Not the document.

## Scope note

Only the **best-effort** paths are affected. The atomic batch paths return the error to `internal/httpapi`, which applies the normal disclosure rules, and every batch handler in `internal/httpapi/handlers.go` sets `Atomic: true`. The exposed surface is the Go API and `cmd/sample/csv_importer.go`.

## Already fixed in #392, recorded so the history is not confusing

- **Credential scrubbing on the published message.** `internal/entity_batch_error.go` now runs `redact.ConnStringPassword`, the same helper `internal/httpapi` wraps.
- **Raw disclosure of non-publishing errors.** The fallback no longer returns `err.Error()`; a withheld failure reports `internal error`, mirroring what the HTTP path emits for the same class, with the full error kept in the log.

The correlation problem that second change exposes — a withheld batch failure has no id to join body and log — is #398.

## Acceptance

- The failure log no longer contains caller-supplied document content.
- A test drives a best-effort batch whose operation fails, and asserts the log line carries the schema name, the row id and the error but not the payload.

PR #392, opened 2026-08-10
fix(internal): #318 strip the whole relation subtree, not just the exact root

iceboundrock:
Closes #318

## The bug

`RelationIndex.StripComputedFields` matched relation roots by **exact key**. So `{"contactSnapshot": {...}}` was discarded before validation, while the registered dotted attribute `"contactSnapshot.name"` sailed through and was persisted to the EAV table. Two spellings of one attribute, disagreeing.

The strip now removes the whole subtree — the root and every dotted descendant, both spellings — by the prefix rule in `coversRelationSubtree` (`internal/relation_index.go`). The `"."` in `root+"."` is load-bearing: a bare `HasPrefix` would also delete a sibling named `contactSnapshotX`, silently dropping an ordinary caller attribute. `TestStripKeepsSamePrefixSibling` is the anchor for that boundary.

Dropping stays **silent**. Neither spelling produces a new `4xx`: the nested spelling has always been dropped without a rejection, and this brings the dotted spelling into line rather than adding a rejection to payloads accepted today.

## Why the persisted value was not behaviour anyone could rely on

Two independent reasons, both now written down in `docs/error-handling.md`:

1. **Unreadable wherever enrichment applies.** Relation enrichment replaces the entire `contactSnapshot` object with the parent entity's fragment on read, so a caller-written value there is overwritten before it reaches the caller.
2. **Deleted by the next update.** The update path rebuilds the dotted attribute name into a nested object (`FromPersistentRecord` → `FromAttributes`), the strip removes that object from the merged document, and the update's scoped EAV replace deletes every attribute id registered for the schema before writing back only what survived. The dotted value is not written back.

## The startup guard

`internal.ValidateRelationSchemas` (`internal/relation_index.go`) rejects, at startup, a schema that lists a relation root in root-level `required`. Such a schema describes an entity that cannot be written at all: the root is stripped before the validator sees it, so every create fails with a missing-required error the caller cannot fix by sending the field, and every update fails too under strict update validation.

**What it was chosen over:** resolving the conflict by validating *before* stripping. That defeats the guard's purpose — it judges a document no row will ever hold — and it would newly `4xx` payloads that are accepted today, which is precisely what this change set is trying not to do.

[...]

Issue #317, opened 2026-07-26
Report-only schema validation has no aggregate signal, and it is the default mode


iceboundrock:
Follow-up from #314 (PR #316).

## The gap

`Entity.ValidateUpdatesStrict` defaults to **false**, so every deployment ships in report-only mode for updates. That mode exists for exactly one purpose: let an operator find and repair rows that violate their schema, then flip `VALIDATE_UPDATES_STRICT` once the violations are gone.

Its only output is a per-write `Warn` (`internal/entity_write_validation.go`), carrying the schema name, schema id, row id, and the violation. That is enough to *grep*, and deliberately excludes the payload. What it cannot answer is the question the rollout actually turns on:

> Is it safe to flip strict mode yet?

There is no count, no per-schema breakdown, and no way to see the trend. An operator has to scrape logs and aggregate by hand, on the default configuration.

## Direction

A counter keyed by schema (and ideally by whether the violation was a `required` failure versus a constraint failure) is a small, behaviour-neutral addition. `internal/telemetry` already exists.

Two things worth deciding at the same time:

- **Where the aggregate lives.** A Prometheus-style counter is the obvious answer if the deployment scrapes; a periodic summary log line is the cheaper answer if it does not.
- **Whether to cap or sample the `Warn`.** On a corpus with many violating rows, report-only produces one line per write. That is fine for diagnosis and unpleasant at volume.

## Note

The `Warn` line is intentionally payload-free but the validation message it carries does embed the offending value (`enum: banana does not equal any of: [...]`). For shipped schemas that can mean `contact.email` or `primaryPhone` values reach logs. That is the deliberate choice for 400 bodies — worth confirming it is also the intended choice for log lines before adding anything that increases log volume.

iceboundrock:
**Artifact: implementation plan** (working copy at `docs/superpowers/plans/2026-08-18-issue-317-report-only-aggregate-signal.md`, gitignored). Posted per the repo's non-code-artifact rule.

# Issue #317: Aggregate Signal for Report-Only Schema Validation — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

[...]

PR #606, opened 2026-09-21
Give a failed best-effort batch operation a correlation id shared with its log line (#398)

### LINKED ISSUE #398: A withheld best-effort batch failure carries no correlation handle, unlike the HTTP redacted body
Follow-up from #318 / PR #392.

## The gap

PR #392 stopped the best-effort batch path publishing raw error text: when nothing resolves a published message, `forma.OperationError.Error` now reports `internal error` and the full error goes to the log only.

That closed a disclosure defect, and it made an older asymmetry start to matter.

When `internal/httpapi` withholds a message it emits a redacted body carrying `error_class` and `error_id`, and logs the same id alongside the full error, so an operator can join the two. `forma.OperationError` has no id field — it never has (`types.go:104-109`) — and the batch failure log (`internal/entity_batch_service.go`) emits none either.

So before #392 a caller could at least read the raw text; now they read `internal error` with nothing to quote to an operator, and the operator has nothing to grep for beyond a timestamp and a schema name.

This is a **pre-existing** asymmetry, not a regression introduced by #392 — but withholding the text is what turns it from cosmetic into an operational problem.

## What to do

Give a withheld batch failure the same correlation handle the HTTP path has:

- add an id to `forma.OperationError` — a public type change, so it wants a deliberate name and a decision about whether it is always present or only when the message is withheld;
- emit the same id in the failure log next to the full error;
- keep `Code` as it is: it classifies, it does not correlate.

Worth checking whether `internal/httpapi`'s id generation can be reused directly rather than grown a second time — the two surfaces should not disagree about the shape of an id.

## Acceptance

- A best-effort batch operation that fails with a non-publishing error yields a result entry whose id appears verbatim in the log line carrying the full error.
- The published-message case is unchanged, or carries the id too if that is judged better — stated either way rather than left implicit.
- `docs/error-handling.md`'s batch-publication section records the correlation contract next to the disclosure one.

iceboundrock:
Closes #398.

## What

Since #392 a best-effort batch failure whose error publishes nothing reports `internal error`, with the full error kept on the failure log line alone. That line carried no id: a caller had nothing to quote to an operator, and the operator nothing to grep for beyond a timestamp and a schema name — the asymmetry with the HTTP redacted body that #398 describes.

- `forma.OperationError` gains `ErrorID` (JSON `error_id`, `omitempty`), the same Go and wire name as `httpapi.APIResponse.ErrorID`.
- `executeBestEffortBatch` mints one id per failed operation and logs it as `error_id` on the `Warnw` line next to the full `error`, and puts the same id on the result entry.
- New `internal/errorid` package holds the one generator; `internal/httpapi` now calls it in place of its two bare `uuid.NewString()` calls, so the two surfaces cannot disagree about the shape of an id.
- `cmd/sample/csv_importer.go`, the one consumer of best-effort results, prints the id beside the message.
- `docs/error-handling.md` records the correlation contract next to the disclosure one in the batch-publication section.

## Decisions (stated, per the issue)

- **The id is unconditional.** A published failure carries it too. The HTTP boundary leaves a detail-less disclosed 4xx id-free because its `Debugw` line does not survive the production Info threshold; the batch path has no such branch — every failure logs the full error at `Warnw` — so every id leads to a line, including the #318 case where a published message withholds operator detail, without growing a second `HasOperatorDetail` branch.
- **The line is never sampled.** zap's production sampler keys on level and message, and every id-carrying line has a constant message, so under a failure storm it would drop lines whose ids callers already hold (review finding). Both surfaces write those lines through `errorid.Logger`, and the sampler is installed through `errorid.SamplerOption`, which wraps it in `errorid.ExemptFromSampler`: entries named `errorid` bypass it, everything else — including the #317 report-only validation line that relies on it — is sampled as before.
- **The exemption is public.** The global logger belongs to the embedding process, so `factory.NewProductionLogger`, `factory.BuildLogger` and `factory.SamplerOption` expose the installation (review finding: an external project cannot import `internal/bootstrap`). `cmd/server`, `cmd/lambda` and `cmd/sample` build their loggers through those same functions; `cmd/tools` calls `bootstrap.NewProductionLogger` directly since it does not otherwise link `factory`. An embedder on stock `zap.NewProduction()` still gets an id on every failure; the join holds up to the sampler's first 100 identical failures per second, and the docs and the `ErrorID` field comment say so.
- **`Code` is untouched.** It classifies; it does not correlate.
- **Published message text is unchanged.** The id sits beside it.

## Acceptance mapping

| Criterion | Where | Verified by |
|---|---|---|
| A withheld failure yields an entry whose id appears verbatim on the log line carrying the full error | `internal/entity_batch_service.go` | `TestBatchResultWithheldFailureCarriesACorrelationID` (zap observer; asserts UUID shape, `error_id` field equality, and that the same line's `error` holds the driver text) |
| Every issued id survives the production sampler | `internal/errorid`, `internal/bootstrap/logger.go`, `factory/logger.go`, `cmd/*` | `TestBatchResultEveryFailureSurvivesProductionSampling`, `TestRespondErrorIDsSurviveProductionSampling`, `TestProductionLoggerNeverSamplesACorrelationLine`, `TestEmbedderBatchFailureIDsSurviveProductionSampling` (150 identical failures under the real 100:100 sampler on a frozen clock; control line still cut to 100; the last one through a factory-built manager and the public logger constructor) |
| Published-message case stated, not implicit | carries the id too; docs + `types_batch.go` comment say so | `TestBatchResultPublishedFailureCarriesACorrelationID` |
| Docs record the correlation contract next to the disclosure one | `docs/error-handling.md`, "Correlation (#398)" | — |
| Reuse the HTTP id generation rather than grow a second one | `internal/errorid` used by both | `TestNewIsAParseableUUID`; existing httpapi tests still `uuid.Parse` every body `error_id` |

Also: `TestBatchResultFailuresCarryDistinctIDs` pins that two failures in one batch never share a handle.

## Validation

- `make fmt-check`, `make lint` — clean
- `./scripts/test_with_container_runtime.sh` (full `make test`) — all packages ok
- `go test ./cmd/...` after the sample change — ok
- `go test -race` over the touched packages' tests — ok

## Out of scope

#396 (the same log line records the caller's payload) is left as is; this change only adds `error_id` to that line.


iceboundrock:
### PR Review

#### Scope & Intent

This PR closes #398: since #392 a best-effort batch failure whose error publishes nothing reports `internal error`, with the full error kept only on the server log line. That line carried no id, so a caller had nothing to quote and an operator had nothing to grep for — the asymmetry with the HTTP redacted body (`error_class` + `error_id` joined to the log line).

The change, as reviewed (single commit `2bb5440`, 9 files):

- `forma.OperationError` gains `ErrorID string` (JSON `error_id,omitempty`) in `types_batch.go`.
- `executeBestEffortBatch` (`internal/entity_batch_service.go`) mints one id per failed operation via new `internal/errorid.New()`, logs it as `error_id` on the `Warnw` failure line next to the full `error`, and puts the same id on the result entry. Unconditional: published failures carry it too. `Code` and published message text are untouched.
- New `internal/errorid` package (one `New()` = `uuid.NewString()`); `internal/httpapi/error_response.go` replaces its two bare `uuid.NewString()` calls with it — pure refactor, no behavior change.
- `cmd/sample/csv_importer.go` (the only in-tree consumer of best-effort results) prints `(error_id …)` beside `Code: message`.
- `docs/error-handling.md` records the correlation contract next to the disclosure one; comments in `internal/entity_batch_error.go` updated.
- Tests: withheld-case, published-case, and distinct-ids tests in `internal/entity_batch_error_test.go` (zap observer join); shape/uniqueness tests in `internal/errorid/errorid_test.go`.

Behavior intended to stay unchanged: published message resolution (`resolveBatchErrorMessage`), `Code` classification, atomic batch paths (which return the error itself, not per-operation entries), HTTP disclosure/redaction branching and log levels, wire shape for hand-built `OperationError` (via `omitempty`). #396 (payload on the log line) is explicitly out of scope.

I reviewed the full diff, the linked issue #398 (including the implementation-plan comment), surrounding code (`entity_batch_service.go`, `error_response.go:336-393`, `handlers.go` batch handlers, `types_batch.go`, `csv_importer.go`), and the affected call sites.

Evaluated against, in order:

1. **#398 acceptance** (explicit): withheld failure yields an entry whose id appears verbatim on the log line carrying the full error; published-message case stated either way; docs record the contract next to the disclosure one; reuse HTTP id generation rather than growing a second one.
2. **#398 design decisions** (explicit, owner-approved in the plan comment): field name `ErrorID`/`error_id` matching `httpapi.APIResponse.ErrorID`; id unconditional because every batch failure logs at `Warnw` (unlike HTTP's id-free detail-less 4xx whose `Debugw` line dies at the production `Info` threshold); `Code` classifies, does not correlate; published text unchanged.
3. **Existing contracts**: `APIResponse.ErrorID` is `error_id,omitempty`, canonical UUID string (httpapi tests `uuid.Parse` every body id); batch HTTP handlers all set `Atomic: true` (verified in `handlers.go:63,234,291`), so the only producer of `OperationError` is `executeBestEffortBatch` — confirmed by grep, plus empty-slice constructors that never enter the failure loop.
4. **Repo conventions** (AGENTS.md / `coding-standard.md` / `docs/error-handling.md`): error-semantics split (this is a correlation handle, not a disclosure change); file ≤500 lines / function ≤100 lines; shared-shape helpers live in `internal/` (cf. `internal/redact`); log-level contract (`Warnw` must clear `Info`).
5. **General correctness**: UUID uniqueness/criticality, zap global-logger test pattern, JSON `omitempty` backward compatibility.

#### Findings

Specifically checked and cleared:

- **Unconditional id is sound, not a leak.** The concern would be an id correlating to nothing (the reason HTTP omits it on detail-less 4xx). Here every failure — published or withheld — emits the same `Warnw` line with the full `error`, so every id lands on a line. The #318 stripped-root case (published message that withholds operator detail) is therefore covered without a second `HasOperatorDetail` branch. The divergence from HTTP is real, stated in docs and in the `types_batch.go` comment, and justified by the differing log-level branch structure.
- **`internal/errorid` indirection is justified.** A one-line wrapper looks gratuitous in isolation, but the issue explicitly asks that the two surfaces not disagree about id shape, and the repo already sets this precedent (`internal/redact`). The httpapi swap is behavior-identical (`uuid.NewString()` → `errorid.New()`), and the shape is pinned on both sides (`TestNewIsAParseableUUID` + existing httpapi `uuid.Parse` assertions).
- **`omitempty` + always-set is the right combination.** Wire-compat concern (old readers, hand-built `OperationError` in `cmd/sample/main.go:341`): producers in this tree always set the id, consumers tolerate its absence. `TestOperationError_JSON` (hand-built, no id) still passes — verified by execution.
- **Atomic paths correctly untouched.** They return the error itself with no per-operation entry, so there is nothing to correlate; best-effort is the only writer of `OperationError.Error`.
- **Log line joins correctly.** `error_id` sits beside `error` (not interpolated into any message), matching the HTTP convention ("its own field"). Tests assert verbatim equality between entry and log field plus UUID-parseability, and that the joined line's `error` holds the driver text.
- **No contract break on `Code` or message text.** Both preserved; importer change is display-only in the sample binary.

#### Observations

One non-blocking note for the author to consider or disregard:

- `cmd/sample/csv_importer.go:235` formats `(error_id %s)` unconditionally. Today `ErrorID` is always non-empty for a live best-effort result (the only path that reaches this loop), so this is fine as-is. If an `OperationError` ever arrives without an id (hand-built, deserialized pre-#398 payload), it would render `(error_id )` with an empty handle. A conditional suffix or `%q`-style guard would make the display robust to that case, but it is not reachable in-tree now and is not worth a blocking change.

Deliberately not raised: the failure line logs the raw `err` (unscrubbed) and the full `operation` payload — both pre-existing behavior, the latter tracked by #396 and explicitly out of scope here. This PR neither improves nor worsens either; no new exposure is introduced by adding `error_id`.

Coverage is well-targeted at the contract, not at percentages:

- `TestBatchResultWithheldFailureCarriesACorrelationID` — the exact #398 scenario: asserts `Error == "internal error"`, UUID shape, `error_id` equality between entry and the single `Warnw` line, and that that line's `error` carries the driver text. Would fail without the fix (no id on either side).
- `TestBatchResultPublishedFailureCarriesACorrelationID` — pins the "unconditional" decision rather than leaving it implicit; asserts the published text is preserved and the join holds.
- `TestBatchResultFailuresCarryDistinctIDs` — two failures never share a handle.
- `TestNewIsAParseableUUID` / `TestNewDoesNotRepeat` — pin the cross-surface shape contract.
- Pre-existing guards (`TestBatchResultWithholdsAnUnpublishedFailure`, `TestBatchResultScrubsCredentialsFromAPublishedMessage`, httpapi `uuid.Parse` assertions) still pass, confirming no disclosure or shape regression.

Validation actually executed on the PR branch:

- `go test ./internal/errorid -v` — PASS (2/2).
- `go test ./internal -run 'TestBatchResult|TestOperationError' -v` — PASS (includes the 3 new tests plus `TestBatchResultPublishesOnlyThePublishedMessage`).
- `go test ./internal/httpapi -run 'TestError|TestRedact|TestWithheld|TestDisclosed|TestBody'` — PASS.
- `go test . -run 'TestOperationError|TestBatchResult'` — PASS.
- `make fmt-check`, `make lint` — clean.
- CI on the PR (read from `statusCheckRollup`, not re-run): Lint, Test, E2E Smoke, k6 Smoke, Benchmark Smoke — all SUCCESS.

Not verified: the full federated `-tags=e2e` suite and nightly production harness (container-dependent, long-running). The change touches no federated, CDC, compaction, or sqlgen code, so the risk of silent interaction there is negligible; the smoke suites that did run cover the affected HTTP/batch surface indirectly.

Minor coverage note (not a gap worth acting on): the correlation tests drive `BatchCreate` only. `BatchUpdate`/`BatchDelete` share the identical `executeBestEffortBatch` helper with only `operationName`/`errorCode`/`executor` varying, so coverage transfers by construction; parametrizing the join test across all three would add little.

#### Overall Assessment

The PR does exactly what #398 asks, no more: it closes the caller↔operator join for withheld batch failures with the same handle shape the HTTP path already uses, deduplicates the generator so the shapes cannot drift, states the one deliberate divergence (unconditional vs. HTTP's id-free branch) with its log-level rationale, and documents it where the disclosure contract lives. Implementation follows established repo patterns, stays within size/style gates, preserves wire compatibility, and leaves disclosure, classification, and atomic paths untouched. Test design is strong — it asserts the join (id equality + full-error presence), not just the presence of a field.

No blocking or high-severity issues. The single observation above is optional polish.

#### Summary

- Findings: 0 blockers, 0 high, 0 medium, 0 low.
- Observations: 1 non-blocking (importer display when id absent — not reachable in-tree).
- Risks: none material; out-of-scope items (#396 payload logging) correctly untouched.
- Verification: targeted unit tests + httpapi tests + `fmt-check` + `lint` all executed and passing; PR CI checks all green; full e2e federated suite not re-run (unaffected subsystem, stated as limitation).

iceboundrock:
Addressed the review at https://github.com/Lychee-Technology/forma/pull/606#issuecomment-5763414849. The PR carries no inline threads; the review reports no findings and one observation.

**Observation — `cmd/sample/csv_importer.go` renders `(error_id )` for an entry without an id: fixed** in 01aad62.

The underlying point is that `OperationError.ErrorID` is declared `error_id,omitempty`, so the wire contract makes the id optional even though the one in-tree producer (`executeBestEffortBatch`) always sets it. A consumer that formats the field should follow the declared contract, not the producer's current behaviour. The formatting moved into `failureReason(opErr)`, which appends ` (error_id …)` only when the entry carries one; `TestFailureReasonShowsTheCorrelationID` and `TestFailureReasonOmitsAnAbsentCorrelationID` pin both cases (first test file in `cmd/sample`; `make test` already covers `./cmd/...`).

Nothing else changed: no behaviour of the batch service, the id generator, or the HTTP path moved.

Validation run on the updated branch:
- `make fmt-check`, `make lint` — clean
- `go test ./cmd/... ./internal/errorid/` — ok

iceboundrock:
### PR Review

#### Scope & Intent

This PR implements #398: it adds `OperationError.ErrorID` (`error_id,omitempty`) for each failed best-effort operation, writes the same UUID as structured `error_id` on the full-error failure log line, shares generation with HTTP through `internal/errorid`, displays the handle in the CSV importer, and documents the correlation contract.

The review treated these as unchanged: publication/redaction of `OperationError.Error`, `Code` classification, atomic batch behavior, HTTP disclosure branches, and the optional wire shape for hand-built values. The implementation scope is limited to the public batch result, the shared UUID generator, the sample consumer, tests, and error-handling documentation.

1. **Explicit #398 requirement:** every returned best-effort failure ID must identify the `Warnw` line containing the full error, including withheld and published messages. `Code` remains a classifier, not a correlation key.
2. **Explicit PR decision:** IDs are unconditional because every batch failure is intended to log at `Warnw`, unlike the detail-less HTTP 4xx `Debugw` branch.
3. **Existing contracts:** `OperationError` is a public JSON type; `error_id` follows the HTTP field name and uses `omitempty`; HTTP handlers use atomic batches, while the Go API and CSV importer use the best-effort path.
4. **Repository conventions:** operator detail must remain discoverable at the production log threshold; error correlation is a concrete operational contract, not only a response-format concern.
5. **Library semantics:** the pinned Zap v1.27.1 production configuration samples by log level and message.

#### Findings

**[High] Production log sampling leaves many returned IDs with no error record**

* **Location:** `internal/entity_batch_service.go:161-162`; production logger setup in `cmd/server/main.go:33`, `cmd/lambda/main.go:193`, and `cmd/sample/main.go:38`
* **Issue:** Every failure of a given operation type uses the identical `Warnw` message, for example `BatchCreate operation failed`. `zap.NewProduction()` and `zap.NewProductionConfig()` enable the default 100:100 sampler. The sampler keeps the first 100 entries with a given level and message in a one-second interval, then drops entries 101 through 199, retaining only every 100th thereafter. Structured fields, including `error_id`, do not participate in that sampling key.
* **Impact:** A best-effort `BatchCreate` with 101 rapidly failing operations returns 101 distinct IDs, but the 101st ID has no line holding the full error. This is allowed by the default `MaxBatchSize` of 1000. It is also reachable in the CSV importer: it deliberately uses non-atomic batches, defaults to 100 rows, and can begin a second failing batch within the same second. The caller is again left with an unusable correlation handle, which directly violates the PR and documentation claim that every ID leads to the full-error line.
* **Evidence:** `executeBestEffortBatch` creates an ID and returns it regardless of whether `Warnw` is emitted. The pinned Zap source documents the 100:100 production sampler (`config.go:141-169`) and `zapcore/sampler.go:214-228` keys its counter only by `ent.Level` and `ent.Message`. The new observer tests install `zap.New(core)` without this sampler, so their single-failure assertions cannot expose the production behavior.
* **Recommended direction:** Preserve the invariant that each issued ID has an unsampled, structured log record containing the full error. Configure or route these correlation records so the production sampler cannot discard them; merely adding fields or changing the level does not solve this because both are part of the sampler key. Add a regression test using production-equivalent sampling and more than 100 identical batch failures that verifies every returned ID joins exactly one full-error record.

#### Observations

* Adding the exported `ErrorID` field is intentionally additive on the JSON wire, but Go consumers using unkeyed `OperationError` literals will need a source update. This is an inherent public-struct compatibility consideration for the release, not a request to remove the field required by #398.
* `TestOperationError_JSON` still exercises only an empty `ErrorID` round trip. A direct assertion that a populated value serializes as exactly `error_id`, and an empty value is absent, would protect the newly documented wire contract.

The new tests correctly cover one withheld failure, one published failure, UUID shape, distinct IDs, and CSV display with and without an ID. They do not cover the production sampler interaction described above, so they pass while the primary correlation guarantee fails under a reachable load pattern.

Validation executed during this review:

* `git diff --check origin/main...HEAD` and commit whitespace checks: passed.
* `make fmt-check`: passed.
* `go test ./internal -run ...` covering the batch publication and new correlation tests: passed.
* `go test ./internal/errorid ./internal/httpapi ./cmd/sample -count=1`: passed.
* `go test . -run TestOperationError_JSON|TestBatchResult_JSON -count=1`: passed.
* PR CI status was inspected: Test, Lint, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all pass.

`make test`, `make lint`, and a targeted `-race` run were attempted locally but could not complete because concurrent compilation exhausted the review container `/tmp` quota; the failures were `disk quota exceeded` during Go build/vet output, not test or lint findings. Full nightly federated and production E2E suites were not run locally.

#### Overall Assessment

The field design, shared generator extraction, redaction boundaries, atomic-path separation, importer handling, and documentation are otherwise consistent with the issue and existing HTTP conventions. However, the core caller-to-operator join is not reliable in the actual production logger configuration. This should be corrected before merge because the feature explicitly promises every issued handle can locate its full error.

#### Summary

* Findings: 0 blocker, 1 high, 0 medium, 0 low.
* Major risk: repeated best-effort failures can return correlation IDs whose full-error log entries were sampled out.
* Verification: formatting, diff checks, targeted tests, and all reported PR CI checks pass; full local test/lint/race gates were limited by temporary-disk exhaustion.

iceboundrock:
### Response to the review (a3e3543)

**[High] Production log sampling — fixed.** Confirmed against the pinned zap v1.27.1: `zapcore/sampler.go` keys its counter on `(level, message)` only, and `NewProductionConfig` installs 100:100. So the batch line dropped the 101st identical failure in a second, and — same root cause, not batch-specific — `respondErrorWithStatus`'s `Errorw`/`Warnw` lines did the same to the HTTP `error_id` (#301/#361) under a burst of identical failures. The PR's "every id leads to a line" claim was false in production for both surfaces.

**Root cause:** correlation records went through the sampled global logger, and nothing in the system distinguished a line a caller holds a handle to from any other line.

**Fix — route the correlation records around the sampler, keep sampling for everything else.** I chose this over `cfg.Sampling = nil` because `internal/entity_write_validation.go` (#317) documents relying on the sampler as the volume bound for the report-only validation line, and dropping sampling process-wide is an operational decision that does not belong in a batch-id PR. The mechanism is by logger name, because that is the only thing a core sees at `Check` time (fields arrive at `Write`, after the sampler has decided — which is also why adding a field or changing the level cannot help, as the review noted).

- `internal/errorid`: `LoggerName = "errorid"`, `Logger()` (the global logger under that name, so observer-based tests keep working unchanged and it follows `zap.ReplaceGlobals`), and `ExemptFromSampler(sampled, unsampled)` — a `zapcore.Core` that routes entries named `errorid` to the unsampled core and everything else to the sampled one (`With`, `Level`, `Sync` delegated to both).
- `internal/bootstrap/logger.go`: `NewProductionLogger` / `BuildLogger` / `SamplerOption` — `zap.Config.Build` with `cfg.Sampling` installed through the exemption, in the same position zap's own `WrapCore` would sit. `cmd/server`, `cmd/lambda`, `cmd/sample` (its console/verbose config too) and `cmd/tools` build their loggers there, so the contract holds in every binary that can mint an id.
- The batch failure line and the two id-minting HTTP branches write through `errorid.Logger()`. The id-free `Debugw` branch and every other line stay on the sampled global logger. Visible change: id-carrying lines gain `"logger":"errorid"`, which an operator can filter on.

**Regression tests, production-equivalent sampling:**
- `TestBatchResultEveryFailureSurvivesProductionSampling` (`internal`): real `BatchCreate`, 150 failing operations, observer under `bootstrap.SamplerOption(zap.NewProductionConfig().Sampling)`; every returned id joins exactly one line holding the withheld driver text. Verified it fails with the batch line back on `zap.S()`.
- `TestRespondErrorIDsSurviveProductionSampling` (`httpapi`): the same for both HTTP id-minting branches.
- `TestProductionLoggerNeverSamplesACorrelationLine` (`bootstrap`): through the real `NewProductionConfig` (JSON, Info, sampler) with a memory sink and a frozen clock: 150 `errorid.Logger` lines all written, 150 plain lines cut to 100 — the sampler is live and the exemption, not its absence, is what keeps the ids joinable. Plus `Sampling: nil` builds plain, the Info threshold is unchanged, and the core's `With`/`Level`/`Sync` in `internal/errorid`.

**Observations:**
- Unkeyed `OperationError` literals — replied, no change: none exist in the repo (grepped), and the field is what #398 asks for.
- `TestOperationError_JSON` — fixed: now asserts a populated id serialises as exactly `"error_id":"<uuid>"` and round-trips, and an empty one omits the key.

**Docs/comments:** `docs/error-handling.md` gains "The line is never sampled" next to the correlation paragraph; the #317 comment names the new constructor and says why that line is deliberately *not* exempt; the `respondErrorWithStatus` level-contract comment records the routing.

**Validation:** `make fmt-check`, `make lint` (root + infra) clean; `./scripts/test_with_container_runtime.sh` (full `make test`) — all 36 packages ok; `go test -race` over `internal`, `internal/errorid`, `internal/bootstrap`, `internal/httpapi`, `cmd/...`, root for the touched tests — ok. Nightly federated/production E2E not run locally.

iceboundrock:
### PR Review

#### Scope & Intent

This PR adds `OperationError.ErrorID` (`error_id`) to every failed best-effort batch operation, writes the same UUID beside the full error on the failure log record, presents it in the CSV importer, and documents the correlation contract. It also centralizes UUID generation with the existing HTTP error-response IDs and routes ID-bearing records around zap's production sampler.

The review basis was the #398 acceptance criteria and PR description, the public `forma.OperationError` / `EntityManager` and `factory.NewEntityManagerWithConfig*` contracts, the repository's error-publication documentation and Go standards, and zap's sampler semantics. The intended invariant is that every returned correlation ID identifies exactly one retained record containing the withheld detail, including failure bursts.

- #398 and the PR explicitly require the result ID and full-error log record to be joinable, and state that no issued ID may be sampled away.
- `OperationError`, `EntityManager`, and the factory are public API. The factory documents itself as the primary construction path for external projects; external modules cannot import this repository's `internal/*` packages.
- zap's production sampler counts entries by level and message per tick, not by logger name or fields; it preserves the first 100 matching records and drops subsequent non-milestone records.
- Repository documentation distinguishes caller-safe publication from operator-visible detail, so a batch result returning `internal error` must retain a dependable operator lookup path.

#### Findings

**[High] Public factory integrations can still return unjoinable IDs during normal production sampling**

- **Location:** `internal/entity_batch_service.go:164-170`; `internal/bootstrap/logger.go:12-61`; `types_batch.go:66-75`; `docs/error-handling.md:1755-1777`
- **Issue:** The sampler exemption is installed only by `internal/bootstrap.NewProductionLogger` / `BuildLogger`, but best-effort batches are a public library API. An external Forma user can create a manager through `factory.NewEntityManagerWithConfig*` and configure the global logger with ordinary `zap.NewProduction()`, but cannot import either internal package to install the exemption.
- **Impact:** In that supported integration, 101 identical failed non-atomic operations in one sampler tick each return an `OperationError.ErrorID`, while the 101st log record is dropped. The caller then holds the exact dangling handle this PR is intended to eliminate. The default maximum batch size is 1000, so this is reachable within one request during a repeated storage outage.
- **Evidence:** `errorid.Logger()` only calls `zap.S().Named("errorid")`; without the wrapper it remains behind the application's ordinary zap sampler. zap's `sampler.Check` indexes its counter using only `ent.Level` and `ent.Message`, so the logger name cannot change that result. The tests install `bootstrap.SamplerOption` themselves and therefore cover bundled entrypoints, not an external factory consumer using stock production zap.
- **Recommended direction:** Make the required logging/sampler installation available to supported external embedders through a public API and document that setup, or narrow the public `OperationError`/documentation promise so an ID is issued only when the library can ensure a retained correlation record. Add a regression test for the external-style `zap.NewProduction()` configuration chosen by the final contract.

#### Observations

- The two call-site sampling tests (`TestBatchResultEveryFailureSurvivesProductionSampling` and `TestRespondErrorIDsSurviveProductionSampling`) use wall-clock time. If the 150-record loop crosses a sampler tick, it can split into groups of 100 or fewer and pass even if the exemption is no longer applied at the call site. The lower-level sampler and bootstrap tests already use a frozen clock; applying the same approach here would make the integration regression checks deterministic.
- The shipped server, Lambda, sample, and production tool logger construction paths were audited and do install the wrapper. The issue is the separately supported external embedding path, not those binaries.

- The new tests cover UUID shape, distinct batch IDs, published and withheld batch messages, CSV rendering, the sampler wrapper itself, bootstrap production configuration, and both HTTP ID-bearing branches.
- Missing coverage: an external factory consumer using an ordinary production zap logger, plus deterministic wall-clock control in the two call-site sampler tests noted above.
- Passed: `git diff --check origin/main...origin/pr-606`
- Passed: `make fmt-check`
- Passed: `make lint`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/errorid ./internal/bootstrap -v`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal -run 'TestBatchResult.*Correlation|TestBatchResultEveryFailure' -v`
- Passed: `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/httpapi -run TestRespondErrorIDsSurviveProductionSampling -v`
- Passed: `TMPDIR=$PWD/.tmp-review GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./cmd/sample -run '^TestFailureReason' -v`
- `go test ./cmd/... -v` was not fully verifiable: `cmd/tools` passed, but linking the CGO-heavy `cmd/server`, `cmd/sample`, `cmd/lambda`, and `cmd/benchmark` test binaries failed because `/tmp` reached its disk quota. This was an environment capacity failure, not a reported test failure. Full unit, race, containerized, and E2E suites were not run.

#### Overall Assessment

The implementation is sound for the repository's packaged production binaries: UUID generation is shared, batch results and log records carry the same handle, published text remains unchanged, and the custom core correctly preserves ordinary sampling for non-correlation records. However, the feature elevates the same behavior to a public `OperationError` contract while making the mechanism required to uphold it inaccessible to public factory consumers. That leaves a reachable high-volume failure path with IDs that cannot be correlated, so the contract should be resolved before merge.

#### Summary

- Findings: 0 Blocker, 1 High, 0 Medium, 0 Low.
- Major risk: external embedders using standard `zap.NewProduction()` receive IDs for records that zap samples away.
- Verification: formatting, lint, targeted correlation/sampler suites, and the touched CSV helper pass; broad command-package coverage was limited by `/tmp` disk quota.

iceboundrock:
### Response to the review (ba953b4)

Addresses https://github.com/Lychee-Technology/forma/pull/606#issuecomment-5764271946. No inline threads on the PR; the review is one finding and two observations.

**[High] Public factory integrations can still return unjoinable IDs — fixed.** Confirmed the reading: Forma logs through zap's global logger, the embedder owns that global, and the exemption was reachable only from `internal/bootstrap`. An embedder on `factory.NewEntityManagerWithConfig*` + `zap.NewProduction()` had the 101st identical failure in a second return an id whose line was dropped.

Root cause: the contract was elevated to a public type (`forma.OperationError.ErrorID`) while the mechanism it depends on stayed internal. The two directions the review offered were weighed:

- *Narrow the promise so an id is issued only when a retained record is guaranteed.* Not implementable: zap's sampler type is unexported and a core is opaque, so the library cannot tell whether the global logger samples. The nearest approximation is a documented caveat, which is now stated, not the whole answer.
- *Expose the installation publicly.* Done, in `factory`, the documented composition root for external projects. It already imports `internal/bootstrap` and zap, and it is where an embedder already looks.

Changes:

- `factory/logger.go`: `NewProductionLogger`, `BuildLogger` (an embedder's own `zap.Config`), `SamplerOption` (a hand-assembled core). Doc comments state the failure they prevent and, for `SamplerOption`, the ordering constraint (it must stand where zap's sampler would, not above one).
- `cmd/server`, `cmd/lambda`, `cmd/sample` now build their loggers through those `factory` functions, so the shipped binaries run the path embedders are told to use. `cmd/tools` stays on `bootstrap.NewProductionLogger` — it does not otherwise link `factory`, and the implementation is the same function.
- `errorid.SamplerOption` (moved from `bootstrap`) is the single place the sampler is wrapped; `bootstrap.BuildLogger` and `factory.SamplerOption` both delegate to it. No second copy of the wrapping.
- `NewEntityManagerWithConfig`'s doc, `OperationError.ErrorID`'s comment and `docs/error-handling.md` name the setup and state what stock `zap.NewProduction()` still gets: an id on every failure, joinable up to the sampler's first 100 identical failures per second.

Regression test at the public boundary, `factory.TestEmbedderBatchFailureIDsSurviveProductionSampling`: a manager built through the factory fixture, the global logger installed via `factory.BuildLogger` over the real production config (JSON, Info, 100:100 sampler) writing to a file, 150 failed best-effort operations inside one sampler tick. Every returned id appears on exactly one written line named `errorid`; 150 ordinary warnings logged alongside are cut to 100, so the sampler is provably live. Verified it fails with `cfg.Build` in place of `BuildLogger` (`id … must join exactly one written line`).

**Observation: wall-clock time in the two call-site sampler tests — fixed.** New `internal/errorid/erroridtest` package: `FrozenClock()` and `ObserveUnderProductionSampler(t, level)` (observer under `zap.NewProductionConfig().Sampling` on the frozen clock, installed as the global for the test). `TestBatchResultEveryFailureSurvivesProductionSampling`, `TestRespondErrorIDsSurviveProductionSampling`, the `errorid` and `bootstrap` sampler tests and the new factory test all use it; the two private `frozenClock` copies are gone. `errorid`'s tests moved to `package errorid_test` to use the helper without a cycle.

**Observation: shipped binaries install the wrapper — no change needed** beyond the switch to the `factory` names above; audited the same four paths.

**Validation:** `make fmt-check`, `make lint` (root + infra) clean; `./scripts/test_with_container_runtime.sh` (full `make test`) all packages ok; `go test -race` over `internal` (`TestBatchResult*`), `internal/errorid/...`, `internal/bootstrap`, `factory`, `internal/httpapi` for the sampler/correlation tests — ok. Nightly federated/production E2E not run locally.

iceboundrock:
### PR Review

#### Scope & Intent

This PR implements #398 by adding `forma.OperationError.ErrorID` (`error_id`) to each failed best-effort operation, logging the same UUID beside the full error, reusing the generator for HTTP error responses, displaying the handle in the CSV importer, and documenting the caller-to-operator correlation contract. It also adds a Zap sampler exemption and exposes logger construction through `factory` so error IDs survive the production 100:100 sampler.

The review treated publication/redaction of `OperationError.Error`, `Code` classification, atomic batch behavior, HTTP disclosure branches, and the optional wire shape for hand-built values as unchanged. The intended invariant is explicit in #398 and the PR: every issued ID must identify one retained operator log record containing the full detail, including burst failures.

1. **Explicit #398 / PR contract:** a best-effort result ID must join to the `Warnw` record containing the full error; IDs are intentionally unconditional for batch failures; `Code` remains classification only.
2. **Public API and wire contract:** `OperationError` is exported; `ErrorID` deliberately matches the HTTP `error_id` name and uses `omitempty` for manually constructed values.
3. **Repository error-handling convention:** a caller-safe `internal error` response needs a dependable operator-visible lookup path.
4. **Zap semantics:** production sampling is keyed by level and message; `Logger.Named` appends name segments with periods; `zapcore.Core.Sync` is a lifecycle operation without an idempotence guarantee.
5. **Repository standards:** code size limits, wrapped errors, and targeted behavioral tests. Changed Go files and functions remain within the documented limits.

#### Findings

**[Medium] Named global loggers lose the sampler exemption**

* **Location:** `internal/errorid/errorid.go:52`; `internal/errorid/sampler.go:64-78`
* **Issue:** `errorid.Logger()` calls `zap.S().Named("errorid")`, while the exemption only accepts `entry.LoggerName == "errorid"`. A valid embedder setup such as `zap.ReplaceGlobals(logger.Named("service"))` therefore emits correlation records as `service.errorid`, which follows the production sampler.
* **Impact:** With `factory.BuildLogger(zap.NewProductionConfig(), ...)`, a named global logger, and 150 identical failures in one sampler tick, result entries and redacted HTTP bodies still receive 150 IDs but only the first 100 correlation records are retained. This reintroduces the dangling-handle behavior that this PR explicitly promises to prevent.
* **Evidence:** Zap v1.27.1 `Logger.Named` joins an existing name and child name with `.`. The exact-name checks in both `Check` and `Write` consequently route `service.errorid` to `sampled`. Existing factory, bootstrap, errorid, batch, and HTTP sampling tests install an unnamed global logger only, so they do not exercise this public integration.
* **Recommended direction:** Preserve the exemption when `errorid` is the terminal logger-name segment, and add a public-factory regression test that installs a named global logger and verifies every one of more than 100 returned IDs joins exactly one full-error record. Update the documentation that currently promises `logger: errorid`, since the correct named-root output will be `service.errorid`.

**[Low] Logger shutdown syncs the shared core twice**

* **Location:** `internal/errorid/sampler.go:24-25,81-85`
* **Issue:** `SamplerOption` passes the same core as the sampler input and `unsampled` core. Zap sampler embeds that core, so `c.sampled.Sync()` already calls `core.Sync()`; `c.unsampled.Sync()` calls it a second time.
* **Impact:** Every logger created through `factory.NewProductionLogger`, `factory.BuildLogger`, or bootstrap invokes the sink lifecycle operation twice at shutdown. The comment calls the second call a no-op, but `zapcore.Core.Sync` does not require that property; custom sinks can flush, block, or fail on each call.
* **Evidence:** Zap v1.27.1 `sampler` embeds `Core` directly, inheriting its `Sync`. The normal production construction here always shares the same underlying core. The existing sync test only checks propagation from two distinct cores and does not assert the normal shared-core behavior.
* **Recommended direction:** Make the standard `SamplerOption` path sync the underlying core once, while retaining distinct-core behavior only if it is an intentional supported use case. Add a counting-core test for one `logger.Sync()` call through the public constructor.

**[Low] CSV output can attach a new correlation ID to the wrong row**

* **Location:** `cmd/sample/csv_importer.go:218-235`
* **Issue:** The importer builds every operation with the same schema name, then maps each `OperationError` back to the first batch operation with that schema. Thus every returned failure is rendered as the first row of the batch, irrespective of the failed operation.
* **Impact:** A batch where row 2 succeeds and row 3 fails reports the row-3 error ID against row 2. The operator lookup itself succeeds, but the newly displayed handle misleadingly identifies the wrong source CSV row.
* **Evidence:** `ImportFromReader` assigns one `schemaName` to every batch operation at `cmd/sample/csv_importer.go:162-168`; the inner lookup at lines 222-228 breaks on the first matching schema. Existing tests cover only `failureReason`, not `processBatch` with failures after the first row.
* **Recommended direction:** Correlate returned failures to their original operation/index rather than schema name, and cover a mixed success/failure batch with at least one failure after the first input row. If the public result needs an index to make duplicate operations unambiguous, preserve that invariant at the API boundary rather than guessing in the importer.

#### Observations

The shared UUID generator, structured `error_id` field, unconditional batch rule, and frozen-clock sampler tests are sound choices. The public `OperationError` field is an intentional additive contract change required by #398; downstream Go users with unkeyed literals will need to account for the additional field, while JSON producers without an ID remain compatible through `omitempty`.

The added tests meaningfully cover withheld and published batch failures, UUID shape and uniqueness, HTTP ID-bearing branches, production sampling for an unnamed global logger, custom sampler hooks, and CSV rendering with and without an ID. They do not cover the named-global integration, one-sync lifecycle invariant, or multi-row importer mapping described above.

Validation executed on PR head `ba953b47da8af767daf7a41c7d657d6baf740ba8`:

* `git diff --check main...HEAD` - passed.
* `make fmt-check` - passed.
* `go vet ./internal/errorid ./internal/httpapi ./internal ./factory ./cmd/sample` - passed.
* `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test ./internal/errorid ./internal/httpapi ./internal ./factory ./cmd/sample` - passed.
* `./scripts/test_with_container_runtime.sh` - passed full `make test` using the detected rootless Podman socket.
* `make lint` - passed, including the infra module.
* `GOCACHE=/tmp/opencode/gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal ./internal/bootstrap ./internal/errorid ./internal/httpapi ./factory ./cmd/sample` - passed.
* Submitted CI is green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

The nightly full federated and production E2E suites were not rerun locally; this PR does not change their implementation paths.

#### Overall Assessment

The core feature is well scoped: IDs are minted once, appear verbatim beside the retained full error in the ordinary factory setup, reuse the HTTP format, preserve publication boundaries, and are backed by focused tests and documentation. However, the stated no-sampling guarantee is incomplete for a normal named-global Zap setup, which is part of the public embedding surface. That contract gap should be corrected before merge. The lifecycle and importer mapping issues are narrower but worth resolving with the same change set because both affect the newly introduced operational path.

#### Summary

* Findings: 0 blocker, 0 high, 1 medium, 2 low.
* Major risk: named global loggers can return IDs whose correlation records were sampled away during failure bursts.
* Verification: formatting, lint, full containerized unit suite, targeted tests, targeted race tests, static analysis, and all submitted CI checks passed.

iceboundrock:
### Response to the review (b6812aa)

Addresses https://github.com/Lychee-Technology/forma/pull/606#issuecomment-5764956577. No inline threads; the review is three findings. Two commits: `310cd03` (the two sampler findings) and `b6812aa` (the importer finding).

**[Medium] Named global loggers lose the sampler exemption — fixed.** Confirmed: `Logger.Named` joins with a period, so under `zap.ReplaceGlobals(logger.Named("svc"))` the line arrives as `svc.errorid` and the exact-name check routed it through the sampler. Reproduced first with the new `TestEmbedderBatchFailureIDsSurviveProductionSamplingUnderANamedGlobal` (`factory`): under exact matching it fails with the correlation lines cut to 100; under the fix all 150 ids join exactly one line while the ordinary lines are still cut to 100.

Root cause: the exemption keyed on the whole logger name, but the only part of that name Forma controls is the segment `Logger` appends; the prefix is the embedder's. The fix keys on the last segment (`errorid.IsCorrelationLogger`: `errorid` or `*.errorid`). The alternative — having `Logger` produce an absolute name — does not exist in zap; `Named` only appends. `TestSamplerOptionRoutesByTheLastNameSegment` (`internal/errorid`) pins the boundary: `errorid`, `svc.errorid`, `a.b.errorid` exempt; `""`, `svc`, `myerrorid`, `errorid.child` sampled. Docs and the `LoggerName` comment now say the `logger` field is `errorid` or `svc.errorid` under a named global.

**[Low] Logger shutdown syncs the shared core twice — fixed.** Confirmed: zap's sampler embeds the core, so `sampled.Sync()` was already `core.Sync()`, and the wrapper ran it again. The wrapper now syncs the core once. The "two distinct cores" arrangement was never a supported use — `SamplerOption` was the only constructor that built one — so `ExemptFromSampler` is folded into it rather than kept public with a contract the type cannot enforce. `TestSamplerOptionLevelAndSync` builds through `SamplerOption` over a counting sink and asserts one `logger.Sync()` is one sink `Sync`, with the sink's error propagated.

**[Low] CSV output can attach a new correlation id to the wrong row — fixed, at the API boundary.** Confirmed, and the arithmetic around it was also wrong: `startRowNum + idx` assumed every CSV row entered the batch, but a row that fails mapping never does, so any such row inside the batch window shifted the reported rows after it.

Root cause: `BatchResult.Failed` gave a caller a copy of the operation and nothing that says which one it was, so the importer guessed by schema name — and every operation in its batch has the same schema. Per the review's direction, `forma.OperationError` gains `Index int` (JSON `index`, no `omitempty`: zero is a position), set by `executeBestEffortBatch` from the loop index; `TestBatchResultFailureCarriesItsOperationIndex` sends three identical operations, fails the middle one, and asserts `Index == 1`. The importer now keeps each batched operation's CSV row in a parallel `batchRows` slice and reports `batchRows[opErr.Index]`, which retires the `rowNum-len(batch)+1` / `rowNum-len(batch)` arithmetic in both import loops. `TestImportReportsAFailureAgainstItsOwnRow` drives `ImportFromReader` through a stub `EntityBatchCreator`: rows 2–6 with an unmappable row 4 and failures at 5 and 6, batch size 3, and asserts each error lands on its own row with its own id. `TestOperationError_JSON` pins `index` on the wire.

This is an additive change to a public type, beyond what #398 listed; it is the general fix rather than an importer-local heuristic, and HTTP never returns `Failed` entries (every handler sets `Atomic: true`), so the wire change reaches Go embedders only. Say if you would rather it went to a follow-up.

**Validation** (all run locally on `b6812aa`):

- `make fmt-check`, `make lint` (root and infra) — clean
- `./scripts/test_with_container_runtime.sh` (full `make test`) — all packages ok
- `go test -race ./internal ./internal/bootstrap ./internal/errorid ./internal/httpapi ./factory ./cmd/sample` — ok
- The named-global test was confirmed to fail against the previous exact-name check before the fix.

iceboundrock:
### PR Review

#### Scope & Intent

This PR implements #398: failed best-effort batch operations now return `OperationError.ErrorID`, log the same UUID beside the full error, share UUID generation with the HTTP error boundary, and show the handle in the CSV importer. It additionally ensures ID-bearing records bypass Zap's production sampler, exposes that logger construction through `factory`, adds the operation input index needed to map importer failures to source rows, and documents the correlation contract.

The review treated caller-safe error publication, `Code` classification, atomic batch behavior, HTTP disclosure decisions, and raw-error/payload handling tracked by #396 as unchanged. The load-bearing invariant is that a returned ID identifies a retained operator record with the full detail.

- #398 and the PR description require a caller-to-operator join between each best-effort failure ID and its full-error log record, including failure bursts.
- `OperationError`, `EntityBatchCreator`, and the `factory` package are public API. `factory.BuildLogger` explicitly supports a caller-owned `zap.Config`, including custom levels.
- The repository error-handling contract distinguishes caller-safe messages from operator detail; `internal error` therefore needs a dependable operator lookup path.
- Zap core semantics apply: a logger's enabled level is evaluated before the sampler/exemption can write an entry. The sampler exemption cannot make a `Warn` entry pass an `Error` threshold.
- Repository size, error-handling, formatting, and validation conventions in `AGENTS.md` and `coding-standard.md` were applied. Changed source files and functions are within the documented size limits.

#### Findings

**[Medium] Custom error-level factory loggers still issue dangling correlation IDs**

- **Location:** `internal/entity_batch_service.go:157-172`, `internal/httpapi/error_response.go:362-370`, `factory/logger.go:33-39`, `internal/errorid/sampler.go:52-70`
- **Issue:** `factory.BuildLogger` accepts a custom Zap configuration and preserves its level, but a best-effort failure always mints and returns an ID while logging the corresponding record at `Warn`. With a supported configuration such as `cfg := zap.NewProductionConfig(); cfg.Level = zap.NewAtomicLevelAt(zap.ErrorLevel)`, the `Warn` record is disabled before `samplerExemption` can route it around sampling. The result still contains `ErrorID`.
- **Impact:** An embedder using `factory.BuildLogger` at `Error` level receives the same dangling handle #398 is meant to remove. The same gap affects the disclosed-4xx-with-operator-detail HTTP branch, which also returns an ID for a `Warn` record. Redacted HTTP failures logged at `Error` are unaffected.
- **Evidence:** `executeBestEffortBatch` calls `errorid.New()` before `errorid.Logger().Warnw`; `samplerExemption.Enabled` and `Check` delegate to the underlying core and do not lower its threshold. I reproduced this with the public factory API, an `Error`-level production config, and the correlation logger: the `Warn` line was not written. The existing sampling tests correctly prove sampler bypass under an Info-level core, but none exercises an elevated level.
- **Recommended direction:** Define and enforce the invariant for custom-level configurations. Either document and enforce a `Warn`-or-lower minimum wherever batch IDs or withheld-detail HTTP IDs are issued, or avoid issuing an ID when no correlation record can be retained. Add a factory-level regression test for an `Error`-level config that protects the selected contract.

#### Observations

- The final head addresses the earlier high-signal risks well: the sampler bypass is public to embedders, named global loggers match the terminal `errorid` segment, shared-core `Sync` occurs once, frozen clocks make burst tests deterministic, and CSV failures use the returned input index rather than schema-name guessing.
- `Index` is an intentional public Go/JSON contract extension beyond the original ID-only scope. It is appropriate for the importer fix, but external users of unkeyed `OperationError` literals need a source update, and hand-built values now serialize an `index` field. This is a release-compatibility consideration rather than a request to remove the field.

Existing and added tests meaningfully cover:

- withheld and published batch failures, UUID shape/uniqueness, and ID-to-full-error joins;
- burst behavior for batch and both ID-bearing HTTP branches under production-equivalent sampling;
- public factory construction, a named global logger, sampler hooks, level propagation, and single-sync behavior;
- CSV ID presentation and correct row mapping through an unmappable row.

The missing behavioral coverage is the elevated-level configuration in the finding above.

Validation executed on final head `b6812aa`:

- `git diff --check origin/main...HEAD` and clean-worktree check: passed.
- `make fmt-check`: passed.
- `go vet ./factory ./internal/errorid ./internal/bootstrap ./internal/httpapi ./cmd/sample`: passed.
- Focused unit tests across `internal`, `internal/errorid`, `internal/bootstrap`, `internal/httpapi`, `factory`, and `cmd/sample`: passed.
- Focused `-race` tests for the touched logger, factory, HTTP, and importer paths: passed.
- Public-factory `Error`-threshold reproduction described in the finding: passed, confirming the dropped `Warn` line.
- Reported PR CI is green: Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke.

`make lint` was attempted locally but could not complete because the review environment exhausted its `/tmp` quota while compiling linter dependencies. This was an environment-capacity failure, not a lint diagnostic; the submitted CI lint check is successful. Full nightly federated and production E2E suites were not rerun locally; this PR does not modify those subsystems.

#### Overall Assessment

The implementation is otherwise well scoped and carefully tested. It preserves publication boundaries, uses one UUID generator across batch and HTTP surfaces, correctly protects correlation records from the production sampler without disabling sampling for unrelated logs, and fixes the CSV row-association defect at the public result boundary. The remaining logger-level gap means the central correlation guarantee is not true for every custom configuration that the new public factory API accepts, so it should be resolved before merge.

#### Summary

- Findings: 0 blocker, 0 high, 1 medium, 0 low.
- Major risk: `Error`-level `factory.BuildLogger` configurations return IDs whose `Warn` correlation records are not retained.
- Verification: focused unit, race, vet, formatting, diff checks, and submitted CI passed; local lint was limited by `/tmp` disk quota.

iceboundrock:
Response to the [review](https://github.com/Lychee-Technology/forma/pull/606#issuecomment-5767124554), addressed in 1324019.

### [Medium] Custom error-level factory loggers still issue dangling correlation IDs: **fixed**

**Root cause.** Each id site minted its id unconditionally and then wrote the line. zap checks the level in `Logger.check` before any core sees the logger name, so `samplerExemption` cannot help a line the level has already dropped. That covers more than the Error-level `BuildLogger` config. zap's default no-op global, in a process that never installs a logger, returned ids for every line, the redacted `Errorw` branch included.

**Options considered.**
- *Document and enforce a Warn-or-lower minimum.* `BuildLogger` would have to reject or clamp a caller's level. That overrides a supported config, and it still misses the no-op global and an `AtomicLevel` raised at run time.
- *Log correlation lines at Error.* This gives caller-triggerable 4xx and per-operation batch failures Error's alerting weight, which the HTTP boundary deliberately avoids. It also still fails above Error.
- *Issue no id when the line will not be written* (chosen). This is the rule the HTTP boundary already applies to the detail-less 4xx ("an error_id that correlates to nothing is worse than none"), applied at run time. It holds under every configuration without a policy on the embedder's level.

**Change.**
- New `errorid.Log(level, msg, kv...) string` checks `Core().Enabled(level)`. If the check fails it writes nothing and returns `""`. Otherwise it mints the id, writes the line on `errorid.Logger` with `error_id`, and returns the id.
- `New` is now unexported (`newID`), so `Log` is the only way to get an id. Any future id-bearing surface therefore inherits the rule instead of re-deriving it.
- All three id sites use `Log`: the batch failure (`Warn`), the withheld-detail 4xx (`Warn`) and the redacted response (`Error`). The response types already carry `omitempty`, and `cmd/sample` already omits the suffix for an empty id (01aad62), so no consumer changes were needed.
- Contract docs updated: `OperationError.ErrorID`, `factory.BuildLogger` (the level decides which ids are issued), `respondErrorWithStatus`, and a new "No line, no id" paragraph in `docs/error-handling.md`.

One remaining gap: a level raised between the check and the write can lose one line. Only an `AtomicLevel` changed under live traffic can cause this, and the `Log` doc comment says so.

**Tests.**
- `factory.TestEmbedderBatchFailuresIssueNoIDAboveWarn`: through the public `BuildLogger` with `cfg.Level = ErrorLevel`, a 150-failure batch returns no `ErrorID`, while `Index` and `Error` are still set, and no failure line is written. It fails on the previous head.
- `httpapi.TestRespondErrorIssuesNoIDForADroppedLine`: under an Error-level logger, a withheld-detail 400 carries no `error_id` and logs nothing. A redacted 5xx still carries an id that joins its Error line.
- `errorid.TestLogIssuesNoIdForALineItWillNotWrite`: covers both the level-above-line case and the no-op global. `TestLogWritesTheIdItReturns` pins the join: level, logger name and fields.
- The httpapi tests that installed `zap.NewNop()` to silence output were relying on the dangling id to assert `error_id` shape. They now install `discardLogger()`, which is enabled at every level and discards its output, like a deployed logger.

### Observation: `Index` public contract extension: **acknowledged, no change**
Agreed that this is a release-compatibility note, not a defect. `OperationError` literals in the repo are all keyed, and `index` is additive on the wire. It should be called out in the release notes.

### Validation (on 1324019)
- `make fmt-check`: passed
- `make lint` (root + infra): passed
- `./scripts/test_with_container_runtime.sh` (the full `make test` unit suite): passed
- `go test -race ./internal/errorid/... ./internal/bootstrap ./internal/httpapi ./factory ./cmd/sample ./internal/`: passed

Not rerun locally: the nightly federated and production E2E suites. This change does not touch those subsystems.

iceboundrock:
### PR Review

#### Scope & Intent

This PR adds an `error_id` to failed best-effort batch results and writes the same ID beside the full error in the warning log. It reuses the ID path for HTTP errors, adds `OperationError.Index` so repeated batch operations map back to their original inputs, updates the CSV sample, and introduces public logger builders that exempt correlation lines from zap sampling.

I evaluated it against issue #398’s correlation acceptance criteria, the existing disclosure rule that `OperationError.Error` contains only a published message, the HTTP error-ID contract, and zap’s level/message sampling behavior.

- **Explicit (#398):** a withheld failure must have an ID on both the result and the full-error log line; the published-message case must be explicit; `Code` remains classification rather than correlation.
- **Existing contract/docs:** operator-only detail stays out of caller-visible messages; IDs are useful only when their corresponding log line is retained. `OperationError` is public Go/JSON API, so its new fields and wire names need consistent behavior.
- **Repository conventions:** preserve best-effort result semantics, map batch failures by operation position when operations can be identical, and protect observable contracts with behavioral tests.

#### Findings

**[Low] The sampler exemption also exempts host-application loggers named `errorid`**

* **Location:** `internal/errorid/sampler.go` — `IsCorrelationLogger` / `samplerExemption.Check`
* **Issue:** The exemption identifies Forma’s correlation logger solely by whether the final logger-name segment is `errorid`. Any logger in the embedding process with that name is routed around sampling as well.
* **Impact:** When an embedder installs `factory.SamplerOption` or `factory.BuildLogger`, its shared process logger is wrapped. A host component logging through `zap.S().Named("errorid")` (or `Named("svc").Named("errorid")`) therefore bypasses the sampler, even when the line has no correlation ID. This weakens the documented volume bound for non-correlation logs.
* **Evidence:** `IsCorrelationLogger("svc.errorid")` returns true, and `samplerExemption.Check` sends every matching entry directly to the unsampled core. The named-global test intentionally verifies that same suffix behavior for Forma’s logger, so the routing cannot distinguish the two sources.
* **Recommended direction:** Make the exemption use a collision-resistant, explicitly reserved logger namespace or otherwise distinguish Forma’s correlation path from host loggers; document the reserved name so the sampling guarantee is clear to embedders.

#### Observations

- The correlation guarantee for embedded applications depends on using `factory.NewProductionLogger`, `factory.BuildLogger`, or `factory.SamplerOption`. With stock `zap.NewProduction()`, repeated identical failures can still produce IDs whose lines are sampled away after the first 100 per second. The PR documents this integration requirement and tests the supported logger setup.
- `Index` is zero-based and identifies the failed operation in the original request. The CSV importer now maps it through a parallel row-number slice, including when an earlier row fails mapping; this addresses the duplicate-operation ambiguity without changing the existing per-operation failure order.

- Tests cover withheld and published failures joining the same full-error line, distinct IDs, JSON field names/omission, production-sampler bursts for batch and HTTP errors, logger-level thresholds, named globals, and CSV row mapping.
- `make fmt-check` — passed.
- `./scripts/test_with_container_runtime.sh` — passed the full `make test` suite using the configured Podman socket.
- Targeted tests for `internal/errorid`, `internal/bootstrap`, `internal/httpapi`, `internal`, `factory`, and `cmd/sample` — passed.
- `make lint` — did not pass locally: golangci-lint reported unused `nolint:staticcheck` directives at `factory/factory.go:386-387`, which are unchanged by this PR. The PR’s CI Lint check passes.
- PR CI checks reported passing for Lint, Test, E2E Smoke Tests, Benchmark Smoke, and k6 Smoke.

#### Overall Assessment

The main correlation flow, its HTTP reuse, and the new input-index mapping are well covered and behave consistently with the documented error-disclosure contract. No blocker or higher-severity correctness issue was identified. The remaining review finding is the narrow logger-name collision that can bypass sampling for host-application logs.

#### Summary

**Findings:** 0 Blocker, 0 High, 0 Medium, 1 Low. **Major risks:** none identified; one low-impact sampling-bound concern for host loggers named `errorid`. **Verification:** formatting and full unit tests passed; CI checks pass; local lint reported diagnostics at unchanged lines.

iceboundrock:
Response to [the review](https://github.com/Lychee-Technology/forma/pull/606#issuecomment-5804593672): fixed in f2371e9.

**[Low] Sampler exemption also exempts host loggers named `errorid`: fixed.**

**Root cause.** The exemption keyed on the logger name, and the name is shared with every logger in the embedding process. A reserved name would only make a collision less likely. The real problem is that a name cannot tell Forma's correlation logger apart from anyone else's.

**Alternatives considered.**
- *Reserved, namespaced name* (e.g. `forma.errorid`), which is the review's first suggestion. It's a smaller diff, but it's still spoofable, it still needs the last-segment matching for named globals, and it becomes a convention every embedder has to be told about.
- *Unwrap to the unsampled core by type assertion in `errorid.Logger`.* This is identity-based, but it breaks silently as soon as an embedder puts another core on top of the exemption (a tee, their own `WrapCore` after `BuildLogger`'s sampler).
- *Private marker delivered through `With`*: **chosen.** A core sees only level, message and name at `Check`, and per-line fields arrive only after the sampler has decided. `With` fields, however, reach every core in the chain when the child logger is built, and wrappers forward them as the `zapcore.Core` contract requires.

**Change.**
- `errorid.Logger()` is now `zap.L().Named("errorid").With(correlationField)`. `correlationField` is a `SkipType` field whose value has an unexported type, so no code outside `internal/errorid` can build it, and a core that does not know it encodes nothing.
- `samplerExemption.With` recognises the marker, strips it, and returns the core beneath the sampler. Children of that logger stay unsampled. Every other logger, including a host `Named("errorid")` or `Named("svc").Named("errorid")`, goes through the sampler. `Check`/`Write` no longer look at the name at all.
- `IsCorrelationLogger` is removed. `LoggerName` stays as the operator-facing `logger` label and is documented as a label, not a reserved namespace.
- Docs (`docs/error-handling.md`, "The line is never sampled") and the package docs now describe the marker instead of the name rule.

**Tests.**
- `TestSamplerOptionExemptsOnlyTheCorrelationLogger` (`internal/errorid`) replaces `...RoutesByTheLastNameSegment`. Under an unnamed global and under globals named `svc` and `a.b`, Forma's logger (and a child of it with fields) is exempt, and host loggers named `errorid` / `svc.errorid` are sampled.
- `TestSamplerOptionStripsTheMarker`: the sink receives only the caller's fields.
- `TestSamplerOptionExemptsCorrelationLinesOverAHandBuiltCore` (`factory`, public `SamplerOption`) is extended: 150 host lines under `Named("errorid")` are cut to 100, while all 150 `errorid.Log` lines survive.
- I ran the two new tests against the previous `sampler.go`/`errorid.go` and both failed, then passed with the fix. The existing embedder storm tests (plain and named global) still pass unchanged.

**Validation.** `make fmt-check` and `make lint` are clean. `./scripts/test_with_container_runtime.sh` (full `make test`) passes. `go test -race` over `internal/errorid/...`, `internal/bootstrap`, `factory` and `internal/httpapi` passes. On the lint note in the review: `make lint` passed locally on this branch with the pinned v1.64.8 run through the Makefile's `GOTOOLCHAIN=go1.26.0+auto`. The unused-`nolint` diagnostics at `factory/factory.go:386-387` suggest a different golangci-lint or toolchain version was used.

**Observations.** No change. The stock `zap.NewProduction()` limitation is still documented as before, and the `Index` mapping is unchanged.

iceboundrock:
### PR Review

#### Scope & Intent

This PR adds `OperationError.ErrorID` (`error_id`) to best-effort batch failures and logs the same ID beside the full error. It also adds the zero-based operation `Index` used to map repeated operations back to source rows, updates the CSV importer, centralizes HTTP/batch ID logging, and provides logger builders that exempt correlation lines from zap sampling.

The intended behavior is to preserve the existing disclosure boundary (`OperationError.Error` contains only published text), keep `Code` as the failure classification, and make an ID useful only when its corresponding line is written. Published failures intentionally carry IDs too.

- **Explicit, issue #398:** a withheld failure’s result ID must match the ID on the full-error log line; the published-message behavior must be explicit; `Code` remains classification, not correlation.
- **Existing contract:** `docs/error-handling.md` and the HTTP error boundary require operator-only details to stay out of published messages. IDs are correlation handles and must match the corresponding log record.
- **Inferred repository/API conventions:** `OperationError` is public Go and JSON API; new fields need stable wire names and documented omission/default behavior. Batch failures must be mapped by operation position when operations may be identical.
- **Library semantics:** zap 1.27.1 checks logger enablement when logging, and its production sampler groups by level and message. The PR’s logger wrapper and opt-in factory API are therefore part of the behavior under review, not just plumbing.

#### Findings

**[Low] A live level increase can still publish an ID for a line zap drops**

* **Location:** `internal/errorid/errorid.go`, `Log` (lines 80–86)
* **Issue:** `Log` checks `Core().Enabled(level)`, mints the ID, then calls `SugaredLogger.Logw`, which performs another enablement check before writing.
* **Impact:** If an embedder raises a zap `AtomicLevel` between those checks (for example, `Warn` to `Error` while a failure is being handled), the second check suppresses the log line but `Log` still returns the ID. The batch result or HTTP response can then contain an ID that has no matching record, violating the documented “no line, no ID” correlation invariant. This is a narrow runtime-configuration race, not a normal static-level failure.
* **Evidence:** zap v1.27.1 `SugaredLogger.log` checks `s.base.Core().Enabled` and then calls `s.base.Check`; `Log` performs an earlier independent `Enabled` call. Existing tests cover fixed thresholds but not a threshold change between these checks. The PR’s latest review response also records this as the remaining gap.
* **Recommended direction:** Make the write decision once and use that accepted checked entry to write the generated ID and fields, so a second level check cannot discard the line after the ID is issued; add a regression test for a level change in that window.

#### Observations

- Correlation records intentionally bypass zap’s production sampler, while other records remain sampled. As a result, full-error HTTP and batch log volume now scales with correlated failures rather than the prior 100-identical-records-per-second cap. This matches the PR’s stated guarantee, but the reviewed request paths contain no application-level rate limiter; sustained failure traffic can therefore increase log volume and encoding/sink work substantially. Deployment ingress and log-sink limits should account for this trade-off.
- The operation index, row-number side table, `error_id,omitempty` behavior, and preservation of published error text are consistent with their documented contracts.

- Tests cover withheld and published failure correlation, distinct IDs, ID omission when a static logger level drops the line, production-sampler bursts, named globals, JSON field names/omission, and CSV row mapping across unmappable rows.
- The missing case is the live `AtomicLevel` transition described in the finding.
- Local checks run:
  - `make fmt-check` — passed.
  - `make lint` — passed (root and infra module).
  - `./scripts/test_with_container_runtime.sh` — passed the full unit suite; it configured the rootless Podman socket and all packages passed.
  - `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/errorid ./internal/httpapi ./internal ./factory ./cmd/sample` — passed.
  - `git diff origin/main...HEAD --check` — passed.
- A direct `make test` attempt initially failed because testcontainers had no `$XDG_RUNTIME_DIR`; the repository-provided container-runtime script configured Podman and the full suite then passed. GitHub CI checks for Lint, Test, E2E Smoke, Benchmark Smoke, and k6 Smoke are all passing.

#### Overall Assessment

The core batch correlation and CSV index changes are consistent with the issue and have strong behavioral coverage. The sampler wrapper addresses the production sampling behavior and keeps unrelated records sampled. One low-severity edge remains around a dynamically changed zap level; the intentional increase in correlated log volume is the main operational trade-off.

#### Summary

**Findings:** 0 Blocker, 0 High, 0 Medium, 1 Low. **Major risks:** none; one narrow AtomicLevel correlation race and the documented log-volume trade-off. **Verification:** formatting, lint, full unit suite, targeted race tests, diff checks, and all reported CI checks passed.

iceboundrock:
Response to the [review above](https://github.com/Lychee-Technology/forma/pull/606#issuecomment-5804965155), addressed in 0c92381.

### [Low] A live level increase can still publish an ID for a line zap drops — **fixed**

**Root cause.** `errorid.Log` made the write decision twice: its own `Core().Enabled(level)` probe, then `SugaredLogger.Logw`, which runs its own `Enabled` + `Logger.Check`. The id was minted between the two, so an `AtomicLevel` raised in that window dropped the line after the id existed.

**Fix.** `Log` now makes the decision once and writes through the entry that decision returned:

```go
checked := Logger().With(keysAndValues...).Desugar().Check(level, msg)
if checked == nil {
	return ""
}
id := newID()
checked.Write(zap.String("error_id", id))
```

`CheckedEntry.Write` hands the entry to the cores that accepted it and does not check the level again. So a raise either lands before `Check` (no line, no id) or after it (the line is committed and carries the id). The sugared pairs go through `With`, which uses the same `sweetenFields` conversion as `Logw`, so a malformed pair is handled the same way as before, and the fields still come ahead of `error_id`. Call sites (`internal/entity_batch_service.go` and both `internal/httpapi/error_response.go` paths) keep their signature and behavior. This was the only place the id was minted, so no other call site had the gap.

**Alternatives considered.** (a) Mint the id first and have the caller throw it away when the line is dropped. That can't work, because `Logw` doesn't report whether it wrote. (b) Convert the pairs into `zap.Field`s inside `errorid`. That would add a second copy of zap's sugaring rules, and `With` already does the conversion.

**Regression test.** `TestLogIdMatchesALineWhenTheLevelRisesMidWrite` (`internal/errorid`) wraps an observer core that raises its `AtomicLevel` from Warn to Error right after its first consult. One subtest raises it after `Enabled`, the other after `Check`. Both assert the invariant: an id is returned exactly when a line carrying it was written. On the previous `Log`, the "raised after enabled" subtest fails (id returned, 0 lines). With the fix, both pass.

`docs/error-handling.md` ("No line, no id") now describes the single check and names the test. I also removed the "a level raised … can still lose one line" caveat from the `Log` doc comment.

### Observation: log volume from unsampled correlation lines — **documented, no code change**

This is the intended trade-off. The exemption exists so that no line whose id a caller holds is ever dropped. Putting a rate limit inside the logger would bring back exactly that failure, so any limit has to sit upstream of the logger. I added a paragraph to the "The line is never sampled" section of `docs/error-handling.md`. It says that correlation-line volume scales with the failure rate (at most `MaxBatchSize` lines per best-effort batch), that Forma applies no rate limit of its own on these paths, and that ingress limits and log-sink capacity should be sized with that in mind.

### Validation
- `make fmt-check`: passed
- `make lint` (root + infra): passed
- `go test -race ./internal/errorid/... ./internal/httpapi ./internal ./factory ./internal/bootstrap ./cmd/sample`: passed
- `./scripts/test_with_container_runtime.sh` (full `make test`): all packages passed
- New test run against the pre-fix `errorid.go`: fails as expected (`expected: 1, actual: 0`)

iceboundrock:
### PR Review

#### Scope & Intent

PR #606 addresses issue #398 by adding `OperationError.ErrorID` (JSON `error_id`) and writing the same UUID next to the full error on each best-effort failure log line. It reuses that ID path for HTTP error responses, adds `OperationError.Index` so callers such as the CSV importer can map failures back to input rows, and provides logger constructors that keep ID-bearing lines out of zap production sampling.

The review basis was the issue acceptance, the published-message/redaction rules in `docs/error-handling.md`, the best-effort batch contract and public result types, and zap 1.27.1 level, `Check`, and sampling behavior. The intended invariants are that published text remains unchanged, withheld detail stays out of results, IDs are only returned when the associated line is accepted, and ordinary log lines retain sampling.

- Issue #398 requires the result ID to match the log line carrying the full error, reuses one ID shape across batch and HTTP paths, and explicitly states whether published failures also carry IDs. The PR states and tests the unconditional-when-written behavior.
- The repository error contract keeps operator-only detail in logs, publishes only authored messages, and scrubs credentials at the boundary.
- `OperationError.Index` is the zero-based position in `BatchOperation.Operations`; `Code` remains classification, not correlation.
- The logger integration follows the repository factory/bootstrap composition pattern. Zap production sampling keys on level and message, and `CheckedEntry.Write` avoids a second level check after acceptance.

#### Findings

**[Low] Avoid encoding fields before checking whether the line is enabled**

* **Location:** `internal/errorid/errorid.go:88` (`Log`)
* **Issue:** `Logger().With(keysAndValues...)` builds a child logger before `Check` determines whether the requested level is enabled. Zap `With` passes those fields to `Core.With`; the production JSON core encodes contextual fields there. Thus an `ErrorLevel` logger—which this PR supports and tests—still serializes the complete operation and error for every disabled `Warn` batch line, then returns no ID and writes nothing.
* **Impact:** A large operation payload or repeated batch failures can consume avoidable CPU and allocations even though logging is filtered. The prior `Warnw` path checked `Enabled` before converting its fields.
* **Evidence:** Zap 1.27.1 `SugaredLogger.With` calls `Logger.With`, and `zapcore.ioCore.With` eagerly adds fields to its encoder. The current no-ID tests assert output only, so they do not detect this work.
* **Recommended direction:** Keep the single `Check`/`Write` acceptance path, but attach event fields to the accepted `CheckedEntry` (or defer field evaluation until acceptance) instead of putting per-event payload fields on a logger before checking.

**[Low] Preserve caller attribution when `AddCaller` is enabled**

* **Location:** `internal/errorid/errorid.go:88` (`Log`)
* **Issue:** The `Logger.Check` call now occurs inside the shared `errorid.Log` helper. With a custom zap config or option that enables `AddCaller`, the entry caller therefore points into `internal/errorid/errorid.go`, rather than to `executeBestEffortBatch` or `respondErrorWithStatus` as the previous direct log calls did.
* **Impact:** Embedders using caller metadata lose the source location that distinguishes the batch and HTTP error paths.
* **Evidence:** Zap captures caller metadata during `Logger.Check`; the production defaults disable caller metadata, so this affects custom caller-enabled loggers only.
* **Recommended direction:** Preserve the original callsite through caller-skip handling in the helper and add a caller-enabled logger assertion.

#### Observations

The sampler exemption deliberately removes the production sampler’s 100-identical-lines-per-second bound for every ID-bearing line. This can make error-log volume scale with failing request volume; the PR documents that trade-off, and Forma has no in-process request-rate limiter. Deployments therefore need ingress/load-shedding and log-sink capacity controls for these paths.

The sample importer’s index-based row mapping and conditional ID rendering are exercised by regression tests. The result/log correlation, UUID shape, disabled-level behavior, mid-write `AtomicLevel` change, named globals, marker stripping, and production sampler behavior also have focused coverage.

- `make fmt-check` — passed.
- `make lint` — passed, including the infra module.
- `./scripts/test_with_container_runtime.sh` — passed; the complete `make test` suite ran with the configured rootless Podman socket.
- `GOCACHE=$PWD/.gocache GOFLAGS=-buildvcs=false GOTOOLCHAIN=go1.26.0+auto go test -race ./internal/errorid/... ./internal/httpapi ./internal/bootstrap ./internal ./factory ./cmd/sample ./cmd/server ./cmd/lambda ./cmd/tools` — passed.
- GitHub CI checks visible on the PR — Lint, Test, E2E Smoke Tests, k6 Smoke, and Benchmark Smoke all succeeded.

The field-evaluation and caller-attribution behaviors above are not currently pinned by tests. The full diff was re-read after the implementation and checks; no other material regressions were identified.

#### Overall Assessment

The core correlation contract, error disclosure boundary, JSON shape, index mapping, and sampler routing appear sound. No confirmed data-integrity, credential-disclosure, or ID/log-join defect was found. The two low-severity findings affect filtered custom loggers and caller-enabled custom loggers respectively; the uncapped correlation-log volume is an explicit operational trade-off.

#### Summary

**Findings: 0 Blocker, 0 High, 0 Medium, 2 Low.** Main residual risk is unbounded ID-bearing log volume under sustained failures, documented for deployment-level controls. Formatting, lint, full tests, targeted race tests, and reported GitHub CI checks passed.

iceboundrock:
Closed in favour of #608, which re-implements #398 from `origin/main`. This PR had turned a correlation-id feature into a guarantee that zap would keep every correlated log line: a sampler-exemption core, public logger constructors in `factory`, and id issuance keyed on level. That guarantee is about the embedder's logging infrastructure, which the issue never asked for. #608 keeps the contract at emission instead: 9 files and +220/−6 lines, against 33 files and +1667/−79 here. The full analysis, including the review-by-review root causes, is at https://github.com/Lychee-Technology/forma/issues/398#issuecomment-5807349278. The `OperationError.Index` and importer change moved to #607. The branch `398-batch-error-id` is left in place for reference.

